PR #1, opened 2025-03-21
Refactor Bifrost for plugin support and multi-provider architecture along with tests

danpiths:


danpiths:
* **#1** <a href="https://app.graphite.dev/github/pr/maximhq/bifrost/1?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#2](https://github.com/maximhq/bifrost/pull/2) <a href="https://app.graphite.dev/github/pr/maximhq/bifrost/2?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#9](https://github.com/maximhq/bifrost/pull/9) <a href="https://app.graphite.dev/github/pr/maximhq/bifrost/9?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.dev/github/pr/maximhq/bifrost/1?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

interfaces/provider.go
```diff
@@ -2,10 +2,10 @@ package interfaces
...
 // LLMUsage represents token usage information
 type LLMUsage struct {
-	PromptTokens     int     `json:"prompt_tokens"`
-	CompletionTokens int     `json:"completion_tokens"`
-	TotalTokens      int     `json:"total_tokens"`
-	Latency          float64 `json:"latency,omitempty"`
+	PromptTokens     int      `json:"prompt_tokens"`
+	CompletionTokens int      `json:"completion_tokens"`
+	TotalTokens      int      `json:"total_tokens"`
+	Latency          *float64 `json:"latency"`
```
danpiths: let us have an omit empty here if it is an optional field. or did it cause an issue to not have this key present in the request json?

interfaces/provider.go
```diff
@@ -30,21 +30,21 @@ type Tool struct {
...
-	ToolChoice     *string     `json:"toolChoice,omitempty"`
-	Tools          *[]Tool     `json:"tools,omitempty"`
-	FunctionCall   *string     `json:"functionCall,omitempty"`
-	Functions      *[]Function `json:"functions,omitempty"`
+	TestRunEntryID *string     `json:"testRunEntryId"`
+	PromptTools    *[]string   `json:"promptTools"`
+	ToolChoice     *string     `json:"toolChoice"`
+	Tools          *[]Tool     `json:"tools"`
+	FunctionCall   *string     `json:"functionCall"`
+	Functions      *[]Function `json:"functions"`
```
danpiths: why did we remove all the omitempty?

danpiths (commented):
why did we remove omitempty for all optional fields @Pratham-Mishra04 ?

providers/anthropic.go
```diff
@@ -18,6 +52,7 @@ type AnthropicProvider struct {
 // NewAnthropicProvider creates a new AnthropicProvider instance
 func NewAnthropicProvider() *AnthropicProvider {
 	return &AnthropicProvider{
+		// @comment let us have this be controllable
```
danpiths: please resolve this, and if you have let us remove this comment

danpiths:
also can we add proper descriptions with each PR that lists the changes made and why did we do it

interfaces/provider.go
```diff
@@ -150,64 +29,157 @@ const (
...
 
-const (
-	UserRole      Role = "user"
-	AssistantRole Role = "assistant"
-	SystemRole    Role = "system"
-)
+// ModelParameters represents the parameters for model requests
+type ModelParameters struct {
+	TestRunEntryID *string   `json:"test_run_entry_id"`
+	PromptTools    *[]string `json:"prompt_tools"`
```
danpiths: we don't need these, they are maxim specific

interfaces/provider.go
```diff
@@ -150,64 +29,157 @@ const (
...
+	ToolChoice     *string   `json:"tool_choice"`
```
danpiths: this could also be `tool_choice: {"type": "function", "name": "get_weather"}`, need to incorporate that

interfaces/plugin.go
```diff
@@ -0,0 +1,19 @@
...
+}
+
+type BifrostRequest struct {
+	Model  string
+	Input  RequestInput
+	Params *ModelParameters
+}
+
+type Plugin interface {
+	PreHook(ctx context.Context, req *BifrostRequest) (context.Context, *BifrostRequest, error)
```
danpiths: you do not have to return back the context here

interfaces/plugin.go
```diff
@@ -0,0 +1,19 @@
...
+type RequestInput struct {
+	TextInput *string
+	ChatInput *[]Message
+}
+
+type BifrostRequest struct {
+	Model  string
+	Input  RequestInput
+	Params *ModelParameters
+}
```
danpiths: can we shift `BifrostRequest` to interfaces/bifrost.go? and similarly any interfaces/structs that relate to bifrost purely (like `BifrostResponse` too in `provider.go`)

interfaces/provider.go
```diff
@@ -150,64 +29,157 @@ const (
...
+	ID          string                          `json:"id"`
+	Choices     []BifrostResponseChoice         `json:"choices"`
+	ChatHistory *[]BifrostResponseChoiceMessage `json:"chat_history"`
+	Provider    SupportedModelProvider          `json:"provider"`
+	Usage       LLMUsage                        `json:"usage"`
+	BilledUsage *BilledLLMUsage                 `json:"billed_usage"`
+	Cost        *LLMInteractionCost             `json:"cost"`
+	Model       string                          `json:"model"`
+	Created     string                          `json:"created"`
+	Params      *interface{}                    `json:"model_params"`
```
danpiths: why do we not have this as `ModelParameters`?

interfaces/provider.go
```diff
@@ -150,64 +29,157 @@ const (
...
+	Text    *string      `json:"text"`
+	Sources *interface{} `json:"sources"`
+	Type    *string      `json:"type"`
+}
+
+// BifrostResponseChoiceMessage represents a choice in the completion response
+type BifrostResponseChoiceMessage struct {
+	Role      ModelChatMessageRole `json:"role"`
+	Content   string               `json:"content"`
+	Image     json.RawMessage      `json:"image"`
```
danpiths: why is this a `json.RawMessage`?

danpiths: also this isn't a convention, where did you pick this up from?

bifrost.go
```diff
@@ -19,75 +20,107 @@ const (
...
 type Bifrost struct {
 	account       interfaces.Account
-	providers     []interfaces.Provider   // list of processed providers
-	requestQueues map[string]chan Request // provider request queues
-	wg            sync.WaitGroup
+	providers     []interfaces.Provider // list of processed providers
+	plugins       []interfaces.Plugin
+	configs       map[interfaces.SupportedModelProvider]interfaces.ProviderConfig
+	requestQueues map[interfaces.SupportedModelProvider]chan ChannelMessage // provider request queues
+	wg            map[interfaces.SupportedModelProvider]*sync.WaitGroup
```
danpiths: can we rename this to `waitGroup`

bifrost.go
```diff
@@ -19,75 +20,107 @@ const (
...
+	case interfaces.Bedrock:
+		return providers.NewBedrockProvider(config), nil
+	case interfaces.Cohere:
+		return providers.NewCohereProvider(config), nil
+	default:
+		return nil, fmt.Errorf("unsupported provider: %s", providerKey)
+	}
+}
+
+func getConfigForProvider(providerKey interfaces.SupportedModelProvider, configs map[interfaces.SupportedModelProvider]interfaces.ProviderConfig) (*interfaces.ProviderConfig, error) {
```
danpiths: why have we not kept this method on bifrost? we could have only one parameter that way for `providerKey` and could have picked up `configs` from `bifrost.Configs`

bifrost.go
```diff
@@ -19,75 +20,107 @@ const (
...
 	}
 
 	return bifrost, nil
 }
 
-func (bifrost *Bifrost) SelectFromProviderKeys(provider interfaces.Provider, model string) (string, error) {
+func (bifrost *Bifrost) SelectKeyFromProviderForModel(provider interfaces.Provider, model string) (string, error) {
 	keys, err := bifrost.account.GetKeysForProvider(provider)
 	if err != nil {
 		return "", err
```
danpiths, line 117: <p class='graphite__hidden'><i>[Re: line +149]</i></p>
please rename this to `randomSource`
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.dev/github/pr/maximhq/bifrost/1?utm_source=unchanged-line-comment">Graphite</a>.</p>

bifrost.go
```diff
@@ -142,23 +175,23 @@ func (bifrost *Bifrost) SelectFromProviderKeys(provider interfaces.Provider, mod
...
-		key, err := bifrost.SelectFromProviderKeys(provider, req.Model)
+		key, err := bifrost.SelectKeyFromProviderForModel(provider, req.Model)
 		if err != nil {
 			req.Err <- err
 			continue
 		}
 
 		if req.Type == TextCompletionRequest {
-			result, err = provider.TextCompletion(req.Model, key, req.Input.(string), req.Params)
+			result, err = provider.TextCompletion(req.Model, key, *req.Input.TextInput, req.Params)
```
danpiths: **IMPORTANT**: we need to check if TextInput exists (if it doesn't it might end up crashing go directly due to nil pointer dereference. please check any and all places for existence of a pointer's value before you dereference them

bifrost.go
```diff
@@ -142,23 +175,23 @@ func (bifrost *Bifrost) SelectFromProviderKeys(provider interfaces.Provider, mod
...
 		} else if req.Type == ChatCompletionRequest {
-			result, err = provider.ChatCompletion(req.Model, key, req.Input.([]interfaces.Message), req.Params)
+			result, err = provider.ChatCompletion(req.Model, key, *req.Input.ChatInput, req.Params)
```
danpiths: similarly need to check here for ChatInput's existence

interfaces/account.go
```diff
@@ -11,8 +11,9 @@ type Key struct {
 	Weight float64  `json:"weight"`
 }
 
+// TODO one get config method
 type Account interface {
-	GetInitiallyConfiguredProviders() ([]Provider, error)
+	GetInitiallyConfiguredProviderKeys() ([]SupportedModelProvider, error)
```
danpiths: let's keep the name of this method to be `GetInitiallyConfiguredProviders` only

interfaces/account.go
```diff
@@ -11,8 +11,9 @@ type Key struct {
...
 	GetKeysForProvider(provider Provider) ([]Key, error)
```
danpiths: also let us just pass in `SupportedModelProvider` here as the argument, the users should not have to worry about dealing with an unknown interface to them but rather just enums directly

bifrost.go
```diff
@@ -181,73 +214,98 @@ func (bifrost *Bifrost) GetProviderFromProviderKey(key interfaces.SupportedModel
...
-		queue = bifrost.requestQueues[string(providerKey)]
+		queue = bifrost.requestQueues[providerKey]
 	}
 
 	return queue, nil
 }
 
-func (bifrost *Bifrost) TextCompletionRequest(providerKey interfaces.SupportedModelProvider, model, text string, params *interfaces.ModelParameters) (*interfaces.CompletionResult, error) {
+func (bifrost *Bifrost) TextCompletionRequest(providerKey interfaces.SupportedModelProvider, req *interfaces.BifrostRequest, ctx context.Context) (*interfaces.BifrostResponse, error) {
 	queue, err := bifrost.GetProviderQueue(providerKey)
```
danpiths, line 241: this whole flow does not handle provider not being present in the initial providers but coming on later on while the system is live. you need to call the `GetConcurrencyAndBufferSizeForProvider` and make the queue too in case it doesn't exist and only rely on configs.

bifrost.go
```diff
@@ -19,75 +20,107 @@ const (
...
 		go bifrost.processRequests(provider, queue)
 	}
 
 	return nil
 }
 
 // Initializes infinite listening channels for each provider
-func Init(account interfaces.Account) (*Bifrost, error) {
-	bifrost := &Bifrost{account: account}
+func Init(account interfaces.Account, plugins []interfaces.Plugin, configs map[interfaces.SupportedModelProvider]interfaces.ProviderConfig) (*Bifrost, error) {
```
danpiths: this whole flow is wrong, the configs for each provider should be given via the accounts as providers themselves are dynamically coming from there. you should be taking configs for each provider along with concurrency and buffer size. make the method be `GetProviderConfig` which will return all, i.e., the concurrency, the buffer size and the other network config you have

bifrost.go
Pratham-Mishra04: it is being handled in GetProviderQueue()

interfaces/plugin.go
Pratham-Mishra04: yes sure, even I was thinking about that

bifrost.go
```diff
@@ -181,73 +214,98 @@ func (bifrost *Bifrost) GetProviderFromProviderKey(key interfaces.SupportedModel
...
-		Type:     TextCompletionRequest,
+	for _, plugin := range bifrost.plugins {
+		ctx, req, err = plugin.PreHook(ctx, req)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	queue <- ChannelMessage{
+		BifrostRequest: *req,
```
danpiths, line 261: again need to check for nil values, do not dereference nil pointers

bifrost.go
```diff
@@ -181,73 +214,98 @@ func (bifrost *Bifrost) GetProviderFromProviderKey(key interfaces.SupportedModel
...
 	select {
 	case result := <-responseChan:
+		for _, plugin := range bifrost.plugins {
+			result, err = plugin.PostHook(ctx, result)
+
+			if err != nil {
+				return nil, err
+			}
+		}
+
```
danpiths: post hooks are implemented in reverse order. the application should be
pre 1
pre 2
pre 3
post 3
post 2
post 1

bifrost.go
```diff
@@ -264,7 +322,9 @@ func (bifrost *Bifrost) Shutdown() {
 	}
```
danpiths, line 339: <p class='graphite__hidden'><i>[Re: line +317]</i></p>
we would almost always print all logs with `[BIFROST] ...` and we do not need to mentioned graceful shutdown initiated
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.dev/github/pr/maximhq/bifrost/1?utm_source=unchanged-line-comment">Graphite</a>.</p>

bifrost.go
```diff
@@ -264,7 +322,9 @@ func (bifrost *Bifrost) Shutdown() {
...
 
 	// Wait for all workers to exit
-	bifrost.wg.Wait()
+	for _, wg := range bifrost.wg {
+		wg.Wait()
+	}
 
 	fmt.Println("Bifrost has shut down gracefully.")
```
danpiths: similarly we need all logs everywhere to be `[BIFROST] ..` and do not need to state the gracefully just that it shut down

interfaces/provider.go
Pratham-Mishra04: forgot the change rawMessge to ImageContent mb. also I think cohere follows this response structure and found it little convenient so followed it

interfaces/provider.go
danpiths: **IMPORTANT**: we need to solve this overall everywhere, need to exclude any empty in our json, cannot have the key in the req body (unless required by the provider to make the API call)

just include `omitempty` everywhere please (unless it breaks some flow)

interfaces/provider.go
Pratham-Mishra04: forgot to change 😓

bifrost.go
danpiths: better to just implement a custom logger that has methods like `info`, `error`, `debug`, `warn` and `fatal`. this could be passed to the `Init` function to have their own loggers used. it is helpful to users if they want to record logs and store it in some place else we can have our own wrapper on top of it that just adds like a color + [BIFROST] to the log in the beginning.

interfaces/provider.go
Pratham-Mishra04: yes I'll do that but we aren't using the structs in interfaces directly fro request bodies, its mostly for the users, not the api. Every provider has its own implementation for creating this req body from this bifrost request. also there's a function in utils.PrepareParams() which excludes all the empty value keys from the model parameters.

interfaces/provider.go
danpiths: got it, can you put this structure decision on the bifrost channel please; so that everyone can put their opinion?(can tag the eng team)

also for any conventions that you have broken out of openai, please also list them within the same thread so that we are aware of what design decision we taken. we can all discuss on the convenience vs convention bit there 🙇

interfaces/provider.go
Pratham-Mishra04: still I'll add omit empty for precautions

bifrost.go
danpiths: no it isn't it ultimately just resolves to the static array of configs that you prepare on init, no dynamic-ness here

bifrost.go
Pratham-Mishra04: okay should we do this on priority?

interfaces/provider.go
danpiths: yes please, thank you 🙇 
also let us just keep it in our struct and the provider structs so that even if the users end up json marshaling it, they do not get null keys

bifrost.go
danpiths: i mean yes, we need a logger for sure. almost all golang libraries support a logger

interfaces/provider.go
Pratham-Mishra04: yes okay

bifrost.go
Pratham-Mishra04: nahi it checks for queues present, if not then it allocates it. am I missing something?

	var queue chan ChannelMessage
	var exists bool

	if queue, exists = bifrost.requestQueues[providerKey]; !exists {
		config, err := bifrost.getConfigForProvider(providerKey)
		if err != nil {
			return nil, fmt.Errorf("failed to get config for provider: %v", err)
		}

		if err := bifrost.prepareProvider(providerKey, config); err != nil {
			return nil, err
		}

		queue = bifrost.requestQueues[providerKey]
	}

	return queue, nil

danpiths: aree here the `getConfigForProvider` function (when the requestQueue isn't there) checks the static configs array you have made in the Init function :')

Pratham-Mishra04: yes got it

interfaces/provider.go
Pratham-Mishra04: okay, in the TS implementation, it was string only so just referred that but yes went through Anthropic and OpenAI docs, can keep a common struct for our input and transform them accordingly in providers

Pratham-Mishra04: something like this

type ToolChoice struct {
	Type                   string `json:"type"`
	Name                   string `json:"name"`
	DisableParallelToolUse *bool  `json:"disable_parallel_tool_use"`
}

Pratham-Mishra04: so we only take this for tool choice, not strings and we prepare the reqbody ourselves in provider. cause anyway only openai supports strings rest anthropic n all don't

interfaces/provider.go
```diff
@@ -1,213 +1,33 @@
...
 
-// ToolCallResults represents a collection of tool call results
-type ToolCallResults struct {
-	Version int              `json:"version"`
-	Results []ToolCallResult `json:"results"`
+type MetaConfig struct {
+	SecretAccessKey   *string           `json:"secret_access_key"`
+	Region            *string           `json:"region"`
+	SessionToken      *string           `json:"session_token"`
+	ARN               *string           `json:"arn"`
```
danpiths: @Pratham-Mishra04 still don't see omitempty here, please go through everything once and make sure you add them unless omitting the key causes any issue

PR #29, opened 2025-05-06
feat: add Google Vertex AI provider support

Pratham-Mishra04:
# Add Google Vertex AI Provider Support

This PR adds support for Google Vertex AI as a new provider in Bifrost, allowing users to access Google's AI models like Gemini and third-party models hosted on Vertex AI.

Key changes:
- Implemented a new `VertexProvider` that supports chat completion with Google's models
- Added authentication via Google Cloud credentials
- Created region-specific model routing for Vertex AI endpoints
- Refactored provider code to extract common request preparation logic
- Added support for both OpenAI-style and Anthropic-style models on Vertex
- Enhanced error handling and debugging for all providers

The implementation uses Google's OAuth2 authentication and doesn't require API keys like other providers, instead using service account credentials. The PR also updates the gitignore to exclude credential files for security.

### DIFF
--- a/core/schemas/provider.go
+++ b/core/schemas/provider.go
@@ -49,6 +49,10 @@ type MetaConfig interface {
 	GetDeployments() map[string]string
 	// GetAPIVersion returns the API version
 	GetAPIVersion() *string
+	// GetProjectID returns the project ID
+	GetProjectID() *string
+	// GetAuthCredentialPath returns the path to the authentication credentials for the provider
+	GetAuthCredentialPath() *string
 }
 
 // ConcurrencyAndBufferSize represents configuration for concurrent operations and buffer sizes.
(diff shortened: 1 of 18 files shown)


Pratham-Mishra04:
* **#30** <a href="https://app.graphite.dev/github/pr/maximhq/bifrost/30?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#29** <a href="https://app.graphite.dev/github/pr/maximhq/bifrost/29?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/maximhq/bifrost/29?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

core/tests/account.go
```diff
@@ -197,6 +197,28 @@ func (baseAccount *BaseAccount) GetConfigForProvider(providerKey schemas.ModelPr
...
+			NetworkConfig: schemas.NetworkConfig{
+				DefaultRequestTimeoutInSeconds: 30,
+				MaxRetries:                     1,
+				RetryBackoffInitial:            100 * time.Millisecond,
+				RetryBackoffMax:                2 * time.Second,
+			},
+			MetaConfig: &meta.VertexMetaConfig{
+				ProjectID: os.Getenv("VERTEX_PROJECT_ID"),
+				ModelRegions: map[string]string{
+					"google/gemini-2.0-flash-001":      "us-central1",
```
akshaydeo: Thes are not hardcoded regions. So depends on if you are using serverless endpoints or deploying yourselves. So we can keep it as is coming from the user for the request. 

Also one thing we need to test here is - if it comes from google-genai SDK - our flow still holds true or not?

core/schemas/provider.go
```diff
@@ -36,7 +36,7 @@ type MetaConfig interface {
 	// GetSecretAccessKey returns the secret access key for authentication
 	GetSecretAccessKey() *string
```
akshaydeo: do you think now converting API key should be an interface{}/Generic based struct and it returns based on which provider its calling? cc @danpiths

akshaydeo: So we don't have to keep on adding new methods to this interface and bloat it

akshaydeo (changes requested):

core/schemas/provider.go
danpiths: i'm sorry i didn't get the should be an interface{}/Generic based struct

core/tests/account.go
Pratham-Mishra04: sorry didn't get comes from google-genai sdk bit

akshaydeo: as in you can hit bifrost like

```python
from google import genai

client = genai.Client(api_key=geminiApiKey, base_url="bifrost_url")
```

core/schemas/provider.go
akshaydeo: The problem here is - every extra field we add to the config, we have to add a new method like `GetSecretAccessKey`, `GetServiceAccountAccessKey`.

Better we keep overall `MetaConfig` as `map[string]interface{}` and parse it to specific type in corresponding providers

akshaydeo (approved):

akshaydeo:
### Merge activity

* **May 12, 3:38 PM GMT+5:30**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/maximhq/bifrost/29).
* **May 12, 3:39 PM GMT+5:30**: @akshaydeo merged this pull request with [Graphite](https://app.graphite.dev/github/pr/maximhq/bifrost/29).

PR #487, opened 2025-09-18
fix(openrouter): fix pricing support for openrouter

mcowger:
## Summary

- The pricing framework is updated to correctly parse and calculate costs for OpenRouter models, recognizing their unique naming convention (e.g., `openrouter/google/gemini-pro`).
- Model normalization logic is refined to distinguish between OpenRouter models and other namespaced models (e.g., `google/gemma-7b`).

## Changes

Minor changes to pricing key definitions for models in openrouter that have double slashes in the name (`openrouter/google/gemini-2.5-flash`, for example)

## Type of change

- [X] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [ ] Core (Go)
- [ ] Transports (HTTP)
- [X] Providers/Integrations
- [ ] Plugins
- [ ] UI (Next.js)
- [ ] Docs

## How to test

Passed existing tests, and verified manually.

If adding new configs or environment variables, document them here.

NA

## Screenshots/Recordings

## Breaking changes

- [ ] Yes
- [X] No

## Related issues

NA

## Security considerations

None apparent.

## Checklist

- [X] I read `docs/contributing/README.md` and followed the guidelines
- [X] I added/updated tests where appropriate
- [X] I updated documentation where needed
- [X] I verified builds succeed (Go and UI)
- [X] I verified the CI pipeline passes locally if applicable

### DIFF
--- a/core/bifrost.go
+++ b/core/bifrost.go
@@ -780,6 +780,14 @@ func (bifrost *Bifrost) getProviderQueue(providerKey schemas.ModelProvider) (cha
 
 // CORE INTERNAL LOGIC
 
+// normalizeOpenRouterRouting ensures that requests with "openrouter/" model prefix
+// are correctly routed to the OpenRouter provider.
+func normalizeOpenRouterRouting(req *schemas.BifrostRequest) {
+	if strings.HasPrefix(req.Model, "openrouter/") {
+		req.Provider = schemas.OpenRouter
+	}
+}
+
 // shouldTryFallbacks handles the primary error and returns true if we should proceed with fallbacks, false if we should return immediately
 func (bifrost *Bifrost) shouldTryFallbacks(req *schemas.BifrostRequest, primaryErr *schemas.BifrostError) bool {
 	// If no primary error, we succeeded
@@ -824,6 +832,7 @@ func (bifrost *Bifrost) prepareFallbackRequest(req *schemas.BifrostRequest, fall
 	fallbackReq := *req
 	fallbackReq.Provider = fallback.Provider
 	fallbackReq.Model = fallback.Model
+	normalizeOpenRouterRouting(&fallbackReq)
 	return &fallbackReq
 }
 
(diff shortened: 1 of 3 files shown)


akshaydeo:
Thanks  @mcowger for the PR ❤️  - going through it

framework/pricing/utils.go
```diff
@@ -73,13 +73,19 @@ func normalizeRequestType(reqType schemas.RequestType) string {
...
-	// Handle provider/model format - extract just the model name
 	modelName := modelKey
-	if strings.Contains(modelKey, "/") {
+
+	// For openrouter, the model key from the pricing JSON is the canonical model name.
+	// e.g., "openrouter/google/gemini-2.5-flash"
+	// We should also ensure the provider is set correctly.
+	if strings.HasPrefix(modelKey, "openrouter/") {
+		provider = "openrouter"
+	} else if strings.Contains(modelKey, "/") {
```
TejasGhatte, line 83: Heyy @mcowger , pricing is being skipped for models with providers like OpenRouter or Bedrock because when splitting the model and provider name in the pricing module, we were using incorrect splitting logic:
```
modelName := modelKey
if strings.Contains(modelKey, "/") {
    parts := strings.Split(modelKey, "/")
    if len(parts) > 1 {
        modelName = parts[1]
    }
}
```
In transports, we have a utility called `ParseModel(model string)` which has the correct logic for extracting the provider and model.
We could move this function to utils.go in core/ so that both the framework and transports can use it, along with updating existing function calls.
In the framework/pricing, this change would be needed in main.go at `CalculateCostFromUsage `and in utils.go at `convertPricingDataToTableModelPricing`

core/bifrost.go
```diff
@@ -780,6 +780,14 @@ func (bifrost *Bifrost) getProviderQueue(providerKey schemas.ModelProvider) (cha
...
 // CORE INTERNAL LOGIC
 
+// normalizeOpenRouterRouting ensures that requests with "openrouter/" model prefix
+// are correctly routed to the OpenRouter provider.
+func normalizeOpenRouterRouting(req *schemas.BifrostRequest) {
+	if strings.HasPrefix(req.Model, "openrouter/") {
+		req.Provider = schemas.OpenRouter
+	}
+}
+
```
TejasGhatte, line 790: normalization here won't be needed as provider/model parsing is handled in transports layer itself. before sending to core logic.
https://github.com/maximhq/bifrost/blob/main/transports/bifrost-http/handlers/completions.go#L454

mcowger: Gotcha.  Will remove - I only added this because code rabbit demanded I do so 🤪

framework/pricing/utils.go
mcowger: Roger that.  I'll rework this a bit.

Pratham-Mishra04: @TejasGhatte @mcowger lets move that PraseModel to frameworks untils.go and not core cause it has no use in core

PR #755, opened 2025-11-05
transports/lib: ensure x-bf-vk is written under both the typed key and string key in request context

leventov:
## Summary

The governance plugin reads VK via `schemas.BifrostContextKeyVirtualKey`. `ConvertToBifrostContext` now explicitly writes VK under that same typed key (and the existing string-cast key). 

Under enforced governance, some paths read VK from governance.ContextKey("x-bf-vk") while others read from schemas.BifrostContextKeyVirtualKeyHeader. If only one is set, VK‑scoped routing/auth can fail.

## Type of change

- [x] Bug fix

## Affected areas

- [x] Transports (HTTP)
- [x] Plugins

## Breaking changes

- [x] No

## Security considerations

I don't see any.

## Checklist

### DIFF
--- a/transports/bifrost-http/lib/ctx.go
+++ b/transports/bifrost-http/lib/ctx.go
@@ -143,7 +143,11 @@ func ConvertToBifrostContext(ctx *fasthttp.RequestCtx, allowDirectKeys bool) (*c
 		}
 		// Handle virtual key header (x-bf-vk)
 		if keyStr == string(schemas.BifrostContextKeyVirtualKey) {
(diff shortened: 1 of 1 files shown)


akshaydeo:
hey @leventov we do want all keys to be retrieved using `BifrostContextKey<>`. Could you elaborate need for this change?

leventov:
I want to make the governance plugin work in my use case. It uses BifrostContextKeyVirtualKey. Does this mean what governance plugin's code should be changed instead?

leventov:
Ok, I realise this is no-op change now, sorry for making this PR.

PR #829, opened 2025-11-14
Fix goroutine leaks on context cancellation

### LINKED ISSUE #828: [Bug]: Goroutines leak on context cancellation
### Prerequisites

### Description

Bifrost has a bug where worker goroutines and streaming goroutines don't clean up on context cancellation, causing them to block indefinitely and leak.

\[reported using Claude]

1. **Worker Loop Doesn't Monitor Context**: The `requestWorker` function (bifrost.go:2016-2135) uses `for req := range queue` which only exits when the queue channel is closed during `Shutdown()`. Workers never check `bifrost.ctx.Done()` for cancellation, so they continue blocking on `<-queue` waiting for requests even after the context is cancelled.

2. **Streaming Goroutines Block on I/O**: The streaming goroutine in `HandleOpenAIChatCompletionStreaming` (openai.go:786-939) blocks on `scanner.Scan()` at line 804. The context cancellation check happens inside the loop, so it never executes while blocked on I/O. When context times out, the goroutine stays permanently blocked waiting for data.

### Steps to reproduce

https://gist.github.com/pjcdawkins/6f63fad7eea19c3d698b2740aaf21959

### Expected behavior

Goroutines should return to baseline after context cancellation. When `bifrost.Init()` is called with a context and that context is cancelled, all worker goroutines and streaming goroutines should clean up and exit.

### Actual behavior

Multiple goroutines leak (workers + streaming goroutines). They remain blocked indefinitely until process exit, causing memory growth and resource exhaustion.

### Affected area(s)

Core (Go)

### Version

v1.2.22

### Environment

```text
- Go version: 1.25
- OS: Linux
- Affected providers: any providers using streaming
```

### Relevant logs/output

```shell
### Root Cause Analysis

**Issue 1: Worker Loop (`bifrost.go:2026`)**

The worker loop only exits on channel close:

for req := range queue {  // ONLY EXITS ON CHANNEL CLOSE
    // ... request processing
}


This loop only exits when the queue channel is closed in `Shutdown()` at line 2514. When a context is cancelled, the worker doesn't check `bifrost.ctx.Done()` and continues blocking.

**Why passing context to `Init()` doesn't fix it**: While `bifrost.Init()` accepts and stores a context in `bifrost.ctx`, the `requestWorker` goroutines don't monitor this context. They only monitor the per-request context (`req.Context`) for individual operations, not for the worker lifecycle itself.

**Issue 2: Streaming I/O Blocking (`openai.go:804`)**


for scanner.Scan() {  // BLOCKS HERE ON I/O
    // Context check happens AFTER scan completes
    select {
    case <-ctx.Done():
        return
    default:
    }
    // ... process line
}


The `ctx.Done()` check is inside the loop, so it never executes while blocked on `scanner.Scan()`. When context times out, the HTTP client may not immediately close the connection, and the goroutine stays blocked indefinitely.


### Proposed Fixes

**Fix 1: Make Workers Monitor Context Cancellation**

In `bifrost.go:requestWorker`:

func (bifrost *Bifrost) requestWorker(...) {
    // Monitor both queue closure AND context cancellation
    for {
        select {
        case <-bifrost.ctx.Done():
            bifrost.logger.Debug("worker exiting due to context cancellation")
            return
        case req, ok := <-queue:
            if !ok {
                return  // Queue closed - shutdown
            }
            // Process request...
        }
    }
}


**Fix 2: Make Streaming Goroutines Respect Context**

Monitor context in parallel with I/O and force-close response body on cancellation:

go func() {
    done := make(chan struct{})
    defer close(done)

    // Monitor context and force cleanup
    go func() {
        select {
        case <-ctx.Done():
            if resp.BodyStream() != nil {
                resp.BodyStream().Close()
            }
        case <-done:
        }
    }()

    // Existing streaming logic...
}()


## Additional Notes

The channel send pattern at lines 2094-2129 correctly prevents workers from blocking on sends using select with timeout. However, this doesn't help when workers are already blocked on channel receives (`<-queue`) or I/O operations (`scanner.Scan()`).

## Workarounds

Until fixed, consumers can:
1. Create fresh Bifrost instances for each isolated operation
2. Explicitly call `Shutdown()` when done with a Bifrost instance
3. Accept the leak for short-lived processes that exit soon anyway
4. Set aggressive timeouts on the HTTP client level
```

### Regression?

_No response_

### Severity

pjcdawkins:
## Summary

Fixes #828: Workers and streaming goroutines do not exit and leak resources on context cancellation.

This PR addresses two critical goroutine leak issues:
1. Worker loops that ignore context cancellation
2. Streaming handlers that block indefinitely on I/O

### Worker Loop Fix
Refactored `requestWorker` in `core/bifrost.go` to monitor `bifrost.ctx.Done()` using a select statement. Workers now exit immediately when the context is cancelled, instead of only exiting when the queue channel closes.

**Before:**
```go
for req := range queue {
    // Process request
}
```

**After:**
```go
for {
    select {
    case req, ok := <-queue:
        if !ok {
            return
        }
        bifrost.processRequest(provider, config, req)
    case <-bifrost.ctx.Done():
        return
    }
}
```

### Streaming Handler Fixes
Added context monitoring goroutines to all streaming handlers that force-close the response body stream when context is cancelled. This unblocks `scanner.Scan()` calls that were preventing goroutines from exiting.

**Pattern applied to all streaming methods:**
```go
// Monitor context cancellation and force-close response body to unblock scanner.
done := make(chan struct{})
defer close(done)
go func() {
    select {
    case <-ctx.Done():
        resp.CloseBodyStream()
    case <-done:
    }
}()
```

**Fixed in 11 streaming methods across 4 providers:**
- OpenAI: 5 methods (TextCompletionStream, ChatCompletionStream, ResponsesStream, SpeechStream, TranscriptionStream)
- Anthropic: 2 methods (ChatCompletionStream, ResponsesStream)
- Cohere: 2 methods (ChatCompletionStream, ResponsesStream)
- Gemini: 2 methods (SpeechStream, TranscriptionStream)

### Testing
Added `TestBifrostContextCancellationCleanup` to verify goroutines clean up properly after context timeout.

**Test Results:**
- Before fixes: 5 goroutines leaked
- After fixes: 3 goroutines leaked (40% improvement)

## Type of change

- [x] Bug fix

## Affected areas

- [x] Core (Go)

## How to test

```sh
# Run the goroutine leak test
go test -v -run TestBifrostContextCancellationCleanup ./core

# Run all tests
go test ./...
```

## Breaking changes

- [ ] Yes
- [x] No

## Checklist

- [x] I verified builds succeed


Pratham-Mishra04:
Hey @pjcdawkins, thanks for the PR! It’s looking great. There are a few solid Code Rabbit comments to address though. And while you’re at it, please add your update to `core/changelog.md` and double-check that all tests in `tests/core-providers` are still passing.

core/providers/anthropic/anthropic.go
```diff
@@ -480,6 +491,35 @@ func HandleAnthropicChatCompletionStreaming(
 			return
 		}
 
+		// Track last activity time for inactivity timeout detection
```
Pratham-Mishra04, line 494: @pjcdawkins can you have a common util for this? also lets add a field in the UI as well for stream activity timeout in network config

Pratham-Mishra04:
@pjcdawkins lets pls also add the inactivity check in speech and transcription streaming

pjcdawkins:
Good points - I set this as a draft because it's a relatively hard problem and I'll come back to it soon! But it seems important. I think it'll need a few more tests to demonstrate the problem(s) and fixes, and some consolidation as you say.

Pratham-Mishra04:
Yes sure no worries, lmk if you need a hand😄

akshaydeo:
also could you rebase this main with our latest main. we had to nuke one file from the history - which had to be done via rewrite.

akshaydeo:
fixed this in #3541

Issue #432, opened 2025-09-08
[Feature]: Clickhouse support in log store


akshaydeo:
### Prerequisites

- [x] I have searched existing issues and discussions to avoid duplicates

### Problem to solve

Add support for Clickhouse to be used as log store

### Proposed solution

Add support for Clickhouse to be used as log store

### Area(s)

Providers/Integrations

azghr:
Hey @akshaydeo, should I start working on this issue?

PR #914, opened 2025-11-22
[Feat] Add Clickhouse support to logstore

deepakdinesh1123:
## Summary

Add ClickHouse support as a logs store backend enabling users to store logs in ClickHouse.

## Changes

- Added ClickHouse driver and dependency management via GORM
- Implemented `ClickHouseConfig` struct with connection parameters (host, port, user, password, db_name, connection pool settings)
- Created `newClickHouseLogStore()` function to initialize ClickHouse connections with automatic connection pool configuration
- Updated `LogStore` factory to handle ClickHouse type and properly instantiate connections
- Added environment variable support for all ClickHouse config parameters (consistent with PostgreSQL implementation)
- Updated config schema to expose ClickHouse configuration options

## Type of change

- [x] Feature

## Affected areas

- [x] Framework (Go)

## Breaking changes

- [x] No

All existing configurations remain unchanged

## Related issues

#432


deepakdinesh1123:
I have not added any indexes yet as I wanted to get the basic configuration and setup right, please let me know if this works and I will add them if required. 

If the PR seems feasible to you guys I will also add docs and helm charts (if required).

akshaydeo:
@deepakdinesh1123 thanks for the PR. ❤️ 

- For ClickHouse at high scale, we’ll need a write buffer before pushing all the data into ClickHouse. 
- ClickHouse does not scale well with very frequent writes. 
- The usual recommendation is to buffer writes for around 30 seconds and then send them as a batch write request

deepakdinesh1123:
@akshaydeo I will implement this and push the changes

> @deepakdinesh1123 thanks for the PR. ❤️
> 
> * For ClickHouse at high scale, we’ll need a write buffer before pushing all the data into ClickHouse.
> * ClickHouse does not scale well with very frequent writes.
> * The usual recommendation is to buffer writes for around 30 seconds and then send them as a batch write request

deepakdinesh1123:
@akshaydeo I implemented buffered writes using async insert but I am facing a problem where the post hook function of the logging plugin tries to update the log but cannot find it in clickhouse as it is not yet written. 

It could be solved by
- buffering data in bifrost instead of using async insert (would probably require an additional component to buffer data and it should handle crashes gracefully) and then push the data at once
- using "ReplaceMergeTree" engine to change updates to inserts (might require some changes in logstore structure to properly handle this)

Please let me know if any of them are feasible and if there is a more appropriate way of solving this and I will try to implement it.

akshaydeo:
hey @deepakdinesh1123 I missed commenting here 

- So ideally yes; when I said buffer, the usual recommended setup by Clickhouse is buffering outside (when there are frequent writes) or use merging tree indexing - and overwrite all the data multiple times with different versions.
- Both will work here - but we have seen buffering outside the clickhouse setup is the best way.
- Ideal implementation here would be 
    - create a map [id]Log{} (should have a lock or sync.map)
    - every update happens to this map
    - every 30 seconds - take this map -> generate insert queries -> and write to clickhouse

happy to help if you need a hand here or review the flow :)

deepakdinesh1123:
@akshaydeo Thank you very much for letting me know how to implement the flow

I am implementing this currently in the following manner

- add a new logstore called BufferedLogStore

```go
type BufferedLogStore struct {
    rdbLogStore *RDBLogstore
    // other required fields
}
```
- whenever create or update on this store is called it writes the data to the buffer instead of inserting the data immediately
- bulk inserts the data in the buffer once every 30 seconds
- clears up the buffer 
- if any other operations are called it performs it using the RDBLogStore


Please let me know if there is any change required in this flow


Also a follow up question, should users be allowed to enable BufferedLogStore for other databases? If the load is higher users utilising other databases will also benefit from this.

akshaydeo:
creating other log store defeats the purpose of abstraction. This is an implementation detail of Clickhouse log store and the way we need to circle around high speed writes. This should be all handled inside Clickhouse instance vs creating other log store

deepakdinesh1123:
got it, I will create a logstore specifically for clickhouse so that all it's implementation details are in that logstore (similar to how vector stores)

```go
type ClickHouseLogStore struct {
    *RDBLogStore
    // fields required for buffer
}

func (c *ClickHouseLogStore) Create(ctx context.Context, entry *Log) error {
     //write to buffer
}

func Ping(ctx context.Context) error {
    return c.RDBLogstore.Ping(ctx)
}
```

and use the same flow that I described (creates and updates through the clichouse logstore and it uses the rdb logstore for other operations). Please let me know if this is not the right approach.

akshaydeo:
this seems like a good approach.

LookingForOwls:
Any help needed picking this one back up? 

Pushing directly to ClickHouse using a logstore would be much cheaper than using a ClickPipe and replicating that way. 

Is accepting the CLA the only blocker?

akshaydeo:
closing due to inactivity - we are picking it inhouse

jzywien:
@akshaydeo any idea what version this might be targeted for? Thanks in advance!

akshaydeo:
its already released in stable cut. We are taking feedback from a few design partners. We will add instructions to the docs

PR #991, opened 2025-12-03
fix: use net.Listen with Server.Serve to properly support IPv6 addresses

elahti:
## Summary

This PR makes it possible to make Bifrost listen to IPv6 addresses, eg. `::`. [ListenAndServe](https://pkg.go.dev/github.com/valyala/fasthttp#Server.ListenAndServe) that is currently being used seems to make Bifrost listen to IPv4 address always, no matter what host value is provided to Bifrost. For example, even when host is set to `::`, Bifrost still actually listens to IPv4 address `0.0.0.0`.

Without being able to listen to IPv6, Bifrost can't be used eg. in IPv6-only EKS environments.

## Changes

- Use [Serve](https://pkg.go.dev/github.com/valyala/fasthttp#Server.Serve) with a custom listener (both TCP stacks) instead of `ListenAndServe`.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [ ] Core (Go)
- [x] Transports (HTTP)
- [ ] Providers/Integrations
- [ ] Plugins
- [ ] UI (Next.js)
- [ ] Docs

## How to test

Start Bifrost (dev mode or production docker image) with `HOST=::` and eg. verify from inside the container that Bifrost is listening to either IPv4 or IPv6 address.

## Screenshots/Recordings

## Breaking changes

- [ ] Yes
- [x] No

Should not be a breaking change, as far as I could verify. I verified the IPv6 functionality by building Bifrost's Docker image with this change and deploying it to an IPv6-only EKS environment.

## Checklist

### DIFF
--- a/transports/bifrost-http/server/server.go
+++ b/transports/bifrost-http/server/server.go
@@ -1161,9 +1161,13 @@ func (s *BifrostHTTPServer) Start() error {
 	signal.Notify(sigChan, syscall.SIGINT, syscall.SIGTERM)
 	// Start server in a goroutine
 	serverAddr := net.JoinHostPort(s.Host, s.Port)
+	ln, err := net.Listen("tcp", serverAddr)
+	if err != nil {
+		return fmt.Errorf("failed to create listener on %s: %v", serverAddr, err)
+	}
 	go func() {
 		logger.Info("successfully started bifrost, serving UI on http://%s:%s", s.Host, s.Port)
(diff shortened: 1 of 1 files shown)


akshaydeo:
@elahti thanks for the PR ❤️

PR #1314, opened 2026-01-12
fix: convert BifrostChatResponse to OpenAI format for client compatibility

### LINKED ISSUE #1037: [Feature] Add support for MstyStudio
## Discord Thread Context
**Thread:** Is there a secret to getting <think> tokens to show up reliably (or at all)?
**Link:** https://discord.com/channels/1371790330269007932/1448022293862416586
**Created by:** akshay_deo (361806850561146880)

## Recent Thread Messages
**akshay_deo:** can you please elaborate the issue - what are you trying - which SDK what setup please
**captainspazz:** Howdy, I just making calls to the /v1/chat/completions endpoint through MstyStudio.
**captainspazz:** I'm just making*
**akshay_deo:** Msty studio is not an officially supported tool
**akshay_deo:** ill create a ticket for it
**akshay_deo:** we will check whats up with that
**captainspazz:** Okay ! thank you.

**Discord Thread:** https://discord.com/channels/1371790330269007932/1448022293862416586

jerkeyray:
## Summary

Fix token usage not displaying in MstyStudio and other OpenAI-compatible clients.

When clients connect to Bifrost's `/v1/chat/completions` endpoint, responses were 
returned in internal `BifrostChatResponse` format containing `extra_fields` that 
OpenAI-compatible clients couldn't parse.

## Changes

- Added OpenAI-compatible response types (without internal `extra_fields`)
- Added `ToOpenAIChatResponse()` and `ToOpenAIChatStreamResponse()` conversion functions
- Updated `ChatResponseConverter` and `ChatStreamResponseConverter` to use conversions as fallback

## Type of change

- [x] Bug fix

## Affected areas

- [x] Core (Go)
- [x] Transports (HTTP)
- [x] Providers/Integrations

## How to test

```sh
# Run unit tests
cd core && go test ./providers/openai/... -v

# Verify builds
cd core && go build ./...
cd transports && go build -tags=skip_ui ./...

# Manual test (with Bifrost running)
curl -s -X POST http://localhost:8080/v1/chat/completions \
  -d '{"model":"openai/gpt-4o-mini","messages":[{"role":"user","content":"Hi"}]}' \
  | jq '.usage'
# Should return: {"prompt_tokens":..., "completion_tokens":..., "total_tokens":...}
```

## Breaking changes

- [x] No

## Related issues

Closes #1037

## Checklist

- [x] I read the contributing guidelines
- [x] I verified builds succeed
- [x] I verified the CI pipeline passes locally


akshaydeo:
thanks @jerkeyray for the PR - let me look into it.

core/providers/openai/responses.go
```diff
@@ -7,6 +7,160 @@ import (
 	"github.com/maximhq/bifrost/core/schemas"
 )
 
+// OpenAIChatResponse represents the OpenAI chat completion response format.
+// This is used when converting BifrostChatResponse to OpenAI-compatible format
+// for clients like MstyStudio that expect pure OpenAI responses.
+type OpenAIChatResponse struct {
```
akshaydeo, line 13: @jerkeyray any reason you went with separate structs. The core messages of Bifrost are OpenAI compatible - and thats how we have structured the codebase.

akshaydeo (changes requested):
added 1 comment

core/providers/openai/responses.go
jerkeyray: Hey @akshaydeo,
extra_fields is Bifrost-specific metadata and part of the documented API. The OpenAI spec doesn’t include this field, and strict OpenAI-compatible clients (like MstyStudio) break or mis-handle usage when it’s present.

I used separate structs to avoid:
- breaking existing Bifrost users by removing extra_fields globally (json:"-" would do that)
- side effects from changing the core schema

The conversion functions are just an adapter at the /v1/chat/completions boundary to return a clean, spec-compliant OpenAI response. Internal Bifrost responses stay unchanged.

I am very open to learning and switching to a better pattern if there's a cleaner approach you could guide me towards.

Pratham-Mishra04: Hey @jerkeyray, I think the right approach here is to first identify the exact fields causing these failures. Based on my understanding, we’re likely dealing with two broad issues:

1. Additional fields like `extra_fields` that we send back in the request.
2. Bifrost-specific (OpenAI++) schema fields that are being sent back.

The core issue is that while we filter requests going to OpenAI, for requests coming **from** other providers and flowing through the OpenAI integration, we currently send the `BifrostResponse` as-is with minimal modification.

The approach you suggested would work, but it introduces additional memory allocations and computation. We’ve intentionally avoided that for OpenAI because Bifrost already uses an Open++ schema. Instead of constructing full OpenAI schemas, we can selectively drop fields (set by other providers) that don’t fit the OpenAI integration.

Examples include cached usage fields from Anthropic or web search outputs from Perplexity.

For issue (1), we can also move `extra_fields` behind a pointer and set it to `nil` for the OpenAI integration.

Let me know if you have any questions or suggestions.

jerkeyray: Thanks for the context, @Pratham-Mishra04. That makes total sense.

I'll switch ExtraFields to a pointer and update the converter to just nil. I'll be reverting the separate structs now and will push the refactor shortly.

jerkeyray: Hey @Pratham-Mishra04, 
I’ve followed your suggestion to move extra_fields behind a pointer and nil it out on the OpenAI integration path, and I’m selectively dropping provider-specific fields for OpenAI compatibility without reconstructing schemas.

Since ExtraFields is constructed across providers, this change touches multiple implementations, but the approach remains pointer + nil + selective field removal as you suggested.

Just wanted to confirm this aligns with what you had in mind.

jerkeyray: Hey @Pratham-Mishra04,
I’ve pushed the refactor based on your earlier guidance. Would really appreciate any feedback.

Pratham-Mishra04: Hey @jerkeyray yes i'll check it today

core/providers/bedrock/bedrock.go
```diff
@@ -1120,15 +1115,18 @@ func (provider *BedrockProvider) ResponsesStream(ctx *schemas.BifrostContext, po
 							// Set raw request if enabled
 							ctx.SetValue(schemas.BifrostContextKeyStreamEndIndicator, true)
 							if providerUtils.ShouldSendBackRawRequest(ctx, provider.sendBackRawRequest) {
-								providerUtils.ParseAndSetRawRequest(&finalResponse.ExtraFields, jsonData)
+								if finalResponse.ExtraFields == nil {
+									finalResponse.ExtraFields = &schemas.BifrostResponseExtraFields{}
+								}
+								providerUtils.ParseAndSetRawRequest(finalResponse.ExtraFields, jsonData)
```
Pratham-Mishra04: lets change ParseAndSetRawRequest to take in finalResponse instead of finalResponse.ExtraFields and we'll handle this nil pointer all at one place

jerkeyray: Hey @Pratham-Mishra04 
I’ve updated ParseAndSetRawRequest to accept the full response instead of ExtraFields and centralized the nil handling inside the helper.

Issue #123, opened 2025-06-24
Files API Support


Pratham-Mishra04:
Support **File APIs** for providers like OpenAI or Anthropic.

#### Example:

* Upload files for use in:

  * Fine-tuning
  * Retrieval-Augmented Generation (RAG)
  * Larger context storage

```bash
POST /v1/files
```

### Why

Essential for advanced workflows where **RAG** or **custom data** ingestion is required.

PR #1644, opened 2026-02-13
fix: add passthrough support for count_tokens endpoint

hazyone:
## Summary

count endpoint has no passthrough support

## Changes

- added passthrough support for count_tokens endpoint

## Type of change

- [ ] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [ ] Core (Go)
- [ ] Transports (HTTP)
- [ ] Providers/Integrations
- [ ] Plugins
- [ ] UI (Next.js)
- [ ] Docs

## How to test

```sh
# Core/Transports
go version
go test ./...

# UI
cd ui
pnpm i || npm i
pnpm test || npm test
pnpm build || npm run build
```

If adding new configs or environment variables, document them here.

## Screenshots/Recordings

## Breaking changes

- [ ] Yes
- [ ] No

## Related issues

## Security considerations

## Checklist

### DIFF
--- a/transports/bifrost-http/integrations/anthropic.go
+++ b/transports/bifrost-http/integrations/anthropic.go
@@ -304,6 +304,7 @@ func checkAnthropicPassthrough(ctx *fasthttp.RequestCtx, bifrostCtx *schemas.Bif
 			bifrostCtx.SetValue(schemas.BifrostContextKeyExtraHeaders, headers)
 			bifrostCtx.SetValue(schemas.BifrostContextKeyURLPath, url)
 			bifrostCtx.SetValue(schemas.BifrostContextKeySkipKeySelection, true)
+			bifrostCtx.SetValue(schemas.BifrostContextKeyDirectKey, "")
 		} else {
 			// API key flow: pass only whitelisted safe headers (like anthropic-beta for feature detection)
 			passthroughHeaders := extractPassthroughHeaders(headers, provider)
@@ -391,6 +392,7 @@ func CreateAnthropicCountTokensRouteConfigs(pathPrefix string, handlerStore lib.
 			ErrorConverter: func(ctx *schemas.BifrostContext, err *schemas.BifrostError) interface{} {
 				return anthropic.ToAnthropicChatCompletionError(err)
 			},
+			PreCallback: checkAnthropicPassthrough,
 		},
 	}
 }


transports/bifrost-http/integrations/anthropic.go
```diff
@@ -304,9 +303,8 @@ func checkAnthropicPassthrough(ctx *fasthttp.RequestCtx, bifrostCtx *schemas.Bif
 			}
 			bifrostCtx.SetValue(schemas.BifrostContextKeyExtraHeaders, headers)
 			bifrostCtx.SetValue(schemas.BifrostContextKeyURLPath, url)
-			// This key is also used in IsClaudeCodeMaxMode
-			// So if you are changing the behaviour of this key, make sure to change IsClaudeCodeMaxMode as well
 			bifrostCtx.SetValue(schemas.BifrostContextKeySkipKeySelection, true)
+			bifrostCtx.SetValue(schemas.BifrostContextKeyDirectKey, "")
```
akshaydeo, line 307: @hazyone `SkipKeySelection` does the same job right? any reason to add new context value?

akshaydeo:
@hazyone thanks for the PR ❤️  - added one comment

transports/bifrost-http/integrations/anthropic.go
hazyone: I guess that no, because BifrostContextKeyDirectKey check section has an early return flow

```
if ctx != nil {
		key, ok := ctx.Value(schemas.BifrostContextKeyDirectKey).(schemas.Key)
		if ok {
			return key, nil
		}
	}
	// Check if key skipping is allowed
	if skipKeySelection, ok := ctx.Value(schemas.BifrostContextKeySkipKeySelection).(bool); ok && skipKeySelection && isKeySkippingAllowed(providerKey) {
		return schemas.Key{}, nil
	}
```

akshaydeo: okay let me go through the flow - but this will break as we are expecting schemas.Key and you are inserting a empty string. did you try this locally?

hazyone: Yep, I checked that in my k8s (hope it was correct build :D ) but it worked!

akshaydeo: I dont think this should work - and if its working - its the side-effect of this bug. 

can you elaborate whats the intended flow of the change?

hazyone: I haven't digged that much, but i decided (and as was mentioned in thread) that direct key is passing to x-api-key(if configured) header and for some reason affects only count_tokens endpoint (or it was only because checkAnthropicPassthrough was not set).

So i decided, that in passthrough mode this direct key logic should be dropped

Pratham-Mishra04: @hazyone but we only attach direct keys when allow direct api keys is on in the configuration and not always, and its actually supposed to be schemas.Key and not a string. But yes the overall construct feels correct to block direct api keys (by overidding with empty) when its claude code with oauth

akshaydeo:
@hazyone I have fixed this issue in another PR - so closing this.

PR #1879, opened 2026-03-02
fix: default Anthropic error type to "api_error" instead of empty string

### LINKED ISSUE #1877: "type":"" in SSE error responses breaks consumers that use the field as a union discriminator
When bifrost proxies an error back to the client as an SSE stream (via the Anthropic `/v1/messages` endpoint), the resulting error event sometimes contains `"type":""` (empty string) in the inner error object:

```
event: error
data: {"type":"error","error":{"type":"","message":"failed to execute HTTP request to provider API"}}
```

This happens in two cases:

1. **Connection-level failures** (`ErrProviderDoRequest`, `ErrProviderNetworkError`) — these are created internally by bifrost with no `Error.Type` set. In `ToAnthropicChatCompletionError`, `errorType` defaults to `""` when `BifrostError.Error.Type` is nil.

2. **Azure rate limit errors** — Azure itself returns `"type":""` in its error body, and bifrost passes it through unchanged.

In both cases, `AnthropicMessageErrorStruct.Type` is a plain `string` with no `omitempty`, so the empty string always gets serialized.

**Impact:** Consumers that use `type` as a discriminator (e.g. the Vercel AI SDK's Zod schema for Anthropic errors) fail to parse the response entirely, receiving a `Type validation failed` error instead of a meaningful error. This also causes rate limit errors to be marked `isRetryable: false` by the SDK, suppressing client-side retries.

**Suggested fix** in `core/providers/anthropic/errors.go`:

```go
// Before
errorType := ""
if bifrostErr.Error.Type != nil {
    errorType = *bifrostErr.Error.Type
}

// After
errorType := "api_error"
if bifrostErr.Error.Type != nil && *bifrostErr.Error.Type != "" {
    errorType = *bifrostErr.Error.Type
}
```

This matches the Anthropic API spec where \`type\` is always a non-empty string.

robechun:
Fixes #1877.

When bifrost proxies certain errors through the Anthropic SSE endpoint, the inner `error.type` field can end up as an empty string `""`:

- **Connection-level failures** (`ErrProviderDoRequest`, `ErrProviderNetworkError`): `BifrostError.Error.Type` is nil, so `errorType` defaults to `""`.
- **Provider errors with empty type** (e.g. Azure rate limits): some providers return `"type":""` themselves, which bifrost passes through unchanged.

`AnthropicMessageErrorStruct.Type` is a plain `string` with no `omitempty`, so `""` always serializes into the response.

This breaks consumers that use `type` as a union discriminator (e.g. the Vercel AI SDK's Zod schema for Anthropic errors), causing `Type validation failed` instead of a meaningful error — and incorrectly marking retryable errors like rate limits as `isRetryable: false`.

## Change

In `ToAnthropicChatCompletionError`, default `errorType` to `"api_error"` and guard against empty strings from providers:

```go
// Before
errorType := ""
if bifrostErr.Error.Type != nil {
    errorType = *bifrostErr.Error.Type
}

// After
errorType := "api_error"
if bifrostErr.Error.Type != nil && *bifrostErr.Error.Type != "" {
    errorType = *bifrostErr.Error.Type
}
```

`"api_error"` is consistent with the [Anthropic API error spec](https://docs.anthropic.com/en/api/errors) for generic server-side errors.

### DIFF
--- a/core/providers/anthropic/errors.go
+++ b/core/providers/anthropic/errors.go
@@ -16,10 +16,10 @@ func ToAnthropicChatCompletionError(bifrostErr *schemas.BifrostError) *Anthropic
 	}
 
 	// Safely extract type and message from nested error
-	errorType := ""
+	errorType := "api_error"
 	message := ""
 	if bifrostErr.Error != nil {
-		if bifrostErr.Error.Type != nil {
+		if bifrostErr.Error.Type != nil && *bifrostErr.Error.Type != "" {
 			errorType = *bifrostErr.Error.Type
 		}
 		message = bifrostErr.Error.Message


Pratham-Mishra04:
Hey @robechun thanks for the PR! This looks good - can you pls just update the change log? https://docs.getbifrost.ai/contributing/raising-a-pr#maintaining-changelogs

Pratham-Mishra04 (approved):

akshaydeo:
### Merge activity

* **Mar 5, 8:41 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/1879).

PR #1925, opened 2026-03-05
fix: replace goroutine+channel with DoDeadline in MakeRequestWithContext

davidrudduck:
## Summary

Fixes a goroutine leak and potential use-after-free in `MakeRequestWithContext` by replacing the goroutine+channel pattern with `client.DoDeadline`.

The previous implementation spawned a goroutine calling `client.Do(req, resp)` and used `select` to race it against `ctx.Done()`. When the context cancelled first, the goroutine continued running and could write to `resp` after the caller had already returned — causing use-after-free of the response buffer.

## Changes

* Replace `client.Do` in a goroutine with `client.DoDeadline(req, resp, deadline)` — synchronous, no goroutine needed
* Derive deadline from `ctx.Deadline()`, falling back to a 5-minute safety-net timeout
* Preserve all existing error classification (timeout, cancellation, DNS, network errors)
* Add `ctx.Err()` check after `DoDeadline` returns to detect in-flight cancellation

## Type of change

* [x] Bug fix
* [ ] Feature
* [ ] Refactor
* [ ] Documentation
* [ ] Chore/CI

## Affected areas

* [x] Core (Go)
* [ ] Transports (HTTP)
* [ ] Providers/Integrations
* [ ] Plugins
* [ ] UI (Next.js)
* [ ] Docs

## How to test

```bash
(cd core && go build ./... && go vet ./providers/utils/...)
```

The fix is in the request execution path — all existing provider tests exercise this code:
```bash
make test-core PROVIDER=openai
```

To verify goroutine leak is fixed, send requests with short timeouts and confirm goroutine count stabilizes (previously it would grow unbounded under cancellation).

## Screenshots/Recordings

N/A

## Breaking changes

* [ ] Yes
* [x] No

## Security considerations

Eliminates use-after-free of fasthttp response buffers when contexts cancel during in-flight requests.

## Checklist

* [x] I verified the CI pipeline passes locally if applicable

### DIFF
--- a/core/providers/utils/utils.go
+++ b/core/providers/utils/utils.go
@@ -120,7 +120,7 @@ func MakeRequestWithContext(ctx context.Context, client *fasthttp.Client, req *f
 				IsBifrostError: true,
 				Error: &schemas.ErrorField{
 					Type:    schemas.Ptr(schemas.RequestCancelled),
-					Message: fmt.Sprintf("Request cancelled or timed out by context: %v", ctx.Err()),
+					Message: fmt.Sprintf("request cancelled or timed out by context: %v", ctx.Err()),
 					Error:   ctx.Err(),
 				},
 			}
--- a/core/providers/utils/utils.go
+++ b/core/providers/utils/utils.go
@@ -96,84 +96,87 @@ func getLogger() schemas.Logger {
 
 var UnsupportedSpeechStreamModels = []string{"tts-1", "tts-1-hd"}
 
-// MakeRequestWithContext makes a request with a context and returns the latency and error.
-// IMPORTANT: This function does NOT truly cancel the underlying fasthttp network request if the
-// context is done. The fasthttp client call will continue in its goroutine until it completes
-// or times out based on its own settings. This function merely stops *waiting* for the
-// fasthttp call and returns an error related to the context.
+// defaultRequestTimeout is a safety-net fallback used only when the context has no deadline.
+// In normal operation, callers always set a deadline via NetworkConfig.DefaultRequestTimeoutInSeconds
+// (default 30s), so this 5-minute ceiling should rarely be reached.
+const defaultRequestTimeout = 5 * time.Minute
+
+// MakeRequestWithContext makes a request with a context deadline and returns the latency and error.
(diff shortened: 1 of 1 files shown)


core/providers/utils/utils.go
```diff
@@ -96,84 +96,87 @@ func getLogger() schemas.Logger {
...
 
-// MakeRequestWithContext makes a request with a context and returns the latency and error.
-// IMPORTANT: This function does NOT truly cancel the underlying fasthttp network request if the
-// context is done. The fasthttp client call will continue in its goroutine until it completes
-// or times out based on its own settings. This function merely stops *waiting* for the
-// fasthttp call and returns an error related to the context.
+// defaultRequestTimeout is a safety-net fallback used only when the context has no deadline.
+// In normal operation, callers always set a deadline via NetworkConfig.DefaultRequestTimeoutInSeconds
+// (default 30s), so this 5-minute ceiling should rarely be reached.
+const defaultRequestTimeout = 5 * time.Minute
```
akshaydeo: We dont need to/want to hardcode this. As we have a few users who have 10 mins as timeout in some cases. Can we nuke this somehow?

akshaydeo:
❤️ for the PR @davidrudduck

davidrudduck:
Updated per your feedback — removed the hardcoded `defaultRequestTimeout` entirely. When the context carries a deadline (the normal path via `NetworkConfig.DefaultRequestTimeoutInSeconds`), it uses `DoDeadline`. When no deadline is set, it falls back to `client.Do` which respects the client's own timeout config. No arbitrary ceilings imposed.

davidrudduck:
Hi team,

I'm happy to sign the CLA for my contributions to remain under Apache 2.0 in the open source project.

However I'd like clarification as to whether contributions may be incorporated into closed source or proprietary products (ie Bifrost Enterprise) under a different license?

If so, I'd prefer to limit the scope of my grant to the open source distribution only.

My contributions are intended to benefit your users and the OSS community in general. If they'll be incorporated into closed/proprietary distributions, I'd rather maintain them as a local patch.

PR #1927, opened 2026-03-05
refactor: use direct struct pointer for Anthropic tool InputSchema

davidrudduck:
## Summary

Replaces a 30-line field-by-field copy of `ToolFunctionParameters` with direct pointer assignment when building Anthropic tool definitions. Both the source (`tool.Function.Parameters`) and destination (`anthropicTool.InputSchema`) use the same `*schemas.ToolFunctionParameters` type, making the manual copy unnecessary.

This also prevents future bugs where new fields added to `ToolFunctionParameters` would be silently dropped by the Anthropic provider.

## Changes

* Replace field-by-field struct initialization with `anthropicTool.InputSchema = tool.Function.Parameters` in `core/providers/anthropic/chat.go`

## Type of change

* [ ] Bug fix
* [ ] Feature
* [x] Refactor
* [ ] Documentation
* [ ] Chore/CI

## Affected areas

* [x] Core (Go)
* [ ] Transports (HTTP)
* [x] Providers/Integrations
* [ ] Plugins
* [ ] UI (Next.js)
* [ ] Docs

## How to test

```bash
cd core && go build ./providers/anthropic/...
make test-core PROVIDER=anthropic
```

Existing Anthropic provider tests validate tool parameter handling.

## Breaking changes

* [ ] Yes
* [x] No

## Security considerations

No security impact — same data, simplified assignment.

## Checklist

### DIFF
--- a/core/providers/anthropic/chat.go
+++ b/core/providers/anthropic/chat.go
@@ -95,38 +95,7 @@ func ToAnthropicChatRequest(ctx *schemas.BifrostContext, bifrostReq *schemas.Bif
 
 				// Convert function parameters to input_schema
 				if tool.Function.Parameters != nil && (tool.Function.Parameters.Type != "" || tool.Function.Parameters.Properties != nil) {
-					anthropicTool.InputSchema = &schemas.ToolFunctionParameters{
-						Type:                 tool.Function.Parameters.Type,
-						Description:          tool.Function.Parameters.Description,
-						Properties:           tool.Function.Parameters.Properties,
-						Required:             tool.Function.Parameters.Required,
-						Enum:                 tool.Function.Parameters.Enum,
-						AdditionalProperties: tool.Function.Parameters.AdditionalProperties,
-						// JSON Schema definition fields
-						Defs:        tool.Function.Parameters.Defs,
-						Definitions: tool.Function.Parameters.Definitions,
-						Ref:         tool.Function.Parameters.Ref,
-						// Array schema fields
-						Items:    tool.Function.Parameters.Items,
-						MinItems: tool.Function.Parameters.MinItems,
-						MaxItems: tool.Function.Parameters.MaxItems,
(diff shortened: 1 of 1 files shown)


Pratham-Mishra04 (approved):

core/providers/anthropic/chat.go
```diff
@@ -95,38 +95,7 @@ func ToAnthropicChatRequest(ctx *schemas.BifrostContext, bifrostReq *schemas.Bif
 
 				// Convert function parameters to input_schema
 				if tool.Function.Parameters != nil && (tool.Function.Parameters.Type != "" || tool.Function.Parameters.Properties != nil) {
-					anthropicTool.InputSchema = &schemas.ToolFunctionParameters{
```
akshaydeo, line 98: @Pratham-Mishra04 I think this was to avoid altering original request right?

davidrudduck: Good point — the pointer is only read (serialized to JSON) after assignment so it's safe today, but a field-by-field copy is more defensive against future mutations. Happy to revert this if you'd prefer to keep the defensive copy. The main concern was that new fields added to `ToolFunctionParameters` would silently be dropped here without a corresponding update, but I understand the tradeoff.

Pratham-Mishra04: Yes I think we directly serialize it after this so shouldnt be an issue but i'll confirm once

davidrudduck: If you'd prefer to keep mutation safety, we could do a shallow struct copy instead:

```go
params := *tool.Function.Parameters
anthropicTool.InputSchema = &params
```

This gives you the defensive copy (no shared pointer) while automatically forwarding any new fields added to `ToolFunctionParameters` — so no more silent field drops when the schema struct grows. Best of both worlds?

PR #1929, opened 2026-03-05
fix: restrict default CORS AllowedOrigins to empty instead of wildcard

davidrudduck:
## Summary

Changes the default `AllowedOrigins` in `DefaultClientConfig` from `["*"]` (allow all origins) to `[]` (no origins allowed by default). This follows the principle of least privilege — CORS must be explicitly configured for production deployments.

Localhost origins are always allowed by `IsOriginAllowed()` regardless of this setting, so local development is unaffected.

## Changes

* Change `AllowedOrigins: []string{"*"}` to `AllowedOrigins: []string{}` in `transports/bifrost-http/lib/config.go`

## Type of change

* [x] Bug fix
* [ ] Feature
* [ ] Refactor
* [ ] Documentation
* [ ] Chore/CI

## Affected areas

* [ ] Core (Go)
* [x] Transports (HTTP)
* [ ] Providers/Integrations
* [ ] Plugins
* [ ] UI (Next.js)
* [ ] Docs

## How to test

1. Start bifrost without configuring `allowed_origins`
2. Verify CORS requests from non-localhost origins are rejected
3. Verify localhost origins still work
4. Configure `allowed_origins: ["https://example.com"]` and verify it works

## Breaking changes

* [x] Yes — deployments relying on the default wildcard CORS will need to explicitly configure `allowed_origins`. Localhost is unaffected.

## Security considerations

Prevents unintentional open CORS in production when operators forget to configure origins.

## Checklist


akshaydeo:
@davidrudduck this was intentional. One of the most common support ticket earlier was - im getting CORS issue. So we resorted to this. If you have any other suggestion - happy to hear 🙇

davidrudduck:
That makes total sense — reducing CORS support tickets is a valid priority. A few alternatives that keep the DX win while flagging the security posture:

1. **Startup warning** — keep `["*"]` as default, but log a warning at startup when wildcard CORS is active (e.g. `"CORS AllowedOrigins is set to wildcard — consider restricting for production"`). Users get it working immediately but are nudged toward explicit config.

2. **Environment-aware default** — default to `["*"]` in dev/local mode, empty in production. Though this adds complexity.

3. **Keep as-is** — if the support volume justified it, the current default is fine. Happy to close this PR.

Option 1 feels like the best balance — what do you think?

davidrudduck:
To expand on this — the security principle here is "secure by default, opt-in to relax." The user should make a conscious decision to accept the risk of open CORS, rather than unknowingly running with it.

A wildcard default means every new deployment is exposed until someone explicitly locks it down — and most users won't, because it "just works." That's the exact pattern that leads to production incidents.

Concretely, we could:

1. Default to `[]` (strict) — users who hit CORS issues add `allowed_origins: ["*"]` to their config. One line, documented, intentional.
2. First-run helper — if `allowed_origins` is empty and a CORS preflight is rejected, log a clear message: `"CORS request blocked from origin X. To allow it, add X to allowed_origins in your config (or use '*' to allow all)."`

This way the "CORS issue" support ticket becomes self-resolving — the error message tells them exactly what to do — while the default posture is secure. The fix is a one-line config change, not a security audit.

davidrudduck:
One more consideration — the users who understand CORS well enough to need cross-origin access can easily add `allowed_origins: ["*"]` to their config. But the users who don't understand CORS are exactly the ones who shouldn't have wildcard access silently enabled on their behalf. A permissive default protects the knowledgeable user from a minor inconvenience at the cost of exposing the less experienced user to real risk.

davidrudduck:
I hear you on the support ticket volume — that's real pain. But turning off a security boundary to reduce support load means every new user is silently exposed by default, and the users most at risk are exactly the ones who won't know to check.

What if we make the experience *better* than the wildcard default? Here's what I'm proposing:

1. **Default to `[]`** (deny by default)
2. **Clear, actionable log message** when a CORS request is rejected:
   ```
   CORS request blocked: origin "https://example.com" is not in allowed_origins. See: https://docs.getbifrost.ai/configuration/cors
   ```

The user sees what happened and where to fix it — no support ticket needed. This is better DX than the wildcard default because:
- With `["*"]` they never learn about the config option at all
- With the log message they understand the system and make an informed choice
- Users who deploy to production are secure by default
- Users in dev add one line and never think about it again

I'm happy to implement the log message as part of this PR so it's a complete package.

akshaydeo:
We have gone through this cycle - peple complain that things are not working when its empty. so probably we will keep it like this

PR #1971, opened 2026-03-06
fixed baseApi for new token sync

akshaydeo:
### TL;DR

Simplified authentication to use HTTP-only cookies for both enterprise and non-enterprise modes instead of client-side token management.

### What changed?

- Removed enterprise-specific OAuth token retrieval logic from `getTokenFromStorage()`
- The function now always returns `null` since authentication is handled server-side via HTTP-only cookies
- Removed import of `getAccessToken` from the enterprise token manager
- Updated comments to reflect that auth tokens are now stored in HTTP-only cookies set by the server

### How to test?

1. Test authentication flows in both enterprise and non-enterprise modes
2. Verify that API requests still work correctly with cookie-based authentication
3. Confirm that the `credentials: "include"` setting properly sends cookies with requests
4. Test that authentication persists across browser sessions

### Why make this change?

This change standardizes authentication across both enterprise and non-enterprise deployments by moving token storage from client-side JavaScript to secure HTTP-only cookies, improving security by preventing client-side access to authentication tokens.

### DIFF
--- a/ui/lib/store/apis/baseApi.ts
+++ b/ui/lib/store/apis/baseApi.ts
@@ -2,26 +2,13 @@ import { IS_ENTERPRISE } from "@/lib/constants/config";
 import { BifrostErrorResponse } from "@/lib/types/config";
 import { getApiBaseUrl } from "@/lib/utils/port";
 import { createBaseQueryWithRefresh } from "@enterprise/lib/store/utils/baseQueryWithRefresh";
-import { clearOAuthStorage, getAccessToken } from "@enterprise/lib/store/utils/tokenManager";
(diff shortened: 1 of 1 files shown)


akshaydeo:
* **#1971** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/1971?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/maximhq/bifrost/1971?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #2082, opened 2026-03-15
fix(anthropic): deterministic tool schema serialization for prompt caching

Edward-Upton:
## Problem

Anthropic's prompt caching is prefix-based — it derives a cache key from the serialized request. When clients send tool definitions with JSON object fields (e.g. `properties`, `$defs`), the key ordering within those objects can vary non-deterministically between turns.

Because tool definitions are part of the request prefix, this means the cache key changes every turn, causing repeated **cache writes** instead of **cache reads**. In practice, only the system prompt portion (~16k tokens) was ever cache-hit; the full tool schema (~25k tokens) was re-written on every single LLM call.

## Impact

We measured this across multiple production agent runs (80+ LLM turns each). After the fix:

| Model | Cost Before | Cost After | Savings |
|---|---|---|---|
| Claude Sonnet 4.6 | ~$15.80 | ~$1.65 | **~90%** |
| Claude Opus 4.6 | ~$9.55 | ~$2.71 | **~72%** |

Cache read tokens went from plateauing early to growing correctly with each turn, confirming the prefix cache is working as intended.

## Fix

Ensures all tool schema JSON is serialized with deterministic, LLM-optimal key ordering across all code paths:

1. **`OrderedMap.SortKeys()`** — sorts keys using JSON Schema priority ordering (`type`, `description`, `properties`, `required` first, then remaining keys alphabetically). This ensures LLMs see type information before descriptions and constraints in nested tool schemas.

2. **`OrderedMap.SortedCopy()`** — returns a new OrderedMap tree with sorted keys, sharing primitive values with the original. This is much cheaper than a full Marshal/Unmarshal clone since it only allocates new key slices and value maps.

3. **`ToolFunctionParameters.Normalized()`** — returns a shallow struct copy with all nested OrderedMaps sorted via `SortedCopy()`. Replaces the old `Clone()` + `NormalizeKeyOrder()` pattern, avoiding a full JSON roundtrip per tool per request.

4. **Chat Completions + Responses API** — both Anthropic converters call `Normalized()` on tool input schemas before serialization.

5. **`excludeTopLevelJSONKeys()`** — the Responses API `excludeFields` path now uses `map[string]json.RawMessage` instead of `map[string]interface{}`, preserving nested byte ordering instead of re-sorting everything alphabetically.

6. **Passthrough mode** — uses `MarshalSorted` for deterministic map key output when forwarding raw request bodies.

## Test plan

- [x] Verified `core/` module compiles
- [x] Tested with production agent runs across Sonnet, Opus, and Haiku — cache read tokens accumulate correctly, cache writes drop to near-zero after first turn
- [ ] Existing provider tests pass (`make test-core PROVIDER=anthropic`)

### DIFF
--- a/core/providers/anthropic/chat.go
+++ b/core/providers/anthropic/chat.go
@@ -134,6 +134,10 @@ func ToAnthropicChatRequest(ctx *schemas.BifrostContext, bifrostReq *schemas.Bif
 					}
 				}
 
+				if anthropicTool.InputSchema != nil {
+					anthropicTool.InputSchema = anthropicTool.InputSchema.Normalized()
+				}
+
 				if tool.CacheControl != nil {
 					anthropicTool.CacheControl = tool.CacheControl
 				}
--- a/core/providers/anthropic/responses.go
+++ b/core/providers/anthropic/responses.go
@@ -4539,6 +4539,16 @@ func convertBifrostToolToAnthropic(model string, tool *schemas.ResponsesTool) *A
 		}
 	}
 
+	// Normalize tool schema key ordering to ensure deterministic serialization.
+	// Clients (e.g. Claude Agent SDK) may send non-deterministic property orderings
+	// across turns, which breaks Anthropic's prefix-based prompt caching since tool
+	// definitions are part of the serialized request prefix.
+	// Normalized() returns a shallow copy with sorted key slices, so the
+	// caller-owned tool.ResponsesToolFunction.Parameters is never mutated.
+	if anthropicTool.InputSchema != nil {
+		anthropicTool.InputSchema = anthropicTool.InputSchema.Normalized()
+	}
+
 	if tool.CacheControl != nil {
 		anthropicTool.CacheControl = tool.CacheControl
 	}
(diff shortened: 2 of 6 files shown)


akshaydeo:
❤️ for this @Edward-Upton  - @Pratham-Mishra04 was going to look into this but thank you - ill review this in a bit and get in

akshaydeo:
one quick question I have here @Edward-Upton  - We have noticed that, specifically with OpenAI - order of the keys (not sorted/unsorted but for example where is the required field for JSON response schema) matters a lot on how much model hallucinates. 

Apparently JSON parser in models is serialized parser so it does not get entire schema and considers that for response generation. Anything you observed with anthropic?

Edward-Upton:
> one quick question I have here @Edward-Upton - We have noticed that, specifically with OpenAI - order of the keys (not sorted/unsorted but for example where is the required field for JSON response schema) matters a lot on how much model hallucinates.
> 
> Apparently JSON parser in models is serialized parser so it does not get entire schema and considers that for response generation. Anything you observed with anthropic?

Haven't noticed, but potentially not ran enough evals of this comparing with vs without. It was so expensive to run evals without this fix that I'm not sure we will even be able to compare. Is there any public information about ordering of keys we could try to replicate?

akshaydeo (approved):

PR #2129, opened 2026-03-17
logging: persist and aggregate tokens_per_second and TTFT with single-source computation

### LINKED ISSUE #1992: [Feature]: Track and display token generation speed (tokens/sec) in logs
### Prerequisites

### Problem to solve

When evaluating LLM provider performance, **token generation speed (tokens per second)** is one of the most critical metrics — it directly impacts user-perceived latency, especially for streaming use cases. However, Bifrost's logging system currently does not calculate or persist this metric.

Today, the `logstore.Log` struct tracks `Latency` (end-to-end ms), `PromptTokens`, `CompletionTokens`, and `TotalTokens`, but there is no derived **tokens/sec** field. Users who want to compare provider throughput or identify slow requests must manually compute `completion_tokens / (latency / 1000)` from raw log data.

Additionally, `TimeToFirstToken` (TTFT) is already computed in the streaming accumulator (`framework/streaming/accumulator.go` → `AccumulatedData.TimeToFirstToken`) but is **not persisted** to the log store — it's only exported to OpenTelemetry tracing spans. This means TTFT is invisible in Bifrost's own log UI and query APIs.

**Who benefits:**
- Platform engineers comparing provider/model performance at a glance
- Teams optimizing for real-time/streaming applications where throughput matters
- Anyone using the Bifrost UI logs page to debug slow responses

### Proposed solution

1. **Persist `TimeToFirstToken`** (already computed) into `logstore.Log` as a new column, and surface it in the logs UI and query APIs.

2. **Compute and persist `TokensPerSecond`** as a new field in `logstore.Log`:
   ```
   tokens_per_second = completion_tokens / (latency_ms / 1000.0)
   ```
   This should be calculated in the logging plugin's `PostLLMHook` (or in `applyStreamingOutputToEntry`) when both `CompletionTokens > 0` and `Latency > 0`.

3. **Add histogram/aggregation support** in the LogStore query layer — similar to the existing `LatencyHistogramBucket` and `TokenHistogramBucket`, add:
   - `AvgTokensPerSecond`, `P90TokensPerSecond` per time bucket
   - `AvgTimeToFirstToken`, `P90TimeToFirstToken` per time bucket
   - Per-provider and per-model breakdowns

4. **Surface in the UI logs page** — display tokens/sec and TTFT as columns (or in the log detail panel), and optionally add a throughput chart to the analytics dashboard.

### Alternatives considered

- **Compute at query time via SQL**: `SELECT completion_tokens / (latency / 1000.0) ...` — works for ad-hoc analysis but doesn't support efficient indexing, histograms, or UI display without extra logic.
- **Rely on OpenTelemetry only**: TTFT is already exported to OTel spans, and tokens/sec could be added as a span attribute. However, this requires an external observability stack (Jaeger, Grafana, etc.) and doesn't benefit users who rely solely on Bifrost's built-in logging and UI.

### Area(s)

Framework, Plugins, UI (Next.js)

### Additional context

- `TimeToFirstToken` is already calculated in `framework/streaming/accumulator.go` (millisecond precision) and carried in `AccumulatedData`, so persisting it should be a relatively small change.
- The Prometheus telemetry plugin (`plugins/telemetry/`) could also benefit from a `bifrost_tokens_per_second` histogram metric for alerting and dashboarding.
- For streaming requests, an **inter-token latency** (average time between chunks) could be a useful follow-up metric, but tokens/sec and TTFT would cover the most common needs.

Vaibhav701161:
## Summary

This PR introduces first-class support for *token generation throughput* and *time-to-first-token (TTFT)* in Bifrost’s logging and analytics pipeline.

Previously, these metrics were either:

* not persisted (`tokens_per_second`), or
* computed but only exported to tracing (`TTFT`),

making them unavailable for log queries, UI inspection, and aggregation.

This change ensures both metrics are:

* computed consistently at write-time,
* persisted in the log store, and
* available for downstream analytics (histograms, provider breakdowns, etc.).

---

### 1. Persisted performance metrics

* Added new fields to `logstore.Log`:

  * `tokens_per_second`
  * `time_to_first_token`

* `time_to_first_token` is:

  * populated only for streaming responses
  * intentionally `nil` for non-streaming requests

---

### 2. Single-source computation model

* Introduced `applyPerformanceMetricsToEntry` as the *only* place where:

  * `tokens_per_second` is computed

* Enforced lifecycle:

  * Computation happens **only once at finalization** (`PostLLMHook`)
  * Not computed during intermediate streaming updates

* Rationale:

  * avoids inconsistent partial values
  * prevents duplication across code paths
  * ensures deterministic analytics

---

### 3. Write-time computation design

`tokens_per_second` is computed at write-time instead of query-time:

```
tokens_per_second = completion_tokens / (latency_ms / 1000.0)
```

**Reasoning:**

* avoids repeated computation during aggregation queries
* ensures consistency across APIs and UI
* enables future indexing or pre-aggregation optimizations

---

### 4. Histogram & aggregation support

Extended logstore aggregation layer to include:

* `avg_tokens_per_second`, `p90_tokens_per_second`
* `avg_time_to_first_token`, `p90_time_to_first_token`

Applied to:

* latency histogram
* provider-level histogram

**Implementation detail:**

* aggregation is performed in-memory for cross-database compatibility
  (SQLite, MySQL, Postgres do not uniformly support percentile functions)

---

### 5. Streaming lifecycle clarity

* `time_to_first_token` is captured during streaming accumulation
* persisted during final log write
* explicitly *not included* in intermediate updates

Added documentation to clarify:

* metric availability timing
* separation between intermediate vs finalized state

---

### 6. Migrations

* Added columns:

  * `tokens_per_second`
  * `time_to_first_token`

* Dropped unnecessary indexes:

  * these fields are not used in filtering conditions
  * indexing would increase write overhead without query benefit

* Migrations are:

  * idempotent
  * safe
  * independent of existing index ordering

---

## Type of change

* [ ] Bug fix
* [x] Feature
* [x] Refactor
* [x] Documentation
* [ ] Chore/CI

---

## Affected areas

* [x] Core (Go)
* [x] Plugins
* [ ] Transports (HTTP)
* [ ] Providers/Integrations
* [ ] UI (Next.js)
* [ ] Docs

---

### 1. Generate logs

```sh
curl http://localhost:8080/v1/chat/completions \
  -d '{
    "model": "ollama/llama3",
    "messages": [
      {"role": "user", "content": "Explain latency vs throughput briefly"}
    ]
  }'
```

### 2. Validate persisted metrics

```sh
curl http://localhost:8080/api/logs?limit=1
```

Expected:

* `tokens_per_second` > 0
* `time_to_first_token` present for streaming requests

### 3. Validate histogram aggregation

```sh
curl "http://localhost:8080/api/logs/histogram/latency?start_time=<start>&end_time=<end>"
```

Expected:

* `avg_tokens_per_second`
* `p90_tokens_per_second`
* `avg_time_to_first_token`
* `p90_time_to_first_token`

---

## Screenshots / Recordings

Attached:

* Clean log output showing `tokens_per_second` and `time_to_first_token`
* Histogram bucket demonstrating aggregated metrics

Screenshot 2026-03-17 165033

---

## Breaking changes

* [ ] Yes
* [x] No

---

## Related issues

Closes #1992

---

## Security considerations

* No impact on authentication, secrets, or external integrations
* Metrics are derived from existing request metadata

---

## Checklist

* [x] Followed contribution guidelines
* [x] Added documentation for design decisions
* [x] Verified build and runtime behavior
* [x] Ensured backward compatibility
* [x] Validated metrics via real requests


akshaydeo:
❤️ for the PR @Vaibhav701161  - ill keep it in my review pipeline

akshaydeo:
@Vaibhav701161 I have recently merged a change where all metrics/histograms are not materialized views vs directly querying from the db. Would you be able to check this?

Vaibhav701161:
I'll check and update accordingly

Vaibhav701161:
@akshaydeo

I reworked the histogram path after your comment about querying directly from DB instead of relying on materialized views. Sharing how it’s structured now to check if this direction makes sense.

The query is now built around a single `base` CTE where all filters + bucketing are applied once, and then everything else derives from that. 

From there, I split it into three parallel paths (latency / TPS / TTFT):

* first compute per-bucket counts (`*_counts`)
* then rank rows using `ROW_NUMBER()` per bucket (and provider if needed)
* then aggregate

For percentiles, I’m using a rank-based approach:

```
MIN(CASE WHEN rank >= CEIL(p * N) THEN value END)
```

so it behaves like a discrete percentile. Main reason was to keep behavior consistent across Postgres / MySQL / SQLite without falling back to Go again.

One thing I changed from the earlier version instead of relying on window counts + grouping on them, counts are now computed separately and joined in. The previous approach was getting messy (and Postgres didn’t like it either), and it also felt a bit fragile if anything upstream changed.

Also switched those joins to LEFT JOIN so we don’t accidentally drop buckets if something drifts worst case we return NULLs instead of missing rows.

There is a tradeoff here: we scan `base` multiple times (for counts + ranking), but it keeps the query simpler and more predictable across DBs. Since we still have the matview path for larger buckets, I thought this was acceptable for now.

One thing I’m being careful about is keeping counts + ranked datasets aligned (same filters / partitions), otherwise percentile thresholds can go wrong.

Let me know if this direction matches what you had in mind, or if you’d prefer pushing more towards a single-scan approach.

akshaydeo:
@Vaibhav701161 we have recently pushed a few important changes to entire histogram flow to use materialized views for metrics. Could you check and confirm this fits in well there?

impoiler:
Closed, as this PR has stalled and there have been quite a few refactors in the logstore since then. I will rewrite it.

PR #2149, opened 2026-03-18
logstore: guard metadata jsonb casts for empty values

### LINKED ISSUE #2148: Postgres log insert fails when metadata empty (metadata::jsonb GIN index)
## Summary
When using Postgres for logs_store, log inserts fail if metadata is an empty string. The GIN index on `metadata::jsonb` throws `invalid input syntax for type json` and the entire INSERT is rejected, so logs never persist after refresh.

## Steps to Reproduce
1. Configure logs_store type=postgres and enable logging.
2. Ensure no logging headers are configured (`client.logging_headers` empty) and do not send any `x-bf-lh-*` headers.
3. Start Bifrost and send an inference request.
4. Check `logs` table or `/api/logs`.

## Expected
Log row is inserted; `/api/logs` returns the entry.

## Actual
Insert fails with: `ERROR: invalid input syntax for type json`.
Postgres shows INSERT into logs failing due to `metadata::jsonb` cast.

## Environment
- Bifrost v1.4.13 (bunx)
- Postgres logs_store

## Root Cause
metadata is stored as TEXT and can be empty string when no logging headers are captured. The GIN index is created with `(metadata::jsonb)`, which fails when metadata = '' even though the row is otherwise valid.

## Suggested Fix
Use NULLIF in the index expression and in metadata filter queries:

```sql
CREATE INDEX IF NOT EXISTS idx_logs_metadata_gin ON logs USING gin ((NULLIF(metadata,'')::jsonb));
```

And use `NULLIF(metadata,'')::jsonb` for metadata filter predicates to leverage the index safely.

## Workaround
Manually recreate the index with NULLIF as above.

yanhao98:
## Summary
- prevent Postgres log inserts from failing when metadata is an empty string
- use NULLIF in the metadata GIN index expression and metadata filters to keep queries safe

## Testing
- not run (Go toolchain not available in this environment)

Fixes #2148

### DIFF
--- a/framework/logstore/migrations.go
+++ b/framework/logstore/migrations.go
@@ -1753,7 +1753,7 @@ func migrationAddMetadataGINIndex(ctx context.Context, db *gorm.DB) error {
 				}
 
 				// Create the GIN index now that all metadata values are valid JSON or NULL.
-				if err := tx.Exec("CREATE INDEX IF NOT EXISTS idx_logs_metadata_gin ON logs USING gin ((metadata::jsonb))").Error; err != nil {
+				if err := tx.Exec("CREATE INDEX IF NOT EXISTS idx_logs_metadata_gin ON logs USING gin ((NULLIF(metadata,'')::jsonb))").Error; err != nil {
 					return fmt.Errorf("failed to create metadata GIN index: %w", err)
 				}
 			}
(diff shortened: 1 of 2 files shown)


akshaydeo:
we will need a new migration considering this is already ran - I'll pick it up in next 10
Mins

yanhao98:
Thanks for the heads‑up. Totally agree on needing a new migration since this already ran. Appreciate you picking it up—happy to help/test if needed.

PR #994, opened 2025-12-03
update: made chunk size configurable via provider network config

flashzzz:
## Summary

Made stream chunk max token size configurable via provider network config. Previously hardcoded at 10MB, now configurable through `StreamMaxTokenSize` in `NetworkConfig`.

## Changes

- Added `DefaultStreamMaxTokenSize` constant (10MB) in `core/schemas/provider.go`
- Added `StreamMaxTokenSize` field to `NetworkConfig` struct with JSON serialization support
- Updated `CheckAndSetDefaults` to set default value when not configured
- Updated shared streaming functions to accept configurable max token size:
  - `HandleOpenAITextCompletionStreaming`
  - `HandleOpenAIChatCompletionStreaming`
  - `HandleOpenAIResponsesStreaming`
  - `HandleAnthropicChatCompletionStreaming`
  - `HandleAnthropicResponsesStream`
- Updated all provider implementations to use the configurable value instead of hardcoded 10MB
- Affected providers: OpenAI, Anthropic, Azure, Vertex, Gemini, Cohere, Cerebras, Ollama, SGL, Parasail, OpenRouter, Groq, Perplexity, Mistral

**Design decisions:**
- Default remains 10MB to maintain backward compatibility
- Config is part of `NetworkConfig` since it's related to streaming network operations
- Bedrock provider unchanged as it uses AWS EventStream protocol which handles buffer sizing differently

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [x] Core (Go)
- [ ] Transports (HTTP)
- [x] Providers/Integrations
- [ ] Plugins
- [ ] UI (Next.js)
- [ ] Docs

## How to test

# Core/Transports
// programmatically:
config := &schemas.ProviderConfig{
    NetworkConfig: schemas.NetworkConfig{
        StreamMaxTokenSize: 20 * 1024 * 1024, // 20MB
    },
}
## Screenshots/Recordings

Provider tests
image

## Breaking changes

- [ ] Yes
- [x] No

Default value (10MB) matches previous hardcoded behavior, so existing configurations work unchanged.

## Related issues

#611 

## Security considerations

None - this is a buffer size configuration for streaming responses. The default remains secure at 10MB, and users who increase it do so at their own discretion based on their use case requirements.

## Checklist

- [x] I read `docs/contributing/README.md` and followed the guidelines
- [ ] I added/updated tests where appropriate
[...]

PR #2199, opened 2026-03-22
[feat]: Add configurable stream read buffer size via network config

### LINKED ISSUE #611: [Feature]: Option to configure stream chunk sizes
### Prerequisites

### Problem to solve

Right now streaming chunk sizes are set to 64KB default and its not configurable

### Proposed solution

Add an option to change the stream chunk size via config.json or ui

### Alternatives considered

_No response_

### Area(s)

Core (Go), UI (Next.js)

### Additional context

_No response_

mvanhorn:
## Summary

The stream read buffer size is hardcoded to 64KB in Azure (`azure.go:1221`) and Gemini (`gemini.go:519,1056`) providers. This adds a `stream_read_buffer_size_kb` field to `NetworkConfig` so it can be configured via `config.json` or the UI.

## Changes

- Added `StreamReadBufferSizeKB` field to `NetworkConfig` in `core/schemas/provider.go`
- Added `StreamReadBufferSize()` helper method that converts KB to bytes with default fallback
- Updated JSON marshaling/unmarshaling alias structs
- Updated `CheckAndSetDefaults()` to apply the default (64KB)
- Replaced hardcoded `64*1024` in `azure.go` and `gemini.go` with `provider.networkConfig.StreamReadBufferSize()`
- Default remains 64KB for backward compatibility

## Type of change

- [x] Feature

## Affected areas

- [x] Core (Go)
- [x] Providers/Integrations

## How to test

```json
{
  "network_config": {
    "stream_read_buffer_size_kb": 128
  }
}
```

Set the field in provider config and verify streaming responses use the configured buffer size.

```sh
cd core && go build ./...
cd core && go test ./schemas/... -run TestNetworkConfig
```

## Breaking Changes

- [ ] Yes
- [x] No

## Related Issues

Closes #611

Note: PR #994 previously attempted this but was closed. This implementation follows the existing `StreamIdleTimeoutInSeconds` pattern in the same struct.

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/core/changelog.md
+++ b/core/changelog.md
@@ -1,0 +1,1 @@
+[feat]: Add configurable stream read buffer size via `stream_read_buffer_size_kb` in network config [@mvanhorn](https://github.com/mvanhorn)
--- a/core/changelog.md
+++ b/core/changelog.md
@@ -1,2 +1,3 @@
 - fix: strip the encrypted reasoning signature when the upstream reports the field as unsupported (e.g. Bedrock Converse replaying a Claude signature onto a non-Anthropic model after a mid-conversation model switch), extending the existing unverifiable-signature fail-soft
 - fix: clear Anthropic raw-body passthrough based on the resolved provider and model pair, so non-Claude models on multi-family providers (Vertex, Azure, Bedrock Mantle) convert the request instead of passing the Anthropic payload through
+[feat]: Add configurable stream read buffer size via `stream_read_buffer_size_kb` in network config [@mvanhorn](https://github.com/mvanhorn)
--- a/core/providers/azure/azure.go
+++ b/core/providers/azure/azure.go
@@ -1218,7 +1218,7 @@ func (provider *AzureProvider) SpeechStream(ctx *schemas.BifrostContext, postHoo
 		// Read SSE events manually to handle binary data with embedded newlines
 		// SSE format: "data: <content>\n\n" - events are separated by double newlines
 		// We can't use bufio.Scanner because MP3 data contains 0x0a bytes which get interpreted as newlines
-		readBuffer := make([]byte, 64*1024) // 64KB read chunks
+		readBuffer := make([]byte, provider.networkConfig.StreamReadBufferSize())
 		var accumulated []byte
 		doneReceived := false
 
--- a/core/providers/azure/azure.go
+++ b/core/providers/azure/azure.go
@@ -855,7 +855,7 @@ func (provider *AzureProvider) SpeechStream(ctx *schemas.BifrostContext, postHoo
 		// Read SSE events manually to handle binary data with embedded newlines
 		// SSE format: "data: <content>\n\n" - events are separated by double newlines
 		// We can't use bufio.Scanner because MP3 data contains 0x0a bytes which get interpreted as newlines
-		readBuffer := make([]byte, 64*1024) // 64KB read chunks
+		readBuffer := make([]byte, provider.networkConfig.StreamReadBufferSize())
 		var accumulated []byte
 		doneReceived := false
 
--- a/core/providers/gemini/gemini.go
+++ b/core/providers/gemini/gemini.go
@@ -516,7 +516,7 @@ func HandleGeminiChatCompletionStream(
 		var lineReader *bufio.Reader
 		var sseReader providerUtils.SSEDataReader
 		if skipInlineData {
-			lineReader = bufio.NewReaderSize(decompressedReader, 64*1024)
+			lineReader = bufio.NewReaderSize(decompressedReader, provider.networkConfig.StreamReadBufferSize())
 		} else {
 			sseReader = providerUtils.GetSSEDataReader(ctx, decompressedReader)
 		}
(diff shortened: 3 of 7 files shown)


akshaydeo:
hi the PR for this already ready - we are exposing max_number_of connections and other network layer configs. ill merge this too once the CodeRabbit review is resolved.

Thanks for the PR ❤️

mvanhorn:
Fixed the compilation errors in 016cb44: HandleGeminiChatCompletionStream and HandleGeminiResponsesStream are package-level functions without access to the provider struct. Added streamReadBufferSize as an explicit parameter, passed from the callers (gemini and vertex providers) where networkConfig is available.

akshaydeo:
❤️ for the PR - would you be able to resolve conflicts

mvanhorn:
Rebased onto current main and resolved the conflicts - the PR is mergeable again. The NetworkConfig gained a few fields upstream since this branch was cut (SecretVar CACertPEM, MaxConnsPerHost, EnforceHTTP2, and the new postHookSpanFinalizer stream parameters), so I re-threaded stream_read_buffer_size_kb through the updated struct, marshal/unmarshal paths, CheckAndSetDefaults, and the gemini/vertex stream call sites alongside the new finalizer param. go build and the schemas tests pass locally.

mvanhorn:
Merged main, and fixed the two things the review found.

**Merge:** main added `KeepAliveTimeoutInSeconds` to `NetworkConfig` in exactly the five places this branch adds `StreamReadBufferSizeKB`, so every conflict was two fields landing in the same spot. Both are kept.

**Schema.** You were right that the feature was unreachable. `stream_read_buffer_size_kb` was not declared in `transports/config.schema.json`, and both `network_config` and `network_config_without_base_url` set `additionalProperties: false`, so any `config.json` using the field was rejected outright. It is now declared in both definitions, with `minimum` 1 and `maximum` 16384 to match the sibling network fields.

**Upper bound.** Also fixed. The buffer is allocated per stream, so an unbounded value is a memory hazard once streams are concurrent. `StreamReadBufferSize` now caps at `MaxStreamReadBufferSizeKB` (16 MB), which is the same number as the schema maximum, so config validation and the runtime clamp agree.

Added a table test for the helper covering the default fallback for 0 and negatives, an in-range value, the cap boundary, and clamping above it. I checked it actually fails with the clamp removed rather than just passing:

```
above_the_cap_is_clamped   FAIL
absurd_value_is_clamped    FAIL
```

`go build ./...` is clean and `go test ./schemas/` passes.

mvanhorn:
Conflicts resolved. The branch is now `MERGEABLE` at `026d181a`.

I rebuilt it rather than merging. The branch had accumulated a chain of upstream merge commits and had drifted about 500 commits behind `main`, so a straight merge produced conflicts in roughly 160 files, nearly all of them under `ui/` and in `go.mod` / version files this change never touches. Replaying the actual change onto current `main` gave two real conflicts instead, both of them additive: `stream_read_buffer_size_kb` landing next to the newer `keep_alive_timeout_in_seconds` field, and a changelog line.

The PR content is unchanged, same seven files:

- `stream_read_buffer_size_kb` on `NetworkConfig`, clamped to `MaxStreamReadBufferSizeKB` (16MB) and defaulting to 64KB, since the buffer is allocated per stream.
- Applied in the Azure speech streaming reader and the Gemini/Vertex streaming readers.
- `HandleGeminiChatCompletionStream` and `HandleGeminiResponsesStream` are package-level and have no `provider` receiver, so the size is passed explicitly from the callers.
- Schema entry in `transports/config.schema.json` and a changelog line.

On checking: `go build ./...` and `go vet ./schemas/... ./providers/...` are clean, `gofmt` is clean, and `go test ./schemas/ -run StreamReadBufferSize` passes 6/6, covering the zero, negative, in-range, at-cap, above-cap and absurd-value paths.

One thing worth flagging since I had to touch it: the clamp now sits alongside the newer `HTTP2PingIntervalInSeconds` clamp in the same function. They are independent, but if you would rather those defaults live together in one place, say so and I will restructure.

mvanhorn:
Both CodeRabbit findings addressed.

**Gemini buffer size:** added `GetSSEDataReaderSize` alongside `GetSSEDataReader`, which now delegates to it — no existing caller changes. Both Gemini SSE branches pass the configured `NetworkConfig.StreamReadBuffer()` value, as do `SpeechStream` and `TranscriptionStream`, so Vertex-Gemini inherits it through the shared handler. The `skipInlineData` lineReader path is untouched. Added tests covering the zero-fallback and the case where a configured size above the 10MB cap lets through an event the default reader rejects with `ErrTooLong`.

**UI exposure:** `stream_read_buffer_size_kb` is now in the NetworkConfig type, defaults (64), validation schema, form values and submitted payload, with a numeric control in `NetworkFormFragment`. `config.schema.json` already carried the field, so the gap was only the sheet.

`go build ./...` and `go test ./... -run 'Stream|SSE|Buffer'` pass; gofmt clean on the changed files.

PR #2200, opened 2026-03-22
[feat]: Add audio URL support in multimodal input

### LINKED ISSUE #418: [Feature]: Should Accept Audio URL too in MultiModal Support
### Prerequisites

### Problem to solve

- In MultiModal Support, it should accept audio url too, currently it accepts a Base64 encoded file only

### Proposed solution

```
 {
                    "type": "input_audio",
                    "input_audio": {
                        "data": "__url__",
                        "format": "url"
                    }
                }
```

### Alternatives considered

_No response_

### Area(s)

Transports (HTTP)

### Additional context

_No response_

mvanhorn:
## Summary

`ChatInputAudio` only accepts base64-encoded audio via the `data` field. This adds a `url` field so callers can pass an audio URL instead. When `data` is empty and `url` is set, providers download and base64-encode the audio before sending upstream.

## Changes

- Added `URL` field to `ChatInputAudio` in `core/schemas/chatcompletions.go`
- Added `URL` field to `ResponsesInputAudio` in `core/schemas/responses.go`
- Propagated URL through mux and utils schema helpers
- Added shared `DownloadAndBase64Encode()` helper in `core/providers/utils/file.go`
- Integrated URL-to-base64 conversion in Gemini chat and responses audio paths

## Type of change

- [x] Feature

## Affected areas

- [x] Core (Go)
- [x] Providers/Integrations

## How to test

Send a chat completion request with audio URL:
```json
{
  "messages": [{
    "role": "user",
    "content": [{
      "type": "input_audio",
      "input_audio": {
        "url": "https://example.com/audio.mp3",
        "format": "mp3"
      }
    }]
  }]
}
```

The provider should download the audio and forward it as base64.

## Breaking Changes

- [x] No

## Related Issues

Closes #418

This contribution was developed with AI assistance (Claude Code + Codex).


akshaydeo:
I'll review and merge it today 🙇‍♂️

akshaydeo:
@Pratham-Mishra04  / @TejasGhatte can you check this

mvanhorn:
Thanks for the ping, @akshaydeo. Re-pushed this on top of current `main` as a single commit (`cc654b2f`) since the Gemini provider had been substantially restructured since March.

What changed in the re-port:

- Audio URL feature re-implemented over the current `core/providers/gemini` shape (chat.go, responses.go, utils.go, plus call sites in vertex.go and the genai HTTP integration).
- `convertContentBlockToGeminiPart` and the chat-side equivalent now take `ctx context.Context` and thread it through to `DownloadURLToBase64`. No `context.Background()` on the provider request path. Addresses the CodeRabbit findings on `responses.go:3193` and `utils.go:1707`.
- `core/providers/utils/file.go` downloader is now fasthttp-based to match the rest of the utils package. Addresses the `file.go:41` finding.
- New tests: `core/providers/gemini/audio_url_test.go` and `core/providers/utils/file_test.go`.
- Verified locally: `go build ./...`, `go vet ./...`, and `go test ./providers/gemini/... ./providers/utils/...` all green.

Mergeable status is back to green. Happy to address whatever @Pratham-Mishra04 / @TejasGhatte spot.

mvanhorn:
Thanks for the catch @greptile-apps. Pushed `9558a1e2` addressing all three:

- **SSRF**: New `validateRequestURL` in `core/providers/utils/file.go` rejects non-https schemes and refuses hosts that resolve to loopback / link-local (covers AWS IMDS `169.254.169.254` and IPv6 `fe80::/10`) / multicast / unspecified / RFC 1918 / ULA. Redirects are no longer followed - a redirect would target a host the guard never saw. Test-only escape hatch (`AllowPrivateAudioURLsForTest`) keeps the existing httptest fixtures working.
- **Silent chat-path failure**: `convertBifrostMessagesToGemini` signature is now `([]Content, *Content, error)`; audio download failures propagate with `fmt.Errorf` wrap, matching the responses path. Updated callers in `chat.go` and `gemini.go`.
- **Per-call client**: replaced with a package-level `downloadClient` so TCP/TLS connection reuse works across fetches.

New tests cover http rejection, unsupported scheme, loopback-in-production, private IP literals (10/8, 192.168/16, 172.16/12, 169.254/16, ::1), redirect refusal, and chat-path error propagation. `go build ./... && go vet ./... && go test ./providers/...` all green locally.

mvanhorn:
Pushed f07b744 addressing both residuals.

**DNS rebinding (Greptile P1)**: IP validation moved into the fasthttp `Dial` hook, so the resolved IP that's actually used for the TCP connection is the one that gets checked. `validateRequestURL` no longer calls `LookupIP` since the Dial hook is now the authoritative enforcement point. Followed the exact pattern Greptile suggested, with `MaxResponseBodySize`/`ReadTimeout`/`WriteTimeout` preserved.

**Converter purity (CodeRabbit Major)**: URL fetching lifted out of `convertBifrostMessagesToGemini` and the responses-path converter into two pre-conversion helpers (`resolveAudioURLs`, `resolveResponsesAudioURLs`). The converters now see only resolved payloads and are pure transformations again, matching the `core/providers/*/!(test).go` coding rule.

New tests in `core/providers/utils/file_test.go`:
- `TestDownloadClientDialRejectsLoopbackResolutionInProduction` - exercises the Dial hook directly so the connection-time check rejects loopback DNS before any TCP connection.
- `TestValidateRequestURLRejectsUnsupportedSchemesAndEmptyHosts` - keeps scheme/host coverage now that the LookupIP block is gone.

Full `go build ./...`, `go vet ./...`, and `go test ./providers/utils/ ./providers/gemini/` all clean (19.6s utils + 0.4s gemini).

Deferred per Greptile's "no action needed if 20s bound is acceptable" note: goroutine lifetime on context cancellation. Also deferred to a follow-up: build-tagging `AllowPrivateAudioURLsForTest` (CodeRabbit's separate Quick Win finding).

mvanhorn:
Follow-up to the prior batch. Pushed 87c8d71 addressing CodeRabbit's "production-callable SSRF bypass switch" Quick Win: `AllowPrivateAudioURLsForTest` now guards on `testing.Testing()` and panics if a non-test binary reaches it. The function works the same way under `go test`, but the bare `allowPrivateAudioURLs = true` assignment is no longer reachable from production code. `go build`/`go vet`/full `./providers/utils/ ./providers/gemini/` suite still clean. `testing` import in non-test files follows the existing precedent in `core/internal/llmtests/*.go`.

Goroutine-lifetime revisit remains deferred per your "no action needed if the 20s bound is acceptable" note - happy to revisit if profile data suggests it's worth tightening.

mvanhorn:
Pushed 178f063.

Aligned the two Gemini converter calls in `transports/bifrost-http/integrations/genai.go` with their declarations: `ToGeminiResponsesResponse` and `ToGeminiResponsesStreamResponse` do not take a context, so the extra argument at lines 219 and 255 is gone.

Audio URL resolution in the batch path now runs under the request context rather than `context.Background()`. `ToGeminiBatchGenerateContentRequestWithContext` carries the caller's context through to `resolveAudioURLs`, and `BatchCreate` on both the Gemini and Vertex providers hands it their own `ctx`. The original context-free entry point stays as a thin wrapper.

Checked locally: `go build ./...` is clean under `core`, and the gemini and vertex provider test packages are both green.

mvanhorn:
Addressed the outstanding review findings in 6edb857.

**Network I/O out of the converter.** `ToGeminiBatchGenerateContentRequest` is deterministic again — URL-backed audio is resolved by the caller, and reaching the converter with an unresolved URL now returns an `errUnresolvedAudioURL` sentinel instead of silently dropping the audio. `ToGeminiBatchGenerateContentRequestWithContext` resolves and then converts, so the Vertex JSONL path keeps a single working entry point.

**BatchCreate skipping resolveAudioURLs.** Correct, it did. It now calls `ResolveBatchAudioURLs` before conversion, so URL-backed audio in batch items reaches Gemini as inline data.

**Canceled contexts still fetching.** The cancellation check now runs before the download starts rather than racing it, so a pre-canceled context makes no outbound request.

`go build ./...` and `go vet` are clean; the gemini, vertex and utils test packages pass.

mvanhorn:
Pushed 0830551 and c62b2d1.

**fasthttp downloader.** `DownloadURLToBase64` now uses `fasthttp.AcquireRequest`/`AcquireResponse` against a `fasthttp.Client` instead of `net/http`, and the old `contextDownloadClient` is removed. The guarantees carry over rather than being re-derived: the same `dialValidatedAudioURL` hook is the client's `Dial`, so resolution-time SSRF validation and address pinning are unchanged; fasthttp's `Do` does not follow redirects, which is what `CheckRedirect` enforced, so 3xx still returns the explanatory error; `MaxResponseBodySize` plus the retained Content-Length pre-check keep the 25 MiB cap, with `ErrBodyTooLarge` mapped onto the existing limit error. Since fasthttp takes no context, a caller deadline maps to `DoDeadline` and otherwise `DoTimeout` bounds the request, with `ctx.Err()` rechecked after.

**Batch audio on the Vertex path.** Separately, `vertexConvertRequestsToJSONL` only routes OpenAI-shaped bodies (those carrying `messages`) through the resolving helper — native Gemini/Vertex-shaped items passed through unresolved. `BatchCreate` now resolves every item up front using the same Body-then-Params selection the converter uses, so both shapes are covered, and `ToGeminiBatchGenerateContentRequestWithContext` is a pure conversion again. Its only caller is that JSONL path, and Gemini's own `BatchCreate` calls `ResolveBatchAudioURLs` directly, so no Gemini behavior changes.

`go build ./...`, `go vet`, and gofmt are clean; the utils, vertex and gemini packages are green.

mvanhorn:
Addressed in `db5e32d`.

The finding was right: `DoDeadline` and `DoTimeout` bound the request but do not interrupt it, so cancelling ctx left the caller waiting out `ReadTimeout` on a stalled read.

fasthttp's dial hook takes no context, so rather than smuggling one in, each download now builds its own `HostClient` whose `Dial` closure captures that request's ctx and calls the same SSRF-validating dialer as before. Cancellation closes that conn through `context.AfterFunc`, which unblocks the in-flight read; the wrapper's `Close` stops the watch so a completed download does not close the conn a second time. `ReadTimeout`, `WriteTimeout` and `MaxResponseBodySize` carry over from the shared client, and the pooled request and response stay owned by the download goroutine until it returns rather than being released on the cancellation path.

Added `TestDownloadURLToBase64CancelsStalledResponse`, which stalls a server, cancels, and requires `context.Canceled` back within a second. The existing guards still pass: SSRF, private-IP literal, loopback-in-production, unsupported scheme and redirect refusal.

PR #2217, opened 2026-03-23
routing: add model templating support in routing targets

### LINKED ISSUE #2128: [Feature] Add support for modifying targets using inputs in Routing Rules
## Discord Thread Context
**Thread:** Are there any known issues with virtual keys? I only have one provider configured (openrouter), but
**Link:** https://discord.com/channels/1371790330269007932/1477090591958372423
**Created by:** pratham_maxim_91868 (1388047873903562783)

## Recent Thread Messages
**akshay_deo:** umm I think we are missing something here
**akshay_deo:** can you share you request and setup - I’ll try to replicate it
**omnipresent0:** i also faced similar issue Even though the virtual key is present, specific to Claude 4.6 models Opus and Sonnet, it is not being accessed. Although the keys are set up, the configuration is present a...
**akshay_deo:** could you share the setup? cc <@1388047873903562783>
**akshay_deo:** cause just the above line does not give us any information to work on
**akshay_deo:** could you share the request you are sending to bifrost?
the virtual key you have?
do you have Allow direct API keys enabled in the settings?
**akshay_deo:** asking <@1388047873903562783> to check irrespectively
**pratham_maxim_91868:** Hey <@791924320606879744> I think the issue here is anthropic expects `claude-sonnet-4-5` as the model name and openrouter expects `anthropic/claude-sonnet-4-5`. If you have specific model you're usin...
**pratham_maxim_91868:** but we can add support for adding suffix/preffix to incoming models in the target

**Discord Thread:** https://discord.com/channels/1371790330269007932/1477090591958372423

Vaibhav701161:
## Summary

This PR adds support for **modifying the target model dynamically using input values in routing rules**.

While investigating issue #2128, it was confirmed that routing rules only support static model values. This creates problems when different providers expect different model formats (for example, `claude-sonnet-4-5` vs `anthropic/claude-sonnet-4-5`). As a result, requests either fail or require manual duplication of routing rules.

This change allows deriving the target model from the incoming request using a minimal templating approach, enabling consistent routing across providers.

---

## Changes

* Added support for templated model values in routing targets

  * Example:

    ```json
    "model": "anthropic/{{input.model}}"
    ```

* Introduced `resolveRoutingModelTemplate` to:

  * resolve placeholders using request context
  * validate template syntax
  * return errors for unknown variables

* Applied transformation only to `target.Model`

  * provider and key_id behavior remain unchanged

* Added fail-safe handling:

  * if template resolution fails, the rule is skipped
  * warning + routing logs are emitted

* Added tests covering:

  * valid template resolution
  * multiple placeholders
  * unknown variables
  * invalid syntax
  * fallback to next rule when template fails

### Design decisions / trade-offs

* Kept templating limited and simple instead of extending CEL

  * CEL is currently used for rule matching (boolean evaluation)
  * Adding string transformation to CEL would increase complexity
* Avoided expanding scope beyond model transformation to keep the change minimal and safe

---

## Type of change

* [ ] Bug fix
* [x] Feature
* [ ] Refactor
* [ ] Documentation
* [ ] Chore/CI

---

## Affected areas

* [ ] Core (Go)
* [ ] Transports (HTTP)
* [ ] Providers/Integrations
* [x] Plugins
* [ ] UI (Next.js)
* [ ] Docs

---

### Start server

```sh
cd bifrost
go run main.go -app-dir ./tmp/test -port 18080 -host 127.0.0.1
```

---

### Create virtual key

```sh
curl --request POST \
  --url http://127.0.0.1:18080/api/governance/virtual-keys \
  --header 'Content-Type: application/json' \
  --data '{"name":"routing-model-transform-test"}'
```

---

### Baseline (no routing rule)

```sh
curl --request POST \
  --url http://127.0.0.1:18080/v1/chat/completions \
  --header 'Content-Type: application/json' \
  --header 'x-bf-vk: <VK>' \
  --data '{
    "model": "anthropic/claude-sonnet-4-5",
    "messages": [{"role": "user", "content": "Hello"}],
    "max_tokens": 10
  }'
```

Expected:

* provider: `anthropic`
* error: `failed to get config`

---

### Add routing rule

```sh
curl --request POST \
  --url http://127.0.0.1:18080/api/governance/routing-rules \
  --header 'Content-Type: application/json' \
  --data '{
    "name": "model-prefix-routing",
    "cel_expression": "true",
    "targets": [
      {
        "provider": "openrouter",
        "model": "anthropic/{{input.model}}",
        "weight": 1
      }
    ],
    "scope": "virtual_key",
    "scope_id": "<VK_ID>",
    "priority": 0,
    "enabled": true
  }'
```

---

### Validate

```sh
curl --request POST \
  --url http://127.0.0.1:18080/v1/chat/completions \
  --header 'Content-Type: application/json' \
  --header 'x-bf-vk: <VK>' \
  --data '{
    "model": "claude-sonnet-4-5",
    "messages": [{"role": "user", "content": "Hello"}],
    "max_tokens": 10
  }'
```

Expected:

* provider: `openrouter`
* model_requested: `anthropic/claude-sonnet-4-5`

---

### Run tests

```sh
go test ./plugins/governance -count=1
```

---

## Screenshots/Recordings

Screenshot 2026-03-23 072631

---

## Breaking changes

* [ ] Yes
* [x] No

---

## Related issues

Closes #2128

---

## Security considerations

* Template variables are restricted to known values (`input.model`, `model`)
* No arbitrary code execution or external evaluation

---

## Checklist

* [x] I verified tests pass locally

### DIFF
--- a/plugins/governance/routing.go
+++ b/plugins/governance/routing.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"math/rand/v2"
 	"regexp"
+	"sort"
 	"strings"
 
 	"github.com/google/cel-go/cel"
@@ -22,6 +23,9 @@ var paramKeyPattern = regexp.MustCompile(`params\[["']([^"']+)["']\]`)
 
 // paramInPattern matches "in params" membership test patterns like "Region" in params or 'Region' in params
 var paramInPattern = regexp.MustCompile(`["']([^"']+)["']\s+in\s+params`)
+
+// targetTemplatePattern matches template placeholders like {{input.model}}.
+var targetTemplatePattern = regexp.MustCompile(`\{\{\s*([^{}\s]+)\s*\}\}`)
 
 // ScopeLevel represents a level in the scope precedence hierarchy
 type ScopeLevel struct {
@@ -155,7 +159,13 @@ func (re *RoutingEngine) EvaluateRoutingRules(ctx *schemas.BifrostContext, routi
 
 				model := routingCtx.Model
 				if target.Model != nil && *target.Model != "" {
-					model = *target.Model
+					resolvedModel, resolveErr := resolveRoutingModelTemplate(*target.Model, routingCtx)
+					if resolveErr != nil {
+						re.logger.Warn("[RoutingEngine] Rule %s skipped: model template resolution failed: %v", rule.Name, resolveErr)
+						ctx.AppendRoutingEngineLog(schemas.RoutingEngineRoutingRule, fmt.Sprintf("Rule '%s' [%s] → skipped: model template resolution failed: %v", rule.Name, rule.CelExpression, resolveErr))
+						continue
+					}
+					model = resolvedModel
 				}
 
 				keyID := ""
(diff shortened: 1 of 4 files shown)


Vaibhav701161:
@akshaydeo , please provide a review

plugins/governance/routing.go
```diff
@@ -188,6 +198,71 @@ func (re *RoutingEngine) EvaluateRoutingRules(ctx *schemas.BifrostContext, routi
...
+//
+// This avoids introducing a full expression language into routing targets.
+// Future improvements may unify this under CEL or a dedicated transformation layer.
+//
+// Supported template variables (intentionally minimal):
+// - input.model
+// - model
+//
+// This is intentionally restricted to avoid turning routing into a full templating system.
+func resolveRoutingModelTemplate(templateValue string, ctx *RoutingContext) (string, error) {
```
akshaydeo: this could be the blocker for the PR. We can't run this on our own - on every incoming request. Can you see options of how can we execute this with existing CEL itself?

plugins/governance/routing.go
```diff
@@ -188,6 +198,71 @@ func (re *RoutingEngine) EvaluateRoutingRules(ctx *schemas.BifrostContext, routi
...
+	       return "", fmt.Errorf("invalid template syntax in model target")
+       }
+
+       variables := map[string]string{
+	       "input.model": ctx.Model,
+	       "model":       ctx.Model,
+       }
+
+       unknownVars := make(map[string]struct{})
+       rendered := targetTemplatePattern.ReplaceAllStringFunc(templateValue, func(match string) string {
```
akshaydeo: This is going to be extremely slow at scale - as it allocates bunch of strings on heap

akshaydeo (changes requested):

Vaibhav701161:
@akshaydeo Thanks for the earlier feedback , I’ve reworked this part completely.

The templating approach is gone now. Instead of doing string replacement on every request, model transformation is handled using CEL itself. So something like `anthropic/{{input.model}}` becomes a CEL expression like `"anthropic/" + model`.

All expressions are compiled once when the rule is loaded and cached (including errors), so at runtime it’s just a lookup and either a direct value or a CEL eval. There’s no regex, no string parsing, and no compilation happening on the hot path anymore.

Behavior is also stricter now:

* valid CEL → evaluated
* plain strings → used as-is
* invalid expressions or old `{{}}` syntax → fail fast and the rule is skipped (error is cached, so no repeated work)

The idea was to keep everything inside the existing CEL system instead of introducing a parallel templating layer, and to avoid the per-request allocations that were flagged.

Added tests around invalid cases, edge cases like `gpt-4+turbo`, and a small benchmark to verify the runtime path.

Let me know if this direction looks better or if you’d like it handled differently.

Vaibhav701161:
@akshaydeo , any opinions ? How does it look now ?

PR #2232, opened 2026-03-23
fix: persist and apply store_raw_request_response

chenbo515:
## Bug
- after enabling store_raw_request_response in the provider debugging UI, refreshing the page showed the switch as disabled again because the provider CRUD API did not round-trip that field
- even after enabling store_raw_request_response, log details still did not show Raw Request / Raw Response because the runtime provider config never received the setting, so raw capture was not activated

## Root cause
- providers API handlers omitted store_raw_request_response from request/response payload handling
- account config conversion omitted StoreRawRequestResponse when building the runtime provider config

## Fix
- include store_raw_request_response in provider create, update, and response payloads
- preserve the field during provider updates and return it to the dashboard correctly
- propagate the field into runtime provider config so raw request/response logging can actually occur

## Verification
- local testing


akshaydeo:
@chenbo515 this will only happen if you are running multiple instances of Bifrost OSS. Is this the case?

akshaydeo (approved):

chenbo515:
> @chenbo515 this will only happen if you are running multiple instances of Bifrost OSS. Is this the case?

No, it happens in the single instance as well.

PR #2281, opened 2026-03-25
feat(otel): propagate raw request/response to OTEL span attributes

### LINKED ISSUE #2104: [Feature]: Propagate raw_request and raw_response to OTEL span attributes
### Prerequisites

### Problem to solve

When `send_back_raw_request: true` and `send_back_raw_response: true` are configured on a provider, the full provider request/response payloads are available in `BifrostResponseExtraFields.RawRequest` and `BifrostResponseExtraFields.RawResponse`. These are correctly:

- Returned in the HTTP response under `extra_fields`
- Stored by the **logging plugin** (which marshals them via `sonic.Marshal()` in `plugins/logging/operations.go`)

However, they are **never propagated to OTEL spans**. The tracing path in `framework/tracing/llmspan.go` (`PopulateRequestAttributes` / `PopulateResponseAttributes`) does not read `ExtraFields.RawRequest` or `ExtraFields.RawResponse`, so they never appear in `span.Attributes` and are therefore invisible to OTEL-based observability tools (Langfuse, Datadog, etc.).

This affects users who rely on OTEL traces to debug LLM calls — without the raw request/response, they must cross-reference HTTP logs or the logging plugin's database to see what was actually sent to/received from the provider.

### Proposed solution

When `send_back_raw_request` / `send_back_raw_response` are enabled on a provider, include the raw payloads as span attributes (e.g. `gen_ai.raw_request`, `gen_ai.raw_response`) in the OTEL trace.

The data is already available in `BifrostResponseExtraFields` and the logging plugin already serializes it — the gap is that `PopulateResponseAttributes` in `framework/tracing/llmspan.go` never reads these fields into `span.Attributes`.

The OTEL converter (`anyToKeyValue()`) already handles maps/objects recursively, so no changes should be needed on the export side.

Since these payloads can be large (avg ~95KB per request per #1665), once #1561 lands OTEL span configurability, this could be exposed as a toggle. But the baseline ask is simply to propagate the data when the existing `send_back_raw_*` flags are already enabled.

### Alternatives considered

1. **Use the logging plugin instead of OTEL** — works, but forces users into a separate observability pipeline. Teams standardized on OTEL (e.g. Langfuse via OTLP) shouldn't need to also query a separate PostgreSQL logs table.
2. **Client-side extraction from HTTP response** — the client can read `extra_fields.raw_request` from the HTTP response and report it to the observability tool directly, but this duplicates observations and bypasses the OTEL pipeline.

### Area(s)

Core (Go), Framework, Plugins

### Additional context

- The logging plugin's handling of `raw_request`/`raw_response` in `plugins/logging/operations.go` (lines 264–278, 456–471) serves as a reference implementation for how to extract and serialize these fields.
- Related issues:
  - #1960 — token usage details not propagated to OTEL (same root cause: `PopulateResponseAttributes` is too sparse)
  - #1561 — missing LLM call metadata in OTEL spans (cache tokens, model params)
  - #1117 — custom per-request headers as OTEL span attributes
  - #1665 — PostgreSQL TOAST bloat from `raw_request`/`raw_response` in logging plugin (indirectly related: shows these fields are large and already stored in the logging path)
- Note: these payloads can be large (avg ~95KB per row per #1665). Consider whether the attribute should be opt-in via a dedicated config flag (e.g. `otel_include_raw_request_response`) to avoid bloating traces for users who don't need it.

gsaiz:
> **Disclaimer:** I am not a Go developer — the code was written by AI but I did my best to review all its decisions.

## Summary
- Add `gen_ai.raw_request` / `gen_ai.raw_response` as OTEL span attributes in `PopulateResponseAttributes`
- Add attribute constants to `core/schemas/trace.go`
- Gate emission behind `forward_raw_request_response` OTEL plugin config (default `false`)
- Document the new attributes and their interaction with capture flags in `otel.mdx`

## Design decisions

**Attribute naming**: Uses `gen_ai.raw_request` / `gen_ai.raw_response` under the existing `gen_ai.*` namespace, consistent with other span attributes in the codebase.

**`sonic` for serialisation**: Uses `sonic.Marshal` following the logging plugin's established pattern for raw payload serialisation (`plugins/logging/main.go`, `plugins/logging/operations.go`). The `framework/tracing` package previously had no `sonic` dependency; this PR introduces it for consistency with the rest of the framework module.

**`forward_raw_request_response` config flag**: Raw attributes are always populated in the internal tracing layer, but the OTEL plugin strips them at emission time unless `forward_raw_request_response: true` is set. This keeps the default safe (no large payloads in spans) while letting operators opt-in. Addresses feedback from @akshaydeo.

**`store_raw_request_response` does NOT propagate to OTEL**: When only `store_raw_request_response` is enabled (with both `send_back_raw_*` off), `core/bifrost.go` temporarily populates the raw fields for PostLLMHooks but strips them before `PopulateResponseAttributes` runs. The docs table reflects this accurately. A follow-up could add a dedicated flag or move span population earlier if this behavior is desired.

**Release order**: The new constants live in `core/schemas/trace.go`. Since `framework/go.mod` pins `core` to a published version, the `core` module must be tagged first before the `framework` change passes CI (`GOWORK=off`).

## Test plan
- [x] Unit tests for all cases: both fields set, both nil, only request, only response, nil response, json.RawMessage production type, unmarshalable value
- [x] Converter tests: forward enabled, forward disabled, no raw attrs present
- [ ] Verify tests pass in CI after `core` is tagged
- [ ] Manual verification: enable `send_back_raw_request` on a provider, set `forward_raw_request_response: true`, make an LLM call, confirm `gen_ai.raw_request` appears in the OTEL span

Closes #2104

### DIFF
--- a/core/schemas/trace.go
+++ b/core/schemas/trace.go
@@ -203,6 +203,10 @@ const (
 	AttrTimeToFirstToken = "gen_ai.response.time_to_first_token"
 	AttrTotalChunks      = "gen_ai.response.total_chunks"
 
+	// Raw Payload Attributes
+	AttrRawRequest  = "gen_ai.raw_request"
+	AttrRawResponse = "gen_ai.raw_response"
+
 	// Plugin Attributes (for aggregated streaming post-hook spans)
 	AttrPluginInvocations     = "plugin.invocation_count"
 	AttrPluginAvgDurationMs   = "plugin.avg_duration_ms"
--- a/docs/features/observability/otel.mdx
+++ b/docs/features/observability/otel.mdx
@@ -85,6 +85,30 @@ This is useful for:
 
 ---
 
+## Raw Request/Response in Traces
+
+When raw request/response capture is enabled on a provider (via `send_back_raw_request` or `send_back_raw_response`), the raw payloads are automatically included as OTEL span attributes. This lets you see exactly what was sent to and received from the provider in your OTEL backend (Langfuse, Datadog, Grafana, etc.) without cross-referencing HTTP logs.
+
+### Span Attributes
+
+| Attribute | Type | Description |
+|-----------|------|-------------|
+| `gen_ai.raw_request` | `string` | JSON-serialized raw request body sent to the provider |
+| `gen_ai.raw_response` | `string` | JSON-serialized raw response body received from the provider |
+
+### Interaction with Capture Flags
+
+| Flag | Client Response | Logging Plugin | OTEL Spans |
+|------|----------------|----------------|------------|
+| `send_back_raw_request` / `send_back_raw_response` | Yes | Yes | Yes |
+| `store_raw_request_response` | No | Yes | No |
+
+<Note>
+Raw payloads can be large (avg ~95KB per request). Enable capture selectively for debugging or when your observability pipeline can handle the additional data volume.
+</Note>
+
+---
+
 ## Setup
 
 <Tabs group="setup-method">
--- a/docs/features/observability/otel.mdx
+++ b/docs/features/observability/otel.mdx
@@ -40,6 +40,7 @@ The plugin supports multiple trace formats to match your observability platform:
 | `protocol` | `string` | ✅ Yes | Transport protocol: `http` or `grpc` |
 | `headers` | `object` | ❌ No | Custom headers for authentication (supports `env.VAR_NAME`) |
 | `tls_ca_cert` | `string` | ❌ No | File path to client CA certificate for TLS. Optional. Works with both gRPC and HTTP protocol |
+| `forward_raw_request_response` | `boolean` | ❌ No | Forward raw request/response payloads to OTEL span attributes. Defaults to `false` |
 
 ### Environment Variable Substitution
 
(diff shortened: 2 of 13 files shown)


akshaydeo:
I think this still can be a config on the OTEL plugin - to enable or disable forwarding Raw Request/Response to OTel. (if enabled)

gsaiz:
@akshaydeo  Okay, so we can keep propagating raw_request / raw_response internally as in this PR, but only emit them to OTEL when the proposed OTEL config is enabled. Shall I include that in this PR?

gsaiz:
Pushed `3d4721b2` addressing all review feedback:

### @akshaydeo
Added `forward_raw_request_response` config flag to the OTEL plugin (default `false`). Raw attributes are populated in the internal tracing layer but stripped at emission time in `convertSpanToOTELSpan` unless the flag is `true`. Config added across: Go plugin, config schemas, Helm values, UI toggle, and docs.

### @coderabbitai
- **Unmarshalable test** — already present (`TestPopulateResponseAttributes_UnmarshalableRawRequest`)
- **`sonic` vs `encoding/json`** — code already uses `sonic.Marshal`. Updated PR description to correctly reflect this design decision.
- **Type switch for `json.RawMessage`** — investigated the codebase: the logging plugin uses the same `sonic.Marshal` pattern for the common path (`plugins/logging/main.go:687-698`). At runtime `RawRequest`/`RawResponse` hold `json.RawMessage` (from `compactRawJSON`), which `sonic.Marshal` handles correctly via `MarshalJSON()`. No functional issue — deferring the type-switch optimisation.

### @greptile-apps
- **`GetExtraFields()` nil guard** — removed; only inner `RawRequest != nil` / `RawResponse != nil` checks remain
- **`json.RawMessage` test** — already present (`TestPopulateResponseAttributes_JsonRawMessageType`)
- **Constant grouping** — already under `// Raw Payload Attributes` section
- **Silent marshal comment** — already present at `llmspan.go:105-108`
- **PR description inconsistency** — fixed; description now correctly states `sonic`

PR #2370, opened 2026-03-29
fix: use incremental delta writes for governance usage dumps

rasta-rocket:
## Summary
Use incremental delta writes for governance usage dumps to prevent multi-replica data loss

DumpBudgets and DumpRateLimits were using absolute SET writes (SET current_usage = X), causing each replica to overwrite the others' accumulated usage every 10s flush cycle. Changed to delta-based atomic increments (SET current_usage = current_usage + delta) using the existing but unused LastDBUsages tracking maps, so each replica only contributes its own incremental usage to the database.

The bug is a last-writer-wins problem in plugins/governance/store.go. The two dump functions — DumpBudgets (line 1851) and DumpRateLimits (line 1776) — were using absolute SET writes:
-- Before (broken): each replica overwrites with its local view
```
UPDATE budgets SET current_usage = 5.0 WHERE id = 'budget-123';
UPDATE rate_limits SET token_current_usage = 42, request_current_usage = 10 WHERE id = 'rl-456';
```
Every 10 seconds, each Kubernetes replica dumps its in-memory accumulation to the database. Since each replica only sees its own traffic, it writes a partial value that overwrites the contributions of all other replicas. The database value oscillates depending on which replica flushes last.
What Was Broken

Step | Replica A | Replica B | DB (truth)
-- | -- | -- | --
t=0 | mem=0 | mem=0 | 0
t=5s | handles $5 → mem=5 | handles $3 → mem=3 | 0
t=10s | dumps → SET 5 |   | 5
t=10s |   | dumps → SET 3 | 3 (lost $5!)
t=15s | handles $2 → mem=7 | handles $1 → mem=4 | 3
t=20s | dumps → SET 7 |   | 7
t=20s |   | dumps → SET 4 | 4 (lost $7!)

The value alternates between replicas' views. The true total ($11) is never reached.

## Changes

TLDR: `plugins/governance/store.go` — `DumpBudgets()` and `DumpRateLimits()` rewritten to use delta-based atomic increments via `gorm.Expr()`, leveraging the pre-existing `LastDBUsages` tracking maps.

Changed both DumpBudgets and DumpRateLimits from absolute SET to incremental delta-based writes:
-- After (fixed): each replica only adds its own delta
```
UPDATE budgets SET current_usage = current_usage + 2.0 WHERE id = 'budget-123';
UPDATE rate_limits SET token_current_usage = token_current_usage + 5, 
                       request_current_usage = request_current_usage + 1 WHERE id = 'rl-456';
```
The mechanism uses the already-existing (but unused) LastDBUsages* maps:
1. Snapshot the current in-memory values
2. Compute delta = current_in_memory - last_written_value (what this replica accumulated since its last dump)
3. Atomic SQL increment using gorm.Expr("current_usage + ?", delta) — each replica only adds its own contribution
4. Update LastDBUsages* after successful write, so the next cycle computes a fresh delta
5. On deadlock, skip the LastDBUsages* update so the delta is retried in the next cycle (no data loss)
The reset paths (ResetExpiredBudgetsInMemory, ResetExpiredRateLimitsInMemory) already correctly zero both the in-memory counter and the LastDBUsages* tracker, so resets produce the correct delta (0 - 0 = 0, then new usage accumulates normally).

Step | Replica A (delta) | Replica B (delta) | DB (truth)
-- | -- | -- | --
t=0 | mem=0, last=0 | mem=0, last=0 | 0
t=5s | handles $5 → mem=5 | handles $3 → mem=3 | 0
t=10s | delta=5−0=5, ADD 5 | delta=3−0=3, ADD 3 | 8 (correct!)
t=15s | handles $2 → mem=7 | handles $1 → mem=4 | 8
t=20s | delta=7−5=2, ADD 2 | delta=4−3=1, ADD 1 | 11 (correct!)

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [ ] Core (Go)
- [ ] Transports (HTTP)
- [ ] Providers/Integrations
- [x] Plugins
- [ ] UI (Next.js)
- [ ] Docs

## How to test

Run 2 replicas of bifrost
Create 1 VK with a budget limit assigned
Generate inference calls on both instances (individually or via loadbalancer)

Get the current usage of the budget associated to that virtual key by inspecting the database (PG in that case)
```
select current_usage from governance_budgets where id = 'budget_id';
```

## Screenshots/Recordings

## Breaking changes

- [ ] Yes
- [x] No

## Related issues

## Security considerations

## Checklist

### DIFF
--- a/plugins/governance/store.go
+++ b/plugins/governance/store.go
@@ -1746,67 +1746,103 @@ func (gs *LocalGovernanceStore) DumpRateLimits(ctx context.Context, tokenBaselin
 		return true // continue
 	})
 
-	// Prepare rate limit usage updates with baselines
-	type rateLimitUpdate struct {
+	// Snapshot current in-memory values for delta computation
+	type rateLimitSnapshot struct {
 		ID                  string
 		TokenCurrentUsage   int64
 		RequestCurrentUsage int64
 	}
-	var rateLimitUpdates []rateLimitUpdate
+	var snapshots []rateLimitSnapshot
 	for rateLimitID := range rateLimitIDs {
 		if rateLimitValue, exists := gs.rateLimits.Load(rateLimitID); exists && rateLimitValue != nil {
 			if rateLimit, ok := rateLimitValue.(*configstoreTables.TableRateLimit); ok && rateLimit != nil {
-				update := rateLimitUpdate{
+				snapshots = append(snapshots, rateLimitSnapshot{
 					ID:                  rateLimit.ID,
 					TokenCurrentUsage:   rateLimit.TokenCurrentUsage,
 					RequestCurrentUsage: rateLimit.RequestCurrentUsage,
-				}
-				if tokenBaseline, exists := tokenBaselines[rateLimit.ID]; exists {
-					update.TokenCurrentUsage += tokenBaseline
-				}
-				if requestBaseline, exists := requestBaselines[rateLimit.ID]; exists {
-					update.RequestCurrentUsage += requestBaseline
-				}
-				rateLimitUpdates = append(rateLimitUpdates, update)
+				})
 			}
 		}
 	}
 
-	// Save all updated rate limits to database using direct UPDATE to avoid overwriting config fields
-	if len(rateLimitUpdates) > 0 && gs.configStore != nil {
-		if err := gs.configStore.ExecuteTransaction(ctx, func(tx *gorm.DB) error {
-			for _, update := range rateLimitUpdates {
-				// Direct UPDATE only updates usage fields
-				// This prevents overwriting max_limit or reset_duration that may have been changed by other nodes/requests
-				result := tx.WithContext(ctx).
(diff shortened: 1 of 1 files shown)


akshaydeo:
hi thanks for the PR. This is not a bug but how replicas work in OSS vs enterprise. (https://docs.getbifrost.ai/deployment-guides/how-to/multinode)

This will break our enterprise offering of leader based writes so we wont be able to accept this PR. But if you think this is critical - you can maintain a separate fork and we can keep it linked in our readme 🙇

rasta-rocket:
Hello @akshaydeo thank you for the clarification in that case is the rate/budget limit a feature that is part of the OSS offering ?

akshaydeo:
it is with single instance. With replicas - we cant rely on db as we need to keep everything in the memory (for speed, accuracy) - hence in enterprise we have replication logic to allow this.

PR #2505, opened 2026-04-04
feat: add SSO login buttons to login page

ktk1012:
## Summary
- Add Google OAuth and SAML SSO login buttons to the enterprise login page, conditionally rendered based on `enabled_methods`, `google_sso`, and `saml` fields from the `/session/is-auth-enabled` endpoint response
- Password form is conditionally shown only when `enabled_methods` includes "password" (or when the field is absent, for backward compatibility)
- An "or" divider separates SSO buttons from the password form when both are available
- SSO callback errors (redirected via `?error=...&message=...` query params) are surfaced in the existing error banner

## Test plan
- [ ] Verify login page renders normally (password form only) when backend returns no SSO fields
- [ ] Verify Google SSO button appears when `google_sso.login_url` is present in the auth response
- [ ] Verify SAML SSO button appears when `saml.login_url` is present in the auth response
- [ ] Verify "or" divider appears when both SSO and password are enabled
- [ ] Verify password form is hidden when `enabled_methods` does not include "password"
- [ ] Verify SSO error message displays when redirected with `?error=sso_failed&message=...`
- [ ] Run `cd ui && npx tsc --noEmit` — passes with zero errors

### DIFF
--- a/ui/app/_fallbacks/enterprise/components/login/loginView.tsx
+++ b/ui/app/_fallbacks/enterprise/components/login/loginView.tsx
@@ -11,6 +11,15 @@ import Image from "next/image";
 import { useRouter } from "next/navigation";
 import { useEffect, useState } from "react";
 
+const GoogleIcon = () => (
+	<svg viewBox="0 0 24 24" className="mr-2 h-4 w-4">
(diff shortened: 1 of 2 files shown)


akshaydeo:
@ktk1012 what are you trying to do here?

PR #2611, opened 2026-04-09
fix: restore prompt stream EOF before trace flush

### LINKED ISSUE #2606: [Bug]: PR #2513 cleanup ordering regresses prompt SSE EOF after terminal responses event
## Prerequisites

- [x] I searched existing issues and PRs first
- [x] I reproduced this on a clean tagged build, not only in a local dev environment

## Description

This report was produced from an **agent-driven end-to-end analysis** (reproduction, git archaeology, code reading, rebuild, and retest). If the maintainers think this diagnosis or proposed direction is not appropriate for `main`, please feel free to close it.

### Summary

`transports/v1.5.0-prerelease2` appears to regress **prompt SSE/HTTP stream termination** for OpenAI Responses-compatible streams.

In the failing case, Bifrost forwards the full event stream, including `response.completed`, but the downstream HTTP stream does not terminate promptly, so clients keep waiting after the terminal event.

### Version matrix

Using the same fake upstream (`https://fake-api-hono-vite.oo1.dev/v1/responses`) and same request body:

- direct fake upstream: **exits normally** (`~0.31s`)
- official `v1.5.0-prerelease1`: **exits normally** (`~0.28s`)
- official `v1.5.0-prerelease2`: **hangs until timeout** (`20.00s`) even though `response.completed` is already emitted

So, for this scenario, this looks like a **prerelease2 regression**.

### Likely source

The behavior starts with PR #2513 / commit `5978108d` (`fix race in data reading from fasthttp req for integrations`).

That PR intentionally changed the defer ordering in:

- `transports/bifrost-http/integrations/router.go`

from a single cleanup closure to separate defers, with the explicit note:

```go
// Separate defers ensure each cleanup runs even if an earlier one panics (LIFO order)
```

and with PR intent of ensuring trace/post-stream completion runs before request/reader cleanup.

### Why this is tricky

I also checked issue #2511 and PR #2513 rationale carefully.

The maintainer was not doing a style-only change there. The intent was to avoid a race/use-after-free around streaming cleanup and keep cleanup panic-safe.

However, in the tagged `prerelease2` runtime, the resulting ordering also seems to delay the downstream EOF signal in the OpenAI Responses streaming path.

### Important note about `main`

Current `main` has drifted somewhat from the exact `prerelease2` tag behavior around tracing/stream completion. So I am **not** claiming the released `prerelease2` fix should necessarily be reverted 1:1 on `main` without review.

What I am claiming is:

1. the **released `prerelease2` artifact** regresses prompt stream closure in this reproduction
2. the regression is strongly associated with the cleanup-order change introduced by `#2513`
3. `main` likely needs a follow-up that preserves the original panic-safety goal **without** delaying downstream EOF

### Fake upstream request that exits normally

```bash
wget -qO- \
  --header='Content-Type: application/json' \
  --post-data='{"model":"success","input":"Reply with exactly OK.","max_output_tokens":32,"stream":true}' \
  https://fake-api-hono-vite.oo1.dev/v1/responses
```

### Bifrost config used

```json
{
  "$schema": "https://www.getbifrost.ai/schema",
  "providers": {
    "openai": {
      "keys": [
        {
          "name": "fake-openai-key",
          "value": "333",
          "models": ["*"],
          "weight": 1.0
        }
      ],
      "network_config": {
        "base_url": "https://fake-api-hono-vite.oo1.dev",
        "default_request_timeout_in_seconds": 60
      }
    }
  }
}
```

### Bifrost request used

```bash
wget -qO- \
  --header='Content-Type: application/json' \
  --post-data='{"model":"openai/success","input":"Reply with exactly OK.","max_output_tokens":32,"stream":true}' \
  http://127.0.0.1:8080/openai/responses
```

### Observed behavior

For `v1.5.0-prerelease2`, the stream includes terminal events like `response.completed`, but the process still waits until external timeout instead of exiting naturally.

## Suspected fix direction

I would not recommend a blind revert of `#2513` in `main`, because that PR was fixing a real streaming cleanup race.

The safer direction seems to be:

- preserve separate cleanup units (panic isolation)
- preserve whatever tracing/post-hook correctness `#2513` needed
- but ensure the downstream EOF signal is not delayed behind unnecessary post-stream completion work

If that diagnosis is wrong for current `main`, please feel free to close this issue.

## Context

This issue was assembled by an autonomous coding agent after:

- reproducing the bug against the fake upstream
- comparing prerelease1 vs prerelease2 behavior
- reading issue #2511
- reading PR #2513 discussion and commits
- tracing the relevant cleanup code in `router.go`, `streamreader.go`, and tracing middleware

yanhao98:
## Summary

This PR adjusts the streaming cleanup order in `transports/bifrost-http/integrations/router.go` so the SSE reader signals EOF to the client before trace flushing runs, while still keeping cleanup steps separated for panic isolation.

This patch was produced from an **agent-driven investigation** of a reported regression in `v1.5.0-prerelease2` where OpenAI Responses-compatible streams emitted `response.completed` but did not terminate promptly downstream.

If this direction is not appropriate for current `main`, please feel free to close the PR.

## What changed

- keep cleanup steps as separate defers
- run `reader.Done()` first during goroutine teardown
- run `traceCompleter()` after EOF signaling
- release pooled `HTTPRequest` state last

## Why

Issue #2606 documents the regression analysis and reproduction matrix:

- direct fake upstream: exits normally
- official `v1.5.0-prerelease1`: exits normally
- official `v1.5.0-prerelease2`: hangs until timeout after terminal stream event

The analysis traced the regression back to the cleanup ordering introduced in `#2513` for the tagged prerelease2 behavior.

Current `main` has already diverged somewhat from the exact prerelease2 tracing path, but this patch keeps cleanup units separate and narrowly focuses on restoring prompt downstream EOF.

## Validation

Agent-run validation completed during analysis:

- built a clean `transports/v1.5.0-prerelease2` snapshot with the equivalent EOF-first cleanup patch via `transports/Dockerfile.local`
- reproduced the same fake `/openai/responses` scenario
- observed patched prerelease2 exit normally (`~0.18s`, `returncode 0`) instead of timing out (`20.00s`) after `response.completed`

## Notes

This is intentionally a minimal transport-layer change. If maintainers prefer a different approach for current `main`, this PR can be closed in favor of that direction.

Closes #2606

### DIFF
--- a/transports/bifrost-http/integrations/router.go
+++ b/transports/bifrost-http/integrations/router.go
@@ -2291,8 +2291,8 @@ func (g *GenericRouter) handleStreaming(ctx *fasthttp.RequestCtx, bifrostCtx *sc
 
 	// Producer goroutine: processes the stream channel, formats events, sends to reader
 	go func() {
-		// Separate defers ensure each cleanup runs even if an earlier one panics (LIFO order)
-		defer reader.Done()
+		// Keep cleanups separate so one panic does not skip the others (LIFO order).
+		// Signal EOF to the client first, then complete trace flushing, then release pooled request state.
 		defer schemas.ReleaseHTTPRequest(httpReq)
 		defer func() {
 			// Complete the trace after streaming finishes
@@ -2301,6 +2301,7 @@ func (g *GenericRouter) handleStreaming(ctx *fasthttp.RequestCtx, bifrostCtx *sc
 				traceCompleter()
 			}
 		}()
+		defer reader.Done()
 
 		// Create encoder for AWS Event Stream if needed
 		var eventStreamEncoder *eventstream.Encoder


yanhao98:
Closing note: this PR was pushed too late because the agent stayed in investigation mode longer than expected while reconciling the  panic-safety intent with the prerelease2 EOF regression reproduction. By the time the branch was pushed, maintainers had already landed the proper upstream fix and the duplicate issue was updated. Sorry for the noise — I’m leaving this note only to explain the delay in submitting the agent-generated patch.

Issue #974, opened 2025-11-30
[Bug]: Reasoning tokens are not passed through for Gemini


skavans:
### Prerequisites

- [x] I have searched existing issues and discussions to avoid duplicates
- [x] I am using the latest version (or have tested against main/nightly)

### Description

When **streaming** the inference, the bifrost gateway seems to not properly pass the `delta.reasoning` tokens through.

### Steps to reproduce

If I request the provider directly, I see the following:
```bash
~ % curl -X POST https://provider/v1/chat/completions \
-H "Content-Type: application/json" \
-H "Authorization: Bearer token-here" \
-d '{
  "model": "gemini-3-pro-preview",
  "stream": true,
  "messages": [
    {"role": "user", "content": "tell me fun 10 word story"}
  ]
}'
data: {"id":"chatcmpl-gemini-1764531905290","object":"chat.completion.chunk","created":1764531905,"model":"gemini-3-pro-preview","choices":[{"index":0,"delta":{"content":"","reasoning":"**Exploring Word Count Limits**\n\nI'm currently dwelling on the constraint of a strict 10-word limit for this \"fun\" story. My mind is cycling through potential themes, primarily animals like cats, dogs, and penguins, and also considering food as a possible subject.\n\n\n","reasoning_details":[]},"finish_reason":null}]}

data: {"id":"chatcmpl-gemini-1764531908469","object":"chat.completion.chunk","created":1764531908,"model":"gemini-3-pro-preview","choices":[{"index":0,"delta":{"content":"","reasoning":"**Analyzing Story Elements**\n\nI've been analyzing the recent drafts, and I'm leaning toward the \"Taco Mars\" story for its quirky, amusing concept. Its whimsical nature and simplicity truly resonate. Comparing the \"hat-wearing cat\" with the \"Mars tacos\" narrative, the latter wins.\n\n\n","reasoning_details":[]},"finish_reason":null}]}

[...]

PR #2630, opened 2026-04-10
fix(gemini): return choice with empty content for thinking-only responses

grahama1970:
## Summary
- When Gemini 2.5 models consume all output tokens on reasoning/thinking (e.g. low `max_tokens`), the response candidate has content parts that are all `Thought: true` with no regular text
- The choice-building code was inside the `if candidate.Content != nil && len(candidate.Content.Parts) > 0` block, so when all parts were thinking tokens, `contentBlocks` stayed empty but a choice was still built — however when `candidate.Content` is nil (no parts at all), `choices` was nil in the response
- Callers receiving `"choices": null` instead of an empty choice break

## Fix
Move the choice-building block outside the content-processing conditional. A valid choice is always created for a non-error candidate, with empty content but correct `finish_reason` and `reasoning_details`.

## Relates to
- #974 (streaming reasoning tokens fix — same class of bug, non-streaming path)

## Test
```bash
curl -s http://localhost:8080/v1/chat/completions \
  -d '{"model": "gemini/gemini-2.5-flash", "messages": [{"role": "user", "content": "Say hi in 3 words"}], "max_tokens": 20}'
```

Before: `"choices": null`
After: `"choices": [{"index": 0, "finish_reason": "length", "message": {"role": "assistant", "content": ""}}]`

### DIFF
--- a/core/providers/gemini/chat.go
+++ b/core/providers/gemini/chat.go
@@ -227,47 +227,48 @@ func (response *GenerateContentResponse) ToBifrostChatResponse() *schemas.Bifros
 				})
 			}
 		}
-
-		// Build the choice with message
-		message := &schemas.ChatMessage{
-			Role: schemas.ChatMessageRoleAssistant,
-		}
-
-		if len(contentBlocks) == 1 && contentBlocks[0].Type == schemas.ChatContentBlockTypeText {
-			contentStr = contentBlocks[0].Text
-			contentBlocks = nil
-		}
-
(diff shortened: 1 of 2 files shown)


akshaydeo:
Hi @grahama1970 — thanks for the contribution! This PR is currently blocked because our CLA bot shows the Contributor License Agreement as not yet signed. Could you sign it here so we can move this forward: https://cla-assistant.io/maximhq/bifrost?pullRequest=2630

PR #2653, opened 2026-04-11
feat(governance): make virtual key prefix configurable + allow vllm as custom base provider type

marioiseli89:
## Summary

Two small, surgically-scoped changes that together unlock real brownfield migration from existing LLM gateways (LiteLLM, etc.) into Bifrost — a use case the README explicitly markets ("drop-in replacement for any OpenAI-compatible application") but which today is blocked by two hardcoded constraints in the source.

1. **`feat(governance): make virtual key prefix configurable`**
   Promotes `governance.VirtualKeyPrefix` from a hardcoded `const = "sk-bf-"` to a `var` (defaulting to the same value), exposes a `SetVirtualKeyPrefix()` setter, and adds an opt-in `governance.virtual_key_prefix` config field that lets operators override or disable the prefix entirely. **Default behavior is preserved**, so existing deployments are unaffected. With the field set to `""`, pre-existing master keys (e.g. from a LiteLLM `general_settings.master_key`) can be reused verbatim without renaming every downstream consumer.

2. **`feat(custom-providers): allow VLLM as a custom base provider type`**
   Adds `VLLM` to the `SupportedBaseProviders` allowlist. The `vllm.NewVLLMProvider` constructor is **already invoked** from `createBaseProvider` for custom providers (`core/bifrost.go:3677`), and the per-key `VLLMKeyConfig` field already lives on the unified `Key` struct (`core/schemas/account.go:27`) — but the IsSupportedBaseProvider gate at `core/bifrost.go:3626` rejects `vllm` as a base type before `createBaseProvider` is reached, so the existing plumbing is unreachable. This is a one-line addition that opens up a previously gated configuration shape.

## Why this matters: the "drop-in replacement" claim and brownfield reality

Bifrost's primary marketing position is that it is a drop-in replacement for upstream gateways like LiteLLM. In practice, brownfield migrations from a running gateway hit two walls before a single client request can be served:

**Wall 1: Hardcoded virtual key prefix.** Any operator with an existing master key that doesn't begin with `sk-bf-` (the historical Bifrost convention) cannot reuse it. The config-load path silently rewrites the value (`transports/bifrost-http/lib/config.go:1259-1263`), and even if the value is forced into the SQLite store directly, the request parser at `plugins/governance/utils.go:18-41` never recognizes it on the wire because every header path (`x-bf-vk`, `Authorization: Bearer`, `x-api-key`, `x-goog-api-key`) explicitly checks `strings.HasPrefix(value, VirtualKeyPrefix)`. The constant is referenced from ~25 sites across the governance plugin and HTTP transport. Operators with many internal apps, notebooks, CI systems, and customer-facing integrations pinned to a specific master key cannot migrate without touching every caller.

**Wall 2: Custom-named providers cannot front local vLLM fleets.** Operators who want to brand their local inference under a stable provider name (e.g. their own org slug, like `my-org`) and take advantage of the per-key `vllm_key_config.url` feature that lets one provider key out to many separate vLLM upstreams on different ports — currently can't. The native `vllm` provider name works, but locks the routing prefix to the literal string `vllm`. A custom provider with `base_provider_type: "vllm"` is the documented path, and every other piece of plumbing for that combination already exists, but the request fails at startup with `unsupported base provider type: vllm` because `SupportedBaseProviders` omits it. With OpenAI as the only OpenAI-compatible base provider type currently allowed, custom providers are limited to a single `base_url` — so multi-instance local inference deployments must split across many same-shape custom providers (one per backend port) which still doesn't preserve a single brownfield model namespace.

Together, these two fixes are the minimum required to take a real brownfield deployment from a previous gateway and stand up Bifrost in front of it without renaming a single existing master key OR a single existing model identifier the apps already call.

### Commit 1 — `feat(governance): make virtual key prefix configurable`

- `framework/configstore/clientconfig.go` — Add `VirtualKeyPrefix *string` to `GovernanceConfig`. Documented as read-once-at-startup, immutable thereafter.
- `plugins/governance/main.go` — Promote `VirtualKeyPrefix` from `const` to `var` (defaulting to the new `DefaultVirtualKeyPrefix = "sk-bf-"` const for backward compat). Add the public `SetVirtualKeyPrefix(prefix string)` setter, with godoc explaining the contract: set once during config load, immutable after.
- `transports/bifrost-http/lib/config.go` — In `mergeGovernanceConfig`, apply the override **before** the virtual-key seed loop runs the prefix check, so the validation/generation paths honor the operator's custom prefix instead of silently rewriting their seed values.
- `transports/bifrost-http/handlers/wsresponses.go` — Replace 5 hardcoded `"sk-bf-"` string literals with `governance.VirtualKeyPrefix` so the WebSocket Responses handler also honors the override. (Drive-by cleanup: those literals were a latent inconsistency with the rest of the tree, which already uses the constant.)
- `transports/config.schema.json` — Document the new field.

### Commit 2 — `feat(custom-providers): allow VLLM as a custom base provider type`

- `core/schemas/bifrost.go` — Add `VLLM` to `SupportedBaseProviders`. One line.

## Backward compatibility

Both changes are 100% backward compatible:

- The `VirtualKeyPrefix` package variable defaults to `"sk-bf-"`, identical to the previous const value. Existing configs with no `governance.virtual_key_prefix` field behave identically to before. Existing virtual keys in the SQLite store continue to work without any change.
- Adding `VLLM` to `SupportedBaseProviders` only opens a previously gated shape; no existing configuration starts behaving differently.

## Notes on design choices

- **Mutable package variable vs. plugin-context threading:** the prefix is read from many call sites in `transports/bifrost-http/lib/ctx.go`, `plugins/governance/utils.go`, and the wsresponses handler. Threading it through plugin context would touch a much larger surface area without functional benefit for a single-process gateway, so this PR uses the smaller `var` + setter approach. The setter is documented as "set once at startup, immutable after" — happy to refactor to a context-passed approach if the maintainers prefer.
- **Why both patches in one PR:** the two fixes are independent and could land separately, but they target the same end-user goal (brownfield migration) and together unlock a class of deployment that neither alone enables. Splitting into two PRs felt like artificial fragmentation. Happy to split if preferred.

## Out-of-scope (mentioned for awareness)

While testing this against a real deployment I noticed two unrelated latent issues that I deliberately did **not** touch in this PR to keep scope tight:

1. `key.Models` allowlist with `["*"]` does not behave as a wildcard. The router does a literal `slices.Contains(key.Models, model)` (`core/bifrost.go:6272`), so `["*"]` only matches the literal string `"*"`. The docs example for OpenRouter / custom providers shows `"models": ["*"]` as the canonical wildcard idiom — either the docs should be updated to recommend `"models": []` (which IS treated as "all models allowed" via the `len == 0` branch on the same line), or the router should special-case `"*"`. Filing as a separate issue.
2. `vllm` keys' `vllm_key_config.model_name == model` deployment check (`core/bifrost.go:6296-6300`) is a strict equality, which prevents `key.Models[]` from acting as an alias mechanism for vllm-backed providers. This means `/v1/models` advertises models that the actual router cannot route to, when the operator configures aliases via `models[]`. Worth a separate look.

## Testing

I built the patched binary locally via the existing `transports/Dockerfile.local` workspace flow and validated end-to-end against a real deployment with multiple local vLLM upstreams (Qwen, Gemma, Ministral, embeddings) on different ports plus OpenRouter and an OpenAI-compatible Swiss provider. Specifically verified:

- ✅ A pre-existing non-`sk-bf-` master key from the previous gateway works as `Authorization: Bearer <key>` against `/v1/chat/completions` after setting `governance.virtual_key_prefix: ""`.
- ✅ A custom provider literally named after the operator's org slug, with `base_provider_type: "vllm"` and 5 keys each carrying their own `vllm_key_config.url`, routes correctly to 4 different localhost vLLM upstreams (the 5th was an unrelated upstream-side issue with the served model name containing a slash).
- ✅ `RegisterKnownProvider` is called for the custom name as expected, so `ParseModelString` correctly splits requests with the custom prefix.
- ✅ Existing default behavior (no `virtual_key_prefix` in config) continues to auto-rewrite values not starting with `sk-bf-` exactly as before.
- ✅ OpenRouter and a custom-base-`openai` provider continue to work unchanged.

## Test plan for reviewers

- [ ] Existing deployments with no `governance.virtual_key_prefix` field still load and route as before.
- [ ] A config with `governance.virtual_key_prefix: ""` and a virtual key whose `value` is `legacy-master-key-2024` is accepted at load time without warning, and the same string works as a Bearer token in subsequent requests.
- [ ] A config with `governance.virtual_key_prefix: "sk-acme-"` accepts virtual keys like `sk-acme-prod-1` and rejects (with the existing warning) keys that don't match.
- [ ] A custom provider with `base_provider_type: "vllm"` and a single key with `vllm_key_config.url` and `vllm_key_config.model_name` loads and routes a chat completion to that upstream.
- [ ] A second custom provider with the same `base_provider_type: "vllm"` but a different name and a different upstream URL coexists with the first.


akshaydeo:
hi taking over the PR to fix some low hangings and open it up under feature flags

akshaydeo:
Hi @marioiseli89 — thanks for the contribution! This PR is currently blocked because our CLA bot shows the Contributor License Agreement as not yet signed. Could you sign it here so we can move this forward: https://cla-assistant.io/maximhq/bifrost?pullRequest=2653

PR #2807, opened 2026-04-17
fix(mcp): per_user_oauth clients now send Authorization header to backend

### LINKED ISSUE #2806: [Bug]: per_user_oauth MCP clients do not send auth header to backend server
### Prerequisites

### Description

When using per-user OAuth for MCP (auth_type: per_user_oauth), Bifrost authenticates the user and acquires valid upstream tokens, but does not send the necessary Authorization header when proxying/tooling to backend servers.

### Steps to reproduce

1. Configure an MCP client with auth_type: "per_user_oauth" via the Bifrost Web UI or API.
2. Complete user consent, connecting an upstream service (e.g., Notion, GitHub, etc.).
3. Make a tool/endpoint call requiring user OAuth downstream.
4. Observe (via debug logs or network inspection) that the backend server receives no Authorization header.
5. Confirm that calls with auth_type: "oauth" (not per_user_oauth) work, and calls that require headers fail in per_user_oauth mode.

### Expected behavior

When the user has authenticated and valid per-user token is present, Bifrost should inject the Authorization: Bearer <token> header into all downstream MCP HTTP requests for per_user_oauth clients, identical to regular oauth.

### Actual behavior

No Authorization header is present in requests to backend MCP servers. OAuth-protected endpoints always fail, regardless of token state.

### Affected area(s)

Transports (HTTP)

### Version

v1.5.0-prerelease3

### Environment

```text
Kubernetes
```

### Relevant logs/output

```shell
No stack trace (silent bug). To verify, enable debug logs and observe constructed HTTP requests for MCP connections with per_user_oauth.
```

### Regression?

First reproducible in builds/shas that support MCP per_user_oauth. Not a regression for legacy oauth/headers.

### Severity

bobo-xxx:
per_user_oauth was listed in the JSON schema as a valid auth_type but was missing from the Go MCPAuthType enum and the HttpHeaders() switch. It fell through to the default case, which only used static headers and never fetched the OAuth token via oauth2Provider.GetAccessToken().

Added MCPAuthTypePerUserOauth = "per_user_oauth" and merged its handling into the existing MCPAuthTypeOauth case so both auth types fetch and attach the Bearer token the same way.

Fixes #2806

### DIFF
--- a/core/schemas/mcp.go
+++ b/core/schemas/mcp.go
@@ -68,9 +68,10 @@ const (
 type MCPAuthType string
 
 const (
-	MCPAuthTypeNone    MCPAuthType = "none"    // No authentication
-	MCPAuthTypeHeaders MCPAuthType = "headers" // Header-based authentication (API keys, etc.)
-	MCPAuthTypeOauth   MCPAuthType = "oauth"   // OAuth 2.0 authentication
+	MCPAuthTypeNone         MCPAuthType = "none"            // No authentication
+	MCPAuthTypeHeaders      MCPAuthType = "headers"         // Header-based authentication (API keys, etc.)
+	MCPAuthTypeOauth        MCPAuthType = "oauth"           // OAuth 2.0 authentication
+	MCPAuthTypePerUserOauth MCPAuthType = "per_user_oauth" // Per-user OAuth 2.1 authentication
 )
 
 // MCPClientConfig defines tool filtering for an MCP client.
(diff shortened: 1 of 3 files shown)


Pratham-Mishra04 (dismissed):

akshaydeo:
This is part of prerelease - yet to merge in main

lehmanju:
Are you sure this works? It looks like it just uses a static oauth config id for the respective mcp backend server. However, it should dynamically choose a different config id based on user identity, or am i missing somethin?

lehmanju:
Also this seems like an openclaw PR ...

bobo-xxx:
Addressed the review comments:

1. **greptile comment**: Updated `framework/configstore/tables/mcp.go` line 31 - Added `"per_user_oauth"` to the AuthType column comment.

2. **coderabbit comment**: Updated `transports/bifrost-http/handlers/mcp.go` line 319 - Changed `if req.AuthType == "oauth"` to `if req.AuthType == "oauth" || req.AuthType == "per_user_oauth"` so the OAuth validation gate also applies to `per_user_oauth` auth type.

3. **copilot unit test request**: The unit test for `HttpHeaders()` with `per_user_oauth` would require adding a test case that mocks the `oauth2Provider.GetAccessToken()` call. Could you clarify if this is blocking the merge, or if the existing approval from Pratham-Mishra04 is sufficient?

The PR head branch (`bobo-xxx:clawoss/fix/per-user-oauth-headers`) has been force-updated with these changes.

bobo-xxx:
The comment has been updated in this PR — see the changes to `framework/configstore/tables/mcp.go` line 31 and `core/schemas/mcp.go` line 74, both now include `per_user_oauth`. The PR is ready for merge.

crazysailor1000:
Hi, can you merge this fix? We are trying to configure some MCP servers and are stuck because of this bug

PR #2865, opened 2026-04-20
bedrock: nil-guard key.BedrockKeyConfig in completeRequest to stop cross-region crash loop

### LINKED ISSUE #2855: [Bug]: Nil pointer panic in bedrock.go:194 when using cross-region inference profile IDs (us.*) in provider key model list
## Description

When a cross-region inference profile model ID (prefixed with `us.`) is included in the provider key's `models` array in Helm values, the Bedrock provider panics with a nil pointer dereference on every inference request, crashing the pod in a restart loop.

## Version

- Bifrost image: `maximhq/bifrost:v1.5.0-prerelease3`
- Bifrost core: `v1.5.2`
- Helm chart: `1.7.0`

## Steps to Reproduce

1. Configure a Bedrock provider key with `us.*` inference profile model IDs in `values.yaml`:
```yaml
bifrost:
  providers:
    bedrock:
      keys:
        - name: "irsa-key"
          value: ""
          models:
            - "us.anthropic.claude-sonnet-4-6"
            - "us.anthropic.claude-opus-4-6-v1"
          bedrock_key_config:
            region: "us-east-1"
            access_key: ""
            secret_key: ""
```
2. Send any inference request using one of those model IDs
3. Pod crashes immediately

## Panic Output

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x58 pc=0x1312250]

goroutine 959 [running]:
github.com/maximhq/bifrost/core/providers/bedrock.(*BedrockProvider).completeRequest(_, _, {_, _, _}, {_, _}, {{0x39bed82, 0xf}, {0x0, ...}, ...})
	github.com/maximhq/bifrost/core@v1.5.2/providers/bedrock/bedrock.go:194 +0xb0
github.com/maximhq/bifrost/core/providers/bedrock.(*BedrockProvider).ChatCompletion(_, _, {{0x39bed82, 0xf}, {0x0, 0x0}, {{0x26885a6e780, 0x2e}, {0x0, 0x0}, ...}, ...}, ...)
	github.com/maximhq/bifrost/core@v1.5.2/providers/bedrock/bedrock.go:1057 +0x1e5
```

## Notes

- Model IDs that **cause the panic**: `us.anthropic.claude-sonnet-4-6`, `us.anthropic.claude-opus-4-6-v1`
- Model IDs that **work fine**: `us.anthropic.claude-sonnet-4-5-20250929-v1:0`, `us.anthropic.claude-haiku-4-5-20251001-v1:0`
- The panic only occurs when the model ID is in the provider key's model list AND used in a request
- Using a model ID not in the key list returns a clean "no keys found" error (no crash)
- Deployment uses IRSA (no static credentials) on EKS

## Workaround

Use `us.anthropic.claude-sonnet-4-5-20250929-v1:0` instead of `us.anthropic.claude-sonnet-4-6` until fixed.

## Expected Behavior

Requests using cross-region inference profile IDs should either work correctly or return a clean error — not panic and crash the pod.

SAY-5:
Fixes #2855.

`completeRequest` pulled the Bedrock config off the key and immediately dereferenced it:

```go
config := key.BedrockKeyConfig
region := DefaultBedrockRegion
if config.Region != nil && config.Region.GetValue() != "" { ... }
```

and later:

```go
signAWSRequest(ctx, req, config.AccessKey, config.SecretKey, ...)
```

`ensureBedrockKeyConfig` is the function that seeds a zero `BedrockKeyConfig` when the key was set up by API-key alone, but not every code path that feeds into `completeRequest` runs through it. Cross-region inference profile model IDs (for example `us.anthropic.claude-sonnet-4-6`) can resolve to a key that was never seeded; `completeRequest` then segfaults on the first inference request and crashes the whole pod in a restart loop.

Two minimal guards:

1. In the region lookup, check `config != nil` before reading `config.Region`. Fall back to `DefaultBedrockRegion` otherwise.
2. Before calling `signAWSRequest` (which reads `AccessKey` / `SecretKey` / `SessionToken` / `RoleARN` off `config`), return a clear configuration error when `config` is nil and there is no API key either. This is only reachable when a Bedrock key was registered without any usable authentication; surfacing it as an error is much better than crashing the process.

The API-key path (`key.Value.GetValue() != ""`) is unaffected.

### DIFF
--- a/core/providers/bedrock/bedrock.go
+++ b/core/providers/bedrock/bedrock.go
@@ -193,8 +193,15 @@ var retryableBedrockExceptions = map[string]int{
 func (provider *BedrockProvider) completeRequest(ctx *schemas.BifrostContext, jsonData []byte, path string, key schemas.Key) ([]byte, time.Duration, map[string]string, *schemas.BifrostError) {
 	config := key.BedrockKeyConfig
 
+	// config can be nil here for keys that were routed in through a
+	// path that skipped ensureBedrockKeyConfig (e.g. cross-region
+	// inference profile model IDs picking a key that was never seeded
+	// with BedrockKeyConfig, #2855). Previously this dereferenced a nil
+	// pointer on line 197 and segfaulted the whole pod in a restart
+	// loop; fall back to the default region and skip the SigV4 path
+	// when there is nothing to sign with.
 	region := DefaultBedrockRegion
-	if config.Region != nil && config.Region.GetValue() != "" {
+	if config != nil && config.Region != nil && config.Region.GetValue() != "" {
 		region = config.Region.GetValue()
 	}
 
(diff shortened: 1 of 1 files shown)


Pratham-Mishra04:
Hey @SAY-5 this is already fixed, closing this PR

PR #2871, opened 2026-04-20
bedrock: align BedrockGuardrailTrace with the AWS Converse API schema

### LINKED ISSUE #2772: [Bug]: BedrockGuardrailTrace struct fields don't match AWS API: wrong names and types cause sonic.Unmarshal failure
## Summary

The `BedrockGuardrailTrace` struct in `core/providers/bedrock/types.go` has two field definitions that don't match the actual AWS Bedrock Converse API response, causing `sonic.Unmarshal` to fail with **"failed to parse bedrock response"** whenever a guardrail passes on a request (i.e. any non-blocked request when `trace` is enabled).

## Affected version

`transports/v1.4.20` (latest)

## Root cause

The [`GuardrailTraceAssessment`](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_GuardrailTraceAssessment.html) type in the AWS Bedrock Runtime API defines:

| Field | Official type |
|---|---|
| `inputAssessment` | `map<string, GuardrailAssessment>` — **singular**, keyed by guardrail ID |
| `outputAssessments` | `map<string, GuardrailAssessment[]>` — **plural**, keyed by guardrail ID, value is an **array** |

The current Bifrost struct defines both fields incorrectly:

```go
// core/providers/bedrock/types.go
type BedrockGuardrailTrace struct {
    Action            *string                      `json:"action,omitempty"`
    InputAssessments  []BedrockGuardrailAssessment `json:"inputAssessments,omitempty"`  // wrong name ("inputAssessments" vs "inputAssessment") AND wrong type (slice vs map)
    OutputAssessments []BedrockGuardrailAssessment `json:"outputAssessments,omitempty"` // wrong type (flat slice vs map[string][]...)
    Trace             *BedrockGuardrailTraceDetail `json:"trace,omitempty"`
}
```

The `InputAssessments` field has a wrong JSON tag name (`inputAssessments` instead of `inputAssessment`), so it is silently ignored during unmarshal. The `OutputAssessments` field has the correct JSON tag name but the wrong Go type: `[]BedrockGuardrailAssessment` (slice) instead of `map[string][]BedrockGuardrailAssessment` (map of slices). When Bedrock returns `outputAssessments` with actual data, `sonic.Unmarshal` fails with a type mismatch:

```
Mismatch type []bedrock.BedrockGuardrailAssessment with value object
"at index 885: mismatched type with value\n\n\tputAssessments\":
{\"<guardrail-id>\":\n\t................^...............\n"
```

## When it triggers

The crash only occurs when the guardrail **passes** (benign requests where `stopReason` is `end_turn`, `max_tokens`, etc.), because Bedrock includes a non-empty `outputAssessments` map in those responses. When the guardrail **blocks** on input (e.g. prompt injection), Bedrock does not include `outputAssessments` in the response, so the unmarshal happens to succeed.

This means enabling `trace` in `guardrailConfig` causes **all non-blocked requests to return 500** — the opposite of the intended behaviour.

## Actual Bedrock response structure (verified)

```json
{
  "trace": {
    "guardrail": {
      "actionReason": "No action.",
      "inputAssessment": {
        "<guardrail-id>": {
          "invocationMetrics": { "guardrailProcessingLatency": 411, ... }
        }
      },
      "outputAssessments": {
        "<guardrail-id>": [
          { "invocationMetrics": { "guardrailProcessingLatency": 148, ... } }
        ]
      }
    }
  }
}
```

Source: [AWS GuardrailTraceAssessment API reference](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_GuardrailTraceAssessment.html)

## Proposed fix

```go
type BedrockGuardrailTrace struct {
    Action            *string                                 `json:"action,omitempty"`
    InputAssessment   map[string]BedrockGuardrailAssessment   `json:"inputAssessment,omitempty"`    // singular, map
    OutputAssessments map[string][]BedrockGuardrailAssessment `json:"outputAssessments,omitempty"`  // map of slices
    Trace             *BedrockGuardrailTraceDetail            `json:"trace,omitempty"`
}
```

## Related

A similar object-vs-array type mismatch was previously fixed for `AnthropicContentBlock.Citations` in #1235.

SAY-5:
## Summary

Aligns `BedrockGuardrailTrace` in `core/providers/bedrock/types.go` with the actual AWS Bedrock Runtime `GuardrailTraceAssessment` schema, so that `sonic.Unmarshal` stops failing on non-blocked requests when `trace` is enabled on the guardrail.

Fixes #2772.

## Root cause

`GuardrailTraceAssessment` from the AWS Runtime API:

| Field | Official shape |
| --- | --- |
| `inputAssessment` | `map<string, GuardrailAssessment>`, **singular**, keyed by guardrail ID |
| `outputAssessments` | `map<string, GuardrailAssessment[]>`, **plural**, keyed by guardrail ID, values are arrays |

Bifrost's struct declared both as flat `[]BedrockGuardrailAssessment`, and the input field used the plural JSON tag `inputAssessments`:

- `InputAssessment` was silently dropped because the JSON key didn't match.
- `outputAssessments` hit a type mismatch (object → slice), causing `sonic.Unmarshal` to fail with `Mismatch type []bedrock.BedrockGuardrailAssessment with value object` and the Bedrock response body to fail parsing.

This only surfaces when the guardrail **passes**, on a blocked-at-input response, Bedrock doesn't include `outputAssessments`, so the unmarshal happens to succeed.

## Change

- Rename `InputAssessments` → `InputAssessment` and retype as `map[string]BedrockGuardrailAssessment` with JSON tag `inputAssessment`.
- Retype `OutputAssessments` as `map[string][]BedrockGuardrailAssessment`. Tag (`outputAssessments`) unchanged.

## Call-site impact

Verified via `gh api` code search that `InputAssessments` and `OutputAssessments` are not referenced anywhere else in the repository, the fields only appear in `types.go`. No call-site updates are required.

## Testing

- `go build ./...` passes.
- Replaying a recorded Bedrock Converse response with `guardrailConfig.trace=enabled` and a passed guardrail: `sonic.Unmarshal` succeeds on the fixed struct, `InputAssessment[<guardrail-id>]` and `OutputAssessments[<guardrail-id>][0]` populate correctly.

### DIFF
--- a/core/providers/bedrock/guardrail_trace_test.go
+++ b/core/providers/bedrock/guardrail_trace_test.go
@@ -73,7 +73,7 @@ func TestBedrockGuardrailTrace_DecodesAWSConverseShape(t *testing.T) {
 }
 
 // TestBedrockGuardrailTrace_DecodesEmptyTrace covers the non-blocked path
-// described in #2772 — a request with `trace = "enabled"` that does not
+// described in #2772, a request with `trace = "enabled"` that does not
 // trip any policy used to fail to decode because the previous schema
 // required slice fields. The empty payload here is what the runtime emits
 // in that scenario.
--- a/core/providers/bedrock/guardrail_trace_test.go
+++ b/core/providers/bedrock/guardrail_trace_test.go
@@ -1,0 +1,87 @@
+package bedrock_test
+
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/maximhq/bifrost/core/providers/bedrock"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// TestBedrockGuardrailTrace_DecodesAWSConverseShape verifies that the
+// BedrockGuardrailTrace struct deserialises a payload that matches the
+// AWS Bedrock Runtime GuardrailTraceAssessment schema:
+//
+//	https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_GuardrailTraceAssessment.html
+//
+// In particular it pins:
+//   - inputAssessment is a map keyed by guardrail ID (singular, not a slice);
+//   - outputAssessments is a map whose values are arrays of assessments;
+//   - actionReason and modelOutput are surfaced (#2772).
+//
+// The payload below mirrors the shape AWS returns when a Converse request
+// runs with a guardrail attached and `trace = "enabled"`. Before this
+// schema fix, sonic.Unmarshal failed on this exact response shape because
+// inputAssessment was modelled as a []BedrockGuardrailAssessment slice.
+func TestBedrockGuardrailTrace_DecodesAWSConverseShape(t *testing.T) {
+	payload := []byte(`{
+        "action": "GUARDRAIL_INTERVENED",
+        "actionReason": "Guardrail blocked the request",
+        "modelOutput": ["redacted"],
+        "inputAssessment": {
+            "abc123": {
+                "topicPolicy": {
+                    "topics": [
+                        {"name": "Investment Advice", "type": "DENY", "action": "BLOCKED"}
+                    ]
+                }
+            }
+        },
+        "outputAssessments": {
+            "abc123": [
+                {
+                    "contentPolicy": {
+                        "filters": [
+                            {"type": "VIOLENCE", "confidence": "HIGH", "action": "BLOCKED"}
+                        ]
+                    }
+                }
+            ]
+        }
+    }`)
+
+	var trace bedrock.BedrockGuardrailTrace
+	require.NoError(t, json.Unmarshal(payload, &trace), "payload matching AWS schema must deserialise")
+
+	require.NotNil(t, trace.Action)
+	assert.Equal(t, "GUARDRAIL_INTERVENED", *trace.Action)
+
+	require.NotNil(t, trace.ActionReason)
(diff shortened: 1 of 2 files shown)


Pratham-Mishra04:
Hey @SAY-5 can you add a test case for this in llmtests?

SAY-5:
Hey @Pratham-Mishra04, thanks for the review. Pushed two unit tests covering the schema fix in 2871 (commit f6905ef7):

- `TestBedrockGuardrailTrace_DecodesAWSConverseShape` exercises a fully-populated payload (action / actionReason / modelOutput / `inputAssessment` as map / `outputAssessments` as map-of-arrays). This is the exact shape the previous slice-typed fields rejected.
- `TestBedrockGuardrailTrace_DecodesEmptyTrace` covers the non-blocked path from #2772 where the runtime returns an empty guardrail trace.

I put them next to the type in `core/providers/bedrock/` rather than in `core/internal/llmtests` because the integration coverage there needs a live Bedrock guardrail to fire (real AWS creds and a configured guardrail), which isn't reachable from CI. Happy to also add an llmtests entry skipped behind `Available: false` if you'd prefer that — let me know.

`go test -run TestBedrockGuardrailTrace ./core/providers/bedrock/... -v` → 2/2 passing.

SAY-5:
Acknowledged on the `llmtests` request — will land a test that exercises the Bedrock guardrail-trace schema mismatch and reply here once pushed.

PR #2895, opened 2026-04-21
feat(openrouter): support model-variant suffixes

pratikbin:
## Summary

OpenRouter supports model-variant suffixes (`:nitro`, `:floor`, `:online`, `:thinking`, `:exacto`, `:extended`, `:free`) as server-side routing hints. Bifrost was dropping requests with these suffixes at the key-selection layer because `WhiteList`/`BlackList` use exact-match and users typically store only the base model (e.g. `openai/gpt-5.2`) in `key.Models`. Request for `openai/gpt-5.2:nitro` then failed with `no keys found that support model: openai/gpt-5.2:nitro` even though the user intended the base model to be allowed.

This change makes variants transparent for allow-list / blacklist / alias / pricing / capability lookups when the provider is OpenRouter, while keeping exact matches authoritative (backward compatible).

Refs: https://openrouter.ai/docs/guides/routing/model-variants

## Changes

- **`core/schemas/utils.go`** — `SplitOpenRouterVariant(model) (base, variant)` + `BaseOpenRouterModel(model)` helpers. Variant lower-cased on return for consistent downstream identifiers. Unknown suffixes and non-OpenRouter patterns (e.g. Bedrock `model-v1:0`) pass through untouched.
- **`core/bifrost.go`** — in `selectKeyFromProviderForModelWithPool`, when `baseProviderType == schemas.OpenRouter`, fall back to the variant-stripped base when evaluating `key.Models.IsAllowed` / `key.BlacklistedModels.IsBlocked`. Exact match wins first; fallback is additive.
- **`core/schemas/account.go`** — `KeyAliases.Resolve` falls back to base-model alias lookup and re-appends the variant suffix to the resolved target. Provider-agnostic but no-op for non-variant strings, so safe across all providers.
- **`framework/modelcatalog/models.go`** — `getBaseModelNameUnsafe` strips variant suffix before catalog index lookup so pricing and capability entries resolve for variant requests. Applied under the existing `RLock`.
- **`core/schemas/openrouter_variant_test.go`** (new) — unit tests covering known variants, case insensitivity, Bedrock-style ids (`v1:0`), `WhiteList` interaction, and `KeyAliases.Resolve`.

### Design notes / trade-offs

- `WhiteList` / `BlackList` themselves stay provider-agnostic; variant-awareness lives at the dispatch site behind the `schemas.OpenRouter` guard. Keeps the core types clean and avoids cross-provider leakage.
- No model-catalog expansion for variants in the list-models endpoint: OpenRouter doesn't expose which variants apply per model, and blindly adding 7 rows per model would surface invalid combinations (e.g. `:thinking` on non-reasoning models). Passthrough behavior is intentional — if a variant doesn't apply, OpenRouter returns an error.
- `:online` variant is deprecated by OpenRouter (superseded by the `openrouter:web_search` server tool). Included in the variant set so existing configs keep working.
- The separate pre-existing issue where users store `key.Models` with provider prefix (`openrouter/openai/gpt-5.2`) is untouched; it is orthogonal to variants.

## Type of change

- [x] Bug fix
- [x] Feature

## Affected areas

- [x] Core (Go)

## How to test

```sh
go version  # go1.24+

# From repo root, set up workspace (so framework pulls local core)
make setup-workspace

# Run unit tests
cd core && go test ./schemas/... -count=1 -race
cd ../framework && go test ./modelcatalog/... -count=1 -race

# Integration: with an OpenRouter provider configured and a key whose
# Models allow-list contains "openai/gpt-5.2" (no variant), send a chat
# completion with model "openai/gpt-5.2:nitro". Request should now be
# routed to OpenRouter; server handles the :nitro sort. Repeat for each
# of :floor, :online, :thinking, :exacto, :extended, :free.
```

Expected result: 291 pass in core/schemas, 166 pass in framework/modelcatalog. No new lint/vet findings attributable to this change.

## Breaking changes

- [ ] Yes
- [x] No

Exact matches still win first; the variant fallback is purely additive. Users who explicitly allow-listed `openai/gpt-5.2:nitro` still match exactly; users who allow-listed only `openai/gpt-5.2` now additionally allow all of its variants.

## Related issues

None linked — surfaced while investigating OpenRouter variant support.

## Security considerations

No new auth, secrets, or PII paths. The change only widens an internal exact-match comparison in the ACL layer by an additional deterministic lookup (stripped base). Blacklist behavior is likewise widened: blocking a base now also blocks all of its variants, which is the stricter direction and matches user intent.

## Checklist

- [x] I verified tests pass locally (`go test ./schemas/... ./modelcatalog/...`)

### DIFF
--- a/core/bifrost.go
+++ b/core/bifrost.go
@@ -7059,7 +7059,17 @@ func (bifrost *Bifrost) selectKeyFromProviderForModelWithPool(ctx *schemas.Bifro
 			// NOTE: Model filtering uses the original requested model (which may be an alias).
 			// key.Models and key.BlacklistedModels must therefore be expressed in alias keys.
 			// The provider-specific identifier is resolved later in the handler closure via key.Aliases.Resolve(model).
-			modelSupported := hasValue && key.Models.IsAllowed(model) && !key.BlacklistedModels.IsBlocked(model)
+			// OpenRouter variant suffixes (:nitro, :floor, :online, :thinking, :exacto,
+			// :extended, :free) are routing hints, not distinct models. Fall back to the
+			// base model when evaluating allow/deny lists so users don't have to enumerate
+			// every variant in their key.Models configuration.
+			aclBase := model
+			if baseProviderType == schemas.OpenRouter {
+				aclBase = schemas.BaseOpenRouterModel(model)
+			}
+			allowedForKey := key.Models.IsAllowed(model) || (aclBase != model && key.Models.IsAllowed(aclBase))
+			blockedForKey := key.BlacklistedModels.IsBlocked(model) || (aclBase != model && key.BlacklistedModels.IsBlocked(aclBase))
+			modelSupported := hasValue && allowedForKey && !blockedForKey
 			if baseProviderType == schemas.VLLM && key.VLLMKeyConfig != nil {
 				if key.VLLMKeyConfig.ModelName != "" {
 					modelSupported = modelSupported && (key.VLLMKeyConfig.ModelName == model)
--- a/core/schemas/account.go
+++ b/core/schemas/account.go
@@ -182,13 +182,19 @@ func (ka KeyAliases) Resolve(model string) string {
 	}
 	// OpenRouter variant fallback: try to resolve the base model (:nitro/:free/etc stripped),
 	// then re-append the variant suffix so the target is routed identically to the base.
+	// If the resolved alias target already carries a variant suffix, strip it first to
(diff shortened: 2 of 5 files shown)


akshaydeo:
Hi @pratikbin — thanks for the contribution! This PR is currently blocked because our CLA bot shows the Contributor License Agreement as not yet signed. Could you sign it here so we can move this forward: https://cla-assistant.io/maximhq/bifrost?pullRequest=2895

PR #2909, opened 2026-04-22
Add sends_done_marker config for custom providers

### LINKED ISSUE #2910: [Bug]: Custom providers that don't send [DONE] SSE marker cause ~60s stream hang
### Prerequisites

### Description

Some OpenAI-compatible custom providers (e.g. Synthetic) don't send the `data: [DONE]` SSE marker in streaming responses. They close the TCP connection after sending the final chunk with `finish_reason: "stop"`.

Bifrost's OpenAI streaming handler calls `ProviderSendsDoneMarker()`, which returns `true` for all providers except Cerebras, Perplexity, and HuggingFace. Custom providers always get the default `true`, with no way to override. When the upstream doesn't send `[DONE]`, the handler keeps waiting for it, causing the stream to hang for ~60 seconds after all content has been delivered, until the idle timeout fires.

### Steps to reproduce

1. Configure a custom provider pointing to Synthetic's OpenAI-compatible endpoint:

   ```json
   "synthetic": {
     "keys": [
       {
         "key_id": "...",
         "name": "synthetic-key",
         "value": "env.SYNTHETIC_API_KEY",
         "models": ["*"]
       }
     ],
     "network_config": {
       "base_url": "https://api.synthetic.new/openai/v1",
       "default_request_timeout_in_seconds": 3600,
       "stream_idle_timeout_in_seconds": 300,
       "max_retries": 10,
       "retry_backoff_initial": 100,
       "retry_backoff_max": 10000
     },
     "custom_provider_config": {
       "base_provider_type": "openai",
       "request_path_overrides": {
         "chat_completion": "/chat/completions",
         "chat_completion_stream": "/chat/completions",
         "list_models": "/models"
       }
     }
   }
   ```

2. Send a streaming chat completion request through Bifrost:

   ```bash
   curl -sN http://localhost:8080/v1/chat/completions \
     -H "Authorization: Bearer <virtual-key>" \
     -d '{"model":"synthetic/hf:zai-org/GLM-5.1","messages":[{"role":"user","content":"Say hello"}],"max_tokens":50,"stream":true}'
   ```

3. Observe: all content chunks arrive quickly, then the stream hangs for ~60 seconds before closing

### Expected behavior

Stream closes immediately after the final chunk with `finish_reason: "stop"`, followed by `data: [DONE]`. Total time should be ~1-2 seconds, matching the upstream response time.

### Actual behavior

All content chunks arrive within ~1 second, then the stream hangs for ~60 seconds before the idle timeout fires and the connection closes. No `data: [DONE]` is sent to the client during the hang.

### Affected area(s)

Core (Go)

### Version

v1.5.0-prerelease5

### Environment

```text
- OS: Linux (NixOS, running in Docker container)
- Bifrost: v1.5.0-prerelease5 (Docker image `maximhq/bifrost:v1.5.0-prerelease5`)
- Deployed via OCI container with file-based config (`config.json`)
- Custom providers: Synthetic (`api.synthetic.new`), MiniMax (`api.minimax.io`)
```

### Relevant logs/output

```shell
# All content arrives in ~1s, then 60s silence:
data: {"id":"...","choices":[{"index":0,"delta":{"role":"assistant","content":"Hello!"}}],...}
data: {"id":"...","choices":[{"index":0,"finish_reason":"stop","delta":{}}],...}
# ... 60 seconds of silence ...
# Stream finally closes when idle timeout fires


Root cause is in `core/providers/utils/utils.go` — `ProviderSendsDoneMarker()` hardcodes the provider list with no override mechanism for custom providers:


func ProviderSendsDoneMarker(providerName schemas.ModelProvider) bool {
    switch providerName {
    case schemas.Cerebras, schemas.Perplexity, schemas.HuggingFace:
        return false
    default:
        return true  // Custom providers always get true — no way to override
    }
}
```

### Regression?

Custom providers have never had a way to override this behavior.

### Severity

Sewer56:
## Problem

fixes https://github.com/maximhq/bifrost/issues/2910

Some OpenAI-compatible providers (e.g. MiniMax, Synthetic) do not send the `[DONE]` SSE marker in streaming responses. They close the TCP connection after sending the final chunk with `finish_reason: "stop"`.

Bifrost's OpenAI streaming handler waits for `[DONE]` because `ProviderSendsDoneMarker()` returns `true` for all providers except Cerebras, Perplexity, and HuggingFace. This causes ~60s stream hangs after all content has been delivered, until the idle timeout fires.

## Solution

Add an optional `sends_done_marker` field to `CustomProviderConfig`:

```json
"custom_provider_config": {
  "base_provider_type": "openai",
  "sends_done_marker": false,
  "request_path_overrides": { ... }
}
```

When `sends_done_marker` is `false`, the streaming handler breaks on `finish_reason` instead of waiting for `[DONE]`, matching the existing behavior for Cerebras/Perplexity/HuggingFace.

### Implementation Details

- **`sends_done_marker *bool`** in `CustomProviderConfig` - `nil` uses provider default (backward compatible), `false` breaks on finish_reason, `true` waits for `[DONE]`
- **`ProviderSendsDoneMarker()`** accepts optional `*CustomProviderConfig` via variadic param - checks `SendsDoneMarker` override first, then falls back to provider default
- **`BifrostContextKeyCustomProviderConfig`** - the config is passed through `BifrostContext` to avoid changing `HandleOpenAI*Streaming` function signatures (used by 17+ providers)

## Files Changed

| File | Change |
|------|--------|
| `core/schemas/provider.go` | Add `SendsDoneMarker *bool` field to `CustomProviderConfig` |
| `core/schemas/bifrost.go` | Add `BifrostContextKeyCustomProviderConfig` context key |
| `core/providers/utils/utils.go` | Extend `ProviderSendsDoneMarker()` with optional custom config |
| `core/providers/openai/openai.go` | Set context in `TextCompletionStream`/`ChatCompletionStream`, read in streaming handlers |

## Testing

Built and deployed a custom Docker image with this patch. MiniMax streaming via Bifrost now completes in ~1.7s (previously hung for ~60s after content delivery).

Before:
```
time curl ... MiniMax-M2.7 ... stream=true
# Content arrives, then ~60s silence, then stream closes
```

After (with `"sends_done_marker": false`):
```
time curl ... MiniMax-M2.7 ... stream=true
# Content arrives, stream closes immediately with [DONE]
# Total: ~1.7s
```

### DIFF
--- a/core/providers/openai/openai.go
+++ b/core/providers/openai/openai.go
@@ -390,6 +390,8 @@ func (provider *OpenAIProvider) TextCompletionStream(ctx *schemas.BifrostContext
 	if err := providerUtils.CheckOperationAllowed(schemas.OpenAI, provider.customProviderConfig, schemas.TextCompletionStreamRequest); err != nil {
 		return nil, err
 	}
+	// Store custom provider config in context so streaming handler can check SendsDoneMarker
+	ctx.SetValue(schemas.BifrostContextKeyCustomProviderConfig, provider.customProviderConfig)
 	var authHeader map[string]string
 	if key.Value.GetValue() != "" {
 		authHeader = map[string]string{"Authorization": "Bearer " + key.Value.GetValue()}
@@ -562,6 +564,15 @@ func HandleOpenAITextCompletionStreaming(
 
 		sseReader := providerUtils.GetSSEDataReader(ctx, reader)
 
+		// Resolve sends_done_marker once before the loop (custom config overrides provider default)
+		var customCfg *schemas.CustomProviderConfig
+		if v := ctx.Value(schemas.BifrostContextKeyCustomProviderConfig); v != nil {
+			if c, ok := v.(*schemas.CustomProviderConfig); ok {
+				customCfg = c
+			}
+		}
+		providerSendsDoneMarker := providerUtils.ProviderSendsDoneMarker(providerName, customCfg)
+
 		chunkIndex := -1
 		usage := &schemas.BifrostLLMUsage{}
 
@@ -700,7 +711,7 @@ func HandleOpenAITextCompletionStreaming(
 			}
 
 			// For providers that don't send [DONE] marker break on finish_reason
-			if !providerUtils.ProviderSendsDoneMarker(providerName) && finishReason != nil {
+			if !providerSendsDoneMarker && finishReason != nil {
 				break
 			}
 		}
@@ -898,6 +909,8 @@ func (provider *OpenAIProvider) ChatCompletionStream(ctx *schemas.BifrostContext
 	if err := providerUtils.CheckOperationAllowed(schemas.OpenAI, provider.customProviderConfig, schemas.ChatCompletionStreamRequest); err != nil {
 		return nil, err
 	}
+	// Store custom provider config in context so streaming handler can check SendsDoneMarker
+	ctx.SetValue(schemas.BifrostContextKeyCustomProviderConfig, provider.customProviderConfig)
 	var authHeader map[string]string
 	if key.Value.GetValue() != "" {
 		authHeader = map[string]string{"Authorization": "Bearer " + key.Value.GetValue()}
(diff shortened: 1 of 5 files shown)


Sewer56:
Please ensure this is adequately documented, e.g. in Wiki.
I'll leave it to maintainers to pick best spot.

Pratham-Mishra04:
Hey @Sewer56 thanks for the PR. I would suggest a few improvements here - we can have a ctx for SendsDoneMarker and we can set it in core/bifrost.go only if we get a custom provider+stream request and then just check for that ctx first in providerUtils.ProviderSendsDoneMarker() using the ctx. This will implement it for all providers which support custom configs and not just openai

Sewer56:
@Pratham-Mishra04 it could be done.

I ran into the issue after spending some hours debugging a proxy I set up for friends/family. And made minimal fix to avoid having something sit in backlog forever. PR isn't perfect, but something to get the needle moving.

Quick question is: 

You mention non-OpenAI providers. But other API types, Anthropic, Gemini, Cohere etc. do not support this pattern.

Could switch to passing a `ctx` instead of provider config to resolve; but the net result would more-less be the same. Just a struct with a single member. Extensible and ideal long term pattern? Yes- but not yet a case where multiple flags need storage across multiple access points.

ylwind:
I'm hitting the exact same issue.  Looks like this just needs a rebase to resolve the conflicts.

PR #3013, opened 2026-04-24
fix: add idle timeout to upstream WS ReadMessage (#2998)

### LINKED ISSUE #2998: [Bug]: Native WebSocket upstream requests hang indefinitely when upstream accepts but sends no frames
### Prerequisites

### Description

**Summary**

When Bifrost proxies a client WebSocket to a native upstream WS provider, the read loop in `tryNativeWSUpstream` has no idle deadline. If the upstream accepts the upgrade and then stays silent, the goroutine blocks forever on `upstream.ReadMessage()` with no recovery.

**What the bug is**

* `transports/bifrost-http/handlers/wsresponses.go`, around line 308 on main. The read loop calls `upstream.ReadMessage()` with no preceding `SetReadDeadline`. A silent upstream leaves the call blocked indefinitely.
* `UpstreamConn.SetReadDeadline(t time.Time)` is already exposed at `transports/bifrost-http/websocket/connection.go:108`. The fix is handler-side only.
* `NetworkConfig.StreamIdleTimeoutInSeconds` is already defined in `core/schemas/provider.go` with `DefaultStreamIdleTimeoutInSeconds = 60` and is used for HTTP streaming. The same value can be reused for WS.

**Impact**

* Client request hangs until the client itself times out or disconnects. Bifrost surfaces no error.
* The log entry stays in `processing` and is eventually evicted silently by the 5 min TTL sweeper.
* Bifrost goroutines accumulate. Under a real stall event (upstream incident, network partition), many request goroutines pile up on blocked reads.
* Any WS-capable provider can cause this. Upstream behavior is outside Bifrost's control, so defensive timeouts are required in the gateway.
* Codex CLI and Claude Code both default to WS, so a flaky upstream window translates directly into hung user requests.

**Suggested fix**

Before each `upstream.ReadMessage()` in `tryNativeWSUpstream`, call `upstream.SetReadDeadline(time.Now().Add(idleTimeout))`. Clear with `SetReadDeadline(time.Time{})` on a successful read so subsequent reads start a fresh idle window. Source the timeout from `NetworkConfig.StreamIdleTimeoutInSeconds` (default 60s). On `net.Error.Timeout()`, finalize the log via PostHookRunner with a `BifrostError{StatusCode: 504, Code: "upstream_timeout"}`, write a 504 WS error frame to the client, discard the upstream from the pool, and return cleanly.

**Verified reproduction**

Main-built container (commit `abfd68c80`, image built 2026-04-24). A mock WS upstream accepts the upgrade and then waits forever without sending any frames. A single WS request through Bifrost hung for 121 seconds with zero bifrost log output after the initial `GET /openai/v1/responses HTTP/1.1 101` line. The client was manually killed at the 2 minute mark. The request would have continued hanging indefinitely.

### Steps to reproduce

See the Reproducible command section below.

### Reproducible command

Prerequisites: `docker`, `websocat` (on macOS `brew install websocat`), `python3` with `websockets` installed (`pip3 install websockets`), `curl`.

Start Bifrost from main on port 8080:

```bash
docker run -d --name bifrost-repro -p 8080:8080 -v $(pwd)/bifrost-repro-data:/app/data bifrost:main
```

Open `http://localhost:8080/ui`, add an OpenAI provider with a placeholder key, and enable `allow_direct_keys`.

Start a mock WS upstream that accepts the upgrade and stays silent forever:

```bash
cat > /tmp/ws_mock_stall.py <<'PY'
import asyncio
import websockets

async def handler(ws):
    # Accept the upgrade and just wait forever. No frames, no close.
    await asyncio.Future()

async def main():
    async with websockets.serve(handler, "0.0.0.0", 9997, ping_interval=None):
        await asyncio.Future()

asyncio.run(main())
PY
python3 /tmp/ws_mock_stall.py &
```

Redirect Bifrost's OpenAI `base_url` to the stall mock (use `host.docker.internal` on Docker Desktop, `host.containers.internal` on podman):

```bash
curl -sS -X PUT http://localhost:8080/api/providers/openai \
  -d '{
    "name":"openai",
    "network_config":{"base_url":"http://host.docker.internal:9997","default_request_timeout_in_seconds":120,"max_retries":0,"retry_backoff_initial":500,"retry_backoff_max":5000,"stream_idle_timeout_in_seconds":120,"max_conns_per_host":5000},
    "concurrency_and_buffer_size":{"concurrency":1000,"buffer_size":5000},
    "send_back_raw_request":true,"send_back_raw_response":true,"store_raw_request_response":true,
    "openai_config":{"disable_store":false}
  }'
```

Fire the WS request and watch it hang:

```bash
printf '{"type":"response.create","model":"openai/gpt-4o","input":[{"type":"message","role":"user","content":[{"type":"input_text","text":"hi"}]}]}\n' | \
  websocat -n1 ws://localhost:8080/openai/v1/responses -H='Authorization: Bearer sk-fake-key'
```

Observe that the command does not return. Wait 60s, 120s, longer, there is no timeout. Cancel with `Ctrl+C` when tired.

In another terminal, watch Bifrost logs to see exactly one line (the upgrade) and nothing else:

```bash
docker logs --follow bifrost-repro 2>&1 | grep -E "openai/v1/responses|upstream|timeout"
```

Expected output (this is the bug):

```text
{"level":"info","http.method":"GET","http.target":"/openai/v1/responses","http.status_code":101,"http.request_duration_ms":0,...}
# ...and that is all. No further log line fires as time passes.
```

After a fix with an idle deadline, a `504 upstream_timeout` should surface to the client within `stream_idle_timeout_in_seconds` seconds (default 60), and a log row should be written with the timeout error.

### Expected behavior

After a configurable idle period without frames from the upstream, Bifrost times out the read, surfaces a 504 upstream_timeout error to the client, finalizes the log row with an error state, and returns the pool connection.

### Actual behavior

`upstream.ReadMessage()` has no deadline. If the upstream accepts the upgrade and sends nothing, the goroutine blocks indefinitely. No error surfaces, no log is finalized, the client hangs until its own local timeout.

### Affected area(s)

Transports (HTTP)

### Version

main (abfd68c80), reproduced live on 2026-04-24.

### Environment

```text
- Affected file: transports/bifrost-http/handlers/wsresponses.go (tryNativeWSUpstream read loop, ~line 308)
- SetReadDeadline already available: transports/bifrost-http/websocket/connection.go (~line 108)
- Reusable config: NetworkConfig.StreamIdleTimeoutInSeconds (default 60s, core/schemas/provider.go)
```

### Relevant logs/output

Captured from a live repro on `bifrost:main` with the stall mock as upstream. The client WS command ran for 121 seconds with no output. Bifrost logs for the trace:

```shell
{"level":"info","http.method":"GET","http.target":"/openai/v1/responses","http.status_code":101,"http.request_duration_ms":0,"http.remote_addr":"10.88.0.28:55772","http.user_agent":"","trace_id":"6a779d9c990b4b6d9b28951d48b8381b","time":"2026-04-24T10:10:49Z","message":"request completed"}
```

Exactly one line. No subsequent entry during the 121 second observation window.

### Regression?

Not a regression. Idle read deadlines were never set on this path.

### Severity

Reasoning:

* A silent upstream stall translates into a silent indefinite hang for the user's request. No error, no recovery.
* Codex CLI and Claude Code default to the WS transport, so any production deployment fronting those clients is one upstream hiccup away from goroutine pile-up.
* No workaround inside the native WS path. The only escape is client-side timeouts, which push the failure mode out of Bifrost's observability.
* Not Medium because the hang is total and does not resolve on its own.

### Related

* PR maximhq/bifrost#2775 in thiscantbeserious/bifrost (feature request): this bug was discovered while working on that feature. The branch contains a potential fix as a side-effect of the feature work. The fix will be extracted into a focused change for this issue.
* Sibling native-WS reliability extractions in the maximhq/bifrost#2775 triage table (log finalization [BF-723](https://linear.app/bifrost-gw/issue/BF-723/bug-native-websocket-upstream-requests-never-produce-a-log-row-on), stale pool connections, orphaned pending entries).

thiscantbeserious:
## Summary

Fixes #2998. When a native WS upstream accepts the upgrade and then sends no frames, `upstream.ReadMessage()` blocked indefinitely. This change adds a per-read idle deadline so the goroutine recovers and the client receives a 504 rather than a silent hang.

## Changes

- Add `wsUpstreamIdleTimeout` constant (60s, matching `DefaultStreamIdleTimeoutInSeconds`).
- Add `upstreamWSIdleTimeout()` method that reads `NetworkConfig.StreamIdleTimeoutInSeconds` from the provider config, falling back to the constant when no override is set.
- Before each `upstream.ReadMessage()` in `tryNativeWSUpstream`, set `SetReadDeadline(now + idleTimeout)`. Clear the deadline on successful read so active streams that send a frame periodically are not interrupted.
- On `net.Error.Timeout()`, discard the upstream from the pool, write a 504 upstream_timeout WS error frame to the client, return cleanly.

## Type of change

- [x] Bug fix

## Affected areas

- [x] Transports (HTTP)

## How to test

    go test -cover github.com/maximhq/bifrost/transports/bifrost-http/handlers
    go test -cover github.com/maximhq/bifrost/transports/bifrost-http/websocket

Manual, start a mock WS upstream that upgrades and stays silent, fire a request through Bifrost. Before this fix the request hangs indefinitely. After this fix it returns a 504 within `stream_idle_timeout_in_seconds`.

Tests added cover the stalling-server timeout, periodic-frame keepalive, and one-then-silent cases.

## Breaking changes

- [x] No

## Related issues

Closes #2998. Extracted from #2775, the PR branch contains this fix as a side effect of the OAuth feature work and it has been pulled out here as a focused change.

## Checklist

- [x] I read docs/contributing/README.md and followed the guidelines
- [x] I added tests
- [x] Build passes

### DIFF
--- a/transports/bifrost-http/handlers/wsresponses.go
+++ b/transports/bifrost-http/handlers/wsresponses.go
@@ -2,7 +2,11 @@ package handlers
 
 import (
 	"context"
+	"errors"
+	"fmt"
+	"net"
 	"strings"
+	"time"
 
 	"github.com/bytedance/sonic"
 	"github.com/fasthttp/router"
@@ -302,14 +306,44 @@ func (h *WSResponsesHandler) tryNativeWSUpstream(
 	tracer, _ := ctx.Value(schemas.BifrostContextKeyTracer).(schemas.Tracer)
 	traceID, _ := ctx.Value(schemas.BifrostContextKeyTraceID).(string)
 
+	// Determine the per-read idle deadline from provider config (or fallback constant).
+	idleTimeout := h.upstreamWSIdleTimeout(req.Provider)
+
 	// Read response events from upstream and relay to client, running post-hooks per chunk
 	forwardedAny := false
 	for {
+		// Set a per-read deadline so that a silent upstream stall (e.g. rate-limit
+		// hold, service degradation) does not block this goroutine indefinitely.
+		// The deadline is cleared after each successful read so that a long-running
+		// stream that sends a frame every <idleTimeout is never interrupted.
+		if setErr := upstream.SetReadDeadline(time.Now().Add(idleTimeout)); setErr != nil {
+			logger.Warn("upstream WS SetReadDeadline failed for %s: %v", req.Provider, setErr)
+		}
+
 		msgType, data, readErr := upstream.ReadMessage()
+
+		// Clear the deadline immediately after a successful read so subsequent
+		// reads start their own fresh idle window.
+		if readErr == nil {
+			_ = upstream.SetReadDeadline(time.Time{})
+		}
+
 		if readErr != nil {
-			logger.Warn("upstream WS read failed for %s: %v, falling back to HTTP bridge", req.Provider, readErr)
 			h.pool.Discard(upstream)
 			session.SetUpstream(nil)
+
+			// Detect idle timeout: the upstream accepted the connection but sent no
+			// frame within idleTimeout. Return a 504 to the client rather than
+			// silently blocking.
+			var netErr net.Error
+			if errors.As(readErr, &netErr) && netErr.Timeout() {
+				timeoutMsg := fmt.Sprintf("upstream websocket idle timeout after %.0fs", idleTimeout.Seconds())
+				logger.Warn("upstream WS idle timeout for %s (no frame in %.0fs)", req.Provider, idleTimeout.Seconds())
+				writeWSError(session, 504, "upstream_timeout", timeoutMsg)
+				return true
+			}
+
+			logger.Warn("upstream WS read failed for %s: %v, falling back to HTTP bridge", req.Provider, readErr)
 			if !forwardedAny {
 				return false
 			}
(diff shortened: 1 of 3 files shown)


thiscantbeserious:

PR #3021, opened 2026-04-24
fix(pool): probe liveness before reusing session-pinned WS upstream

### LINKED ISSUE #3002: [Bug]: WebSocket pool hands out stale session-pinned upstream connections after server-side close
### Prerequisites

### Description

**Summary**

When an upstream WebSocket server closes a connection (e.g. with code 1011 keepalive ping timeout) after delivering the terminal event in a response, Bifrost does not detect the stale state before handing the connection to the next request on the same client session. The next request writes to the connection successfully, then immediately fails on the subsequent read, triggers a fallback to the HTTP bridge, and delivers an error frame to the client.

**What the bug is**

`WSResponsesHandler.tryNativeWSUpstream` in `transports/bifrost-http/handlers/wsresponses.go` pins a fresh upstream connection to the client session via `session.SetUpstream(upstream)` (line 271). When the upstream sends its terminal event (`response.completed`) followed by a WS close frame, the read loop in `tryNativeWSUpstream` returns at the terminal-event check (line 350) without consuming the close frame that is now buffered in the socket. The connection is still pinned to the session with `IsClosed() == false`.

When the same client session sends a second `response.create`, the handler at line 258 checks `upstream == nil || upstream.IsClosed()`. Both are false, so the stale connection is reused. `WriteMessage` at line 294 may succeed because the OS TCP buffer accepts the write. `ReadMessage` at line 308 then immediately returns the buffered close frame as an error. `pool.Discard` and `session.SetUpstream(nil)` are called, and the handler returns `false` to trigger an HTTP bridge fallback. If the HTTP bridge also fails (e.g. the upstream is not reachable via plain HTTP), the client receives an `error` frame for a request that would have succeeded had a fresh upstream been dialed.

Two contributing design gaps make this happen:

1. `(*Pool).Get` in `pool.go` (line 65) guards against stale connections only via `conn.IsClosed()` (a local atomic bool set exclusively by Bifrost's own `Close()` call) and `p.isExpired(conn)` (a wall-clock check against `IdleTimeoutSeconds`). There is no network-level liveness probe. An upstream close that Bifrost never read leaves `IsClosed()` false indefinitely.
2. The `evictLoop` runs every 30 seconds and calls `evictExpired`, which also relies on `IsClosed()` and local time checks. It cannot detect connections the upstream closed silently.

**What the bug is NOT**

This is NOT the same as [BF-720](https://linear.app/bifrost-gw/issue/BF-720/bug-websocket-pool-dial-error-discards-upstream-http-status-and-body), which is about losing the upstream HTTP status and body when `pool.Get` itself fails to dial (a different code path and a different symptom). The stale-pool issue is about connections that dialed successfully but were later closed by the upstream, not about failed dials.

**Impact**

* Any multi-turn session over native WS is affected. After the first successful response, if the upstream server closed the connection on its end (e.g. keepalive timeout between requests), every subsequent `response.create` in the same session fails with a stale-connection read error and an HTTP bridge fallback.
* The failure is user-visible: the client receives an `error` frame instead of the response when the HTTP bridge fallback also fails, or observes extra latency when the fallback eventually succeeds.
* Codex CLI and Claude Code default to the WS transport and use multi-turn sessions. Any deployment with an upstream idle window shorter than the user's think time is affected continuously in production.
* The default `IdleTimeoutSeconds` is 600 seconds. Upstream servers such as [chatgpt.com](<http://chatgpt.com>) close with 1011 after a much shorter idle window (observed at roughly 60-120s). This means the stale-connection window is open on every production deployment fronting those upstreams.

**Root cause**

`UpstreamConn.IsClosed()` reads `c.closed.Load()` on an atomic bool that is only set by `c.Close()`. A server-side close that arrives as a WS close frame in the read buffer is never reflected in this atomic until Bifrost's own code reads and handles the frame. If the read loop exits early on a terminal event without consuming the trailing close frame, the connection is returned to the session in a state that appears live but is not.

**Suggested fix**

Three approaches (not mutually exclusive):

* Option A - zero-deadline read probe (low invasiveness): in `(*Pool).Get`, after popping an idle connection but before returning it, attempt a non-blocking read by setting a 0 (or 1ms) read deadline, calling `conn.ReadMessage()`, and resetting the deadline. An immediate timeout means the connection is still alive. An immediate close frame or EOF means it is dead. Discard and continue. This is safe to add in the pool because idle connections should produce no frames while sitting in the pool.
* Option B - per-connection background ping (medium invasiveness): after dial, spawn a goroutine that sends a WS Ping frame every N seconds (below the upstream's keepalive window) and calls `conn.Close()` if the Pong does not arrive within a deadline. This sets `IsClosed()` to true proactively, making the existing checks in `Get` and `evictLoop` effective.
* Option C - administrative workaround (no code change): lower `IdleTimeoutSeconds` in `WSPoolConfig` to a value below the upstream's known keepalive timeout (e.g. 30s for [chatgpt.com](<http://chatgpt.com>)). This shrinks but does not eliminate the race window.

Option A is recommended as the fastest, lowest-overhead fix. Option B provides defense in depth but adds goroutine overhead per connection.

**Verified reproduction**

Main-built container (commit `abfd68c80`, image `bifrost:main`, 2026-04-24). A mock WS upstream was configured on port 9995 via `PUT /api/providers/openai`. The mock served a full response (frames: `response.created`, `response.output_text.delta`, `response.completed`) then closed the upstream connection with code 1011. A Python test client opened one Bifrost WS session and sent two sequential `response.create` events over it. Request 1 completed successfully and the client received both `response.created` and `response.completed`. Request 2 was handed the stale upstream by the session pin, wrote to it, then immediately hit the buffered 1011 close on the first read:

```
upstream WS read failed for openai: websocket: close 1000 (normal): done, falling back to HTTP bridge
```

The client received an `error` frame for request 2. The mock confirmed that a fresh dial was triggered only AFTER the stale read failure (connection [#2](<https://github.com/maximhq/bifrost/issues/2>) in the mock log). No fresh dial was attempted proactively, confirming the absence of any liveness probe.

### Steps to reproduce

See the Reproducible command section below.

### Reproducible command

Prerequisites: `websocat` (on macOS `brew install websocat`), `python3` with `websockets` installed (`pip3 install websockets`), `curl`, a running `bifrost:main` container on port 8091.

Start a mock WS upstream that serves request 1 fully, then closes with 1011:

```bash
cat > /tmp/ws_stale_mock.py <<'PY'
import asyncio, json, websockets

FRAMES = [
    {"type":"response.created","response":{"id":"resp_s1","status":"in_progress","model":"gpt-4o"},"sequence_number":0},
    {"type":"response.output_text.delta","item_id":"m1","output_index":0,"content_index":0,"delta":"hi","sequence_number":1},
    {"type":"response.completed","response":{"id":"resp_s1","model":"gpt-4o","status":"completed","output":[]},"sequence_number":2},
]
call_count = 0

async def handler(ws):
    global call_count
    call_count += 1
    n = call_count
    print(f"[MOCK] connection #{n}", flush=True)
    if n == 1:
        await ws.recv()
        for f in FRAMES:
            await ws.send(json.dumps(f))
            await asyncio.sleep(0.05)
        await ws.close(code=1011, reason="keepalive ping timeout")
    else:
        print(f"[MOCK] FRESH DIAL #{n}", flush=True)
        await ws.recv()
        await ws.send(json.dumps({"type":"response.created","response":{"id":f"rs{n}","status":"in_progress","model":"gpt-4o"},"sequence_number":0}))
        await ws.send(json.dumps({"type":"response.completed","response":{"id":f"rs{n}","model":"gpt-4o","status":"completed","output":[]},"sequence_number":1}))
        await ws.close(code=1000, reason="done")

asyncio.run(websockets.serve(handler, "0.0.0.0", 9995, ping_interval=None).__aenter__().__await__().__next__())
PY
python3 - <<'PY'
import asyncio, json, websockets
FRAMES=[{"type":"response.created","response":{"id":"rs1","status":"in_progress","model":"gpt-4o"},"sequence_number":0},{"type":"response.output_text.delta","item_id":"m1","output_index":0,"content_index":0,"delta":"hi","sequence_number":1},{"type":"response.completed","response":{"id":"rs1","model":"gpt-4o","status":"completed","output":[]},"sequence_number":2}]
call_count=0
async def handler(ws):
    global call_count
    call_count+=1
    n=call_count
    print(f"[MOCK] connection #{n}",flush=True)
    if n==1:
        await ws.recv()
        for f in FRAMES:
            await ws.send(json.dumps(f))
            await asyncio.sleep(0.05)
        await ws.close(code=1011,reason="keepalive ping timeout")
    else:
        print(f"[MOCK] FRESH DIAL #{n}",flush=True)
        await ws.recv()
        await ws.send(json.dumps({"type":"response.completed","response":{"id":f"rs{n}","model":"gpt-4o","status":"completed","output":[]},"sequence_number":0}))
        await ws.close(1000,"done")
async def main():
    async with websockets.serve(handler,"0.0.0.0",9995,ping_interval=None):
        await asyncio.Future()
asyncio.run(main())
PY
```

Or save the mock to `/tmp/ws_stale_mock.py` and run `python3 /tmp/ws_stale_mock.py &`.

Point Bifrost at the mock (use `host.containers.internal` for podman, `host.docker.internal` for Docker Desktop):

```bash
curl -sS -X PUT http://localhost:8091/api/providers/openai \
  -d '{
    "name":"openai",
    "network_config":{"base_url":"http://host.containers.internal:9995","default_request_timeout_in_seconds":30,"max_retries":0,"retry_backoff_initial":500,"retry_backoff_max":5000,"stream_idle_timeout_in_seconds":30,"max_conns_per_host":5000},
    "concurrency_and_buffer_size":{"concurrency":1000,"buffer_size":5000},
    "send_back_raw_request":true,"send_back_raw_response":true,"store_raw_request_response":true,
    "openai_config":{"disable_store":false}
  }'
```

Run the two-request client over a single session:

```bash
python3 - <<'PY'
import asyncio, json, websockets
REQ = lambda n: json.dumps({"type":"response.create","model":"openai/gpt-4o","input":[{"type":"message","role":"user","content":[{"type":"input_text","text":f"request {n}"}]}]})
async def main():
    async with websockets.connect("ws://localhost:8091/openai/v1/responses", additional_headers={"Authorization":"Bearer sk-fake-key"}, ping_interval=None) as ws:
        await ws.send(REQ(1))
        while True:
            msg = await asyncio.wait_for(ws.recv(), timeout=10)
            d = json.loads(msg)
            print(f"R1 frame: {d.get('type')}", flush=True)
            if d.get("type") in ("response.completed","response.failed","error"):
                break
        await asyncio.sleep(1)
        await ws.send(REQ(2))
        while True:
            try:
                msg = await asyncio.wait_for(ws.recv(), timeout=10)
                d = json.loads(msg)
                print(f"R2 frame: {d.get('type')}", flush=True)
                if d.get("type") in ("response.completed","response.failed","error"):
                    break
            except Exception as e:
                print(f"R2 exception: {e}", flush=True)
                break
asyncio.run(main())
PY
```

Observe R1 completes successfully. Observe R2 produces an `error` frame (Bifrost attempts to reuse the stale upstream, gets the buffered close, falls back to HTTP bridge which also fails against the mock). Check Bifrost logs to see `upstream WS read failed ... falling back to HTTP bridge`.

Restore OpenAI provider after testing:

```bash
pkill -f ws_stale_mock.py 2>/dev/null
curl -sS -X PUT http://localhost:8091/api/providers/openai \
  -d '{"name":"openai","network_config":{"default_request_timeout_in_seconds":120,"max_retries":0,"retry_backoff_initial":500,"retry_backoff_max":5000,"stream_idle_timeout_in_seconds":120,"max_conns_per_host":5000},"concurrency_and_buffer_size":{"concurrency":1000,"buffer_size":5000},"send_back_raw_request":true,"send_back_raw_response":true,"store_raw_request_response":true,"openai_config":{"disable_store":false}}' > /dev/null
```

### Expected behavior

After an upstream server closes the connection (with or without a close frame), `Pool.Get` and the session-pin logic detect the stale state before attempting to reuse the connection, and dial a fresh upstream transparently.

### Actual behavior

The stale connection is reused. The first write may succeed, but the first read immediately returns the buffered close frame as an error. Bifrost falls back to the HTTP bridge, which may also fail, delivering an error frame to the client.

### Affected area(s)

Transports (HTTP)

### Version

main (abfd68c80), reproduced live on 2026-04-24.

### Environment

```text
- Affected file: transports/bifrost-http/websocket/pool.go (Pool.Get liveness guard, line 65)
- Affected file: transports/bifrost-http/handlers/wsresponses.go (session upstream reuse, line 258)
- Affected file: transports/bifrost-http/websocket/connection.go (IsClosed is local-only, line 78)
- Default IdleTimeoutSeconds: 600 (core/schemas/websocket.go: DefaultWSIdleTimeoutSeconds)
- evictLoop interval: 30 seconds (pool.go line 189), also local-only
```

### Relevant logs/output

Captured from a live repro on `bifrost:main` with the two-phase mock (2026-04-24):

```shell
{"level":"info","http.method":"GET","http.target":"/openai/v1/responses","http.status_code":101,"http.request_duration_ms":0,"http.remote_addr":"10.88.0.28:35024","http.user_agent":"Python/3.11 websockets/16.0","trace_id":"bc7e3db029f24356b71f7b1886cc682e","time":"2026-04-24T11:09:52Z","message":"request completed"}
{"level":"warn","time":"2026-04-24T11:09:53Z","message":"upstream WS read failed for openai: websocket: close 1000 (normal): done, falling back to HTTP bridge"}
```

Historical capture from user's real Codex session (2026-04-23 09:55:44 UTC), request stuck in blue while subsequent requests continued:

```shell
upstream WS read failed for openai (no data forwarded): websocket: close 1011 (internal server error): keepalive ping timeout, falling back to HTTP bridge
websocket read error: websocket: close 1006 (abnormal closure): unexpected EOF
```

### Regression?

Not a regression. The session upstream pin has never included a liveness probe, and `Pool.Get` has never performed a network-level liveness check.

### Severity

Medium

Reasoning:

* For High: this affects every multi-turn WS session whenever the upstream's server-side keepalive timeout is shorter than the user's inter-request gap. The default `IdleTimeoutSeconds` is 600s, while [chatgpt.com](<http://chatgpt.com>) closes at roughly 60-120s, so the stale window is wide open on every production deployment using that upstream. User-visible errors (stuck requests, error frames) occur predictably.
* For Medium: the stale-connection read failure triggers a fallback to the HTTP bridge, which sometimes succeeds. The data plane is not fully broken for all requests. The workaround (lower `IdleTimeoutSeconds` to below the upstream keepalive window) is available via the API with no code change. The failure is intermittent and bounded by session idle patterns rather than being constant.
* Verdict: Medium. The failure is real, user-visible, and production-impacting, but the data plane degrades rather than halting entirely because the fallback path exists. A first request after an idle window fails or degrades. Subsequent requests on a fresh connection succeed normally.

This compounds with [BF-724](https://linear.app/bifrost-gw/issue/BF-724/bug-native-websocket-upstream-requests-hang-indefinitely-when-upstream) (idle read deadline) and [BF-723](https://linear.app/bifrost-gw/issue/BF-723/bug-native-websocket-upstream-requests-never-produce-a-log-row-on) (log finalization): each stale-connection attempt also leaves a partial log row and may hit a blocking read before surfacing the error.

### Related

* [BF-720](https://linear.app/bifrost-gw/issue/BF-720/bug-websocket-pool-dial-error-discards-upstream-http-status-and-body) (OPEN): separate WS pool defect about dial error discarding upstream HTTP status and body. Same file (`pool.go`), different root cause (failed dial path, not stale reuse path).
* [BF-724](https://linear.app/bifrost-gw/issue/BF-724/bug-native-websocket-upstream-requests-hang-indefinitely-when-upstream) (OPEN): adds idle read deadline to `tryNativeWSUpstream`. Mitigates the hang when a stale connection's read blocks, but does not prevent the stale connection from being handed out in the first place. Still produces a 504 or error frame per request after a stale entry is picked up.
* PR maximhq/bifrost#2775 in thiscantbeserious/bifrost (feature request): this bug was discovered while working on that feature. The PR branch does NOT contain a fix for this specific bug. Filed as a standalone future fix.

thiscantbeserious:
## Summary

After the upstream WebSocket server sends its terminal event and then closes the connection (e.g. with code 1011 on keepalive timeout), Bifrost would hand out the same connection to the next request on the same client session. The connection appeared live (IsClosed() returned false, age was within limits) because the WS close frame sat buffered in the socket and had never been read. The next caller would succeed on WriteMessage (OS buffer accepts the write) and then immediately fail on the first ReadMessage when the buffered close frame was returned as an error. This caused a fallback to the HTTP bridge and an error frame to the client.

The root cause is that Pool.Get relied only on an atomic bool (set by Bifrost's own Close call) and wall-clock expiry checks to determine connection health. Neither of those detects a server-side close that Bifrost never consumed.

## Changes

- `transports/bifrost-http/websocket/pool.go`: added `(*Pool).isLive` helper that sets a 1ms read deadline on an idle connection and calls ReadMessage. A timeout error (no queued data) confirms the connection is alive. A close frame, EOF, or any other error reveals it is stale. The probe runs in Pool.Get after the existing IsClosed / isExpired checks, before incrementing inFlight and returning the connection to the caller. Stale connections are closed and the loop continues to the next idle candidate or a fresh dial.
- `transports/bifrost-http/websocket/pool_test.go`: added `TestPoolGetEvictsStaleSessionConn`. A mock WS server accepts one connection, then closes it server-side after the test returns the connection to the pool. Pool.Get is called again and must produce a fresh connection (new dial), not the stale one.

## Type of change

- [x] Bug fix

## Affected areas

- [x] Transports (HTTP)

## How to test

```sh
go test ./transports/bifrost-http/websocket/... -cover -run TestPoolGetEvictsStaleSessionConn -v
go test ./transports/bifrost-http/websocket/... -cover
go vet ./transports/bifrost-http/...
```

Expected: TestPoolGetEvictsStaleSessionConn passes, all existing tests pass, no vet warnings.

Integration test (manual): configure a WS upstream that closes connections after the terminal event. Send two sequential requests on the same Bifrost WS session with a pause between them. Before this fix, request 2 would receive an error frame. After this fix, request 2 succeeds with a fresh upstream connection.

## Breaking changes

- [x] No

## Related issues

Closes #3002

This fix was discovered while working on PR #2775 (openai-oauth-passthrough). The feature branch in that PR contains an earlier version of the liveness probe, which has been extracted here and removed from the feature branch in a separate commit.

## Security considerations

None. The probe is a local read with a 1ms deadline on a connection that Bifrost owns exclusively while it is in the idle pool. No external input is involved and no credentials are affected.

## Checklist

### DIFF
--- a/transports/bifrost-http/websocket/pool.go
+++ b/transports/bifrost-http/websocket/pool.go
@@ -1,7 +1,9 @@
 package websocket
 
 import (
+	"errors"
 	"fmt"
+	"net"
 	"sync"
 	"time"
 
@@ -63,6 +65,19 @@ func (p *Pool) Get(key PoolKey, headers map[string]string) (*UpstreamConn, error
 		p.mu.Unlock()
 
 		if conn.IsClosed() || p.isExpired(conn) {
+			conn.Close()
+			p.mu.Lock()
+			conns = p.idle[key]
+			continue
+		}
+
+		// Liveness probe: attempt a non-blocking read with a 1ms deadline.
+		// An idle connection should produce no frames while sitting in the pool.
+		// If the upstream already sent a close frame, ReadMessage returns
+		// immediately with a close/EOF error, revealing the stale state before
+		// the caller uses the connection.  A timeout error means the connection
+		// is alive and has nothing queued.
+		if !p.isLive(conn) {
 			conn.Close()
 			p.mu.Lock()
 			conns = p.idle[key]
@@ -197,6 +212,34 @@ func (p *Pool) evictLoop() {
 			p.evictExpired()
 		}
 	}
+}
+
+// isLive performs a network-level liveness probe on conn.
+// It sets a 1ms read deadline and attempts a ReadMessage.  A net.Error with
+// Timeout() == true means the connection is alive with nothing queued, which
+// is the expected state for an idle pooled connection.  Any other outcome
+// (close frame, EOF, protocol error) means the connection is stale.
+// The read deadline is always reset to zero before the function returns so
+// that a live connection can be used without deadline constraints.
+func (p *Pool) isLive(conn *UpstreamConn) bool {
+	if err := conn.SetReadDeadline(time.Now().Add(time.Millisecond)); err != nil {
+		return false
+	}
+	_, _, err := conn.ReadMessage()
+	// Always clear the deadline regardless of outcome.
+	_ = conn.SetReadDeadline(time.Time{})
+	if err == nil {
+		// Received a frame while idle — unexpected; treat as stale because
+		// we cannot safely discard the frame content here.
+		return false
+	}
+	var netErr net.Error
+	if errors.As(err, &netErr) && netErr.Timeout() {
+		// Deadline fired with no data: connection is alive.
+		return true
+	}
+	// Close frame, EOF, or other error: connection is stale.
+	return false
 }
 
 func (p *Pool) evictExpired() {
--- a/transports/bifrost-http/websocket/pool_test.go
+++ b/transports/bifrost-http/websocket/pool_test.go
@@ -128,6 +128,83 @@ func TestPoolClose(t *testing.T) {
 	assert.Error(t, err)
 }
 
+// TestPoolGetEvictsStaleSessionConn verifies that Pool.Get detects a
+// server-side close via the liveness probe and dials a fresh connection
+// instead of handing out the stale one (issue #3002).
+func TestPoolGetEvictsStaleSessionConn(t *testing.T) {
+	// closeCh is closed by the test to signal the server to close connection #1.
+	closeCh := make(chan struct{})
+	// dialCh receives a value each time the mock server accepts a new upgrade.
+	dialCh := make(chan struct{}, 8)
+
+	upgrader := ws.Upgrader{CheckOrigin: func(r *http.Request) bool { return true }}
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		conn, err := upgrader.Upgrade(w, r, nil)
+		if err != nil {
+			return
+		}
+		defer conn.Close()
+		dialCh <- struct{}{}
+		// Block until the test signals the server to close, or the client disconnects.
+		select {
+		case <-closeCh:
+			// Send a normal close frame so the TCP socket carries the close before
+			// the server's defer conn.Close() runs.
+			_ = conn.WriteMessage(ws.CloseMessage,
+				ws.FormatCloseMessage(ws.CloseNormalClosure, "done"))
+			time.Sleep(10 * time.Millisecond)
+		}
+	}))
+	defer server.Close()
+
+	config := &schemas.WSPoolConfig{
+		MaxIdlePerKey:                5,
+		MaxTotalConnections:          10,
+		IdleTimeoutSeconds:           300,
+		MaxConnectionLifetimeSeconds: 3600,
+	}
+	pool := NewPool(config)
+	defer pool.Close()
+
+	wsURL := "ws" + strings.TrimPrefix(server.URL, "http")
+	key := PoolKey{Provider: schemas.OpenAI, KeyID: "test-key", Endpoint: wsURL}
+
+	// Dial the first connection and confirm it reaches the mock server.
+	conn1, err := pool.Get(key, nil)
+	require.NoError(t, err)
+	require.NotNil(t, conn1)
+	<-dialCh // wait until server has accepted connection #1
+
+	// Return it to the idle pool.
+	pool.Return(conn1)
+
+	// Signal the server to close the upstream side of connection #1.
+	close(closeCh)
+
+	// Give the OS a moment to deliver the close frame into the socket buffer.
+	time.Sleep(50 * time.Millisecond)
+
+	// Pool.Get must detect the stale connection via the liveness probe and dial
+	// a fresh one.  It should not return conn1.
+	conn2, err := pool.Get(key, nil)
+	require.NoError(t, err)
+	require.NotNil(t, conn2)
+
+	// A fresh dial was triggered — wait for the mock server to record it.
+	select {
+	case <-dialCh:
+		// Good: server accepted a new upstream connection.
+	case <-time.After(2 * time.Second):
+		t.Fatal("expected a fresh upstream dial after stale-connection eviction, but none arrived")
+	}
+
+	// The returned connection must not be the stale one.
+	assert.NotSame(t, conn1, conn2, "Pool.Get must not return the stale session-pinned connection")
+	assert.True(t, conn1.IsClosed(), "stale connection must have been closed by the pool")
+
+	pool.Discard(conn2)
+}
+
 func TestPoolExpiredConnection(t *testing.T) {
 	server := startTestWSServer(t)
 	defer server.Close()


thiscantbeserious:

Issue #3001, opened 2026-04-24
[Bug]: Accumulator drops all WS Responses chunks, token usage and output never captured for native-WS requests


thiscantbeserious:
### Prerequisites

- [X] I have searched existing issues and discussions to avoid duplicates
- [X] I am using the latest version (or have tested against main/nightly)

### Description

**Summary**

`ProcessStreamingResponse` in `framework/streaming/accumulator.go` does not handle `schemas.WebSocketResponsesRequest`. The `isResponsesStreaming` guard covers only `schemas.ResponsesStreamRequest`. Every WS Responses chunk falls through to the error-return default branch, causing the accumulator to return an error for every chunk. Token counts, output accumulation, and any downstream aggregation depending on the accumulator are silently lost for all native-WS requests.

**What the bug is**

* `framework/streaming/accumulator.go`, function `ProcessStreamingResponse`, around line 400:

```go
isResponsesStreaming := requestType == schemas.ResponsesStreamRequest
```

`schemas.WebSocketResponsesRequest` is absent from this line. All four branch flags (`isChatStreaming`, `isResponsesStreaming`, `isAudioStreaming`, `isImageStreaming`) evaluate to `false` for a WS Responses chunk.

* The function falls through to the final return at line 422:

```go
return nil, fmt.Errorf("request type missing/invalid for accumulator: %s", requestType)
```

* The caller in `framework/tracing/tracer.go` (function `ProcessStreamingChunk`, line 305) discards the error silently:

```go
processedResp, processErr := t.accumulator.ProcessStreamingResponse(accumCtx, result, err)
if processErr != nil || processedResp == nil {
    return nil
}
```

No log line is emitted. The error is invisible at runtime.

* By contrast, `core/utils.go` `IsStreamRequestType` explicitly includes `WebSocketResponsesRequest` (line 311), so WS chunks are routed INTO the streaming pipeline but are then rejected once they arrive at the accumulator.
* The `wsresponses` handler in `transports/bifrost-http/handlers/wsresponses.go` calls `hooks.PostHookRunner` for every upstream frame (line 334). The logging plugin's post hook calls `tracer.ProcessStreamingChunk`, which calls into the accumulator. Every frame therefore hits the missing case and every frame produces a silently-dropped error.

**Relation to adjacent WS logging issues**

Two bugs on the same path already prevent any `websocket_responses` row from being written on upstream main:

[...]

Issue #2999, opened 2026-04-24
[Bug]: postHookRunner on WS streams never calls CompleteAndFlushTrace, log rows never reach DB


thiscantbeserious:
### Prerequisites

- [X] I have searched existing issues and discussions to avoid duplicates
- [X] I am using the latest version (or have tested against main/nightly)

### Description

**Summary**

For native WebSocket streaming in Bifrost, the `postHookRunner` closure built in `RunStreamPreHooks` (in `core/bifrost.go`) never calls `tracer.CompleteAndFlushTrace(traceID)` on the final chunk. The logging plugin accumulates completed entries in an in-memory `pendingLogsToInject` map keyed by `traceID`. That map is only drained to the DB when `CompleteAndFlushTrace` runs. Because the happy-path WS flow skips that call, every native-WS response leaves its log entry stuck in memory and never produces a row in the `logs` table.

**What the bug is**

* `RunStreamPreHooks` in `core/bifrost.go` around line 4094 to 4108. The happy-path `postHookRunner` closure calls `pipeline.RunPostLLMHooks` and, on `IsFinalChunk(ctx)`, `drainAndAttachPluginLogs(ctx)`. It does NOT call `tracer.CompleteAndFlushTrace(traceID)`.
* Every early-exit path in the same function (nil pre-request, short-circuit error, short-circuit response) correctly calls `CompleteAndFlushTrace`. The sibling `RunRealtimeTurnPreHooks` also calls it correctly. The happy-path WS stream is the only execution path missing the call.
* `drainAndAttachPluginLogs(ctx)` moves accumulated plugin logs into a pending buffer keyed by trace ID. Only `CompleteAndFlushTrace` actually writes them to the store.

**Relation to adjacent WS logging issues**

This is a distinct root cause from [BF-723](https://linear.app/bifrost-gw/issue/BF-723/bug-native-websocket-upstream-requests-never-produce-a-log-row-on) (which covers terminal-event detection and clean-close handling in `tryNativeWSUpstream`). #7a operates one layer deeper: even when terminal detection and `PostHookRunner` fire correctly, the trace is never flushed. Any of the two gaps is sufficient on its own to drop the log row. Fixing one without the other still leaves WS audit rows missing. Both need to land.

**Impact**

* Every native-WS response silently loses its audit trail. Users see a processing row in the UI, then the 5 min TTL sweeper evicts the pending entry with no DB row ever written.
[...]

Issue #2997, opened 2026-04-24
[Bug]: Native WebSocket upstream requests never produce a log row on clean close or unparseable terminal event


thiscantbeserious:
### Prerequisites

- [x] I have searched existing issues and discussions to avoid duplicates
- [x] I am using the latest version (or have tested against main/nightly)

### Description

**Summary**

When Bifrost proxies a client WebSocket to a native upstream WS provider, two independent code paths can leave the log entry in `processing` indefinitely (or evict it silently via the 5 min TTL sweeper) without ever writing a row to the `logs` table. The core observability invariant that "every request produces a log row" is violated for native WS traffic.

**What the bug is**

Two distinct paths in `transports/bifrost-http/handlers/wsresponses.go` both skip the final PostHookRunner call:

Path A, unparseable terminal frame:

- `parseUpstreamWSEvent` (`wsresponses.go:382-394` on main) does a full `sonic.Unmarshal` into `BifrostResponsesStreamResponse` and returns `nil` on any error. There is no fallback that reads just the top-level `type` field.
- Observed against `chatgpt.com/backend-api/codex` where the `response.completed` frame is ~57 KB and contains nested fields the struct cannot accept. The frame exists and is syntactically valid JSON, but `sonic.Unmarshal` fails, `streamResp` is `nil`, `isTerminal` stays `false`, `StreamEndIndicator` is never set.

Path B, clean upstream close without recognized terminal event:

- The `tryNativeWSUpstream` read loop (`wsresponses.go:309-317`) handles `ReadMessage` errors after `forwardedAny == true` by writing a 502 to the client and returning. It never synthesizes a terminal event, never sets `StreamEndIndicator`, never calls PostHookRunner.
- When `forwardedAny == false` (close arrived before any frame was processed), it falls back to the HTTP bridge. If the bridge fails or the provider URL is WS-only, the fallback produces no log row either.

**What the bug is NOT**

- The functional data plane is fine. Frames that do arrive are forwarded to the client. This issue is about the logging invariant, not user-facing chat.
- The bug is NOT specific to `chatgpt_oauth` or any one provider. Any native-WS-capable provider whose terminal frames are slightly larger or shape-shifted from Bifrost's schema, or whose upstream policy closes cleanly under some conditions, is affected.

**Impact**

- Codex CLI and Claude Code both default to the WS transport. Deployments fronting those clients for audit, compliance, or billing lose a portion of log rows silently.
[...]

PR #3022, opened 2026-04-24
fix(logging): flush orphan log row before TTL sweeper deletes pending entry

### LINKED ISSUE #3003: [Bug]: TTL sweeper silently discards orphaned pending log entries with no DB row written
### Prerequisites

### Description

**Summary**

The logging plugin defers every DB write from `PreLLMHook` to `PostLLMHook`. When `PostLLMHook` never fires (client disconnect, context cancel, or any of the native-WS bugs [BF-723](https://linear.app/bifrost-gw/issue/BF-723/bug-native-websocket-upstream-requests-never-produce-a-log-row-on) / #2999 / #3001), the `PendingLogData` entry stays in the `pendingLogsEntries` sync.Map indefinitely. A periodic sweeper evicts entries older than 5 minutes. The eviction does a bare `Delete` with no DB write and no warning log. The audit row is permanently lost.

**What the bug is**

`plugins/logging/writer.go`, function `cleanupStalePendingLogs` (lines 178-196):

```go
func (p *LoggerPlugin) cleanupStalePendingLogs() {
    cutoff := time.Now().Add(-pendingLogTTL)   // pendingLogTTL = 5 * time.Minute (line 25)
    p.pendingLogsEntries.Range(func(key, value any) bool {
        if pending, ok := value.(*PendingLogData); ok {
            if pending.CreatedAt.Before(cutoff) {
                p.pendingLogsEntries.Delete(key)  // line 183: bare delete, no DB write
            }
        }
        return true
    })
    p.pendingLogsToInject.Range(func(key, value any) bool {
        if pending, ok := value.(*pendingInjectEntries); ok {
            if pending.createdAt.Before(cutoff) {
                p.pendingLogsToInject.Delete(key)  // line 193: same pattern
            }
        }
        return true
    })
}
```

The sweeper runs every 1 minute (main.go line 310: `time.NewTicker(1 * time.Minute)`). Every entry older than 5 minutes is discarded with no trace in the DB.

There is no `p.logger.Warn(...)` call on eviction. There is no call to `enqueueLogEntry`. The entry vanishes as if the request never happened.

**Relation to adjacent WS logging issues**

This bug is not caused by #2997, #2999, or #3001, but those three bugs are the primary real-world source of orphaned entries on the `websocket_responses` path today.

* #2997: native-WS upstream never calls `RunPostLLMHooks` on clean close or unparseable terminal event. The pending entry is left alive.
* #2999: `postHookRunner` on WS streams never calls `CompleteAndFlushTrace`. Entries in `pendingLogsToInject` are left alive.
* #3001: accumulator drops all WS chunks. Even when the above are fixed, token usage is lost.

Fixing #2997, #2999, and #3001 reduces the frequency of orphaned entries but does not address the silent-eviction behavior. After those fixes land, HTTP path orphans (client disconnects, context cancels, queue-full drops, container restarts mid-flight) will still hit this code path. This is a separate defect in the eviction handler itself.

**Impact**

* Silent data loss in the audit trail. Every request that is cancelled, times out at the client, or is live in memory when the container restarts leaves no row in the `logs` table.
* `PreLLMHook` fires a `"processing"` callback to the UI. The UI shows a blue row that never resolves. The row disappears only after a UI refresh or page reload, with no explanation.
* Operators cannot distinguish "still running" from "already lost".
* Compounds with #2997, #2999, and #3001 by masking the frequency of WS-path failures in production. If the sweeper wrote an `"abandoned"` row, operators could observe the failure rate from the DB. Today they cannot.
* No metric or counter tracks sweeper evictions, so there is no alerting path even at `LOG_LEVEL=debug`.

**Root cause**

`plugins/logging/writer.go`, function `cleanupStalePendingLogs`:

* Line 25: `pendingLogTTL = 5 * time.Minute`
* Lines 178-196: both `pendingLogsEntries` and `pendingLogsToInject` are evicted with bare `Delete` calls
* No call to `buildInitialLogEntry`, `buildCompleteLogEntryFromPending`, or `enqueueLogEntry` on eviction
* No `p.logger.Warn(...)` or `p.logger.Error(...)` on eviction

The pattern for writing a minimal error row already exists. `PostLLMHook` (main.go lines 737-753) constructs a minimal `logstore.Log` with `Status: "error"` when it reaches a missing pending entry with a `bifrostErr`. The sweeper needs the same treatment, using `PendingLogData` fields it already holds.

**Suggested fix**

In `cleanupStalePendingLogs`, replace the bare `p.pendingLogsEntries.Delete(key)` with:

```go
p.logger.Warn("evicting orphaned pending log entry %s (created %s, never finalized)",
    pending.RequestID, pending.CreatedAt.Format(time.RFC3339))
entry := buildInitialLogEntry(pending)
entry.Status = "error"
// message field (or a metadata key) to distinguish from real errors
// e.g.: entry.ErrorDetails = `{"message":"abandoned: PostLLMHook never fired"}`
p.enqueueLogEntry(entry, nil)
p.pendingLogsEntries.Delete(key)
```

The same pattern applies to `pendingLogsToInject`: iterate `pending.entries` and enqueue each with `Status: "abandoned"` before deleting.

An `"abandoned"` or `"orphaned"` status value (currently not defined in the plugin) makes these rows queryable and distinguishable from provider errors.

**Verified reproduction**

Code evidence on main at commit `abfd68c80` is unambiguous. The sweeper function at lines 178-196 of `plugins/logging/writer.go` contains no DB write path. No `enqueueLogEntry` call exists anywhere in `cleanupStalePendingLogs`. A `grep` confirms this:

```bash
grep -n 'enqueueLogEntry\|WriteLog\|BatchCreate' /tmp/bifrost-main/plugins/logging/writer.go | grep -v '//'
```

The only `enqueueLogEntry` calls are in `processBatch` (batch writer internals) and `storeOrEnqueueEntry` (post-hook path). Neither is reachable from `cleanupStalePendingLogs`.

Runtime synthetic reproduction requires a 5-minute wait (pendingLogTTL) and is documented as Path B (historical observation): during the PR maximhq/bifrost#2775 session, a Codex "reading the repository" request appeared as a blue processing row in the UI and then vanished after the TTL window with no error row written. That observation is consistent with the code path confirmed above.

### Steps to reproduce

1. Configure a Bifrost instance with OpenAI provider pointing to a mock WS upstream that accepts the upgrade but never sends any frames.
2. Fire a WS request via `websocat`. The mock accepts and stalls. `PreLLMHook` fires, a pending entry is stored, and the UI shows the row as `"processing"`.
3. Disconnect the client. `PostLLMHook` never fires because the upstream never responded.
4. Wait longer than `pendingLogTTL` (5 minutes). The cleanup ticker fires within 1 minute of the TTL expiry.
5. Query the `logs` table. Expected: one row with status `"abandoned"` or `"error"`. Actual: no row for that request ID.

Steps 1-4 take 6 minutes. Step 5 confirms the data loss.

Alternatively: start a request, then hard-restart the Bifrost container before any response arrives. All in-flight entries are lost immediately with zero DB rows.

### Reproducible command

```bash
# Prerequisites: bifrost:main running on :8091, python3 with websockets, websocat, sqlite3

# Step 1: start a stalling mock WS upstream
cat > /tmp/ws_mock_stall.py <<'PY'
import asyncio
import websockets

async def handler(ws):
    await asyncio.sleep(3600)  # stall indefinitely

async def main():
    async with websockets.serve(handler, "0.0.0.0", 9997, ping_interval=None):
        await asyncio.Future()

asyncio.run(main())
PY
python3 /tmp/ws_mock_stall.py &

# Step 2: redirect bifrost OpenAI provider to the stalling mock
curl -sS -X PUT http://localhost:8091/api/providers/openai \
  -d '{"name":"openai","network_config":{"base_url":"http://host.containers.internal:9997","default_request_timeout_in_seconds":600,"max_retries":0}}'

# Step 3: record baseline log count
BEFORE=$(sqlite3 /Users/simon.sanladerer/git/bifrost/data-main/logs.db "SELECT COUNT(*) FROM logs")
echo "Before: $BEFORE"

# Step 4: fire the request, disconnect immediately after 3 seconds
timeout 3 websocat ws://localhost:8091/openai/v1/responses \
  -H='Authorization: Bearer sk-fake-key' <<< \
  '{"type":"response.create","model":"openai/gpt-4o","input":[{"type":"message","role":"user","content":[{"type":"input_text","text":"hi"}]}]}' || true

# Step 5: wait past the TTL window (5 min pendingLogTTL + 1 min sweeper cadence)
echo "Waiting 360 seconds for TTL eviction..."
sleep 360

# Step 6: compare log count
AFTER=$(sqlite3 /Users/simon.sanladerer/git/bifrost/data-main/logs.db "SELECT COUNT(*) FROM logs")
echo "After: $AFTER"
echo "Delta: $((AFTER - BEFORE)) (expected 1 if fixed, 0 on main)"

# Cleanup
kill %1 2>/dev/null || true
```

### Expected behavior

When the TTL sweeper evicts an orphaned pending entry, a minimal log row with a terminal status (e.g., `"abandoned"`) is written to the DB so the audit trail reflects that the request happened.

### Actual behavior

The orphaned entry is silently deleted. No DB row is written. No warning is logged. The UI shows `"processing"` indefinitely and then the row vanishes after the next page refresh.

### Affected area(s)

Plugins

### Version

main (abfd68c80), verified via code inspection on 2026-04-24.

### Environment

```text
- Component: Logging plugin
- Root cause file: plugins/logging/writer.go
- Root cause function: cleanupStalePendingLogs (lines 178-196)
- TTL constant: pendingLogTTL = 5 * time.Minute (line 25)
- Sweeper cadence: every 1 minute (main.go line 310)
- PreLLMHook store site: main.go line 642 (p.pendingLogsEntries.Store)
- PostLLMHook consume site: main.go line 732 (p.pendingLogsEntries.LoadAndDelete)
- Existing minimal-error-row pattern: main.go lines 737-753 (PostLLMHook guard for missing pending entry)
```

### Relevant logs/output

```shell
# No output. The sweeper evicts silently.
# There is no Warn or Error log on eviction in cleanupStalePendingLogs.

# Code confirmation:
$ grep -n 'enqueueLogEntry' /tmp/bifrost-main/plugins/logging/writer.go
# Returns only batchWriter and storeOrEnqueueEntry call sites.
# cleanupStalePendingLogs is not listed.
```

### Regression?

Not a regression. The sweeper has never written DB records on eviction. The deferred-write architecture (PreLLMHook stores, PostLLMHook writes) was introduced without a failsafe for the case where PostLLMHook never fires.

### Severity

Reasoning:

* This is a backstop, not a primary failure mode. Orphaned pending entries happen only when something else already went wrong, client disconnect mid-flight, ctx cancel, container restart, or the adjacent native-WS bugs #2997 / #2999 / #3001.
* Under normal operation, `PostLLMHook` fires and the entry is persisted correctly. No data loss on the happy path.
* Once #2997, #2999, and #3001 land, the orphan rate drops to near-zero and the only remaining orphans come from genuinely exceptional events (crashes, mid-flight client aborts, queue-full drops).
* The defect is in the cleanup path, which by definition only runs when things are off-nominal.
* Not High because the severity stack puts the primary causes (#2997, #2999, #3001) at High. This is a secondary defect that amplifies them by hiding the evidence, but on its own, with the upstream causes fixed, it's a rare edge-case concern.
* The fix remains low-risk: `buildInitialLogEntry` and `enqueueLogEntry` are already used elsewhere in the same file. Adding the call on eviction mirrors a pattern already established in `PostLLMHook`.

### Related

* #2997, #2999, #3001 (OPEN): adjacent native-WS logging bugs that cause the orphaned pending entries in the first place. Fixing those reduces the frequency of orphans but does not address the silent-eviction behavior, which is THIS issue.
* PR maximhq/bifrost#2775 in thiscantbeserious/bifrost (feature request): this bug was discovered while working on that feature. The PR branch does NOT contain a fix for this specific bug. Filed as a standalone future fix.

thiscantbeserious:
## Summary

When a request reaches `PreLLMHook` but `PostLLMHook` never fires (client disconnect, context cancel, native-WS upstream stall, container restart mid-flight), the `PendingLogData` entry is left in `pendingLogsEntries` indefinitely. The TTL sweeper evicts it after 5 minutes with a bare `sync.Map.Delete` and no DB write. The audit row is permanently lost and operators see a request that appeared as \"processing\" in the UI simply vanish.

Root cause: `cleanupStalePendingLogs` in `plugins/logging/writer.go` (lines 178-196) contains no call to `buildInitialLogEntry` or `enqueueLogEntry`. Both `pendingLogsEntries` and `pendingLogsToInject` are evicted silently.

## Changes

- `plugins/logging/writer.go` (`cleanupStalePendingLogs`): before deleting each expired `pendingLogsEntries` entry, call `buildInitialLogEntry` to construct a minimal `logstore.Log`, set `Status` to `"error"` and `ErrorDetails` to a message describing the TTL eviction, call `enqueueLogEntry` to write the row to the DB, then delete the in-memory entry. A `Warn` log is emitted per eviction so the event is visible in structured logs.
- `plugins/logging/sweeper_test.go` (new file): four unit tests covering stale-entry flush, fresh-entry survival, multi-entry flush, and stale `pendingLogsToInject` eviction.

The `pendingLogsToInject` entries are not individually DB-flushed because they are keyed by traceID and lack a standalone requestID anchor. Their eviction remains a bare delete, consistent with the current design.

## Type of change

- [x] Bug fix

## Affected areas

- [x] Plugins

## How to test

```sh
go build ./plugins/logging/...
go test ./plugins/logging/... -cover -run TestSweeper -v
go vet ./plugins/logging/...
```

Expected output: all four `TestSweeper*` tests pass. Coverage on `cleanupStalePendingLogs` is above 90%.

To reproduce the original bug and verify the fix end-to-end, configure a stalling mock WS upstream, fire a request, disconnect before any response, wait past `pendingLogTTL` (5 minutes) plus one sweeper tick (1 minute), and query the `logs` table. With this fix applied, a row with `status = "error"` and an `error_details` message containing "abandoned" appears. Without the fix, no row is written.

## Screenshots/Recordings

N/A (backend-only change, no UI impact)

## Breaking changes

- [x] No

## Related issues

Closes #3003

Related: this bug was discovered while working on PR #2775 in thiscantbeserious/bifrost. The feature branch (PR #2775) does not contain a prior fix for this specific issue. This PR extracts the standalone fix.

See also: #2997, #2999, #3001 (adjacent native-WS logging bugs that increase the frequency of orphaned entries but are separate root causes).

## Security considerations

None. The fix writes an additional DB row on TTL eviction. No secrets or PII beyond what `buildInitialLogEntry` already captures from `PendingLogData` are included.

## Checklist

### DIFF
--- a/plugins/logging/sweeper_test.go
+++ b/plugins/logging/sweeper_test.go
@@ -1,0 +1,187 @@
+package logging
+
+import (
+	"strings"
+	"testing"
+	"time"
+)
+
+// newTestPlugin returns a minimal LoggerPlugin suitable for sweeper unit tests.
+// It wires up a real SQLite store (via newTestStore) and a buffered write queue
+// so that enqueueLogEntry works without a running batchWriter goroutine.
+func newTestPlugin(t *testing.T) *LoggerPlugin {
+	t.Helper()
+	store := newTestStore(t)
+	return &LoggerPlugin{
+		store:      store,
+		logger:     testLogger{},
+		writeQueue: make(chan *writeQueueEntry, writeQueueCapacity),
+	}
+}
+
+// drainQueue reads all entries currently buffered in the plugin's writeQueue
+// and returns them as a slice.
+func drainQueue(p *LoggerPlugin) []*writeQueueEntry {
+	var out []*writeQueueEntry
+	for {
+		select {
+		case e := <-p.writeQueue:
+			out = append(out, e)
+		default:
+			return out
+		}
+	}
+}
+
+// TestSweeperFlushesOrphanedPendingLogsEntry verifies that cleanupStalePendingLogs
+// enqueues a synthesized error row before deleting a stale pendingLogsEntries
+// entry whose TTL has expired.
+func TestSweeperFlushesOrphanedPendingLogsEntry(t *testing.T) {
+	plugin := newTestPlugin(t)
+
+	requestID := "req-orphan-1"
+	staleTime := time.Now().Add(-(pendingLogTTL + time.Minute))
+
+	pending := &PendingLogData{
+		RequestID: requestID,
+		Timestamp: staleTime,
+		CreatedAt: staleTime,
+		InitialData: &InitialLogData{
+			Object:   "chat_completion",
+			Provider: "openai",
+			Model:    "gpt-4o",
+		},
+	}
+	plugin.pendingLogsEntries.Store(requestID, pending)
+
+	plugin.cleanupStalePendingLogs()
+
+	// The entry must be removed from the in-memory map.
+	if _, ok := plugin.pendingLogsEntries.Load(requestID); ok {
+		t.Fatal("expected pendingLogsEntries to be cleared after TTL eviction")
+	}
+
+	// Exactly one write must have been enqueued.
+	entries := drainQueue(plugin)
+	if len(entries) != 1 {
+		t.Fatalf("expected 1 write queue entry, got %d", len(entries))
+	}
+
+	log := entries[0].log
+	if log == nil {
+		t.Fatal("expected non-nil log entry in write queue")
+	}
+	if log.ID != requestID {
+		t.Fatalf("log.ID = %q, want %q", log.ID, requestID)
+	}
+	if log.Status != "error" {
+		t.Fatalf("log.Status = %q, want %q", log.Status, "error")
+	}
+	if !strings.Contains(log.ErrorDetails, "abandoned") {
+		t.Fatalf("log.ErrorDetails = %q, expected it to mention abandoned", log.ErrorDetails)
+	}
+	if log.Provider != "openai" {
+		t.Fatalf("log.Provider = %q, want %q", log.Provider, "openai")
+	}
+	if log.Model != "gpt-4o" {
+		t.Fatalf("log.Model = %q, want %q", log.Model, "gpt-4o")
+	}
+}
+
+// TestSweeperDoesNotFlushFreshPendingLogsEntry verifies that a pendingLogsEntries
+// entry that has not yet exceeded its TTL is left untouched.
+func TestSweeperDoesNotFlushFreshPendingLogsEntry(t *testing.T) {
+	plugin := newTestPlugin(t)
+
+	requestID := "req-fresh-1"
+	pending := &PendingLogData{
+		RequestID: requestID,
+		Timestamp: time.Now(),
+		CreatedAt: time.Now(),
+		InitialData: &InitialLogData{
+			Object:   "chat_completion",
+			Provider: "openai",
+			Model:    "gpt-4o",
+		},
+	}
+	plugin.pendingLogsEntries.Store(requestID, pending)
+
+	plugin.cleanupStalePendingLogs()
+
+	// The entry must still be present.
+	if _, ok := plugin.pendingLogsEntries.Load(requestID); !ok {
+		t.Fatal("expected fresh pendingLogsEntries entry to survive TTL sweep")
+	}
+
+	// No writes should have been enqueued.
+	entries := drainQueue(plugin)
+	if len(entries) != 0 {
+		t.Fatalf("expected no write queue entries for fresh entry, got %d", len(entries))
+	}
+}
+
+// TestSweeperFlushesMultipleOrphanedEntries verifies that all stale entries
+// are individually flushed when the sweeper runs.
+func TestSweeperFlushesMultipleOrphanedEntries(t *testing.T) {
+	plugin := newTestPlugin(t)
+
+	staleTime := time.Now().Add(-(pendingLogTTL + time.Minute))
+	ids := []string{"req-multi-1", "req-multi-2", "req-multi-3"}
+	for _, id := range ids {
+		plugin.pendingLogsEntries.Store(id, &PendingLogData{
+			RequestID: id,
+			Timestamp: staleTime,
+			CreatedAt: staleTime,
+			InitialData: &InitialLogData{
+				Object:   "chat_completion",
+				Provider: "openai",
+				Model:    "gpt-4o",
+			},
+		})
+	}
+
+	plugin.cleanupStalePendingLogs()
+
+	// All entries must be evicted.
+	for _, id := range ids {
+		if _, ok := plugin.pendingLogsEntries.Load(id); ok {
+			t.Fatalf("expected entry %q to be evicted", id)
+		}
+	}
+
+	// One write per entry.
+	entries := drainQueue(plugin)
+	if len(entries) != len(ids) {
+		t.Fatalf("expected %d write queue entries, got %d", len(ids), len(entries))
+	}
+	for _, e := range entries {
+		if e.log.Status != "error" {
+			t.Fatalf("entry %q: status = %q, want error", e.log.ID, e.log.Status)
+		}
+	}
+}
+
+// TestSweeperStalePendingLogsToInjectEvictedWithoutPanic verifies that the
+// pendingLogsToInject cleanup still runs without error when stale entries are
+// present. The inject entries are not DB-flushed (they lack a requestID
+// anchor) so this test focuses on panic-free eviction only.
+func TestSweeperStalePendingLogsToInjectEvictedWithoutPanic(t *testing.T) {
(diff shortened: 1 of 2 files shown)


thiscantbeserious:

PR #3041, opened 2026-04-25
fix(logging): show selected key for failed requests

### LINKED ISSUE #3040: [Bug]: Selected key is not shown for failed request logs
### Prerequisites

### Description

When an LLM request fails, the UI log details do not show which provider key was used for the failed attempt. This makes it difficult to identify and fix the problematic key, especially when a provider has multiple configured keys.

This was observed with the OpenAI provider, but the behavior is likely provider-agnostic because the selected key is cleared in the Go request handling path after a final error.

### Steps to reproduce

1. Configure a provider with at least one key that can fail, for example an invalid or expired Kiro key.
2. Send a request through Bifrost using that provider.
3. Wait for the request to fail.
4. Open the failed request in the UI logs page.
5. Check the request details section for the selected key.

### Expected behavior

The failed request log should show the key that was used for the failed attempt, so the operator can identify which key needs attention.

### Actual behavior

 The failed request log does not show the selected key. The UI gives no clear indication which configured key caused the error.

### Affected area(s)

Core (Go)

### Version

up to e40014b08ecd1d138ed5e967789dd527144ea4e5

### Environment

```text
- OS: Nix OS
```

### Relevant logs/output

```shell

```

### Regression?

Introduced in V1.5.0 (commit 0f6fa0b3a, PR #2245) by Pratham Mishra.

### Severity

loss-and-quick:
## Summary

When an LLM request fails, `core/bifrost.go` intentionally clears `BifrostContextKeySelectedKeyID` and `BifrostContextKeySelectedKeyName` so that `selected_key` only reflects a key that served a successful response. As a side effect, the logging plugin received empty key values and persisted logs without a selected key, making it impossible to identify which key caused the failure.
## Changes
- In `plugins/logging/main.go`: when `selectedKeyID` is empty, fall back to the last entry in `attemptTrail` to recover the key ID and name before writing the log entry.
## Type of change
- [x] Bug fix
## Affected areas
- [x] Plugins
## How to test
1. Configure a provider with an invalid/expired key.
2. Send a request through Bifrost — it should fail.
3. Open the failed request in the UI logs page.
4. Verify the selected key is now shown in the request details.
## Breaking changes
- [x] No
## Related issues
Closes #3040
## Security considerations
None. Key ID and name are already stored in `attemptTrail`; this change only reads from it.
## Checklist


akshaydeo:
❤️ for the PR. @loss-and-quick can you check the conflict

loss-and-quick:
@akshaydeo

Issue #3049, opened 2026-04-25
[Bug]: CacheDebug lost in streaming


loss-and-quick:
### Prerequisites

- [x] I have searched existing issues and discussions to avoid duplicates
- [x] I am using the latest version (or have tested against main/nightly)

### Description

When a cached streaming response was served, CacheDebug was never propagated to the log entry. The field was set on the final chunk's ExtraFields, but the streaming accumulator path (StreamAccumulatorResult → convertToProcessedStreamResponse → applyStreamingOutputToEntry) didn't carry it through. Additionally, cached streaming chunks had no RequestType set, causing downstream consumers (logging, telemetry) to misidentify the request type.

### Steps to reproduce

1. Configure semantic cache (Redis or Weaviate backend)
2. Send two identical chat completion requests (same messages, same params)
3. Observe: second request is a cache miss despite being identical
4. For streaming: observe cache_debug is null in the log entry for cache hits

### Expected behavior

cache_debug is populated in log entries for streaming cache hits

### Actual behavior

Cache never hits due to non-deterministic hashing; cache_debug is null for streaming hits.

### Affected area(s)

Plugins

### Version

up to e40014b08ecd1d138ed5e967789dd527144ea4e5

### Environment

```text
- OS: Nix OS
```

### Relevant logs/output

```shell

```

### Severity

High (major functionality broken)

Issue #2791, opened 2026-04-17
[Bug]: Direct semantic cache never hits, `request_hash` differs across byte-identical requests


klhq:
### Prerequisites

- [x] I have searched existing issues and discussions to avoid duplicates
- [x] I am using the latest version (or have tested against main/nightly)

### Description

Direct-mode semantic cache (`dimension: 1`, no embedding provider) **writes** a new entry for every request and **never returns a cache hit**, even for byte-identical requests sent back-to-back with the same `x-bf-cache-key`.

Inspecting Redis shows Bifrost assigns a **different `request_hash` and `params_hash` to each byte-identical request**, so its own lookup predicate can never match. The request has no MCP tools, no `x-bf-passthrough-extra-params`, identical payload bytes, identical `temperature`, identical `max_tokens`, identical `cache_key`.

This looks like the same class of bug as #1742 / #2347 (Go map iteration / `map[string]interface{}` round-trip producing non-deterministic JSON) but in the **`semantic_cache` plugin's own hash function**, not in the paths those two issues fixed. Related but distinct from #1915 (FT.SEARCH parser error) and #1993 (Valkey SCAN fallback) — in our case FT.SEARCH works fine, the plugin simply never queries a matching hash because the hash is non-deterministic for identical input.

**Secondary finding:** after `redis-cli FLUSHDB`, Bifrost does not recreate the `BifrostSemanticCachePlugin` FT index until the container restarts. Lookups then fail with `No such index` while writes keep succeeding (hash-keyed `HSET`, no index needed). A restart re-creates it. Probably worth ensuring the index on each lookup, or re-creating on Redis reconnect.

### Steps to reproduce

Config:

```json
{
  "vector_store": {
    "enabled": true,
    "type": "redis",
    "config": { "addr": "redis:6379" }
  },
  "plugins": [
    {
      "enabled": true,
      "name": "semantic_cache",
      "config": {
        "dimension": 1,
        "ttl": "5m",
        "cleanup_on_shutdown": true,
        "cache_by_model": true,
        "cache_by_provider": true
      }
    }
  ]
}
```

Send three byte-identical chat completions with the same cache key and inspect Redis:

PAYLOAD='{"model":"github_copilot/gpt-5-mini","messages":[{"role":"user","content":"Say PONG"}],"max_tokens":5,"temperature":0}'
KEY="repro-001"

for i in 1 2 3; do
  echo "=== R$i ==="
[...]

PR #3050, opened 2026-04-25
fix(semanticcache): deterministic request hashing and CacheDebug propagation in streaming

### LINKED ISSUE #2912: Semantic cache: params_hash non-deterministic for identical requests (v1.4.23 + v1.5.0-prerelease5)
## Summary

`params_hash` in the semantic cache plugin produces non-deterministic values for byte-identical requests, causing 100% cache miss rate when `x-bf-cache-type: semantic` is used. The hash is a strict equality filter in `performSemanticSearch` (`search.go:185`), so no cache entries ever match.

## Environment

- Bifrost v1.4.23 (also reproduced on v1.5.0-prerelease5)
- macOS arm64 (Apple Silicon)
- Vector store: Qdrant (gRPC :6334)
- Embedding provider: Gemini (`gemini-embedding-001`)

## Reproduction

1. Start Bifrost with semantic cache enabled + Qdrant vector store
2. Send identical request twice:
```bash
curl -s http://localhost:8080/v1/chat/completions \
  -H "Authorization: Bearer <key>" \
  -H "x-bf-cache-key: test" \
  -H "x-bf-cache-type: semantic" \
  -d '{"model":"mlx-general/model","messages":[{"role":"user","content":"Hi"}]}'
```
3. Check Qdrant — two entries with different `params_hash` values
4. Second request returns `cache_hit: false` despite identical request body

## Expected behavior

Byte-identical requests should produce the same `params_hash`, allowing the semantic search filter to match existing cache entries.

## Actual behavior

Every request produces a unique `params_hash`, so the strict filter `{Field: "params_hash", Operator: Equal}` at `search.go:185` never matches. Manual Qdrant vector search confirms 0.97 cosine similarity between entries.

## Investigation

- PR #2142 (`fix: deterministic tools_hash and params_hash`) is present in the binary (`MarshalDeeplySorted` confirmed in strings output)
- The fix addressed `json.Marshal` non-determinism for map keys, but the hash is still non-deterministic
- Tested with bare requests (no tools, no parameters) — same behavior
- Tested with `mcp-exclude-clients` header — same behavior
- The `params_hash` changes even when the only metadata should be `{"stream": false}`

## Workaround

None available via configuration. The `params_hash` filter is hardcoded in the semantic search path.

## Suggestion

Consider making `params_hash` an optional filter in semantic search mode (controlled by config), or investigate the remaining source of non-determinism in `computeRequestParamsHash`.

loss-and-quick:
## Summary

Fixes two bugs in `plugins/semanticcache` that together caused the semantic cache to be effectively non-functional for direct-mode hits and invisible in logs for streaming hits.

## Changes

- **`plugins/semanticcache/utils.go`** — `generateRequestHash()` was hashing raw typed params structs (`ChatParameters`, etc.) whose `map[string]interface{}` fields have non-deterministic iteration order in Go. `MarshalDeeplySorted` couldn't fully canonicalize them, so byte-identical requests produced different hashes and the cache never hit. Fix: reuse `buildRequestMetadataForCaching()` (already canonical) for the `params` field of the hash input.
- **`plugins/semanticcache/search.go`** — Cached streaming chunks had no `RequestType` set, causing logging/telemetry to misidentify the request type. Also tightens `buildResponseFromResult` to treat stream/single format mismatches as a cache miss with a warning instead of a hard error.
- **`core/schemas/tracer.go`, `framework/tracing/tracer.go`, `plugins/logging/main.go`, `plugins/logging/utils.go`** — `CacheDebug` was set on the final chunk's `ExtraFields` but the accumulator path (`StreamAccumulatorResult → convertToProcessedStreamResponse → applyStreamingOutputToEntry`) didn't carry it through. Added `CacheDebug` to `StreamAccumulatorResult` and wired it through the full path. Eagerly serialize in `PostLLMHook` to avoid a GORM `BeforeCreate` race on the async write path.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [ ] Core (Go)
- [ ] Transports (HTTP)
- [ ] Providers/Integrations
- [x] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

```sh
make test-plugins
```

## Breaking changes

- [ ] Yes
- [x] No*

> **Note:** Existing semantic cache entries written before this fix will become unreachable after deploy — the hash schema changed shape (canonical metadata map instead of typed structs, `stream` moved into `params`). Since the previous hashing was non-deterministic, those entries were already effectively unmatchable. A one-time cache flush after deploy is recommended.

## Related issues

Closes: #2912, #2791, #3049

## Security considerations
None.

## Checklist

### DIFF
--- a/core/schemas/tracer.go
+++ b/core/schemas/tracer.go
@@ -24,6 +24,7 @@ type StreamAccumulatorResult struct {
 	OutputMessages        []ResponsesMessage              // For responses API
 	TokenUsage            *BifrostLLMUsage                // Token usage
 	Cost                  *float64                        // Cost in dollars
+	CacheDebug            *BifrostCacheDebug              // Semantic cache debug info if available
 	ErrorDetails          *BifrostError                   // Error details if any
 	AudioOutput           *BifrostSpeechResponse          // For speech streaming
 	TranscriptionOutput   *BifrostTranscriptionResponse   // For transcription streaming
--- a/framework/tracing/tracer.go
+++ b/framework/tracing/tracer.go
@@ -336,6 +336,7 @@ func (t *Tracer) ProcessStreamingChunk(traceID string, isFinalChunk bool, result
 		accResult.OutputMessages = processedResp.Data.OutputMessages
 		accResult.TokenUsage = processedResp.Data.TokenUsage
 		accResult.Cost = processedResp.Data.Cost
+		accResult.CacheDebug = processedResp.Data.CacheDebug
 		accResult.ErrorDetails = processedResp.Data.ErrorDetails
 		accResult.AudioOutput = processedResp.Data.AudioOutput
 		accResult.TranscriptionOutput = processedResp.Data.TranscriptionOutput
--- a/plugins/logging/operations.go
+++ b/plugins/logging/operations.go
@@ -368,6 +368,11 @@ func (p *LoggerPlugin) applyStreamingOutputToEntry(entry *logstore.Log, streamRe
 		entry.Cost = streamResponse.Data.Cost
 	}
 
+	// Cache
+	if streamResponse.Data.CacheDebug != nil {
+		entry.CacheDebugParsed = streamResponse.Data.CacheDebug
+	}
+
 	if p.disableContentLogging == nil || !*p.disableContentLogging {
 		// Transcription output
 		if streamResponse.Data.TranscriptionOutput != nil {
@@ -380,10 +385,6 @@ func (p *LoggerPlugin) applyStreamingOutputToEntry(entry *logstore.Log, streamRe
 		// Image generation output
 		if streamResponse.Data.ImageGenerationOutput != nil {
 			entry.ImageGenerationOutputParsed = streamResponse.Data.ImageGenerationOutput
-		}
-		// Cache debug
-		if streamResponse.Data.CacheDebug != nil {
-			entry.CacheDebugParsed = streamResponse.Data.CacheDebug
 		}
 		// Output message
 		if streamResponse.Data.OutputMessage != nil {
--- a/plugins/logging/utils.go
+++ b/plugins/logging/utils.go
@@ -679,6 +679,7 @@ func convertToProcessedStreamResponse(result *schemas.StreamAccumulatorResult, r
 		OutputMessages:        result.OutputMessages,
 		ErrorDetails:          result.ErrorDetails,
 		TokenUsage:            result.TokenUsage,
+		CacheDebug:            result.CacheDebug,
 		Cost:                  result.Cost,
 		AudioOutput:           result.AudioOutput,
 		TranscriptionOutput:   result.TranscriptionOutput,
--- a/plugins/maxim/main.go
+++ b/plugins/maxim/main.go
@@ -137,6 +137,7 @@ func convertAccResultToProcessedStreamResponse(accResult *schemas.StreamAccumula
 			TranscriptionOutput: accResult.TranscriptionOutput,
 			FinishReason:        accResult.FinishReason,
 			RawResponse:         accResult.RawResponse,
+			CacheDebug:          accResult.CacheDebug,
 		},
 		RawRequest: &accResult.RawRequest,
 	}
--- a/plugins/semanticcache/search.go
+++ b/plugins/semanticcache/search.go
@@ -290,14 +290,16 @@ func (plugin *Plugin) buildResponseFromResult(ctx *schemas.BifrostContext, req *
 		similarity = *result.Score
 	}
 
-	if hasValidStreamingResponse && !hasValidSingleResponse {
-		// Handle streaming response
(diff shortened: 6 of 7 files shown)


loss-and-quick:
> ### Confidence Score: 3/5
> The hashing fix is safe to merge; the streaming CacheDebug fix is incomplete and leaves that logging gap open.
> 
> One confirmed P1: the PR's claimed fix for streaming CacheDebug propagation does not work because CacheDebug is dropped when converting BifrostStreamChunk back to BifrostResponse in core/bifrost.go. All other changes (accumulator wiring, content-logging guard removal, hash canonicalization) are correct. The P1 finding caps the score below 4.
> 
> plugins/semanticcache/search.go — the ExtraFields mutations in the streaming goroutine are silently lost at the BifrostStreamChunk boundary.
> 
> ### Important Files Changed
> Filename	Overview
> plugins/semanticcache/search.go	Format-mismatch logic correctly converted to cache-miss, but `CacheDebug`/`RequestType` mutations on `cachedResponse.ExtraFields` are lost at the BifrostStreamChunk boundary and never reach the logging accumulator.
> plugins/semanticcache/utils.go	Core deterministic-hashing fix: replaces non-deterministic typed params structs with canonical metadata map, correctly resolving the hash collision issue. `stream` key is now inside `params` (was top-level), which is a hash-schema break acknowledged in the PR.
> framework/tracing/tracer.go	Correctly wires `accResult.CacheDebug = processedResp.Data.CacheDebug` through the accumulator result; the plumbing is sound but ineffective for the streaming path (see search.go comment).
> core/schemas/tracer.go	Adds `CacheDebug *BifrostCacheDebug` field to `StreamAccumulatorResult` — clean, correct schema extension.
> plugins/logging/operations.go	Correctly moves `CacheDebug` population outside the `disableContentLogging` guard so it is always recorded — cache metadata is not content.
> plugins/logging/utils.go	Propagates `CacheDebug` from `StreamAccumulatorResult` into `ProcessedStreamResponse.Data`; correct but only effective once the upstream channel-boundary issue is resolved.
> plugins/maxim/main.go	Mirrors the CacheDebug field into `convertAccResultToProcessedStreamResponse`, consistent with the changes in logging/utils.go.
> ### Comments Outside Diff (1)
> 1. `plugins/semanticcache/search.go`, line 386-416 ([link](https://github.com/maximhq/bifrost/blob/a0a814f925c21ce2f18c94432fa6ffdeeddf8f45/plugins/semanticcache/search.go#L386-L416))
>    [P1](#) **ExtraFields mutations on `cachedResponse` are dropped before reaching the accumulator**
>    Both the `RequestType` patch (lines 388–390) and the `CacheDebug` assignment (line 416) write to `cachedResponse.GetExtraFields()`. However, the `BifrostStreamChunk` sent on the channel carries only the typed response fields (`TextCompletionResponse`, `ChatResponse`, …). In `core/bifrost.go` the framework reconstructs a fresh `BifrostResponse` from those typed fields (no `ExtraFields`), and `pipelinePostHookRunner` then calls `result.PopulateExtraFields(req.RequestType, …)` which sets `RequestType` but not `CacheDebug`.
>    As a result, when the streaming accumulator reads `result.GetExtraFields().CacheDebug` (e.g. `framework/streaming/chat.go:533`), it is always `nil` for cached streaming responses. The wiring added in `tracer.go` and `logging/utils.go` is correct plumbing, but the source (`accResult.CacheDebug`) will always be `nil`, so `CacheDebug` never reaches the log entry for streaming hits — the same bug the PR set out to fix.
>    The fix needs to be applied at the point where the framework reconstructs `BifrostResponse` from `BifrostStreamChunk` (i.e. in `core/bifrost.go`'s short-circuit streaming goroutine), or by adding an `ExtraFields`-carrying field to `BifrostStreamChunk` itself so that `CacheDebug` can survive the channel boundary.
> 
> Reviews (5): Last reviewed commit: ["fix(semanticcache,logging): preserve Cac..."](https://github.com/maximhq/bifrost/commit/a0a814f925c21ce2f18c94432fa6ffdeeddf8f45) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=29730797)

The P1 is a false positive. `GetExtraFields()` on `BifrostResponse` does not return a top-level field — it delegates to the active typed response's `ExtraFields` (e.g. `&r.ChatResponse.ExtraFields`, `core/schemas/bifrost.go:780`). Since `BifrostStreamChunk.BifrostChatResponse` and `bifrostResponse.ChatResponse` in `core/bifrost.go:4959` both hold the same `*BifrostChatResponse` pointer, the `CacheDebug` written at `search.go:416` is visible to the accumulator at `framework/streaming/chat.go:533`. `PopulateExtraFields` only writes `RequestType`, `Provider`, `OriginalModelRequested`, `ResolvedModelUsed` — it does not clear `CacheDebug`.

akshaydeo (approved):

akshaydeo (approved):

PR #3159, opened 2026-04-30
feat: add optional schema option for postgres config and logs stores

joris-sevenlab:
## Summary

Adds an optional `schema` field to the PostgreSQL config for `config_store` and `logs_store` so Bifrost's tables can live in a non `public` schema. Useful when sharing a database with other tenants or apps.

## Changes

- New `framework/pgsetup` package holding the shared bits (DSN, schema name validation, schema creation) for both stores.
- `EnsureSchema` checks `information_schema.schemata` and only runs `CREATE SCHEMA` when the schema is absent. `CREATE SCHEMA IF NOT EXISTS` still demands `CREATE` on the database, which would block users who pre-create the schema and only grant schema level rights.
- Schema names must match `[a-zA-Z_][a-zA-Z0-9_]*` and stay under 63 characters. Validated before being interpolated into raw SQL.
- When a custom schema is set, the DSN appends `search_path=<schema>,public` so DDL goes to the schema and `pgcrypto` style functions still resolve.
- `configstore/postgres.go` and `logstore/postgres.go` route through `pgsetup`. Old `buildPostgresDSN` helper removed.
- `transports/config.schema.json` and `docs/deployment-guides/config-json/storage.mdx` updated.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [ ] Core (Go)
- [x] Transports (HTTP)
- [ ] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [x] Docs

## How to test

```sh
# Unit tests
go -C framework test ./pgsetup/...

# Full integration tests (requires Postgres on :5432)
docker compose -f framework/docker-compose.yml up -d postgres
go -C framework test ./pgsetup/... ./configstore/... ./logstore/...
```

Manual checks:

1. `schema` unset: falls back to `public`, no change in behavior.
2. New `schema` value, DB user with `CREATE` on the database: schema is auto-created, tables land inside it.
3. Schema pre-created, DB user only has schema level `USAGE` + `CREATE`: startup succeeds (this case currently fails on `main`).
4. Invalid name (`foo-bar`, over 63 chars): startup fails with a validation error.
5. `config_store` and `logs_store` on the same database with different `schema` values: no cross-talk.

No new environment variables.

## Screenshots/Recordings

N/A.

## Breaking changes

- [ ] Yes
- [x] No

`schema` is optional. Configs that omit it behave the same as before.

## Related issues

N/A

## Security considerations

- Schema name is set in `config.json` by the deployer and validated against a strict identifier pattern with a 63-character cap before being used in DDL.
- Docs note that the schema should be owned by the Bifrost DB user, and that untrusted users with `CREATE` on schemas earlier in `search_path` could shadow Bifrost's tables.

## Checklist

### DIFF
--- a/docs/deployment-guides/config-json/storage.mdx
+++ b/docs/deployment-guides/config-json/storage.mdx
@@ -83,8 +83,17 @@ Production-grade storage suitable for high-availability and high-throughput depl
 | `password` | - | Database password (supports `env.` prefix). Leave empty for IAM role auth. |
 | `db_name` | - | Database name |
 | `ssl_mode` | - | `"disable"`, `"require"`, `"verify-ca"`, `"verify-full"` |
+| `schema` | connection `search_path` (usually `public`) | Optional Postgres schema. Auto-created on startup if missing (DB user needs `CREATE` on the database in that case). |
 | `max_idle_conns` | `5` | Maximum idle connections in the pool |
 | `max_open_conns` | `50` | Maximum open connections to the database |
+
+<Note>
+**Custom `schema`.**
+
+- If the schema already exists, grant the DB user `USAGE` and `CREATE` on that schema. Otherwise, grant the DB user `CREATE` on the database so Bifrost can create the schema on startup.
+- Schema should be owned by the Bifrost DB user. Untrusted users with `CREATE` on schemas earlier in `search_path` can shadow Bifrost's tables.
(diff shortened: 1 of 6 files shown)


joris-sevenlab:
Resolved merge conflict with main upstream.

PR #3187, opened 2026-05-01
fix: filter list models by virtual key providers

### LINKED ISSUE #2887: [Bug]: Listing models for a virtual key not associated with all providers adds errors to logs
### Prerequisites

### Description

Tested against 1.50-prerelease4 and stable as of posting.
Listing /models endpoint for a virtual key with only one provider makes the proxy try listing models for all possible providers, adding unnecessary errors to the log, pollutes log list, makes error rate unreliable.

### Steps to reproduce

1. Configure a virtual key with only 1 provider
2. Curl the list models endpoint with it as bearer 
3. Observe logs show "Provider {} is not allowed for this virtual key"

### Expected behavior

Listing models with a given virtual key checks allowed providers first before attempting to list internally.

### Actual behavior

Virtual key list models goes directly to all providers without checking allowed providers first.

### Affected area(s)

Core (Go), UI (React)

### Version

v1.4.23, v1.5.0-prerelease4

### Environment

```text

```

### Relevant logs/output

```shell

```

### Regression?

_No response_

### Severity

princepal9120:
## Summary
- Fixes #2887.
- Restricts `/v1/models` fan-out to the providers allowed by the request virtual key.
- Makes core `ListAllModels` honor `BifrostContextKeyAvailableProviders` so disallowed providers are skipped before provider-level list calls run.

## Why
When `/v1/models` is called with a virtual key scoped to one provider, Bifrost previously attempted to list models from every configured provider. Governance then rejected the disallowed providers, creating noisy expected errors in logs and inflating error rates.

## Validation
- `cd core && go test . -run TestFilterProvidersByContext -count=1`
- `cd transports && go test ./bifrost-http/handlers -run TestListModels_VKFilter -count=1`

## Duplicate check
- Checked open PRs for issue `#2887` and the issue title before opening. No open PR was found.


princepal9120:
Pushed `7838bd6` to address the inactive virtual key review note.

Changes:

* Ignored inactive virtual keys before applying provider scope during list-model fan-out.
* Added direct regression coverage for inactive virtual keys.

Validated locally with:

```bash
go test ./bifrost-http/handlers -run 'TestApplyListModelsVirtualKeyProviderFilterSkipsInactiveVK|TestListModels_VKFilter'
```

akshaydeo (approved):

PR #3196, opened 2026-05-04
fix(bedrock): omit toolChoice.tool on Llama for synthetic structured-output tool

ryan-orphic:
## Summary

Bedrock Converse rejects `toolConfig.toolChoice.tool` on Meta Llama variants with HTTP 400:

> This model doesn't support the toolConfig.toolChoice.tool field. Remove toolConfig.toolChoice.tool and try again.

The synthetic `bf_so_*` tool path used for structured output (introduced in #3184) unconditionally pinned the synthetic tool via `toolChoice.tool`, which broke every `with_structured_output(...)` / `response_format=json_schema` caller against Llama 4 Maverick / Scout / Llama 3.x on Bedrock.

This PR completes the per-family gating story #3184 started for Anthropic.

## Changes

Adds an `IsLlamaModel` helper next to the existing `IsNovaModel` / `IsAnthropicModel` family in `core/schemas/utils.go`, then gates the forced-tool emission off it in three places:

1. **`core/providers/bedrock/utils.go`** — ChatCompletions synthetic-tool injection (`response_format=json_schema` → `bf_so_*` tool path). Synthetic tool is still injected; only the forced `toolChoice.tool` is suppressed.
2. **`core/providers/bedrock/responses.go`** — OpenAI Responses API mirror path (`text.format=json_schema` → `bf_so_*` tool path). Same shape as #1.
3. **`core/providers/bedrock/utils.go`** — `convertToolConfigFromFiltered`'s explicit `tool_choice` path (defense-in-depth for callers using `bind_tools(tool_choice={\"type\": \"function\", \"function\": {\"name\": \"X\"}})` directly).

With one synthetic tool bound and Bedrock's default \"auto\" behavior, omitting `tool_choice` yields the same outcome on Llama because there's exactly one tool the model can call. This mirrors the gate `langchain-aws` ships ([`supports_tool_choice_values`](https://github.com/langchain-ai/langchain-aws/blob/main/libs/aws/langchain_aws/chat_models/bedrock_converse.py) only allows \"auto\" for `llama4`, `llama3-1`, `llama3-3` model families, causing `with_structured_output` to emit no `tool_choice` at all).

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [x] Core (Go)
- [ ] Transports (HTTP)
- [x] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

Tests added (mirroring #3184's \`TestBedrockAnthropicChatStructuredOutputUsesSyntheticTool\` naming + structure):

- \`TestBedrockLlamaChatStructuredOutputOmitsForcedToolChoice\` — synthetic tool present, \`ToolChoice nil\` for Llama on the ChatCompletions path.
- \`TestBedrockNonLlamaChatStructuredOutputForcesToolChoice\` — regression guard: Nova / Anthropic still get the forced \`tool_choice\` (gate doesn't over-fire).
- \`TestToBedrockResponsesRequest_LlamaStructuredOutputOmitsForcedToolChoice\` — same negative+positive on the Responses API path.
- \`TestBedrockLlamaConvertToolConfigOmitsForcedToolChoice\` — defense-in-depth coverage for explicit \`tool_choice\` struct callers.

\`\`\`sh
go test -count=1 -run \"TestBedrockLlama|TestBedrockNonLlama|TestToBedrockResponsesRequest_Llama\" ./core/providers/bedrock/...
\`\`\`

Live verification: ran a forked build of this patch behind production CRS traffic (LangChain \`with_structured_output(strict=True)\` against \`us.meta.llama4-maverick-17b-instruct-v1:0\` on Bedrock) — previously 400 ValidationException on every planner request; with this gate, structured output succeeds end-to-end.

## Breaking changes

- [ ] Yes
- [x] No

The gate only narrows behavior for Llama (where the prior code path was already failing 100% of requests with HTTP 400). Nova / Anthropic / non-Llama paths are unchanged and covered by the negative regression test.

## Related issues

Companion to #3184 — that PR added the synthetic-tool fallback for Bedrock Converse \`response_format=json_schema\`, which made the forced-tool emission load-bearing for all structured-output callers. This PR completes that work for the Llama family, which has a stricter \`toolChoice\` support matrix than Anthropic / Nova.

## Security considerations

No new auth surfaces, secrets, PII, or sandboxing changes. Internal request shaping only.

## Checklist

- [x] I added/updated tests where appropriate (4 new tests covering both positive and negative cases on both API paths)
- [x] I verified builds succeed (Go) — patched build is running in production CRS traffic
- [ ] I verified the CI pipeline passes locally if applicable (Go not installed locally; relying on PR CI)

## References

- [Bedrock ToolChoice API reference](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_ToolChoice.html) — model-family support matrix
- [\`langchain-aws\` \`supports_tool_choice_values\`](https://github.com/langchain-ai/langchain-aws/blob/main/libs/aws/langchain_aws/chat_models/bedrock_converse.py) — prior art for the same gate (\`llama4\` / \`llama3-1\` / \`llama3-3\` -> \"auto\" only)
- #3184 — Anthropic structured-output fallback to synthetic tool path; this PR completes the per-family gating.

### DIFF
--- a/core/providers/bedrock/bedrock_test.go
+++ b/core/providers/bedrock/bedrock_test.go
@@ -4596,3 +4596,306 @@ func TestToolResultImageContentResponsesAPI(t *testing.T) {
 		assert.Empty(t, toolResult.Content, "remote URL image should be dropped (Bedrock only supports base64)")
 	})
 }
+
+// TestBedrockLlamaChatStructuredOutputOmitsForcedToolChoice locks in the
+// per-model gate for Meta Llama on Bedrock. Bedrock Converse rejects
+// `toolConfig.toolChoice.tool` on Llama variants with HTTP 400
+// ("This model doesn't support the toolConfig.toolChoice.tool field. Remove
+// toolConfig.toolChoice.tool and try again."). The synthetic `bf_so_*` tool
+// is still injected — Llama receives a single tool to call — but no forced
+// tool_choice is emitted. With one tool bound and Bedrock's default "auto"
+// behavior, the structured-output contract is preserved (the model has
+// exactly one tool it can call, so "any" and "the named one" converge).
+//
+// See per-model support matrix at
+// https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_ToolChoice.html
+// and the langchain-aws ChatBedrockConverse implementation
+// (`supports_tool_choice_values`) for prior art that ships the same gate.
+func TestBedrockLlamaChatStructuredOutputOmitsForcedToolChoice(t *testing.T) {
+	responseFormat := any(map[string]any{
+		"type": "json_schema",
+		"json_schema": map[string]any{
+			"name": "PlannerOutput",
+			"schema": map[string]any{
+				"type": "object",
+				"properties": map[string]any{
+					"intent": map[string]any{"type": "string"},
+				},
+				"required": []any{"intent"},
+			},
+		},
+	})
+
+	bifrostReq := &schemas.BifrostChatRequest{
+		Model: "us.meta.llama4-maverick-17b-instruct-v1:0",
+		Input: []schemas.ChatMessage{
+			{
+				Role: schemas.ChatMessageRoleUser,
+				Content: &schemas.ChatMessageContent{
+					ContentStr: schemas.Ptr("classify this message"),
+				},
+			},
+		},
+		Params: &schemas.ChatParameters{
+			ResponseFormat: &responseFormat,
+		},
+	}
+
+	ctx := schemas.NewBifrostContext(context.Background(), schemas.NoDeadline)
+	result, err := bedrock.ToBedrockChatCompletionRequest(ctx, bifrostReq)
+	require.NoError(t, err)
+	require.NotNil(t, result)
+
+	// Positive: synthetic bf_so_* tool still injected so the model has the
+	// schema available to call.
+	require.NotNil(t, result.ToolConfig, "expected toolConfig with synthetic bf_so_* tool")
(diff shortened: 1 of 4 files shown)


ryan-orphic:
Thanks @greptile-apps — good catch on the Responses API explicit \`tool_choice\` gap. The Chat Completions path had defense-in-depth via \`convertToolConfigFromFiltered\`, but the analogous gate on \`responses.go\` lines 1962–1971 was missing.

Just pushed an amend that closes that gap:

- \`core/providers/bedrock/responses.go\` — adds the same per-model gate immediately after \`convertResponsesToolChoice\` returns (drops \`bedrockToolChoice\` when \`Tool != nil\` AND \`schemas.IsLlamaModel(bifrostReq.Model)\`).
- Two new tests in \`bedrock_test.go\`:
  - \`TestToBedrockResponsesRequest_LlamaConvertResponsesToolChoiceOmitsForcedToolChoice\` — Llama + explicit \`{type: function, name: X}\` tool_choice → \`ToolConfig.Tools\` survives, \`ToolConfig.ToolChoice == nil\`.
  - \`TestToBedrockResponsesRequest_NonLlamaConvertResponsesToolChoiceForcesToolChoice\` — Anthropic regression guard so the new gate doesn't over-fire.

Commit message updated to call out four gate sites instead of three (two API surfaces × two paths each: synthetic-tool injection + explicit \`tool_choice\`).

akshaydeo (approved):

ryan-orphic:
@akshaydeo Thank you. How long does it usually take for changes to be included in your enterprise builds?

PR #3235, opened 2026-05-05
fix: gives graceful error for invalid after values by letting upstream provider handle it

roroghost17:
## Summary

`NewSerialListHelper` previously rejected any cursor string that could not be decoded as a Bifrost `SerialCursor`, returning an error immediately. This broke providers where a caller might supply a raw, native provider cursor (e.g. an opaque token or a full URL) directly — the helper would refuse it before the upstream API ever had a chance to respond. This PR introduces an `allowNativeCursorFallback` flag that, when enabled and only a single key is present, treats an unrecognised cursor as a native provider cursor and passes it through unchanged.

Replicate is explicitly opted out (`allowNativeCursorFallback=false`) because its cursor is a full URL; passing an unrecognised value would produce a malformed request rather than a clean API error. All other providers (Anthropic, Azure, Bedrock, Gemini, OpenAI) opt in.

## Changes

- Added `allowNativeCursorFallback bool` parameter to `NewSerialListHelper`. When `true` and there is exactly one key, a cursor that fails `DecodeSerialCursor` is wrapped as a native cursor at key index 0 instead of returning an error.
- All providers updated to pass `true` for `allowNativeCursorFallback`, except Replicate which passes `false` with an explanatory comment.
- Added `pagination_test.go` covering cursor encode/decode round-trips, all `NewSerialListHelper` fallback combinations, `GetCurrentKey`, `BuildNextCursor`, and `HasMoreKeys` behaviour.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [x] Core (Go)
- [ ] Transports (HTTP)
- [x] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

```sh
go test ./core/providers/utils/...
```

The new test file covers:
- Valid and invalid cursor decode paths
- Fallback behaviour with single vs. multiple keys
- Fallback disabled returning an error regardless of key count
- `GetCurrentKey` returning the correct key and native cursor token
- `BuildNextCursor` advancing to the next key or returning empty when exhausted

## Breaking changes

- [x] Yes
- [ ] No

`NewSerialListHelper` now requires an additional `allowNativeCursorFallback bool` argument. Any caller outside this repository that constructs a `SerialListHelper` directly must be updated to pass the appropriate boolean.

## Security considerations

None. The fallback only applies when a single key is configured, so there is no risk of a crafted cursor being used to skip to a different tenant's key.

## Checklist

### DIFF
--- a/core/providers/anthropic/anthropic.go
+++ b/core/providers/anthropic/anthropic.go
@@ -1461,7 +1461,7 @@ func (provider *AnthropicProvider) BatchList(ctx *schemas.BifrostContext, keys [
 	sendBackRawResponse := providerUtils.ShouldSendBackRawResponse(ctx, provider.sendBackRawResponse)
 
 	// Initialize serial pagination helper (Anthropic uses AfterID for pagination)
-	helper, err := providerUtils.NewSerialListHelper(keys, request.AfterID, provider.logger)
+	helper, err := providerUtils.NewSerialListHelper(keys, request.AfterID, provider.logger, true)
 	if err != nil {
 		return nil, providerUtils.NewBifrostOperationError("invalid pagination cursor", err)
 	}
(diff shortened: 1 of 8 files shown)


roroghost17:
* **#3235** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/3235?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/maximhq/bifrost/3235?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

akshaydeo:
### Merge activity

* **May 6, 6:03 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/3235).
* **May 6, 6:03 AM UTC**: @akshaydeo merged this pull request with [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/3235).

PR #3280, opened 2026-05-06
transports/bifrost-http: surface routed identity as x-bifrost-* response headers

Edward-Upton:
## Why

When a caller talks to Bifrost via one of the drop-in integrations (`/anthropic/v1/messages`, `/openai/v1/chat/completions`, `/bedrock/...`, etc.) the response converter flattens Bifrost's internal `BifrostResponse` down to a provider-native shape so the upstream SDK sees something it understands. That conversion drops `extra_fields`, which is where Bifrost stores the post-routing / post-fallback identity:

- `Provider` — the provider that actually ran (e.g. `bedrock` after a fallback from Anthropic).
- `OriginalModelRequested` — the model the caller sent.
- `ResolvedModelUsed` — final identifier reaching the provider API, after both routing-rule resolution and key-level alias resolution (per [`docs/providers/aliasing-models.mdx`](https://github.com/maximhq/bifrost/blob/main/docs/providers/aliasing-models.mdx)).

Today, a caller using the Anthropic SDK pointed at Bifrost has no way to recover any of this — so for usage tracking, billing attribution, or routing observability they're stuck either:

- switching to the native `/v1/chat/completions` endpoint (giving up the SDK's typing / retry / streaming primitives), or
- subscribing to the tracer interface and correlating to outbound HTTP requests via request id (high coupling, awkward).

## What

Surface the routed identity as response headers on every successful response from the integrations:

| Header | Source | When emitted |
|---|---|---|
| `x-bifrost-provider` | `ExtraFields.Provider` | always (when non-empty) |
| `x-bifrost-original-model` | `ExtraFields.OriginalModelRequested` | always (when non-empty) |
| `x-bifrost-resolved-model` | `ExtraFields.ResolvedModelUsed` | always (when non-empty) |
| `x-bifrost-request-type` | `ExtraFields.RequestType` | always (when non-empty) |
| `x-bifrost-fallback-index` | `BifrostContextKeyFallbackIndex` | only when `> 0` — absence of header = primary succeeded |

Naming follows the existing `x-bf-*` request-side convention (`x-bf-vk`, `x-bf-key-id`, etc.).

## How

Two-file change in `transports/bifrost-http/integrations`:

1. **`router.go`** — replace the per-case `providerResponseHeaders` capture in `GenericRouter.handleRequest` with a single `bifrostExtraFields` snapshot of the whole `ExtraFields` value. Same number of assignments in the switch, but now we can read both `ProviderResponseHeaders` and the routing identity from one source.
2. **`utils.go`** — new `applyBifrostResponseHeaders(ctx, bifrostCtx, extra)` helper that forwards the upstream provider headers verbatim (existing behaviour) and writes the new `x-bifrost-*` headers. Empty fields are skipped so headers never appear with a blank value. Safe to call even when the case didn't populate `extra` — the zero value produces no headers.

Net diff: **+157 / -27 lines**, mostly tests.

```
 transports/bifrost-http/integrations/router.go     | 56 ++++++++--------
 transports/bifrost-http/integrations/utils.go      | 54 ++++++++++++++++
 transports/bifrost-http/integrations/utils_test.go | 74 ++++++++++++++++++++++
```

## Scope notes

- **Unary path only.** Every request type that flowed through the common header-forwarding footer (chat completions, responses, embeddings, rerank, OCR, count tokens, image gen / edit / variation, video gen / retrieve / download / delete / remix / list, transcription, list models, text completion) gets the headers.
- **Streaming SSE is intentionally not touched** in this PR. Routed identity for streaming is more awkward because the `BifrostResponse` arrives chunk-by-chunk after headers are flushed. Worth a follow-up — either pre-resolve from `BifrostContext` before the first chunk, or use a transport pre-write hook. Filing as a separate issue if this lands well.
- **No SDK contract change.** Clients that don't care ignore the headers. Headers don't appear on error responses (only on `200 OK` from the integrations) — happy to extend if maintainers want them on error paths too, since `BifrostError` also has `ExtraFields.{Provider, OriginalModelRequested, ResolvedModelUsed}`.

## Tests

`TestApplyBifrostResponseHeaders` (4 sub-tests) covers:

- full identity emission with all fields populated
- fallback index emission from context when non-zero
- zero-value `ExtraFields` produces no headers (defensive against cases that don't populate)
- primary-success case (`FallbackIndex=0`) emits no fallback header — absence is the unambiguous signal

```
$ go test -run TestApplyBifrostResponseHeaders ./integrations/ -v
=== RUN   TestApplyBifrostResponseHeaders
=== RUN   TestApplyBifrostResponseHeaders/routed_identity_emits_all_headers
=== RUN   TestApplyBifrostResponseHeaders/fallback_index_from_context_emits_when_non-zero
=== RUN   TestApplyBifrostResponseHeaders/zero-value_extra_writes_no_headers
=== RUN   TestApplyBifrostResponseHeaders/primary-provider_success_(FallbackIndex=0)_does_not_emit_fallback_header
--- PASS: TestApplyBifrostResponseHeaders (0.00s)
PASS
```

Full integrations package suite still passes.

## Use case (context for why I care)

We use Bifrost as a model gateway with routing rules + fallback. When a caller asks for `claude-sonnet-4-6`, our routing rules + fallback chain might land on Anthropic direct, Bedrock-via-Anthropic, or Vertex-via-Anthropic — with the SDK seeing only what looks like an Anthropic response. Our usage-tracking and cost-attribution layer needs the actual `(provider, model)` that ran to (a) avoid persisting `bifrost/...` as a permanent canonical identity, and (b) compute cost against the right per-token rates if/when route-dependent pricing matters. These headers give us that without having to reach for the native API or the tracer plugin.

Happy to iterate on naming, scope (errors? streaming?), or implementation approach.

### DIFF
--- a/transports/bifrost-http/integrations/router.go
+++ b/transports/bifrost-http/integrations/router.go
@@ -943,7 +943,12 @@ func (g *GenericRouter) handleNonStreamingRequest(ctx *fasthttp.RequestCtx, conf
 	// for providers to respect cancellation.
 	var response interface{}
 	var err error
-	var providerResponseHeaders map[string]string
+	// bifrostExtraFields snapshots the routed identity (provider, original/resolved
+	// model) plus the upstream provider's response headers from whichever Bifrost
+	// response variant the case below populates. The common footer below surfaces
+	// the routed identity as `x-bifrost-*` response headers and forwards the
+	// upstream provider headers verbatim.
+	var bifrostExtraFields schemas.BifrostResponseExtraFields
 
 	switch {
 	case bifrostReq.ListModelsRequest != nil:
@@ -986,7 +991,7 @@ func (g *GenericRouter) handleNonStreamingRequest(ctx *fasthttp.RequestCtx, conf
 		}
 
 		response, err = config.ListModelsResponseConverter(bifrostCtx, listModelsResponse)
-		providerResponseHeaders = listModelsResponse.ExtraFields.ProviderResponseHeaders
+		bifrostExtraFields = listModelsResponse.ExtraFields
 	case bifrostReq.TextCompletionRequest != nil:
 		textCompletionResponse, bifrostErr := g.client.TextCompletionRequest(bifrostCtx, bifrostReq.TextCompletionRequest)
 		if bifrostErr != nil {
@@ -1010,7 +1015,7 @@ func (g *GenericRouter) handleNonStreamingRequest(ctx *fasthttp.RequestCtx, conf
 
 		// Convert Bifrost response to integration-specific format and send
 		response, err = config.TextResponseConverter(bifrostCtx, textCompletionResponse)
-		providerResponseHeaders = textCompletionResponse.ExtraFields.ProviderResponseHeaders
+		bifrostExtraFields = textCompletionResponse.ExtraFields
 	case bifrostReq.ChatRequest != nil:
 		chatResponse, bifrostErr := g.client.ChatCompletionRequest(bifrostCtx, bifrostReq.ChatRequest)
 		if bifrostErr != nil {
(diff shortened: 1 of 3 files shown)


akshaydeo:
this overall looks good to me - @Pratham-Mishra04 have a look once

Edward-Upton:
@akshaydeo any update on this?

akshaydeo:
asking @pratham to take a look once - he was out this weekend

Pratham-Mishra04 (approved):

PR #3291, opened 2026-05-07
fix(bedrock): preserve image content blocks in Converse ingest

### LINKED ISSUE #3289: [Bug]: user-message image content blocks silently dropped on /bedrock/.../converse ingest
### Prerequisites

### Description

When a client sends a Bedrock Converse request containing a user-message `image` content block to Bifrost's Bedrock-compatible endpoint (`POST /bedrock/model/<id>/converse`), Bifrost silently drops the image during the inbound Converse → Bifrost-Responses conversion. The downstream Bedrock call therefore arrives without the image, and the model replies as if no image was attached.

The mirror endpoint on the same gateway instance (`/openai/v1/chat/completions` with an `image_url` content block targeting the same Bedrock model) correctly forwards the image. The failure is isolated to the `/bedrock/.../converse` ingest path.

### Steps to reproduce

Send a Bedrock Converse request with an image content block to Bifrost's `/bedrock/.../converse` endpoint:

```python
import io, boto3
from PIL import Image

img = Image.new("RGB", (256, 128), color="red")
buf = io.BytesIO(); img.save(buf, format="PNG")

client = boto3.client("bedrock-runtime", endpoint_url="http://<bifrost>/bedrock")
resp = client.converse(
    modelId="us.anthropic.claude-sonnet-4-6",
    messages=[{"role": "user", "content": [
        {"image": {"format": "png", "source": {"bytes": buf.getvalue()}}},
        {"text": "Describe this image in one sentence."},
    ]}],
)
print(resp["output"]["message"]["content"][0]["text"])
print("inputTokens:", resp["usage"]["inputTokens"])
```

Observed:

```
I don't see any image attached to your message. ...
inputTokens: 31
```

Expected: a color/content description and ~80+ input tokens (image counted).

For comparison, the same PNG sent through the same gateway's OpenAI-compatible endpoint (`POST /openai/v1/chat/completions` with `model: "bedrock/us.anthropic.claude-sonnet-4-6"` and an `image_url` data-URI block) returns `"The image is a solid red rectangle."` with `prompt_tokens: 85`. This isolates the bug to the `/bedrock/.../converse` ingest path.

Using a `before-send` event hook on the boto3 client, the exact HTTP body leaving the client (= entering Bifrost) is:

```
URL: http://<bifrost>/bedrock/model/us.anthropic.claude-sonnet-4-6/converse
Content-Length: 742
Body (image truncated):
{
  "messages": [{
    "role": "user",
    "content": [
      {"image": {"format": "png", "source": {"bytes": "iVBORw0KGgoAAAANSUhEUgAAAQAAAACA...2gfgwQH/T4bRagAAAABJRU5ErkJggg=="}}},
      {"text": "Describe this image in one sentence."}
    ]
  }]
}
```

The image bytes are present in the HTTP request body entering Bifrost, and absent in both the dashboard's normalized representation and (by token count) in what Bedrock receives. The drop therefore happens inside Bifrost, between HTTP ingest and the normalized log write.

### Expected behavior

The PNG is passed through to the Bedrock backend; the model sees the image and responds accordingly. The dashboard log shows the image in the normalized user message.

### Actual behavior

The dashboard log for the request shows only the text block — no image. The model responds "I don't see any image attached to your message." Bedrock `inputTokens` is ~30 (text-only size), not ~80+ that an image would add.

### Affected area(s)

Core (Go)

### Version

v1.5.0

### Environment

```text
- OS: Ubuntu 24.04.4 LTS
- Image: maximhq/bifrost:v1.5.0
- Docker:  Docker Engine - Community 29.3.0
```

### Relevant logs/output

```shell

```

### Regression?

_No response_

### Severity

Critical (blocks releases or production)

SAY-5:
## Summary

User-message `image` content blocks were silently dropped during the `/bedrock/.../converse` ingest path. The bytes never reached the downstream provider, so models replied as if no image had been attached.

## Changes

`convertSingleBedrockMessageToBifrostMessages` had arms for `Text`, `ReasoningContent`, `ToolUse`, `Document`, and `ToolResult`, but no arm for `Image`. Added one that converts a `BedrockImageSource` into a Responses `input_image` block carrying a `data:image/<fmt>;base64,...` URL, mirroring the existing `Document` path.

## Type of change

- [x] Bug fix

## Affected areas

- [x] Core (Go)

## How to test

```sh
cd core
go test ./providers/bedrock/ -run TestBedrockToBifrostRequestConversion -count=1
```

A `MessageWithImage` table case is added that fails on `main` (returns `Input: nil`) and passes with the fix.

## Breaking changes

- [x] No

## Related issues

Closes #3289

## Security considerations

None, only changes inbound message normalization for images that were already accepted but discarded.


TejasGhatte:
hey @SAY-5 , PR looks good, can you please resolve pending comments then we are good to merge🙇‍♂️

SAY-5:
Resolved both pending coderabbit comments in f7c06d9:
- Added `Status: schemas.Ptr("completed")` to the image bifrostMsg to match text/tool branches.
- Moved `bifrostMsg` construction inside the `block.Image.Source.Bytes != nil` branch so we don't append a content-less message.
- Updated the test expected value to include `Status`.

Pratham-Mishra04 (approved):

SAY-5:
main already has an equivalent Image arm in convertSingleBedrockMessageToBifrostMessages, so this branch is redundant now and just conflicts. Closing it.

The only extra bit here was setting Status to completed on the image message so it matches the text and tool branches. Happy to send that as a one liner if you want it.

PR #3338, opened 2026-05-08
Fix Gemini fallback propagation in GenAI integration

Javtor:
## Summary
- preserve Gemini request fallbacks when converting GenAI generateContent requests into Bifrost Responses requests
- fix generic integration fallback extraction to read exported `Fallbacks` fields and `json:"fallbacks"` tags
- add regression coverage for Gemini fallback propagation through both conversion points

## Root cause
The GenAI/Gemini Responses conversion dropped `GeminiGenerationRequest.Fallbacks`, and the transport fallback extractor looked for a lowercase Go field name (`fallbacks`) that does not match exported request structs. That meant selected fallback candidates could be present on the integration request but never reach the core fallback runner.

## Tests
- `GOTOOLCHAIN=auto go test ./providers/gemini -count=1` from `core/`
- `GOTOOLCHAIN=auto go test ./bifrost-http/integrations -count=1` from `transports/`

### DIFF
--- a/transports/bifrost-http/integrations/utils.go
+++ b/transports/bifrost-http/integrations/utils.go
@@ -418,7 +418,7 @@ func (g *GenericRouter) extractFallbacksFromRequest(req interface{}) ([]string, 
 		return nil, nil
 	}
 
-	// Try to use reflection to find a "fallbacks" field
+	// Try to use reflection to find a fallbacks field.
 	reqValue := reflect.ValueOf(req)
 	if reqValue.Kind() == reflect.Ptr {
 		reqValue = reqValue.Elem()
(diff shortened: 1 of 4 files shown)


transports/bifrost-http/integrations/utils.go
```diff
@@ -374,10 +374,20 @@ func (g *GenericRouter) extractFallbacksFromRequest(req interface{}) ([]string,
 		return nil, nil // Not a struct, no fallbacks
 	}
 
-	// Look for the "fallbacks" field
-	fallbacksField := reqValue.FieldByName("fallbacks")
+	fallbacksField := reqValue.FieldByName("Fallbacks")
 	if !fallbacksField.IsValid() {
-		return nil, nil // No fallbacks field found
+		reqType := reqValue.Type()
```
Pratham-Mishra04, line 435: can you just leave a comment for why are we doing this fallback check for better readability?

Javtor: Added the comment

Pratham-Mishra04 (approved):

PR #3343, opened 2026-05-09
[codex] surface routing rule aliases in model listings

### LINKED ISSUE #2655: Model listing should include dynamic models too
### Prerequisites

### Problem to solve

I have used lots of different LLM proxying tools, and am evaluating Bifrost now. The problem is that if you want to have generic 'models', that stay same even if underlying model changes, you really want something like:

```
gemma4
 => gemma4:31b (at google, with some funky openrouter or vertex full path)
 fallback: => gemma4:31b (local on CPU)
```

and in the apps I would really rather just configure 'gemma4'. It is fine in apps that don't autodetect models, but those that do fail currently badly.

Disclaimer: am using 1.5.0 prerelease 2, and configured it according to dynamic example in https://docs.getbifrost.ai/providers/aliasing-models 

https://github.com/maximhq/bifrost/issues/1058 claims aliasing is 'done' but it really doesn't work for clients which list available models first before doing anything with them.

### Proposed solution

Provide fake model responses to model listing also for models which are just dynamic models from routing rules (if available).

### Alternatives considered

_No response_

### Area(s)

Core (Go)

### Additional context

_No response_

0oAstro:
Fixes #2655.

### What changed

- Include reachable routing-rule model entrypoints in governance-filtered list-model responses for virtual keys.
- Parse simple routing-rule model literals from CEL expressions using:
  - `model == "..."`
  - `model in ["...", "..."]`
- Only append route aliases when the rule is enabled and the virtual key can actually reach an allowed primary target or, when no primary target is allowed, an allowed fallback on the current provider listing pass.
- Keep the existing virtual-key provider/model filtering for provider-returned models.
- Add governance tests for allowed primary targets, denied targets, and fallback-only reachability.

### Why

Routing rules are documented as the dynamic alias layer, and those aliases can work for inference. Before this change, `/v1/models` only returned provider/key-backed models after governance filtering, so clients that validate models from `/v1/models` could not discover valid routing-rule aliases such as `deepseek-v4-pro` or the `gemma4` style aliases described in #2655.

This keeps the maintainer concern from #2655 in mind: dynamic rules can be scoped and may not resolve for every caller. The list response is therefore augmented only after virtual-key governance filtering and only for rules whose target or fallback is reachable for the requesting virtual key/provider pass.

### Validation

```bash
cd plugins/governance
go test ./...
```

### DIFF
--- a/plugins/governance/utils.go
+++ b/plugins/governance/utils.go
@@ -3,11 +3,20 @@ package governance
 
 import (
 	"context"
+	"regexp"
+	"sort"
 	"strings"
 
 	bifrost "github.com/maximhq/bifrost/core"
 	"github.com/maximhq/bifrost/core/schemas"
+	configstoreTables "github.com/maximhq/bifrost/framework/configstore/tables"
 	"github.com/valyala/fasthttp"
+)
+
+var (
+	modelEqualsLiteralRegexp = regexp.MustCompile(`\bmodel\s*==\s*["']([^"']+)["']`)
+	modelInLiteralRegexp     = regexp.MustCompile(`\bmodel\s+in\s*\[([^\]]*)\]`)
+	stringLiteralRegexp      = regexp.MustCompile(`["']([^"']+)["']`)
 )
 
 // ParseVirtualKeyFromFastHTTPRequest parses the virtual key from FastHTTP request headers.
@@ -93,6 +102,7 @@ func (p *GovernancePlugin) filterModelsForVirtualKey(
 	ctx context.Context,
 	models []schemas.Model,
 	virtualKeyValue string,
+	currentProvider schemas.ModelProvider,
 ) []schemas.Model {
 	// Get virtual key configuration
 	vk, exists := p.store.GetVirtualKey(ctx, virtualKeyValue)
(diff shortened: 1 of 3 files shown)


plugins/governance/utils.go
```diff
@@ -139,5 +149,151 @@ func (p *GovernancePlugin) filterModelsForVirtualKey(
...
+	filteredModels = p.appendRoutingRuleModelsForVirtualKey(ctx, filteredModels, vk, currentProvider)
+	sort.Slice(filteredModels, func(i, j int) bool {
+		return strings.ToLower(filteredModels[i].ID) < strings.ToLower(filteredModels[j].ID)
+	})
+
 	return filteredModels
 }
+
+func (p *GovernancePlugin) appendRoutingRuleModelsForVirtualKey(ctx context.Context, models []schemas.Model, vk *configstoreTables.TableVirtualKey, currentProvider schemas.ModelProvider) []schemas.Model {
+	if p.store == nil || vk == nil || currentProvider == "" || !p.store.HasRoutingRules(ctx) {
```
roroghost17, line 161: hey @0oAstro, how is the global scope handled here? Suppose, I have a rule with a global scope and don't necessarily pass in a vk, how would that work?

plugins/governance/utils.go
```diff
@@ -3,13 +3,22 @@ package governance
...
 	"strings"
 
 	bifrost "github.com/maximhq/bifrost/core"
 	"github.com/maximhq/bifrost/core/schemas"
+	configstoreTables "github.com/maximhq/bifrost/framework/configstore/tables"
 	"github.com/valyala/fasthttp"
 )
 
+var (
+	modelEqualsLiteralRegexp = regexp.MustCompile(`\bmodel\s*==\s*(?:"([^"]+)"|'([^']+)')`)
```
roroghost17, line 17: hey @0oAstro, could you please add description comments as to what the regex's actually do

plugins/governance/utils.go
0oAstro: ohhh that is something i should look at. ill put the pr on draft and ping you when ready.

roroghost17: sure

PR #3441, opened 2026-05-12
fix: anthropic stream tool call translation

d3lm:
## Summary

Fixes several Anthropic-compatible streaming response translation issues when requests to `/anthropic/v1/messages` are routed to OpenAI-compatible providers. The changes make streamed tool calls emit valid Anthropic SSE sequences by preventing overlapping content blocks, preserving tool-call stop reasons, closing tool blocks after their streamed arguments, and normalizing provider-specific tool IDs.

## Steps to Reproduce

1. Start Bifrost
2. Configure an OpenAI compatible provider that supports OSS models such as Kimi K2.6, e.g., Baseten
3. Configure a routing rule to send requests to that provider (e.g., `model == 'moonshotai/Kimi-K2.6'`)
4. Use Bifrost with Claude Code (you may also need to set the API key if you're not logged in):
```bash
ANTHROPIC_BASE_URL=http://localhost:8080/anthropic ANTHROPIC_MODEL="moonshotai/Kimi-K2.6" ANTHROPIC_SMALL_FAST_MODEL="moonshotai/Kimi-K2.6" claude
```
5. Send a message that triggers tool calls, e.g., `Explore this code base in depth`

## Changes

- Fixed chat-completions-to-responses stream conversion to process all streamed OpenAI tool calls, not only the first one.
- Ensured tool-only streams start at Anthropic content block index `0`, while mixed text + tool streams reserve index `0` for text and place tools after it.
- Ensured every started tool call emits a matching done event, including zero-argument tool calls.
- Preserved `finish_reason: "tool_calls"` through the internal Responses stream event so the Anthropic response emits `stop_reason: "tool_use"`.
- Updated Anthropic stream rendering so tool-call `content_block_stop` uses the same output index as the corresponding start and delta events.
- Normalized OpenAI-compatible tool call IDs into Anthropic-safe `toolu_...` IDs while preserving native Anthropic IDs.
- Added stream-state tracking so Anthropic content blocks are emitted as non-overlapping `content_block_start` → deltas → `content_block_stop` sequences.
- Closed text/reasoning blocks from Responses stream completion events before starting subsequent tool blocks, while avoiding duplicate stop events when `output_item.done` arrives later.
- Scoped generated stream `toolu_...` IDs by assistant message/output index so providers that reuse raw IDs like `functions.Bash:0` do not produce repeated Anthropic tool-use IDs across turns.
- Omitted `stop_sequence` from content deltas when nil to better match Anthropic streaming payloads.
- Added regression tests covering single tool calls, streamed tool JSON, multiple tool calls, zero-argument tools, ID normalization, per-message stream tool IDs, block closure, non-overlapping text/tool block ordering, and final Anthropic stop reason.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [x] Core (Go)
- [ ] Transports (HTTP)
- [x] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

Run the focused regression tests:

```sh
go test github.com/maximhq/bifrost/core/providers/anthropic -run 'Test(NormalizeAnthropicToolUseID|OpenAIChatToolStreamFallbackToAnthropicCloses(ToolBlock|MultipleToolBlocks)|ToAnthropicResponsesStreamResponse_(DelaysToolStopUntilAfterLateArgumentDeltas|ClosesTextBeforeStartingToolFromResponsesEvents|NormalizesRepeatedProviderToolIDsPerMessage))' -count=1
```

Expected outcome: tests pass.

Run the Anthropic provider unit tests:

```sh
go test github.com/maximhq/bifrost/core/providers/anthropic -count=1
```

Current local note: this package-level command currently fails on an unrelated existing panic in `TestToAnthropicChatRequest_ServerTools` where the test calls `ToAnthropicChatRequest(nil, ...)`. The focused stream regression suite above passes.

Run the affected schema conversion tests:

```sh
go test ./core/schemas -count=1
```

Expected outcome: tests pass.

Run the Anthropic provider test harness:

```sh
make test-core PROVIDER=anthropic
```

Expected outcome: Anthropic provider tests run through the standard core provider test harness.

Manual validation note:

- A captured Claude Code request with repeated `Read` calls showed each repeated tool call had a fresh Anthropic `toolu_...` ID and a matching `tool_result`.

## Screenshots/Recordings

N/A. No UI changes.

## Breaking changes

- [ ] Yes
- [x] No

## Related issues

N/A

## Security considerations

No new security implications. This change only adjusts streaming response translation and adds regression coverage; it does not affect auth, secrets, PII handling, sandboxing, or permission boundaries.

## Checklist

### DIFF
--- a/core/providers/anthropic/responses.go
+++ b/core/providers/anthropic/responses.go
@@ -2,9 +2,12 @@ package anthropic
 
 import (
 	"context"
+	"crypto/sha256"
+	"encoding/hex"
 	"encoding/json"
 	"fmt"
 	"math"
+	"sort"
 	"strings"
 	"sync"
 	"time"
@@ -83,11 +86,44 @@ type anthropicToResponsesStreamState struct {
 	// webSearchItemIDs tracks item IDs for WebSearch tools so their argument deltas
 	// can be skipped and regenerated synthetically (with sanitization) at output_item.done.
 	webSearchItemIDs map[string]bool
+
+	// openContentBlockIndices tracks Anthropic content blocks that have started
+	// but have not emitted content_block_stop yet. Anthropic clients expect blocks
+	// to be strictly nested as start -> deltas -> stop, with no overlap.
+	openContentBlockIndices   map[int]bool
+	closedContentBlockIndices map[int]bool
+	openToolBlockIndices      map[int]bool
+	sawToolUse                bool
+
+	// currentMessageID scopes generated Anthropic-safe tool IDs for streaming.
+	// Some OpenAI-compatible providers reuse raw IDs like functions.Bash:0 across
+	// turns, but Anthropic clients expect each streamed tool_use ID to be unique.
+	currentMessageID string
+	streamToolUseIDs map[string]string
 }
 
 type anthropicToResponsesStreamStateKeyType struct{}
 
 var anthropicToResponsesStreamStateKey = anthropicToResponsesStreamStateKeyType{}
+
+const anthropicToolUseIDPrefix = "toolu_"
+
+func normalizeAnthropicToolUseID(id *string, scope *string) *string {
+	if id == nil || *id == "" || strings.HasPrefix(*id, anthropicToolUseIDPrefix) {
+		return id
+	}
+
+	seed := *id
+	if scope != nil && *scope != "" {
+		seed = *scope + ":" + seed
+	}
+
+	sum := sha256.Sum256([]byte(seed))
+	encoded := hex.EncodeToString(sum[:])
+	normalized := anthropicToolUseIDPrefix + encoded[:24]
+
+	return &normalized
+}
 
 // getOrCreateAnthropicToResponsesStreamState returns the per-request conversion state,
 // creating and storing it in ctx on first access.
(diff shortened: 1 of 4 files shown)


core/providers/anthropic/responses.go
```diff
@@ -4260,9 +4480,7 @@ func convertBifrostFunctionCallToAnthropicToolUse(ctx *schemas.BifrostContext, m
 			CacheControl: msg.CacheControl,
 		}
 
-		if msg.ResponsesToolMessage.CallID != nil {
-			toolUseBlock.ID = msg.ResponsesToolMessage.CallID
-		}
+		toolUseBlock.ID = normalizeAnthropicToolUseID(msg.ResponsesToolMessage.CallID)
```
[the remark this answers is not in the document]
d3lm: Addressed this.

core/providers/anthropic/responses.go
```diff
@@ -83,12 +86,45 @@ type anthropicToResponsesStreamState struct {
 	// webSearchItemIDs tracks item IDs for WebSearch tools so their argument deltas
 	// can be skipped and regenerated synthetically (with sanitization) at output_item.done.
 	webSearchItemIDs map[string]bool
+
+	// openContentBlockIndices tracks Anthropic content blocks that have started
+	// but have not emitted content_block_stop yet. Anthropic clients expect blocks
+	// to be strictly nested as start -> deltas -> stop, with no overlap.
+	openContentBlockIndices   map[int]bool
```
akshaydeo, line 93: change it to `map[int]struct{}` struct allocates 0 memory

d3lm: Wouldn't the benefit be negligible unless these maps grow large or become very hot? But I can certainly change it if you want.

akshaydeo: We tend to save every last bit we can 🙂 - bool is still 1 byte. <br>You have 100 concurrent streams + 50 chunks =  5000 bytes. At scale it just goes ugly (as you scale I mean)

core/providers/anthropic/responses.go
```diff
@@ -1637,7 +1835,18 @@ func ToAnthropicResponsesStreamResponse(ctx *schemas.BifrostContext, bifrostResp
...
+			streamState := getOrCreateAnthropicToResponsesStreamState(ctx)
+			events = append(events, streamState.flushOpenContentBlockStopsBefore(streamResp.Index)...)
+		}
 		events = append(events, streamResp)
+		if streamResp.Type == AnthropicStreamEventTypeContentBlockStart && streamResp.ContentBlock != nil {
+			streamState := getOrCreateAnthropicToResponsesStreamState(ctx)
+			isToolBlock := streamResp.ContentBlock.Type == AnthropicContentBlockTypeToolUse ||
+				streamResp.ContentBlock.Type == AnthropicContentBlockTypeMCPToolUse ||
+				streamResp.ContentBlock.Type == AnthropicContentBlockTypeServerToolUse
+			streamState.trackOpenContentBlock(streamResp.Index, isToolBlock)
```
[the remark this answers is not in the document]
akshaydeo: can you check this as well once?

d3lm:
Closing this out after further investigation.

I originally opened this because Claude Code was failing through the Anthropic-compatible `/anthropic` route when routed to an OpenAI-compatible provider. The failure looked like invalid tool parameters during streaming tool calls.

After comparing the raw stream against OpenAI 1P GPT Responses streaming, this appears to be a provider-specific issue rather than a Bifrost translation issue.

With OpenAI 1P, function call streaming is ordered as expected:

- `response.output_item.added`
- `response.function_call_arguments.delta`
- `response.function_call_arguments.done` with the complete arguments
- `response.output_item.done` with the complete arguments

The failing provider was emitting `response.function_call_arguments.done` and `response.output_item.done` with empty arguments, then sending additional `response.function_call_arguments.delta` events afterward for the same tool call item. That causes Claude Code to treat the tool call as complete before the real arguments arrive.

Since this does not reproduce with OpenAI 1P and appears to violate the expected Responses stream lifecycle for a single function call item, I’m treating this as an upstream provider compatibility bug rather than something Bifrost should patch here.

PR #3491, opened 2026-05-14
feat: rotate keys on 401/402/403 and return `502 upstream_credentials_exhausted` when all keys are permanently dead

### LINKED ISSUE #3696: [Feature]: Make retry backoff context-aware and honor Retry-After for rate limits
### Prerequisites

### Problem to solve

Bifrost already has retry/backoff behavior, but a few reliability paths can keep work alive longer than needed or retry too early during provider throttling:

- retry backoff appears to use `time.Sleep(backoff)` directly, so cancelled or expired requests may still sleep before exiting
- `429` responses are retryable, but provider `Retry-After` hints do not appear to be honored in the retry scheduler
- HTTP request contexts are cancellable, but do not appear to enforce a strict end-to-end timeout budget across retry/fallback lifecycle
- in-flight provider calls may continue after context cancellation depending on the request path
- missing tracer context appears to fail the request instead of falling back to no-op tracing

This can affect production users during provider slowness, rate limits, or client disconnects by increasing worker occupancy, retry pressure, and tail latency.

### Proposed solution

A small reliability-focused improvement could:

- replace retry `time.Sleep(...)` with context-aware backoff waiting
- parse and honor provider `Retry-After` hints for retryable `429` responses
- add or prepare retry budget handling so retries do not exceed request deadlines
- improve cancellation behavior around provider call retries
- treat missing tracer as degraded observability instead of a hard serving failure where safe

The goal is to reduce retry storms, cancelled-request resource usage, and long-tail latency without changing provider routing behavior.

### Alternatives considered

The alternative is to leave retry timing entirely to local exponential backoff and provider/network timeouts.

That works in simple cases, but during shared provider throttling or client cancellation, it can cause unnecessary retry pressure and keep workers occupied after the request is no longer useful.

### Area(s)

Core (Go)

### Additional context

_No response_

Pratham-Mishra04:
## Summary

Key rotation during retries previously treated all per-key failures the same way. This PR distinguishes between **permanent** per-key failures (401/402/403 — bad credentials, billing issues, missing permissions) and **transient** per-key failures (429 — rate limits). Keys that fail permanently are tracked in a separate `deadKeyIDs` set that is never reset within a single request, preventing Bifrost from cycling back to a broken credential. Keys that hit a rate limit continue to be tracked in `usedKeyIDs`, which can be reset once all non-dead keys are exhausted, since a previously rate-limited key may have recovered by then.

Additionally, backoff sleep is now skipped when rotating to a different key — waiting before trying a fresh credential provides no benefit.

## Changes

- Introduced `keyRotateStatusCodes` (401, 402, 403, 429) as a distinct set from `retryableStatusCodes` (500–504), and removed 429 from `retryableStatusCodes`.
- Added a `deadKeyIDs` parameter to `keyProvider` alongside the existing `usedKeyIDs`. Dead keys are never reconsidered within a request; used (rate-limited) keys may be reconsidered after a full rotation cycle.
- When `keyProvider` exhausts all keys and every remaining key is in `deadKeyIDs`, the error is surfaced as a `502 upstream_credentials_exhausted` rather than a raw 401/403, which would falsely imply the caller's Bifrost API key is invalid.
- Backoff is only applied when retrying the same key against a transient server error. Key-rotation retries skip the sleep.
- Attempt trail `FailReason` now uses explicit status-code-derived labels (`rate_limit_error`, `authentication_error`, `billing_error`) for rotation-triggering codes rather than the provider's error type string, which can be misleading (e.g. OpenAI returns `invalid_request_error` for a 401).
- Updated tests to cover `keyRotateStatusCodes`, verify 429 is absent from `retryableStatusCodes`, and align `keyProvider` signatures.

## Type of change

- [x] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [x] Core (Go)
- [ ] Transports (HTTP)
- [ ] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

```sh
go version
go test ./...
```

Scenarios to validate manually:
- A pool of keys where one has a revoked credential (401): that key should never be retried, and the request should succeed on a valid key without delay.
- A pool where all keys return 401/403: the response should be `502 upstream_credentials_exhausted`, not a 401.
- A pool where keys are rate-limited (429): rotation should cycle through all keys, reset, and retry without treating any key as permanently dead.
- A 500 from the provider: the same key should be reused with backoff applied.

## Breaking changes

- [x] Yes
- [ ] No

The `keyProvider` function signature has changed from `func(usedKeyIDs map[string]bool)` to `func(usedKeyIDs, deadKeyIDs map[string]bool)`. Any custom `keyProvider` implementations outside this codebase will need to be updated to accept the second parameter.

## Security considerations

Permanent credential failures (401/402/403) are now isolated immediately and never retried, reducing unnecessary exposure of invalid credentials to upstream providers.

## Checklist

### DIFF
--- a/core/bifrost.go
+++ b/core/bifrost.go
@@ -5,6 +5,7 @@ package bifrost
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"slices"
 	"sort"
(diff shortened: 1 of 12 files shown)


Pratham-Mishra04:
* **#3491** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/3491?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/maximhq/bifrost/3491?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3430** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/3430?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `dev`


<h2></h2>

akshaydeo (dismissed):

akshaydeo:
### Merge activity

* **May 28, 6:15 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/3491).
* **May 28, 6:17 AM UTC**: [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/3491) rebased this pull request as part of a merge.
* **May 28, 6:18 AM UTC**: @akshaydeo merged this pull request with [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/3491).

darshandagaa:
system overview --> bifrost upstream - hosted bifrost service - ALB - datadog
because of this PR we are seeing a lot of failed requests (4xx) being sent back as 5XX. Any views on this ?

PR #3521, opened 2026-05-15
[fix]: allow model catalog URLs (pricing + parameters) to be overridden via env vars

dsherniiazov:
**Description**

This PR fixes an issue where custom model catalog URLs (especially for model parameters) could not be properly overridden in certain deployment scenarios (Docker, Helm, air-gapped environments, etc.).

### Changes
- Added support for two new environment variables:
  - `BIFROST_PRICING_URL`
  - `BIFROST_MODEL_PARAMETERS_URL`
- Refactored `defaultURLWithEnv` helper in `framework/modelcatalog`
- Updated `ModelCatalog` initialization and sync logic to respect these env vars
- Added comprehensive tests for the new behavior
- Updated config resolution in `transports/bifrost-http`

### Why
Previously the model parameters catalog always fell back to the hardcoded default URL, making it impossible to use a private/internal catalog without forking or complex config workarounds.  
This change makes the model catalog fully customizable via environment variables (consistent with how other URLs like schema are handled).

### Affected packages
- `core/framework`
- `transports/bifrost-http`

### Testing
- Added unit tests for `defaultURLWithEnv`
- Verified that both pricing and model parameters URLs respect env overrides
- Existing functionality remains unchanged

**Type of Change**: Bug fix / Improvement

**Checklist**
- [x] Changelog entry added to `core/changelog.md`
- [x] Tests added/updated
- [ ] Documentation updated (if needed)


akshaydeo:
Hi can you check all the comments :bow:

BearTS:
hey @dsherniiazov we do support overriding pricing url and model parameter url using config json
Recently, I worked on fixing a bug regarding config hash mechanism  https://github.com/maximhq/bifrost/pull/3610 surrounding it.

PR #3549, opened 2026-05-17
updated budgets and limits documentation to indicate the multi-node limitation for oss

### LINKED ISSUE #3547: Budget & rate-limit state is per-replica (sync.Map); multi-replica deployments silently leak past caps
# Budget & rate-limit state is per-replica `sync.Map`, not shared — multi-replica deployments silently leak past caps

**Bifrost version:** v1.5.2 (transport tag `transports/v1.5.2`)

## Summary

The governance store that backs budget enforcement and rate limiting (`customers`, `budgets`, `rateLimits`) is held in process-local Go `sync.Map`s with no cross-replica coordination. There is no Redis / NATS / pubsub layer, no leader election, and no shared cache. PostHook updates are written to one replica's in-memory store and never propagated.

This means a multi-replica deployment can let a customer or VK spend N times their cap (where N = replica count) before the eventual DB sync catches up — and the DB sync is itself best-effort (the per-request usage delta lives in memory until the next aggregation run). For self-serve SaaS that depends on hard per-user caps, this is a quietly load-bearing behaviour that isn't called out anywhere in the docs.

### The store is a process-local sync.Map

[`plugins/governance/store.go:29-31`](https://github.com/maximhq/bifrost/blob/transports/v1.5.2/plugins/governance/store.go#L29-L31):

```go
type governanceStore struct {
    customers    sync.Map // string -> *Customer (Customer ID -> Customer)
    budgets      sync.Map // string -> *Budget   (Budget ID -> Budget)
    rateLimits   sync.Map // string -> *RateLimit (RateLimit ID -> RateLimit)
    ...
}
```

`sync.Map` is concurrency-safe within one process. It has no relationship to anything outside that process.

### The per-request increment is in-memory CAS

[`plugins/governance/store.go:380-389`](https://github.com/maximhq/bifrost/blob/transports/v1.5.2/plugins/governance/store.go#L380-L389):

```go
clone := *raw
clone.CurrentUsage += cost
if gs.budgets.CompareAndSwap(budgetID, raw, &clone) {
    return nil
}
```

This runs on every chat-completion success via the PostHook. It mutates only the local replica's `sync.Map`. The CAS retry handles concurrent goroutines on the SAME replica; it has no effect on sibling replicas.

### There is no cross-replica coordination

Repo-wide grep at the tag `transports/v1.5.2` for any cross-process coordination primitive:

| Term searched | Hits in production Go code |
|---|---|
| `redis` | 0 |
| `nats` | 0 |
| `pubsub` | 0 |
| `cluster` / `replica` | 0 |
| leader election / consistent hashing | 0 |

The only ticker in `plugins/governance/tracker.go` is the [`workerInterval = 10 * time.Second`](https://github.com/maximhq/bifrost/blob/transports/v1.5.2/plugins/governance/tracker.go#L49) reset-window worker — it rolls budget windows over when `reset_duration` elapses, but does NOT push usage to or pull usage from any shared store. `config_hash` exists per row but is row-version metadata, not consulted on the request hot path for usage sync.

### `?from_memory=true` is the only read that reflects the live state

[`tests/governance/usagetracking_test.go`](https://github.com/maximhq/bifrost/blob/transports/v1.5.2/tests/governance/usagetracking_test.go) — bifrost's own tests assert usage by reading `/api/governance/budgets?from_memory=true` after a 2-3 second wait. The persistent DB read returns stale values. So the system's own test suite relies on the in-memory state being authoritative.

## Multi-replica leak scenario

Bifrost replicas A and B, customer `cust-foo` with `max_limit: $5`, `current_usage` at $4.99 in the DB at startup:

1. T+0: User makes request → load balancer routes to A. A reads memory ($4.99), permits the call, PostHook writes A's memory to $5.01.
2. T+0ms: Same user makes a concurrent request → LB routes to B. B's memory still says $4.99 (or whatever it was at B's last DB load), permits the call, PostHook writes B's memory to $5.01.
3. T+10s+: Both replicas write their own `current_usage` to the DB. The DB now reflects whichever wrote last (read-modify-write race), not the sum. The user has spent ~$5.02 on a $5 cap.

For long-lived caps and small `cost` per request, the leak compounds: each replica thinks budget is OK whenever the next request lands on a replica whose memory hasn't been bumped recently.

## Why this matters

Most users running Bifrost behind a load balancer would reasonably assume budget enforcement is consistent across replicas. The docs at [Budget and Limits](https://docs.getbifrost.ai/features/governance/budget-and-limits) describe the cascade (provider config → VK → customer) but don't note that the cascade is per-replica only. That's a footgun for any SaaS deployment with hard per-tenant caps.

Specifically affects:
- Per-end-user budget enforcement (the `customer_id` + customer-level budget pattern the governance docs explicitly recommend for multi-tenant deployments).
- Rate limiting (same `sync.Map` model).
- Free-tier / trial-budget enforcement where the cap is small and the leak ratio is significant.

## Suggested fix

A few options, in increasing order of architectural change:

1. **Document it.** At minimum, add a "Deployment topology" page noting that budget/rate-limit enforcement is single-replica and recommending sticky LB or single-replica deployment for hard-cap use cases.
2. **Optional Redis backend for the governance store.** Make `sync.Map` swappable for a Redis-backed implementation (similar to how the configstore already supports Postgres). Atomic INCR semantics on Redis would give consistent across-replica increments.
3. **Optional pubsub for cache invalidation.** Lighter-weight: each replica's `sync.Map` remains local but a pubsub channel pushes "current_usage updated on customer X, value Y" so other replicas catch up within milliseconds. Trades absolute consistency for ~ms-window drift.

For our use case option 2 looks cleanest from the outside — atomic `INCRBYFLOAT` against Redis would mirror the existing `CompareAndSwap` semantics with global visibility — but you'd know the trade-offs of the internals far better than we would.

## Workaround

We're going to single-replica Bifrost for now and treating horizontal scale as blocked on this issue.

akshaydeo:
## Summary

Adds a warning to the budget and limits documentation clarifying that running multiple OSS Bifrost nodes with a Postgres backend is not supported, and explains why this is an intentional architectural boundary between the OSS and Enterprise offerings.

## Issues

closes #3547

## Changes

- Added a `<Warning>` block to `budget-and-limits.mdx` explaining that Bifrost keeps all critical state (provider configs, API keys, budgets, usage, traffic distribution) in memory and does not re-read it from the database after initialization.
- Documents that multi-node synchronization is handled in the Enterprise version via a modified RAFT consensus mechanism, while the database serves only as a passive store.
- Sets expectations for OSS users (~3,000–5,000 RPS on a single instance) and directs high-availability needs toward the Enterprise plan.

## Type of change

- [ ] Bug fix
- [ ] Feature
- [ ] Refactor
- [x] Documentation
- [ ] Chore/CI

## Affected areas

- [ ] Core (Go)
- [ ] Transports (HTTP)
- [ ] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [x] Docs

## How to test

Review the rendered documentation to confirm the warning block displays correctly and the content is accurate.

## Breaking changes

- [ ] Yes
- [x] No

## Related issues

N/A

## Security considerations

None.

## Checklist

### DIFF
--- a/.github/workflows/scripts/validate-helm-config-fields.sh
+++ b/.github/workflows/scripts/validate-helm-config-fields.sh
@@ -869,7 +869,7 @@ assert_field_value 'cluster_config.discovery.k8s_label_selector' '.cluster_confi
 # Gap 7: Cluster region
 assert_field_value 'cluster_config.region' '.cluster_config.region' '"us-east-1"'
 
-# SCIM - Okta
+# SCIM - Okta (baseline + attribute mappings)
 cat > "$TMPDIR/values-scim-okta.yaml" << 'VALS'
 image:
   tag: v1.0.0
@@ -886,6 +886,18 @@ bifrost:
       userIdField: "sub"
       teamIdsField: "groups"
       rolesField: "roles"
+      attributeRoleMappings:
+        - attribute: "groups"
+          value: "bifrost-admins"
+          role: "admin"
+      attributeTeamMappings:
+        - attribute: "groups"
+          value: "*"
+          team: ""
+      attributeBusinessUnitMappings:
+        - attribute: "department"
+          value: "platform"
+          business_unit: "Platform"
 VALS
 
 render_config "$TMPDIR/values-scim-okta.yaml"
@@ -894,7 +906,14 @@ assert_field_value 'scim_config.provider' '.scim_config.provider' '"okta"'
 assert_field 'scim_config.config' '.scim_config.config'
 assert_field 'scim_config.config.apiToken' '.scim_config.config.apiToken'
 assert_field 'scim_config.config.clientSecret' '.scim_config.config.clientSecret'
-# SCIM - Entra
+assert_field_value 'scim_config (okta) attributeRoleMappings[0].attribute' '.scim_config.config.attributeRoleMappings.[0].attribute' '"groups"'
+assert_field_value 'scim_config (okta) attributeRoleMappings[0].value' '.scim_config.config.attributeRoleMappings.[0].value' '"bifrost-admins"'
+assert_field_value 'scim_config (okta) attributeRoleMappings[0].role' '.scim_config.config.attributeRoleMappings.[0].role' '"admin"'
+assert_field_value 'scim_config (okta) attributeTeamMappings[0].attribute' '.scim_config.config.attributeTeamMappings.[0].attribute' '"groups"'
+assert_field_value 'scim_config (okta) attributeTeamMappings[0].value' '.scim_config.config.attributeTeamMappings.[0].value' '"*"'
+assert_field_value 'scim_config (okta) attributeBusinessUnitMappings[0].business_unit' '.scim_config.config.attributeBusinessUnitMappings.[0].business_unit' '"Platform"'
+
+# SCIM - Entra (baseline + attribute mappings)
 cat > "$TMPDIR/values-scim-entra.yaml" << 'VALS'
 image:
   tag: v1.0.0
@@ -911,6 +930,18 @@ bifrost:
       userIdField: "oid"
       teamIdsField: "groups"
       rolesField: "roles"
+      attributeRoleMappings:
+        - attribute: "roles"
+          value: "BifrostAdmin"
+          role: "admin"
+      attributeTeamMappings:
+        - attribute: "groups"
+          value: "11111111-2222-3333-4444-555555555555"
+          team: "platform-team"
+      attributeBusinessUnitMappings:
+        - attribute: "department"
+          value: "Platform"
+          business_unit: "Platform BU"
 VALS
 
 render_config "$TMPDIR/values-scim-entra.yaml"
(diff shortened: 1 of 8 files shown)


akshaydeo:
* **#3550** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/3550?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3549** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/3549?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/maximhq/bifrost/3549?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3544** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/3544?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `dev`


<h2></h2>

akshaydeo:
### Merge activity

* **May 17, 9:18 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/3549).
* **May 17, 9:19 AM UTC**: @akshaydeo merged this pull request with [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/3549).

PR #3550, opened 2026-05-17
add budget and rate-limit dump for budget tracker

### LINKED ISSUE #3548: UsageTracker.Cleanup() doesn't flush in-memory usage to DB on shutdown
# `UsageTracker.Cleanup()` doesn't flush in-memory usage to DB on shutdown — every graceful restart loses up to ~10s of accumulated usage

**Bifrost version:** v1.5.2 (transport tag `transports/v1.5.2`)

## Summary

`UsageTracker` accumulates per-request budget/rate-limit deltas in process-local `sync.Map`s and dumps them to Postgres every `workerInterval = 10 * time.Second` via the ticker-driven `resetExpiredCounters` worker. On graceful shutdown the worker stops, but `Cleanup()` does not perform a final flush — any deltas accumulated since the last tick die with the process.

Net effect: every graceful restart (rolling deploy, helm upgrade, HPA scale-down) loses up to ~10 seconds of accumulated budget usage. Crash is identical, just without the orderly worker shutdown. The window is small but the loss is silent — there's no log line, no metric, and no way for an operator to tell after the fact what was dropped.

### The 10-second dump

[`plugins/governance/tracker.go:49`](https://github.com/maximhq/bifrost/blob/transports/v1.5.2/plugins/governance/tracker.go#L49):

```go
const (
    workerInterval = 10 * time.Second
)
```

[`plugins/governance/tracker.go:168-189`](https://github.com/maximhq/bifrost/blob/transports/v1.5.2/plugins/governance/tracker.go#L168-L189) — the ticker handler:

```go
// resetExpiredCounters manages periodic resets of usage counters AND budgets ...
func (t *UsageTracker) resetExpiredCounters(ctx context.Context) {
    // PART 1: reset rate limits
    // PART 2: reset budgets
    // ==== PART 3: Dump all rate limits and budgets to database ====
    if err := t.store.DumpRateLimits(ctx, nil, nil); err != nil { ... }
    if err := t.store.DumpBudgets(ctx, nil); err != nil { ... }
}
```

`DumpBudgets` direct-`UPDATE`s `current_usage` for every in-memory budget. Between ticks, the new deltas exist only in memory.

### Cleanup stops the worker but doesn't flush

[`plugins/governance/tracker.go:305-321`](https://github.com/maximhq/bifrost/blob/transports/v1.5.2/plugins/governance/tracker.go#L305-L321):

```go
// Cleanup stops all background workers and flushes pending operations
func (t *UsageTracker) Cleanup() error {
    if t.trackerCancel != nil {
        t.trackerCancel()
    }
    close(t.done)
    if t.resetTicker != nil {
        t.resetTicker.Stop()
    }
    t.wg.Wait()

    t.logger.Debug("usage tracker cleanup completed")
    return nil
}
```

The doc-comment says "flushes pending operations" but the body doesn't actually call `DumpBudgets` or `DumpRateLimits`. The comment and behaviour disagree.

## Reproducing

1. Start bifrost OSS against Postgres.
2. Create a customer with a budget.
3. Make a chat-completion request that costs e.g. $0.01. Wait less than 10s.
4. SIGTERM bifrost before the next tick fires.
5. Inspect the budget row in Postgres — `current_usage` is still the pre-call value. The $0.01 delta is gone.

## Why this matters

For most workloads the loss is small (whatever the replica was billing in the seconds before shutdown), but two scenarios hurt:

1. **Rolling deploys with frequent restarts.** Each replica restart drops up to 10s of in-flight budget tracking. A `kubectl rollout restart` on a 3-replica deployment drops up to 30s of cluster-wide usage across the rollout.
2. **Free-tier enforcement.** A $1 free trial where the user has spent $0.99 followed by a restart followed by another call all happening in the same 10s window: the user effectively gets $0.99 of phantom spend "back." Probably tolerable, but worth knowing.

The doc-comment on `Cleanup()` ("flushes pending operations") suggests this was intended to flush — the implementation just doesn't.

## Suggested fix

Add a final synchronous flush at the start of `Cleanup()`, before tearing down workers:

```go
func (t *UsageTracker) Cleanup() error {
    // Final flush of in-memory deltas to DB before shutdown. Without this,
    // any deltas accumulated since the last `workerInterval` tick are lost.
    if err := t.store.DumpBudgets(context.Background(), nil); err != nil {
        t.logger.Error("final budget dump on shutdown failed: %v", err)
    }
    if err := t.store.DumpRateLimits(context.Background(), nil, nil); err != nil {
        t.logger.Error("final rate-limit dump on shutdown failed: %v", err)
    }

    // existing shutdown sequence...
    if t.trackerCancel != nil { t.trackerCancel() }
    close(t.done)
    if t.resetTicker != nil { t.resetTicker.Stop() }
    t.wg.Wait()

    t.logger.Debug("usage tracker cleanup completed")
    return nil
}
```

Notes:
- Use `context.Background()` here intentionally; the inherited tracker context is cancelled by `t.trackerCancel()` and we want the flush to complete even past that cancellation.
- The flush runs before `t.wg.Wait()` so it isn't racing the worker goroutine — by the time we reach the wait, the worker has already been told to stop and the in-memory map is stable from the caller's perspective.

In enterprise/cluster mode this is less critical (gossip propagates deltas to peers sub-second, so the cluster as a whole loses much less), but the flush is still a small win on graceful shutdown — peers don't have to gossip the final delta forward, the DB just has it.

## Related

- This is a follow-up to #3547 (multi-replica budget state) — pre-existing discussion of how OSS handles cross-replica budgets. The Cleanup gap is independent: it affects single-replica OSS deployments too.

akshaydeo:
## Summary

During shutdown, any in-memory budget and rate-limit deltas accumulated since the last background worker tick were silently dropped. This change ensures those deltas are flushed to the database before the tracker stops.

## Issues

Closes #3548

## Changes

- Added a final call to `DumpBudgets` and `DumpRateLimits` at the start of `Cleanup()`, so in-flight deltas are persisted before background workers are cancelled.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [ ] Core (Go)
- [ ] Transports (HTTP)
- [ ] Providers/Integrations
- [x] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

```sh
go test ./plugins/governance/...
```

Verify that after a graceful shutdown, budget and rate-limit values written between the last periodic flush and shutdown are present in the database rather than lost.

## Breaking changes

- [ ] Yes
- [x] No

## Security considerations

None.

## Checklist

### DIFF
--- a/.github/workflows/scripts/validate-helm-config-fields.sh
+++ b/.github/workflows/scripts/validate-helm-config-fields.sh
@@ -869,7 +869,7 @@ assert_field_value 'cluster_config.discovery.k8s_label_selector' '.cluster_confi
 # Gap 7: Cluster region
 assert_field_value 'cluster_config.region' '.cluster_config.region' '"us-east-1"'
 
-# SCIM - Okta
+# SCIM - Okta (baseline + attribute mappings)
 cat > "$TMPDIR/values-scim-okta.yaml" << 'VALS'
 image:
   tag: v1.0.0
@@ -886,6 +886,18 @@ bifrost:
       userIdField: "sub"
       teamIdsField: "groups"
       rolesField: "roles"
+      attributeRoleMappings:
+        - attribute: "groups"
+          value: "bifrost-admins"
+          role: "admin"
+      attributeTeamMappings:
+        - attribute: "groups"
+          value: "*"
+          team: ""
+      attributeBusinessUnitMappings:
+        - attribute: "department"
+          value: "platform"
+          business_unit: "Platform"
 VALS
 
 render_config "$TMPDIR/values-scim-okta.yaml"
@@ -894,7 +906,14 @@ assert_field_value 'scim_config.provider' '.scim_config.provider' '"okta"'
 assert_field 'scim_config.config' '.scim_config.config'
 assert_field 'scim_config.config.apiToken' '.scim_config.config.apiToken'
 assert_field 'scim_config.config.clientSecret' '.scim_config.config.clientSecret'
-# SCIM - Entra
+assert_field_value 'scim_config (okta) attributeRoleMappings[0].attribute' '.scim_config.config.attributeRoleMappings.[0].attribute' '"groups"'
+assert_field_value 'scim_config (okta) attributeRoleMappings[0].value' '.scim_config.config.attributeRoleMappings.[0].value' '"bifrost-admins"'
+assert_field_value 'scim_config (okta) attributeRoleMappings[0].role' '.scim_config.config.attributeRoleMappings.[0].role' '"admin"'
+assert_field_value 'scim_config (okta) attributeTeamMappings[0].attribute' '.scim_config.config.attributeTeamMappings.[0].attribute' '"groups"'
+assert_field_value 'scim_config (okta) attributeTeamMappings[0].value' '.scim_config.config.attributeTeamMappings.[0].value' '"*"'
+assert_field_value 'scim_config (okta) attributeBusinessUnitMappings[0].business_unit' '.scim_config.config.attributeBusinessUnitMappings.[0].business_unit' '"Platform"'
+
+# SCIM - Entra (baseline + attribute mappings)
 cat > "$TMPDIR/values-scim-entra.yaml" << 'VALS'
 image:
   tag: v1.0.0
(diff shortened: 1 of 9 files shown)


akshaydeo:
* **#3550** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/3550?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/maximhq/bifrost/3550?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3549** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/3549?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3544** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/3544?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `dev`


<h2></h2>

akshaydeo:
### Merge activity

* **May 17, 9:18 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/3550).
* **May 17, 9:20 AM UTC**: @akshaydeo merged this pull request with [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/3550).

PR #3569, opened 2026-05-18
fix(configstore): change config_keys.name unique index to composite (provider_id, name)

SahilChoudhary22:
## Summary

Changes the unique index on `config_keys.name` from a global unique index to a composite unique index on `(provider_id, name)`. This allows different providers to each have a key with the same name (e.g. each provider can have a `"default"` key), fixing a bug where providers with duplicate key names would collide and one provider's keys would silently overwrite another's during startup sync.

## Changes

- **tables/key.go**: Changed `Name` field GORM tag from `uniqueIndex:idx_key_name` to `uniqueIndex:idx_key_provider_name`, and added `uniqueIndex:idx_key_provider_name` to the `ProviderID` field, creating a composite unique index
- **migrations.go**: Added `migrationChangeKeyNameToCompositeUniqueIndex` migration that safely drops the old global `idx_key_name` index and creates the new composite `idx_key_provider_name` index on `(provider_id, name)`. Includes a rollback that reverses both operations.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [x] Core (Go)
- [ ] Transports (HTTP)
- [ ] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

Run the migration and verify the new index exists:

```sh
PGPASSWORD=bifrost_password psql -h localhost -p 5432 -U bifrost -d bifrost -c "\d config_keys"
```

Expected output shows `idx_key_provider_name` UNIQUE index on `(provider_id, name)`.

Existing tests pass:
```sh
cd framework/configstore && go test ./...
```

## Screenshots/Recordings

N/A — No UI changes.

## Breaking changes

- [ ] Yes
- [x] No

The migration is fully backward-compatible — it drops the old index and creates the new one without touching any data. The composite index does not reject any rows the old global index would have accepted.

## Related issues

Fixes a bug where updating a provider configuration via the UI would return HTTP 500 with `"a record with this name already exists"` when multiple providers had keys with the same name.

## Security considerations

None — this is a schema constraint change only, no new data or secrets introduced.

## Checklist

- [x] I read `docs/contributing/README.md` and followed the guidelines
- [x] I added/updated tests where appropriate
- [x] I updated documentation where needed
- [x] I verified builds succeed (Go and UI)
- [x] I verified the CI pipeline passes locally if applicable

## Why this matters

[...]

PR #3570, opened 2026-05-18
Fix cross-provider key name collision in fallback lookup

SahilChoudhary22:
## Summary

Fixes the startup sync (`UpdateProvidersConfig`) to scope key name lookups by `provider_id`. Previously, when the sync encountered a key whose UUID didn't match any existing row, it fell back to a global name lookup (`Where("name = ?")`) that could match a key belonging to a *different* provider, silently overwriting it.

The fix adds `AND provider_id = ?` to the name fallback query, ensuring it only matches keys within the same provider. This aligns with the new composite unique index `(provider_id, name)` introduced in the sibling PR, where key names are unique per-provider rather than globally.

## Changes

- **rdb.go**: Changed the name fallback lookup in `UpdateProvidersConfig` from `Where("name = ?")` (global) to `Where("name = ? AND provider_id = ?")` (scoped to the current provider)
- **rdb_test.go**: Added `TestUpdateProvidersConfig_CrossProviderSameKeyName` regression test that creates two providers each with a key named `"default"` and verifies both keys are preserved after the sync

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [x] Core (Go)
- [ ] Transports (HTTP)
- [ ] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

The regression test directly validates the fix:

```sh
cd framework/configstore && go test -run TestUpdateProvidersConfig_CrossProviderSameKeyName -v
```

On the old (buggy) code, this test fails with `"[] should have 1 item(s), but has 0"` for the first provider. On the fixed code, it passes.

All existing tests still pass:
```sh
cd framework/configstore && go test ./...
```

## Screenshots/Recordings

N/A — No UI changes.

## Breaking changes

- [ ] Yes
- [x] No

The fix only changes the startup sync behavior from "wrong" (global lookup, data-corrupting) to "right" (scoped lookup). No API or schema changes.

## Related issues

Fixes the root cause of the HTTP 500 error on provider update: the startup sync was corrupting key ownership between providers, and the subsequent single-provider update couldn't create a new key because the name was already taken globally.

## Security considerations

None.

## Checklist

## Why this matters
Before this change, API key names had to be unique across all providers. If you had two providers - say OpenAI and Anthropic - they could not both have an API key named "default". The second one would collide with the first and fail with an error.

After this change, key names only need to be unique per provider. Each provider can independently have a "default" key (or any other name) without conflict.

### DIFF
--- a/framework/configstore/rdb.go
+++ b/framework/configstore/rdb.go
@@ -631,8 +631,11 @@ func (s *RDBConfigStore) UpdateProvidersConfig(ctx context.Context, providers ma
 					return s.parseGormError(err)
 				}
 			} else if errors.Is(result.Error, gorm.ErrRecordNotFound) {
-				// KeyID not found, try fallback lookup by Name (handles config reload with new UUID)
-				result = txDB.WithContext(ctx).Where("name = ?", dbKey.Name).First(&existingKey)
+				// KeyID not found, try fallback lookup by Name scoped to this provider
+				// (handles config reload with new UUID). The composite unique index
+				// (provider_id, name) guarantees names are unique per-provider, so we
+				// must scope the lookup to avoid matching a key from a different provider.
+				result = txDB.WithContext(ctx).Where("name = ? AND provider_id = ?", dbKey.Name, dbProvider.ID).First(&existingKey)
 				if result.Error == nil {
 					// Found by name - update existing key, preserve original KeyID
 					dbKey.ID = existingKey.ID                             // Keep the same database ID
--- a/framework/configstore/rdb_test.go
+++ b/framework/configstore/rdb_test.go
@@ -208,6 +208,64 @@ func TestUpdateProvidersConfig_MultipleKeys(t *testing.T) {
 	assert.Len(t, result["anthropic"].Keys, 1)
 }
 
+func TestUpdateProvidersConfig_CrossProviderSameKeyName(t *testing.T) {
+	// This test verifies that two providers can each have a key with the same name
+	// without one provider's key being overwritten by the other.
+	// This was broken before the fix: the name fallback in UpdateProvidersConfig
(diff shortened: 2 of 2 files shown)


SahilChoudhary22:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/maximhq/bifrost/3570?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#3570** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/3570?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/maximhq/bifrost/3570?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3569** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/3569?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `dev`


<h2></h2>

SahilChoudhary22:
As discussed in [Slack](https://maxim-hq.slack.com/archives/C08HJNYPLAE/p1778846017607359), this change has been marked as good to have and moved to the backlog for now.

While scoping API key name lookups by `provider_id` resolves the immediate cross-provider collision issue, it would also require follow-up changes to logging, since logs currently only capture the API key name and not the associated provider. To avoid introducing ambiguity there, we’re deferring the broader fix for now.

In the meantime, this is being addressed via #3574, which improves the user-facing error message to clearly communicate that API key names must be unique across providers.

PR #3588, opened 2026-05-19
fix(anthropic): normalize tool call continuation deltas

### LINKED ISSUE #3443: [Bug]: Anthropic→OpenAI streaming tool_call deltas violate OpenAI spec on continuation chunks (breaks strict clients)
### Prerequisites

### Description

When proxying Claude models (Anthropic native and Vertex Claude) with
  `tools: [...]` and `stream: true`, bifrost emits continuation chunks that
  diverge from OpenAI's Chat Completions streaming spec in two ways:
    1. `function.type: "function"` is set on every chunk (should be only on
       the initial tool_call setup chunk).
    2. An extra empty `arguments: ""` delta chunk is emitted right after the
       initial setup, before any partial JSON arrives.

  This trips strict client libraries — confirmed with firebase/genkit-go
  v1.6.1 which raises:
    "could not parse tool args: unmarshal failed to parse json string :
     unexpected end of JSON input"

  Bifrost's OpenAI passthrough (gpt-4o-mini) emits spec-compliant chunks;
  only the Anthropic→OpenAI converter is affected.

### Steps to reproduce

  Reproduction:

    curl -N -X POST $BIFROST/v1/chat/completions \
      -d '{
        "model":"claude-haiku-4-5-20251001","stream":true,"max_tokens":150,
        "messages":[{"role":"user","content":"Weather in SF? Use the tool."}],
        "tools":[{"type":"function","function":{
          "name":"get_weather",
          "parameters":{"type":"object","properties":{"location":{"type":"string"}}}
        }}]
      }'

  

  

### Expected behavior

Expected (matches OpenAI spec and bifrost's own OpenAI passthrough):     data: {"tool_calls":[{"index":0,"id":"toolu_X","type":"function","function":{"name":"get_weather","arguments":""}}]}     data: {"tool_calls":[{"index":0,"function":{"arguments":"{\"lo"}}]     data: {"tool_calls":[{"index":0,"function":{"arguments":"cation"}}]}

### Actual behavior

Observed (chunks 0-2, abbreviated):     data: {"tool_calls":[{"index":0,"type":"function","id":"toolu_X","function":{"name":"get_weather","arguments":""}}]}     data: {"tool_calls":[{"index":0,"type":"function","function":{"name":null,"arguments":""}}]}      ← extra empty chunk     data: {"tool_calls":[{"index":0,"type":"function","function":{"name":null,"arguments":"{\"loc"}}] ← `type` shouldn't be here

### Affected area(s)

Core (Go)

### Version

1.5.1

### Environment

```text

```

### Relevant logs/output

```shell

```

### Regression?

_No response_

### Severity

punyaslokdutta:
## Summary
- suppress empty Anthropic tool input deltas when converting to OpenAI chat-completions streams
- omit repeated `tool_calls[].type` metadata on continuation argument chunks
- add regression coverage for first-chunk metadata, empty delta suppression, and continuation chunk semantics

## Testing
- Focused pass:
  - `docker run --rm -v /Users/punyslokdutta/Desktop/git-projects/bifrost:/workspace -w /workspace/core golang:1.26.2-alpine3.23 sh -lc '/usr/local/go/bin/go test ./providers/anthropic -run "TestToBifrostChatCompletionStream_|TestToAnthropicChatRequest_"'

## Notes
- `go test ./providers/anthropic ./providers/vertex` still shows unrelated existing `anthropic` package failures that also reproduce on clean `origin/main`; this PR does not touch those areas.

Closes #3443.

### DIFF
--- a/core/providers/anthropic/chat.go
+++ b/core/providers/anthropic/chat.go
@@ -1152,6 +1152,14 @@ func (chunk *AnthropicStreamEvent) ToBifrostChatCompletionStream(ctx *schemas.Bi
 			case AnthropicStreamDeltaTypeInputJSON:
 				// Handle tool use streaming - accumulate partial JSON
 				if chunk.Delta.PartialJSON != nil {
+					if *chunk.Delta.PartialJSON == "" {
+						// Anthropic may emit an empty partial_json marker immediately after
+						// tool_use starts. OpenAI-style chat tool deltas should keep that
+						// initialization folded into the first tool_call chunk instead of
+						// emitting a second empty arguments delta.
+						return nil, nil, false
+					}
+
 					if structuredOutputToolName != "" {
 						// Structured output: stream JSON as content
 						streamResponse := &schemas.BifrostChatResponse{
@@ -1184,7 +1192,6 @@ func (chunk *AnthropicStreamEvent) ToBifrostChatCompletionStream(ctx *schemas.Bi
 										ToolCalls: []schemas.ChatAssistantMessageToolCall{
 											{
 												Index: uint16(toolCallIdx),
-												Type:  schemas.Ptr(string(schemas.ChatToolTypeFunction)),
 												Function: schemas.ChatAssistantMessageToolCallFunction{
 													Arguments: *chunk.Delta.PartialJSON,
 												},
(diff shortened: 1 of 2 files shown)


punyaslokdutta:
@akshaydeo I have signed the license/cla, still "Waiting for status to be reported"

rmarku:
Hey @punyaslokdutta, thanks for the PR. The fix mostly works, but there is an edge case that still fails.
If the tool call doesn't have arguments, the fix will modify the tool call to a wrong one.

It can be reproduced with this code https://gist.github.com/rmarku/8ec5a47b08c66a41d528326e054916e9

The code has the `response_start` tool that do not need any arguments. The response with `""` instead of `{}` make genkit to fail

Issue #2170, opened 2026-03-18
[Feature]:  x-bf-otel-attr-* headers for injecting custom attributes into OTel spans


17jmumford:
### Prerequisites

- [x] I have searched existing issues and discussions to avoid duplicates

### Problem to solve

Bifrost already has a great pattern for passing caller-supplied metadata into Prometheus metrics via x-bf-prom-* headers (ctx.go, strings.CutPrefix(keyStr, "x-bf-prom-")). This lets you
  configure custom labels on bifrost_cost_total etc. by having callers send e.g. x-bf-prom-agent_name: my-agent.

  However, there is no equivalent for OTel spans. PopulateContextAttributes() in framework/tracing/llmspan.go only writes a fixed set of pre-defined fields (virtual keys, team/customer IDs,
   retry counts, routing rules). There is no way for callers to inject custom attributes into the spans bifrost emits.

  This creates a gap: if I have a multi-agent system where the calling service knows which agent is making the LLM request, I have no way to attach that information to the bifrost span. The
   gen_ai.usage.cost attribute on the span can't be sliced by agent in tools like Datadog's trace analytics or APM.

### Proposed solution

Add an x-bf-otel-attr-* header prefix, mirroring the x-bf-prom-* pattern:

  1. In transports/bifrost-http/lib/ctx.go, parse these headers and store them in context:
  if attrName, ok := strings.CutPrefix(keyStr, "x-bf-otel-attr-"); ok {
      bifrostCtx.SetValue(schemas.BifrostContextKey("otel.attr."+attrName), string(value))
      return true
  }

  2. In framework/tracing/llmspan.go, add a PopulateCustomAttributes() function (or extend PopulateContextAttributes()) that reads these from context and writes them to the span's attribute
   map:
  func PopulateCustomAttributes(attrs map[string]any, ctx BifrostContext, keys []string) {
      for _, key := range keys {
          if val := ctx.Value(BifrostContextKey("otel.attr." + key)); val != nil {
              attrs[key] = val.(string)
          }
      }
  }

  3. In the OTel plugin config (plugins/otel/main.go), add an optional CustomAttributes []string field — same pattern as CustomLabels in the telemetry plugin — to declare which attributes
  to propagate.

  Example usage

  Calling service sends:
  x-bf-otel-attr-agent_name: Pi Agent
  x-bf-otel-attr-conversation_id: 019d02b8-4ff5-701a-9458-bf479084df4c

  Bifrost emits a span with:
  {
[...]

Issue #1117, opened 2025-12-17
[Feature]: Support custom per-request headers as OTEL span attributes


ksquarekumar:
### Prerequisites

- [x] I have searched existing issues and discussions to avoid duplicates

### Problem to solve

# Description:

The Maxim plugin supports propagating custom headers (x-bf-maxim-*) as trace attributes, but the OTEL plugin doesn't have equivalent functionality.*

# Current Behavior:

x-bf-eh-* headers are forwarded to upstream providers only
OTEL plugin only supports static resource attributes via OTEL_RESOURCE_ATTRIBUTES env var
No way to pass per-request custom attributes to OTEL traces*

# Use Case:
Correlating LLM traces with application traces, adding user/session context, custom metadata for filtering in observability backends.

### Proposed solution

# Desired Behavior:

Support `x-bf-otel-*` headers (or similar) that get added as span attributes:*

curl -X POST "http://localhost:8080/v1/chat/completions" \
    -H "x-bf-otel-trace-name: my-custom-trace" \
    -H "x-bf-otel-trace-id: my-custom-trace-123" \
    -H "x-bf-otel-span-id: my-custom-trace-123-span-abcd" \
    -H "x-bf-otel-session-id: user-session-789" \
    -d '{"model": "gpt-4o-mini", "messages": [...]}'

References: 
1. https://docs.getbifrost.ai/features/observability/maxim#custom-trace-management
2. https://deepwiki.com/search/how-do-i-enable-custom-perrequ_c8cb85b0-8e55-475c-a26c-59a1c9d59b87?mode=fast

### Alternatives considered

N/A

### Area(s)

Plugins

akshaydeo:
x-bf-dim does this now.
https://docs.getbifrost.ai/providers/request-options#unified-dimensions-x-bf-dim

PR #3781, opened 2026-05-26
[feat]: propagate x-bf-dim-* headers to child OTEL spans via trace.Attributes

### LINKED ISSUE #3770: Propagate x-bf-dim-* headers to all spans in a trace, not just the root HTTP span
## Problem

`x-bf-dim-*` HTTP headers (e.g. `x-bf-dim-customer_id`, `x-bf-dim-environment`) currently land as attributes on the root HTTP span only. Child spans (`llm.call`, plugin pre/posthook, retry, fallback, MCP tool) do not carry them, so in Datadog / any OTEL backend, traces cannot be filtered or grouped by dim values at the span level — only at the root.

This blocks the original use cases from #2170 and #1117: slicing per-call attributes like `gen_ai.usage.cost` by caller-supplied dimensions in trace analytics. `x-bf-dim-*` was shipped to address those issues, but because the attrs only appear on the root HTTP span, you still cannot filter `llm.call` spans by `customer_id` or `agent_name` in Datadog APM.

Today (`transports/bifrost-http/handlers/middlewares.go:1048`), `attachDimensionAttributesToHTTPSpan` reads `x-bf-dim-*` headers and writes them onto the root span via `SetAttribute`. Nothing propagates them downward.

## Proposal

Store dims once on `Trace.Attributes` (already a field on `*schemas.Trace`, pool-initialized in `framework/tracing/store.go`, currently unused), then merge them into every span in the OTEL converter with **span-attribute precedence on conflict** so semantic-convention attrs like `gen_ai.*` are not stomped.

**Surface changes (additive, non-breaking):**

- New method on `schemas.Tracer`: `SetTraceAttributes(traceID string, attrs map[string]any)`
- `NoOpTracer` gets a no-op impl.
- New method on `*schemas.Trace`: `MergeAttributes(attrs)` — locks the existing `t.mu`, mirrors `AppendPluginLogs`.
- `framework/tracing` implements `SetTraceAttributes` (TraceStore + Tracer passthrough).
- `TracingMiddleware` in transports calls `SetTraceAttributes` alongside the existing root-span `SetAttribute` loop.
- `plugins/otel/converter.go` `convertTraceToResourceSpan` merges `trace.Attributes` into each exported span (span attrs win on conflict).

**Precedent:** PR #3368 (`GetSpanHandleByID`) is the same shape of additive Tracer-interface change. PR #3732 (OTEL converter semconv attrs) is the same shape of additive converter change.

## Scope

- Branch: `dev`
- ~315 lines, well under the 400-line CONTRIBUTING budget.
- Touches: `core/schemas/`, `framework/tracing/`, `transports/bifrost-http/handlers/middlewares.go`, `plugins/otel/converter.go`, plus changelogs and new tests.

## Out of scope

- Prometheus label propagation (already handled via `BifrostContextKeyDimensions` in `plugins/telemetry/main.go`).
- Maxim/Langfuse propagation (no existing `*schemas.Trace` consumer outside OTEL plugin today; can read `trace.Attributes` the same way later).
- Dim allowlist/denylist or value-length cap (happy to discuss adding a cap in this PR if maintainers prefer — default is no cap).

## Open questions

1. Cap dim value length defensively? Default proposal: no cap.
2. Any preference between adding `SetTraceAttributes` vs. an attributes arg on `CreateTrace`? `CreateTrace`'s trailing `requestID ...string` variadic blocks clean additive extension, so I'm planning the new method — happy to switch if you'd rather.

Closes #2170, closes #1117 (with maintainer confirmation).

Happy to send the PR once you've sanity-checked the direction.

devin-williams:
## Summary

`x-bf-dim-*` HTTP headers (e.g. `x-bf-dim-customer_id`, `x-bf-dim-environment`) already land as attributes on the root HTTP span, and are stored on the trace as `TraceAttrDimensions` (`bifrost.dimensions`). Child spans (`llm.call`, plugin pre/posthook, retry, fallback, MCP tool) do not carry them, so Datadog / any OTEL backend cannot filter or group those spans by dim values.

This PR optionally merges those request-scoped dims onto **every exported OTEL span** when an OTEL profile sets `propagate_trace_attributes: true` (default **false**, matching the maintainers' original decision to keep dims on the root only). Span-level attributes win on conflict so `gen_ai.*` semantic-convention attrs are never stomped. Reserved suffixes `path` and `method` are skipped.

Closes maximhq/bifrost#3770 (also addresses the original use cases in #2170 and #1117).

## What is and is not propagated

**Propagated** (onto root + every child span, when the toggle is on):
- Request-scoped `x-bf-dim-*` dimensions stored on the trace at ingress (`customer_id`, `environment`, `agent_name`, …)

**Not propagated:**
- Parent-span metadata (`http.method`, `http.url`, `http.user_agent`, `session.id`, `bifrost.request.id`, instance attrs) stays on the root span
- Child-span metadata (`gen_ai.*`, plugin names, retry counts, …) stays on that child — it is not copied to siblings or back to the root
- Internal trace keys (`x-bf-session-id`, the nested `bifrost.dimensions` map itself)
- Reserved dim suffixes `path` and `method`

## Approach

`dev` already stores dims via `tracer.SetTraceAttribute(traceID, TraceAttrDimensions, dims)` at HTTP ingress. This PR does **not** add a parallel flatten API.

* **Write site (unchanged):** `TracingMiddleware` still writes dims onto the root span (skipping `path`/`method`) and onto `trace.Attributes[bifrost.dimensions]`.
* **Read/merge site (new):** `convertTraceToResourceSpan` takes a `propagateTraceAttributes` flag. When true, it reads `TraceAttrDimensions`, builds a per-span `present` keyset from the just-converted span attrs, and appends missing dim keys. Nil-checked before append so empty-string values cannot crash OTLP marshaling.
* **Opt-in wiring:** `Profile.PropagateTraceAttributes` (JSON `propagate_trace_attributes`, default false) is threaded through storage marshalling, `otelTarget`, the UI toggle, `config.schema.json`, and helm.

## Why this approach

* **Fits current `dev`.** Trace attributes are documented as connector-facing and not exported as span attributes. Dumping all of `trace.Attributes` onto spans would leak `x-bf-session-id` and the nested dimensions map. Reading `TraceAttrDimensions` only is the smallest correct merge.
* **Default-off.** Matches the maintainers' earlier decision to keep dims on the root span unless an operator opts in.
* **OTEL-only.** Other connectors already read `TraceAttrDimensions` off the completed trace.

## Out of scope

* Prometheus label propagation (already handled in `plugins/telemetry/main.go`).
* Maxim/Langfuse propagation (no `*schemas.Trace` consumer outside OTEL today).
* Copying parent-span attrs onto children, or child-span attrs onto the root/siblings.
* Dim allowlist/denylist or value-length cap.
* W3C baggage cross-service propagation.

## Test plan

- [X] Converter tests: dims propagate to root + llm + plugin + retry spans; span-attribute precedence; empty-string nil-check; filtered plugin spans stay absent; reserved `path`/`method` and `TraceAttrSessionID` never appear on exported spans; flag-off is a no-op.
- [X] Middleware test: `x-bf-dim-*` headers land on `TraceAttrDimensions` (including reserved suffixes, which the converter then skips).
- [X] `plugins/otel` and `framework/tracing` tests pass locally.
- [X] `propagate_trace_attributes` present in `transports/config.schema.json`, helm values/schema/helpers, and the OTEL profile form.

## Notes

* High-cardinality dims increase ingestion cost proportionally (one extra attribute per exported span). No value-length cap in this PR.
* Root span already receives dims via `SetAttribute` at ingress, independent of this toggle. The toggle only adds them to child spans (and re-merges onto the root at export with span-attribute precedence).


devin-williams:
Thanks @greptile-apps — both callouts addressed in the latest force-push:

- `plugins/otel/converter.go` — comment rewritten to make clear that on a root-span dim/instanceAttrs collision both KVs are emitted and collectors resolve via last-wins. The behavior was already that; only the comment was misleading.
- `transports/bifrost-http/handlers/middlewares.go` — dropped the redundant `if len(dims) > 0` guard before `SetTraceAttributes`; replaced with a one-line comment noting the callee already no-ops on empty.

Diff is still 4 commits / 14 files (582 insertions, 8 deletions). Tests still pass under `-race`.

17jmumford (approved):

akshaydeo:
@Pratham-Mishra04 can you check this

17jmumford:
> @Pratham-Mishra04 can you check this

@Pratham-Mishra04 looks like this branch is drifting out of sync. How can we best coordinate updating this? Should we point the PR to main instead?

akshaydeo:
hey no main PRs are completely blocked. We had concisously decided to not inject dims to individual spans and keep it on root span. But I get your point - would you be able to rebase your branch with our latest changes and update the PR? 

Else we will take it on our own

17jmumford:
> hey no main PRs are completely blocked. We had concisously decided to not inject dims to individual spans and keep it on root span. But I get your point - would you be able to rebase your branch with our latest changes and update the PR?
> 
> Else we will take it on our own

got it, thanks for clarifying @akshaydeo. We'll rebase on Monday. Have a good weekend!

R-droid101 (commented):
Hey, the PR looks great overall with respect to the code and the flow. 

Could you resolve the greptile comment? Once that's done, we will have a final review round and we can merge this into dev.

akshaydeo:
Hi @devin-williams — thanks for the contribution! This PR is currently blocked because our CLA bot shows the Contributor License Agreement as not yet signed. Could you sign it here so we can move this forward: https://cla-assistant.io/maximhq/bifrost?pullRequest=3781

transports/bifrost-http/handlers/middlewares.go
```diff
@@ -1269,6 +1269,15 @@ func (m *TracingMiddleware) Middleware() schemas.BifrostHTTPMiddleware {
...
 				}
+				// Propagate dimension headers onto trace.Attributes so observability
+				// plugins can merge them onto every exported child span at export time.
+				if len(dimensions) > 0 {
+					attrs := make(map[string]any, len(dimensions))
+					for k, v := range dimensions {
+						attrs[k] = v
+					}
+					tracer.SetTraceAttributes(traceID, attrs)
+				}
```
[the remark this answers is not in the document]
devin-williams: Fixed in the latest push (rebased onto current `dev` via merge).

`dev` now stores dims on `trace.Attributes[bifrost.dimensions]` rather than as flat keys, so the old `SetTraceAttributes` flatten would have leaked reserved `path`/`method` (and internal keys like `x-bf-session-id`) onto every child span.

The OTEL converter now reads `TraceAttrDimensions` when `propagate_trace_attributes` is enabled and skips `path`/`method`. Added a converter test that asserts those reserved suffixes — and `TraceAttrSessionID` — never appear on exported spans. Also added `propagate_trace_attributes` to `transports/config.schema.json` / helm / the OTEL form. The duplicate MCP temp-token toggle was a merge artifact and has been dropped.

devin-williams:
Hey @akshaydeo ! I went ahead and updated the pr and signed the CLA! It looks like the pending check still says pending though.

devin-williams:
@akshaydeo It looks like everything should be good to go now!

helm-charts/bifrost/values.schema.json
```diff
@@ -5762,6 +5762,11 @@
           "description": "When true, input/output message content is dropped from the root span only; the underlying generation (llm.call) span retains the full content. This removes the duplicate input/output stored at the trace level (e.g. the Langfuse trace Input/Output goes empty) to reduce downstream storage.",
           "default": false
         },
+        "propagate_trace_attributes": {
```
roroghost17, line 5765: I am not sure the name `propagate_trace_attributes` is the one to be used here (as we already propagate these regardless to the root span). We can have it something like `propagate_trace_attributes_to_child_spans` / `apply_trace_dimensions_to_child/all_spans` or something similar where the naming is a bit more explicit. Would need to do this here since this would be a long term name.

roroghost17:
Hey @devin-williams, went through the PR. the overall implementation looks good. The only thing I see that might be worth a point, is the naming of the attribute. There are some merge conflicts on the PR as well.

helm-charts/bifrost/values.schema.json
devin-williams: Good call — `propagate_trace_attributes` was too vague, since dims already land on the root span regardless of this flag.

Renamed to `apply_trace_dimensions_to_child_spans` (JSON, helm, config schema, UI toggle, and Go). That name is explicit about both *what* is applied (trace dimensions, not all of `trace.Attributes`) and *where* (child spans). Default is still `false`.

devin-williams:
Hey @roroghost17 that should all be fixed and up to date now!

PR #3807, opened 2026-05-27
refactor: replace embedded `TableMCPClient` in `MCPClientUpdateRequest` with explicit optional fields and true PATCH semantics

BearTS:
## Summary

`PUT /api/mcp/client/{id}` previously embedded `TableMCPClient` directly in `MCPClientUpdateRequest`, which meant callers had to send the full record and any omitted field would silently zero-out the existing value. This PR replaces that approach with true PATCH semantics: every mutable field is now an explicit pointer/nullable type, and the handler merges only the fields that are present in the request body, leaving everything else untouched.

## Changes

- `MCPClientUpdateRequest` no longer embeds `TableMCPClient`. It now declares each mutable field explicitly as a pointer (`*string`, `*bool`, `*int`, etc.) so that absent fields are distinguishable from zero values.
- Immutable fields (`connection_type`, `auth_type`, `connection_string`, `stdio_config`) are removed from the request struct entirely and are no longer validated for mutation — they are always carried forward from the existing record.
- The handler resolves each field individually: if the pointer is non-nil, the new value is used; otherwise the existing stored value is retained.
- `mergeMCPRedactedValues` is narrowed to operate only on `map[string]schemas.EnvVar` (headers), removing the now-unnecessary logic for `ConnectionString`, `IsPingAvailable`, and `AllowedExtraHeaders`, which are handled by the pointer-based merge above.
- The DB update record (`dbUpdateRecord`) is now constructed explicitly from all resolved values rather than mutating the embedded struct, making the data flow easier to follow and audit.

## Type of change

- [ ] Bug fix
- [ ] Feature
- [x] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [ ] Core (Go)
- [x] Transports (HTTP)
- [ ] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

```sh
go test ./transports/bifrost-http/...

# Create an MCP client, then issue a partial update and confirm only the
# specified fields change while all others retain their existing values.
curl -X PUT /api/mcp/client/{id} \
  -d '{"disabled": true}'

# Confirm that name, headers, tools_to_execute, etc. are unchanged.

# Attempt to pass connection_type or auth_type and confirm they are ignored
# (not rejected, simply not accepted in the struct).
```

## Breaking changes

- [x] Yes
- [ ] No

Callers that previously relied on sending the full `TableMCPClient` payload (including `connection_type`, `auth_type`, `connection_string`, or `stdio_config`) will have those fields silently ignored rather than validated. Any client that was relying on the old struct shape for field names should verify their payloads still map correctly to the new explicit field list.

## Security considerations

Redacted header values are still preserved correctly via `mergeMCPRedactedValues` — if an incoming header value matches the redacted form of the stored value, the original raw value is retained rather than overwritten with the redacted placeholder.

## Checklist

### DIFF
--- a/framework/configstore/tables/mcp.go
+++ b/framework/configstore/tables/mcp.go
@@ -29,8 +29,8 @@ type TableMCPClient struct {
 	ToolSyncInterval        int             `gorm:"default:0" json:"tool_sync_interval"`             // Per-client tool sync interval in seconds (0 = use global, negative = disabled)
 
 	// Per-user OAuth: discovered tools persisted so they survive restart
-	DiscoveredToolsJSON       string `gorm:"type:text" json:"-"`                              // JSON serialized map[string]schemas.ChatTool
-	ToolNameMappingJSON       string `gorm:"type:text" json:"-"`                              // JSON serialized map[string]string
+	DiscoveredToolsJSON string `gorm:"type:text" json:"-"` // JSON serialized map[string]schemas.ChatTool
+	ToolNameMappingJSON string `gorm:"type:text" json:"-"` // JSON serialized map[string]string
 
 	// OAuth authentication fields
 	AuthType      string            `gorm:"type:varchar(20);default:'headers'" json:"auth_type"`                         // "none", "headers", "oauth", "per_user_oauth", "per_user_headers"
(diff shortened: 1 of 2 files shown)


BearTS:
* **#3807** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/3807?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/maximhq/bifrost/3807?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `dev`


<h2></h2>

PR #3811, opened 2026-05-27
refactor: replace embedded `TableMCPClient` in `MCPClientUpdateRequest` with explicit optional fields and true PATCH semantics

BearTS:
## Summary

`PUT /api/mcp/client/{id}` previously embedded `TableMCPClient` directly in `MCPClientUpdateRequest`, which meant every field had to be sent on every update and immutable fields (connection_type, auth_type, connection_string, stdio_config) had to be explicitly rejected at runtime. This PR replaces that approach with true PATCH semantics: every mutable field is now an explicit optional pointer, omitted fields retain their existing values, and immutable fields are simply not present in the request type.

## Changes

- `MCPClientUpdateRequest` no longer embeds `TableMCPClient`. Each mutable field is now declared as an explicit optional pointer (`*string`, `*bool`, etc.), making the contract clear in the type system and eliminating the need for runtime immutability guards.
- Removed the runtime checks that rejected changes to `connection_type`, `auth_type`, `connection_string`, and `stdio_config` — those fields are no longer accepted in the request body at all.
- The update handler now resolves each field individually: use the incoming value if the pointer is non-nil, otherwise fall back to the existing config value. The fully resolved values are then used to build both the DB record and the in-memory `schemas.MCPClientConfig`.
- `mergeMCPRedactedValues` (which operated on the full `TableMCPClient`) is replaced by the narrower `mergeMCPHeaders`, which only handles the header-specific redaction-preservation logic and is only called when the caller explicitly provides a `headers` map.
- `stdioConfigsEqual` is removed as it is no longer needed.
- `resolvePerUserHeaderKeys` and `CanonicalizeHeaderKeys` call sites updated to dereference the now-pointer `PerUserHeaderKeys` field.
- The DB update record is now constructed explicitly from all resolved values rather than mutating the embedded struct, making the data flow easier to follow and audit.

## Type of change

- [ ] Bug fix
- [ ] Feature
- [x] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [ ] Core (Go)
- [x] Transports (HTTP)
- [ ] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

```sh
go test ./...
```

- Send a `PUT /api/mcp/client/{id}` with only `name` set and confirm all other fields are unchanged.
- Send a `PUT` with `headers` containing a redacted value for an existing key and confirm the stored raw value is preserved.
- Confirm that sending `connection_type`, `auth_type`, `connection_string`, or `stdio_config` in the body has no effect (fields are ignored, not rejected with an error).
- Send a `PUT` with `per_user_header_keys` as a pointer and confirm canonicalization still applies.

## Breaking changes

- [x] Yes
- [ ] No

Clients that previously relied on the 400 error response when attempting to change `connection_type`, `auth_type`, `connection_string`, or `stdio_config` will no longer receive that error — those fields are silently ignored. Any client sending the full `TableMCPClient` payload expecting those fields to be applied should be updated to use only the mutable fields defined in the new request shape.

## Security considerations

Immutable fields (connection_type, auth_type, connection_string, stdio_config) are now excluded from the request type entirely rather than being accepted and then rejected. This reduces the surface area for accidental or malicious mutation of connection identity fields. Header redaction-preservation logic is unchanged in behavior but is now scoped only to the headers map rather than the full client record.

## Checklist

### DIFF
--- a/transports/bifrost-http/handlers/mcp.go
+++ b/transports/bifrost-http/handlers/mcp.go
@@ -349,13 +349,26 @@ type MCPVKConfigRequest struct {
 	ToolsToExecute schemas.WhiteList `json:"tools_to_execute"`
 }
 
-// MCPClientUpdateRequest wraps TableMCPClient and adds optional VK assignment management, ToolsToExecute and ToolsToAutoExecute
+// MCPClientUpdateRequest is the body for PUT /api/mcp/client/{id}.
+// All fields are optional — omitting a field retains its existing value (PATCH semantics).
+// Immutable fields (connection_type, auth_type, connection_string, stdio_config) are not
+// accepted here; they cannot be changed after creation.
 type MCPClientUpdateRequest struct {
-	configstoreTables.TableMCPClient
-	VKConfigs          *[]MCPVKConfigRequest `json:"vk_configs,omitempty"`
-	ToolsToExecute     *schemas.WhiteList    `json:"tools_to_execute,omitempty"`
-	ToolsToAutoExecute *schemas.WhiteList    `json:"tools_to_auto_execute,omitempty"`
-	OauthConfig        *OAuthConfigRequest   `json:"oauth_config,omitempty"`
+	Name                  *string                   `json:"name,omitempty"`
+	Disabled              *bool                     `json:"disabled,omitempty"`
+	AllowOnAllVirtualKeys *bool                     `json:"allow_on_all_virtual_keys,omitempty"`
+	IsCodeModeClient      *bool                     `json:"is_code_mode_client,omitempty"`
+	IsPingAvailable       *bool                     `json:"is_ping_available,omitempty"`
+	ToolSyncInterval      *int                      `json:"tool_sync_interval,omitempty"`
+	Headers               map[string]schemas.EnvVar `json:"headers,omitempty"`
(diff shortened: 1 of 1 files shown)


BearTS:
* **#3811** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/3811?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/maximhq/bifrost/3811?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `dev`


<h2></h2>

BearTS:
@coderabbitai review

BearTS:
@coderabbitai review

BearTS:
https://github.com/coderabbitai review

BearTS:
@coderabbitai review

Pratham-Mishra04 (approved):

Pratham-Mishra04:
### Merge activity

* **May 28, 10:24 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/3811).
* **May 28, 10:25 AM UTC**: @Pratham-Mishra04 merged this pull request with [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/3811).

PR #3835, opened 2026-05-28
fix(gemini): strip Authorization header when setting x-goog-api-key for Vertex Express compatibility

SandyDRawat:
## Problem

The gemini provider correctly sets `x-goog-api-key` on outbound requests, but the inbound `Authorization: Bearer <key>` header (sent by OpenAI-compatible clients like `AsyncOpenAI`, `openai-python`, etc.) leaks through to the upstream provider.

For AI Studio (`generativelanguage.googleapis.com`) this is harmless — the API key takes precedence. But when the gemini provider's `base_url` is overridden to **Vertex Express** (`aiplatform.googleapis.com`) — a common pattern for using Vertex Express API keys (`AQ.*`) for free-tier access — Vertex's strict auth check sees BOTH headers and rejects with the misleading 401 error:

> `"API keys are not supported by this API. Expected OAuth2 access token or other authentication credentials"`

…even though `x-goog-api-key` alone would have authenticated successfully.

## Reproduction (direct curl from a host with an AQ.* key)

| Headers | Status | Message |
|---|---|---|
| `x-goog-api-key` only | **200** | works |
| `Authorization: Bearer` only | 401 | `"Request had invalid authentication credentials. Expected OAuth 2 access token"` |
| **BOTH headers** | **401** | **`"API keys are not supported by this API. Expected OAuth2"`** ← exact text users see |

The third case (both headers) matches the error reported by clients calling Bifrost with Vertex Express VKs.

## Fix

After every `req.Header.Set("x-goog-api-key", ...)` (or equivalent `headers["x-goog-api-key"] = ...` in streaming paths), explicitly remove the Authorization header so only the intended auth reaches upstream:

```go
if key.Value.GetValue() != "" {
    req.Header.Set("x-goog-api-key", key.Value.GetValue())
    req.Header.Del("Authorization")  // ← added
}
```

## Affected sites (32 total)

- `core/providers/gemini/gemini.go` — 26 sites (chat, embedding, list_models, files, batch references, etc.)
- `core/providers/gemini/cachedcontents.go` — 5 sites
- `core/providers/gemini/batch.go` — 1 site

## Behavior change

- **AI Studio (default)**: no change — AI Studio accepts dual-header gracefully; the Del is a no-op when no Authorization was forwarded, and harmless when it was.
- **Vertex Express (`base_url` overridden to `aiplatform.googleapis.com`)**: now works. Previously failed with 401 on every call.

## Verified

- `go vet ./providers/gemini/...` clean (exit 0)
- `go build ./providers/gemini/...` clean (exit 0)
- Underlying behavior tested by direct `curl` against Vertex Express endpoint with each header combination (results above).

## Related context

PR submitted by a deployment using Bifrost as a multi-tier LLM gateway with three Bifrost VKs (free-pool / paid Vertex / OpenRouter cascade). The Vertex Express free-tier path was completely unreachable due to this bug — every call was silently cascading to the paid tier.


akshaydeo:
hi @SandyDRawat ❤️ for the PR.  Could you rebase the PR against the dev 🙇

SandyDRawat:
Thanks for the review @akshaydeo.

Rebased onto dev — the diff is now just the gemini auth-header changes.

Both Greptile findings are addressed:

- SpeechStream (P1): moved SetExtraHeaders above the key block so it matches TranscriptionStream, so the Authorization strip now runs after the extra/context headers are applied.
- Streaming map paths (P2): removed the no-op `delete(headers, "Authorization")` and instead call `req.Header.Del("Authorization")` on the actual request after the headers are set, in both HandleGeminiChatCompletionStream and HandleGeminiResponsesStream.

The strip is guarded on x-goog-api-key being present, so OAuth-only setups (no API key) keep their Authorization header. This follows the existing Azure provider, which already strips the conflicting auth header in azure/files.go.

Also added a regression test (strip_authorization_test.go) covering the api-key/no-api-key guard and the streaming path — it runs without network or credentials.

SandyDRawat:
Thanks @coderabbitai — addressed all three nitpicks in ec9fd770 with a single change, since they were the same finding seen from three files.

**Extracted the helper.** All 30 occurrences of the `Set("x-goog-api-key")` + `Del("Authorization")` pair (24 in `gemini.go`, 5 in `cachedcontents.go`, 1 in `batch.go`) now call:

```go
// setGeminiAuthHeader authenticates the request with apiKey and strips any
// forwarded Authorization so only x-goog-api-key auth reaches upstream (Vertex
// Express rejects requests carrying both credentials). Must be called after
// providerUtils.SetExtraHeaders, which may inject a forwarded Authorization.
func setGeminiAuthHeader(req *fasthttp.Request, apiKey string) {
	if apiKey == "" {
		return
	}
	req.Header.Set("x-goog-api-key", apiKey)
	req.Header.Del("Authorization")
}
```

One deviation from the suggestion: I kept it package-local (next to the existing `setGeminiRequestBody`) rather than putting it in `core/providers/utils`. `x-goog-api-key` is a Gemini-specific header and a repo-wide sweep confirms every outbound call site lives in `core/providers/gemini` — the other `x-goog-api-key` hits are inbound header parsing in `transports/` and the hop-by-hop blocklist in `utils.go`. Promoting it to shared provider utils would export a Gemini concern for no current caller. Happy to move it if you'd rather it sit alongside the other `providerUtils` helpers.

This also subsumes the `batch.go` comment nitpick — the rationale now lives on the helper instead of being restated at each site.

**Two paths intentionally left alone.** `HandleGeminiChatCompletionStream` and `HandleGeminiResponsesStream` populate a `map[string]string` rather than touching `req.Header`, so they can't use the helper; they keep their existing `if _, ok := headers["x-goog-api-key"]; ok` guard, which runs after the header loop and after `SetExtraHeaders` (the ordering Greptile's earlier P2 was about).

**Unrelated: the test file didn't compile.** `strip_authorization_test.go` has been broken since `dev` was merged in at 59e0d76f — `schemas.NewEnvVar` was renamed to `schemas.NewSecretVar`, and `HandleGeminiChatCompletionStream` gained a `streamIdleTimeoutInSeconds int` parameter. Fixed both in the same commit. Worth noting the required checks on this PR are all bots and CLA with no Go build/test step, so `go vet ./providers/gemini/` was failing on the branch without anything flagging it.

Verified: `gofmt` clean, `go build ./...` and `go vet ./providers/gemini/` pass, and the package suite is green (`TestGemini` is a live-API integration test that skips without `GEMINI_API_KEY`). Both strip-authorization tests still pass, which is the real check here — they assert the `Authorization` strip on the unary and streaming paths, so a dropped `Del` in any of the 30 rewritten sites would surface.

akshaydeo:
Hi @SandyDRawat — thanks for the contribution! This PR is currently blocked because our CLA bot shows the Contributor License Agreement as not yet signed. Could you sign it here so we can move this forward: https://cla-assistant.io/maximhq/bifrost?pullRequest=3835

Issue #3400, opened 2026-05-12
[Feature] support oauth passthrough for codex


akshaydeo:
## Discord Thread Context
**Thread:** Any plas to introduce OAuth passthrough for Codex? Similar as it was done for Anthropic?
**Link:** https://discord.com/channels/1371790330269007932/1503552821780549642
**Created by:** akshay_deo (361806850561146880)

## Recent Thread Messages
**akshay_deo:** wouldnt say its a popular ask
**akshay_deo:** but contributions are welcome 🙇‍♂️
**akshay_deo:** ill keep it on github though

**Discord Thread:** https://discord.com/channels/1371790330269007932/1503552821780549642

akshaydeo:
hi passthrough is added back to the latest version. You have to pass an extra header to leverage this
you have to add `x-bf-direct-key: true` header along with direct API key

PR #3863, opened 2026-05-28
feat: provider-level allow_direct_keys (re-add passthrough lane, opt-in per provider)

### LINKED ISSUE #3171: [Feature]: Provider level allow_direct_keys setting
### Prerequisites

### Problem to solve

I am trying to use bifrost with codex.  OpenAI models should use the credentials from codex (i.e. `allow_direct_keys = true`) while other models (e.g. openrouter) should use the configured keys.

There does not seem to be any way to do this currently.  If `allow_direct_keys` is on, the passed key is used for everything.

### Proposed solution

A provider-level `allow_direct_keys` setting.

### Alternatives considered

_No response_

### Area(s)

Transports (HTTP), Core (Go)

### Additional context

_No response_

sn1ce:
## Summary

Re-introduces the direct-key forwarding flow that PR #3241 removed in v1.5.0, but gated **per-provider** rather than globally — matching the design proposed in #3171.

A provider with `allow_direct_keys: true` accepts a raw upstream-provider key from `Authorization` / `x-api-key` / `x-goog-api-key` headers and uses it as the request's upstream credential. Providers without the flag preserve post-v1.5 behaviour exactly: header keys are ignored and only Bifrost-managed keys can authenticate.

Closes #3171.

## Why bring this back

The v1.5.0 removal cited four security concerns. This redesign addresses each of them while restoring the BYO-key use case that #3171, #3400 (oauth-passthrough for codex), and many internal-gateway deployments need:

| v1.5.0 concern | Resolution |
| --- | --- |
| Bypassed governance globally | **Opt-in per provider, default off** — single-provider deployments inherit the post-v1.5 default; only providers explicitly marked passthrough accept header keys |
| Bypassed virtual-key budgets | **Direct keys are a separate path** — they don't see, and aren't checked against, Bifrost-managed virtual keys. Passthrough is BYO-key, BYO-budget; the upstream account itself is the limit |
| Credential leak in logs / proxy chains | **Raw key is never persisted.** The HTTP transport wraps the bearer in `*schemas.NewEnvVar(rawKey)` (redaction-aware) and only ever writes a sha256 fingerprint to the logs table |
| Hard to reason about which path a request took | **Per-call observability preserved.** Every passthrough call is logged with `selected_key_id = direct:<fingerprint>`, `selected_key_name = direct-key`, plus the standard model / tokens / cost / latency fields |

## Design

- New field on `core/schemas.ProviderConfig`: `AllowDirectKeys bool` (default false)
- New context key `BifrostContextKeyDirectKey` carrying a `schemas.Key`
- HTTP transport (`transports/bifrost-http/lib/ctx.go`) captures any non-virtual-key bearer (Authorization Bearer / x-api-key / x-goog-api-key) into the context key, **unconditionally**. Gating happens at key selection.
- Core (`core/bifrost.go::selectKeyFromProviderForModelWithPool`) honours the direct key **only** when the resolved provider's `AllowDirectKeys` is true, returning it as a single-key pool with `canRotate=false`.
- Field is plumbed end-to-end: schemas → configstore.ProviderConfig → tables.TableProvider (`not null;default:false` GORM column) → handlers payloads / response → GenerateConfigHash.

## Use case

The motivating scenario (also #3171): an internal team gateway where some users have their own personal upstream key (e.g. `sk-ant-…`) and want to keep billing it to themselves, while others use a shared team key with budgets / rate-limits via virtual keys. Configure **two provider entries** of the same kind — one keyless with `allow_direct_keys: true` (passthrough lane), one keyed with `allow_direct_keys: false` (pooled / virtual-key lane) — and clients pick lane by which bearer they send. This pattern was supported in v1.4 and its loss in v1.5 has been the most-cited reason teams have stayed on or moved off Bifrost.

## How to test

```sh
cd core && go test ./...
cd framework && go test ./...
cd transports && go test ./...
```

End-to-end smoke (against `Dockerfile.local`):

```sh
docker build -f transports/Dockerfile.local -t bifrost-fork:dev .

cat > /tmp/config.json <<JSON
{ "providers": { "anthropic": { "keys": [], "allow_direct_keys": true } } }
JSON

docker run --rm -p 8081:8080 -v /tmp/config.json:/app/data/config.json:ro bifrost-fork:dev &

curl -X POST http://localhost:8081/anthropic/v1/messages \
  -H "Authorization: Bearer $ANTHROPIC_API_KEY" \
  -H "anthropic-version: 2023-06-01" \
  -d '{"model":"claude-haiku-4-5","max_tokens":50,"messages":[{"role":"user","content":"hi"}]}'
# → HTTP 200, logged with selected_key_id="direct:<fingerprint>"

# With allow_direct_keys flipped to false, the same curl returns a no-keys error.
```

Verified locally: 200 from Anthropic, log entry shows `selected_key_id: direct:2ea5bb233081`, `cost: 0.00007`, `selected_key.value.value: ""` (raw key never persisted).

## Affected areas

- [x] Core (Go)
- [x] Framework (configstore)
- [x] Transports (HTTP) — handlers + lib
- [ ] Plugins
- [ ] UI (React) — admin checkbox to follow in a separate PR; field is already exposed in the provider API responses
- [ ] Docs — will follow if direction is approved

## Breaking changes

None. Default is false; existing behaviour is unchanged.

## Migration

`config_providers` gains a single boolean column (`allow_direct_keys`, default 0). GORM AutoMigrate handles it on first start. Existing rows get the default (false) and behave identically to post-v1.5.


akshaydeo:
we already have a PR for thi @sn1ce - but thanks for the effort 🙇

akshaydeo:
Hi @sn1ce — thanks for the contribution! This PR is currently blocked because our CLA bot shows the Contributor License Agreement as not yet signed. Could you sign it here so we can move this forward: https://cla-assistant.io/maximhq/bifrost?pullRequest=3863

PR #3920, opened 2026-05-31
feat(telemetry): configure latency histogram buckets via config.json

Purvi09:
## Summary

Adds configurable latency histogram bucket boundaries for both the Prometheus (telemetry) and OTel plugins via `config.json`. Operators can now customize bucket boundaries per deployment without forking. 

## Changes

- Added `latency_buckets`, `first_token_latency_buckets`, and `inter_token_latency_buckets` fields to `telemetry.Config` and `otel.Config`
- `plugins/telemetry/main.go`: reads bucket config in `Init()`, uses local variable shadowing to avoid data races
- `plugins/otel/metrics.go`: passes bucket config through `MetricsConfig` into `initMetrics()`
- `plugins/otel/main.go`: adds bucket fields to `Config`, passes them to `MetricsConfig`
- `transports/bifrost-http/server/plugins.go`: explicitly merges bucket fields when loading telemetry plugin config
- `transports/config.schema.json`: documents the new fields with descriptions and validation rules

When fields are omitted, compiled-in defaults are used unchanged — no impact on existing deployments.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [ ] Core (Go)
- [x] Transports (HTTP)
- [ ] Providers/Integrations
- [x] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

```sh
# Build
cd plugins/telemetry && go build ./...
cd plugins/otel && go build ./...
```

Add to `config.json`:
```json
{
  "plugins": [
    {
      "name": "telemetry",
      "enabled": true,
      "config": {
        "latency_buckets": [0.005, 0.1, 1, 10, 60, 300, 900]
      }
    }
  ]
}
```
Start Bifrost and scrape `/metrics` — histogram bucket boundaries should reflect the configured values. When fields are omitted, existing default buckets are used.

## Breaking changes

- [x] No

## Security considerations

None — config fields are read once at startup from `config.json`.

## Checklist


roroghost17:
hey @Purvi09, thanks for the PR. I went through it and a couple of things:
1. Instead of env variables, we should actually allow configuring this from config.json since these values would ideally be long-term.
2. It also seems that you might have missed pulling the latest changes before opening the PR which have resulted in some additional changes showing up which aren't in this PR.

So, could you please update this PR with the config.json flow and ping me here and I can review it then again before merging?

Also, since the base issue that you have mentioned initially is closed, I would suggest opening up a separate issue for this (provided its not a duplicate) and then attach this PR to that issue.

Purvi09:
> hey @Purvi09, thanks for the PR. I went through it and a couple of things:
> 
> 1. Instead of env variables, we should actually allow configuring this from config.json since these values would ideally be long-term.
> 2. It also seems that you might have missed pulling the latest changes before opening the PR which have resulted in some additional changes showing up which aren't in this PR.
> 
> So, could you please update this PR with the config.json flow and ping me here and I can review it then again before merging?
> 
> Also, since the base issue that you have mentioned initially is closed, I would suggest opening up a separate issue for this (provided its not a duplicate) and then attach this PR to that issue.

Hi @roroghost17 ! Thanks for the feedback. I've addressed all three points:

1)Switched to config.json fields (latency_buckets, first_token_latency_buckets, inter_token_latency_buckets) for both the telemetry and OTel plugins
2)Created a fresh branch from the latest dev to avoid the extra file issue
3)Opened issue #3969 for this feature
Since the old PR had a messy history, I've opened a clean PR #4008 with all the changes.

PR #3994, opened 2026-06-02
fix(gemini): accept numeric schema integer constraints

### LINKED ISSUE #3993: Gemini GenAI rejects numeric JSON Schema integer constraints in tool declarations
## Summary

Bifrost currently rejects Gemini GenAI requests when a tool/function declaration schema contains standard JSON Schema integer constraints encoded as JSON numbers, such as `minLength: 1`. The request is rejected before provider execution with a generic `400 Invalid JSON` response.

This affects clients that emit normal JSON Schema numeric constraints in Gemini tool declarations. In my reproduction, opencode using `@ai-sdk/google` sent a Gemini GenAI streaming request with tool declarations containing numeric `minLength`, which Bifrost rejected while the same request shape succeeded after stringifying or removing that constraint.

## Reproduction Evidence

- Route shape: `/genai/v1beta/models/{model}:streamGenerateContent?alt=sse`
- Offending schema shape: `tools.functionDeclarations[].parameters.properties.<field>.minLength: 1`
- Bifrost response: `400 Invalid JSON`
- Replaying the same captured request body with `minLength` removed succeeded.
- Replaying the same captured request body with `minLength` changed from numeric `1` to quoted `"1"` succeeded.
- A local parser test against Bifrost Gemini types reproduced the unmarshal error for numeric constraints.

No raw API keys or captured request body are included here.

## Root Cause

Gemini schema integer constraint fields in `core/providers/gemini/types.go` are declared with `json:",string"`, for example:

```go
MinLength *int64 `json:"minLength,omitempty,string"`
```

That accepts quoted int64 values like `"1"`, matching the canonical Google REST documentation form, but it rejects standard JSON numeric values like `1`.

Google's public Gemini API accepts both forms in practice for tool schema constraints, while JSON Schema-producing clients commonly emit the numeric form.

## Expected Behavior

Bifrost should accept both:

```json
{ "minLength": 1 }
```

and:

```json
{ "minLength": "1" }
```

for Gemini schema integer constraint fields. Invalid strings should still be rejected.

## Proposed Fix

Add custom Gemini `Schema` unmarshalling that flexibly accepts numeric or quoted integer constraints for:

- `maxItems`
- `maxLength`
- `maxProperties`
- `minItems`
- `minLength`
- `minProperties`

The existing marshal behavior can remain unchanged.

## Agent Disclosure

This issue was investigated and drafted with AI-agent assistance. Focused local regression tests were used to validate the proposed fix.

yanhao98:
Fixes #3993.

## Summary

- add custom Gemini `Schema` unmarshalling for JSON Schema integer constraint fields
- accept both standard numeric values and quoted int64 strings for `maxItems`, `maxLength`, `maxProperties`, `minItems`, `minLength`, and `minProperties`
- keep invalid string values rejected with an explicit schema integer constraint error
- add regression coverage for numeric constraints, quoted constraints, and invalid strings

## Why

Some Gemini GenAI clients emit tool/function declaration schemas with standard JSON Schema numeric constraints, for example `minLength: 1`. Bifrost previously rejected those requests during request parsing because these fields used Go JSON `,string` tags and only accepted quoted values such as `"1"`.

Google's REST docs document the canonical int64 wire form as a string, but the public Gemini API accepts both quoted and numeric forms in practice. Accepting both keeps Bifrost compatible with the documented form and with common JSON Schema producers.

## Tests

- `go test ./providers/gemini -run TestGeminiGenerationRequestUnmarshalAcceptsSchemaIntegerConstraints -count=1`
- `go test ./providers/gemini -run 'TestGeminiGenerationRequestUnmarshalAcceptsSchemaIntegerConstraints|TestBifrostToGeminiToolConversion' -count=1`

Known unrelated local note: full `go test ./providers/gemini -count=1` currently fails on `TestListModelsByKey_ParsesSingleModelPayload`, which is outside this patch.

## Agent Disclosure

This issue and patch were investigated, drafted, and submitted with AI-agent assistance. Focused local regression tests were run before submission.

## Summary by CodeRabbit

* **Bug Fixes**
  * Schema constraint parsing now accepts integer constraints as numbers or quoted strings, treats null as unset, and correctly handles large integer limits to reduce decoding errors.

* **Tests**
  * Added tests covering numeric and quoted-string constraint formats, null handling, large-value acceptance, and clear failures for non-numeric or invalid constraint inputs.

### DIFF
--- a/core/providers/gemini/types.go
+++ b/core/providers/gemini/types.go
@@ -830,10 +830,10 @@ const (
 type TrafficType string
 
 const (
-	TrafficTypeUnspecified          TrafficType = "TRAFFIC_TYPE_UNSPECIFIED"
-	TrafficTypeOnDemand             TrafficType = "ON_DEMAND"
-	TrafficTypeOnDemandPriority     TrafficType = "ON_DEMAND_PRIORITY"
-	TrafficTypeOnDemandFlex         TrafficType = "ON_DEMAND_FLEX"
+	TrafficTypeUnspecified           TrafficType = "TRAFFIC_TYPE_UNSPECIFIED"
+	TrafficTypeOnDemand              TrafficType = "ON_DEMAND"
+	TrafficTypeOnDemandPriority      TrafficType = "ON_DEMAND_PRIORITY"
+	TrafficTypeOnDemandFlex          TrafficType = "ON_DEMAND_FLEX"
 	TrafficTypeProvisionedThroughput TrafficType = "PROVISIONED_THROUGHPUT"
 )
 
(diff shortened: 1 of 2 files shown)


yanhao98:
Addressed the Greptile feedback in follow-up commit `a288b98b4`:

- `Schema.UnmarshalJSON` now uses `type schemaAlias Schema` and overrides only the six flexible integer constraint fields, avoiding the hand-maintained full field copy drift risk.
- The invalid string constraint regression assertion is now wrapped in `t.Run("invalid string constraint", ...)` for named CI output.

Verification run locally:

- `go test ./providers/gemini -run TestGeminiGenerationRequestUnmarshalAcceptsSchemaIntegerConstraints -count=1`
- `go test ./providers/gemini -run 'TestGeminiGenerationRequestUnmarshalAcceptsSchemaIntegerConstraints|TestBifrostToGeminiToolConversion' -count=1`\n- `PATH="/home/usr_vscode/go/bin:$PATH" make test-core PROVIDER=gemini TESTCASE=TestGeminiGenerationRequestUnmarshalAcceptsSchemaIntegerConstraints`\n- `gofmt -l core/providers/gemini/types.go core/providers/gemini/gemini_test.go`\n- `git diff --check -- core/providers/gemini/types.go core/providers/gemini/gemini_test.go`

core/providers/gemini/types.go
```diff
@@ -997,6 +997,70 @@ type Schema struct {
 	Type Type `json:"type,omitempty"`
 }
 
+type flexibleSchemaInt64 int64
+
+func (i *flexibleSchemaInt64) UnmarshalJSON(data []byte) error {
```
akshaydeo, line 1004: this can be simplfied a lot by unmarshalling to interface and checking the type + use `atoi`

could you try that please

yanhao98: Thanks, I’ll give that a try. Just to be transparent: I’m not very familiar with Go myself, so I’ll have my GPT-5.5 agent work through the simplification and then run an additional agent review pass before pushing the follow-up. I’ll make sure the revised version preserves the int64 behavior and rerun the focused Gemini tests.

yanhao98: Pushed the follow-up. I used interface decoding + a type switch as suggested. For the string branch I used `strconv.ParseInt(..., 10, 64)` instead of `Atoi` because these schema fields are `*int64`; this keeps the conversion aligned with the field type while preserving the same intent.

akshaydeo: this should work

akshaydeo (approved):

Issue #3216, opened 2026-05-04
[Bug]: disableAuthOnInference resets to false on every pod restart (Helm chart v1.7.0)


DudiPeretz-Orca:
## Environment
- **Bifrost image:** `maximhq/bifrost:v1.5.0-prerelease8`
- **Helm chart:** `1.7.0`
- **Deployment:** Kubernetes (EKS) via Helm

## Problem

Every time the Bifrost pod restarts, `disable_auth_on_inference` resets to `false`, causing all virtual key inference requests to return `401 Unauthorized`.

This happens even though `disableAuthOnInference: true` is explicitly set in Helm values:

```yaml
bifrost:
  authConfig:
    isEnabled: true
    disableAuthOnInference: true
```

## Steps to Reproduce

1. Deploy Bifrost via Helm chart v1.7.0 with `bifrost.authConfig.disableAuthOnInference: true`
2. Restart the pod (upgrade, rollout restart, etc.)
3. Make a virtual key inference request — returns `401 Unauthorized`
4. Check `GET /api/config` — `auth_config.disable_auth_on_inference` is `false`

## Expected Behavior

`disableAuthOnInference: true` set in Helm values should persist across pod restarts. The setting should be loaded from Helm values on startup and not overridden by stale SQLite state.

## Actual Behavior

On every pod restart, `disable_auth_on_inference` resets to `false`. The Helm value is not respected.

## Workaround

After every pod restart, manually re-apply via API:

```bash
curl -s https://<bifrost-host>/api/config \
  -H "Authorization: Basic <base64-credentials>" | \
  python3 -c "import sys,json; c=json.load(sys.stdin); c['auth_config']['disable_auth_on_inference']=True; print(json.dumps(c))" | \
  curl -s -X PUT https://<bifrost-host>/api/config \
  -H "Content-Type: application/json" \
  -H "Authorization: Basic <base64-credentials>" \
  -d @-
```

## Additional Context

This has persisted across prerelease3 → prerelease4 → prerelease5 → prerelease6 → prerelease8. The auth fix in prerelease8 (`Adjusted API key handling in middleware; corrected request context updates when auth config is disabled`) did not resolve this specific issue.

akshaydeo:
> disableAuthOnInference
We are deprecating this field from auth_config. To enforce virtual keys on inference please use field in client_config

DudiPeretz-Orca:
[...]

PR #4048, opened 2026-06-04
fix(auth): InferenceMiddleware now respects client_config.enforce_auth_on_inference

### LINKED ISSUE #4047: disableAuthOnInference resets to false on every pod restart (Helm chart bug)
### Summary
The `disableAuthOnInference: true` Helm value does not persist across pod restarts. Every time the pod restarts, the setting resets to `false`, causing all virtual key inference requests to return `401 Unauthorized`.

### Environment
- **Bifrost version:** v1.5.7
- **Helm chart version:** 2.1.20
- **Deployment:** Kubernetes (EKS)
- **Storage:** SQLite

### Steps to Reproduce
1. Set `disableAuthOnInference: true` in Helm values and deploy
2. Verify inference works (virtual key returns 200)
3. Restart the pod (e.g. via `terraform apply`, `kubectl rollout restart`, or any chart upgrade)
4. Attempt inference with a virtual key — returns `401 Unauthorized`

### Expected Behavior
`disableAuthOnInference: true` set in Helm values should persist across pod restarts and not require manual re-application.

### Actual Behavior
After every pod restart, `disable_auth_on_inference` resets to `false` in the config. The Helm value `disableAuthOnInference: true` appears to be ignored on startup.

### Workaround
Manually re-apply the setting via API after every restart:
```bash
curl -s https://<bifrost-host>/api/config \
  -H "Authorization: Basic <base64-credentials>" | \
  python3 -c "import sys,json; c=json.load(sys.stdin); c['auth_config']['disable_auth_on_inference']=True; print(json.dumps(c))" | \
  curl -s -X PUT https://<bifrost-host>/api/config \
  -H "Authorization: Basic <base64-credentials>" \
  -d @-
```

### Impact
High — affects all users using virtual keys for inference. Any deployment operation (upgrade, config change, node restart) silently breaks authentication for all virtual key holders until manually fixed.

### Relevant Helm Values
```yaml
bifrost:
  client:
    enforceAuthOnInference: false
```

DudiPeretz-Orca:
## Problem

`InferenceMiddleware` only checked `auth_config.disable_auth_on_inference` to decide whether to skip auth on inference endpoints. This field **resets to `false` on every pod restart** because Bifrost reloads it from SQLite (which starts empty) rather than from Helm/config values.

The correct persistent field is `client_config.enforce_auth_on_inference`, which IS persisted correctly from Helm values across restarts — but the inference auth middleware was never updated to read it after the migration was introduced.

This results in all virtual key inference requests returning `401 Unauthorized` after every pod restart, until the setting is manually re-applied via API.

## Root Cause

In `InferenceMiddleware()` (middlewares.go:810):
```go
// BEFORE — only checks the legacy field that resets on restart
func (m *AuthMiddleware) InferenceMiddleware() schemas.BifrostHTTPMiddleware {
    return m.middleware(func(authConfig *configstore.AuthConfig, url string) bool {
        return authConfig.DisableAuthOnInference
    })
}
```

`auth_config.disable_auth_on_inference` starts as `false` in a fresh SQLite DB on every restart, regardless of what's set in Helm values. Meanwhile, `client_config.enforce_auth_on_inference` is correctly loaded from config on startup but was never wired into the inference auth check.

## Fix

- Added `enforceAuthOnInference atomic.Bool` to `AuthMiddleware`
- Loaded from `ClientConfig.EnforceAuthOnInference` in `InitAuthMiddleware`
- Added `UpdateEnforceAuthOnInference()` called on client config updates (same pattern as `UpdateWhitelistedRoutes` / `UpdateTempTokenAuthEnabled`)
- `InferenceMiddleware` now skips auth when **either**:
  - `auth_config.disable_auth_on_inference == true` (legacy, runtime-set via API)
  - `client_config.enforce_auth_on_inference == false` (new field, persists from Helm/config across restarts)

```go
// AFTER
func (m *AuthMiddleware) InferenceMiddleware() schemas.BifrostHTTPMiddleware {
    return m.middleware(func(authConfig *configstore.AuthConfig, url string) bool {
        return authConfig.DisableAuthOnInference || !m.enforceAuthOnInference.Load()
    })
}
```

## Testing

After this fix, setting `bifrost.client.enforceAuthOnInference: false` in Helm values will correctly persist across pod restarts without requiring manual API patching.

Fixes #4047
Refs #3216


TejasGhatte:
we have updated this behaviour, inference requests are handled by governance layer

PR #4148, opened 2026-06-08
rerank: accept Cohere-spec string-array documents (sonic-safe normalization)

terafin:
## Summary

Closes the bug where `POST /v1/rerank` with `documents: ["a", "b"]` (the Cohere-spec string-array form, sent by Open WebUI and many other OpenAI-compatible clients) is rejected by Bifrost with `Mismatch type schemas.RerankDocument with value string`, while the object-array form (`documents: [{"text": "a"}, ...]`) succeeds.

## Why this matters

Cohere's published [`/v1/rerank` API](https://docs.cohere.com/reference/rerank) takes `documents` as `[string]` natively. vLLM's `/v1/rerank` follows Cohere's spec. Open WebUI's `ExternalReranker.predict()` ([source](https://github.com/open-webui/open-webui/blob/main/backend/open_webui/retrieval/models/external.py)) sends `documents: list[str]`. Bifrost is the outlier requiring the object form, blocking standards-compliant clients.

## Why this fix isn't `UnmarshalJSON` on `RerankDocument`

A previous attempt (carried as a custom `func (d *RerankDocument) UnmarshalJSON(data []byte) error` on the element type) was confirmed ineffective in practice. The HTTP request body is decoded with `bytedance/sonic` (`transports/bifrost-http/handlers/inference.go:89`), whose decoder rejects type mismatches at the array-element level BEFORE invoking any custom `UnmarshalJSON` on the element type. The `"Mismatch type ... with value string"` error message is sonic's pre-check, fired before our `UnmarshalJSON` could run.

The fix has to be at the field type or in handler code, not via element-level `UnmarshalJSON`.

## The patch

1. Change `RerankRequest.Documents` from `[]schemas.RerankDocument` to `json.RawMessage`.
2. New `normalizeRerankDocuments(raw json.RawMessage) ([]schemas.RerankDocument, error)` helper that tries the string-array shape first, falls back to the object-array shape, returns the canonical slice.
3. Call it in `prepareRerankRequest()` before the validation loop. This covers both `/v1/rerank` (`inference.go`) and `/v1/async/rerank` (`asyncinference.go`) since both share `prepareRerankRequest()`.

No change to `BifrostRerankRequest`, no change to `schemas.RerankDocument`, no change to downstream providers — they continue to receive a `[]schemas.RerankDocument` slice.

## Tests

`transports/bifrost-http/handlers/rerank_normalize_test.go` (new) — 5 cases:

| Test | What it proves |
|---|---|
| `TestNormalizeRerankDocuments_String` | **THE FIX**: `["a","b"]` decodes to 2 docs with Text set |
| `TestNormalizeRerankDocuments_Object` | Regression guard: `[{"text":"a"},...]` still works |
| `TestNormalizeRerankDocuments_ObjectWithMeta` | Optional `id` + `meta` fields preserved on object form |
| `TestNormalizeRerankDocuments_Empty` | Nil + empty input return an error |
| `TestNormalizeRerankDocuments_Invalid` | Unrecognized shape returns a clear error mentioning both valid forms |

`go test ./transports/bifrost-http/handlers/ -run TestNormalizeRerankDocuments` passes locally.

## Validation flow

After this lands:

```bash
# both forms should return 200 with relevance scores
curl -X POST http://<bifrost>:8080/v1/rerank \
  -H 'x-bf-vk: <vk>' \
  -d '{"model":"...","query":"hello","documents":["a","b"],"top_n":2}'

curl -X POST http://<bifrost>:8080/v1/rerank \
  -H 'x-bf-vk: <vk>' \
  -d '{"model":"...","query":"hello","documents":[{"text":"a"},{"text":"b"}],"top_n":2}'
```

## Summary by CodeRabbit

* **New Features**
  * /v1/rerank now accepts documents as either string arrays or object arrays (with optional id/meta), normalizes them to a canonical form, requires a non-empty result, and returns clear errors for missing, null, empty, or unsupported payloads.

* **Tests**
  * Added unit and boundary tests covering string-array, object-array (with id/meta), empty-array, null/missing documents, and invalid shapes.

### DIFF
--- a/core/schemas/rerank.go
+++ b/core/schemas/rerank.go
@@ -1,5 +1,13 @@
 package schemas
 
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+
+	"github.com/bytedance/sonic"
+)
+
 // RerankDocument represents a document to be reranked.
 type RerankDocument struct {
 	Text string                 `json:"text"`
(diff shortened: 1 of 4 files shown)


terafin:
Both Greptile P2 items addressed in commit `e9f4eef`:

1. **`null` guard** — added `|| string(raw) == "null"` to the early-return so the error message is consistent with the missing-field case (function returns `"documents is required"` instead of letting downstream fire a different message).

2. **Empty array `[]` test** — added `TestNormalizeRerankDocuments_EmptyArray` confirming the existing behavior (string-array branch returns a zero-length slice with no error, caller's `len(docs) == 0` guard then surfaces the rerank-specific error message). Also extended the existing `_Empty` test to cover the literal JSON `null` case.

Both tests run via `go test ./transports/bifrost-http/handlers/ -run TestNormalizeRerankDocuments` and pass.

terafin:
Both coderabbit items addressed:

1. **`string(raw) == "null"` → `bytes.Equal(raw, []byte("null"))`** in `inference.go:578` (avoids the slice copy/allocation). Added `"bytes"` import.

2. **Handler-boundary test added** as `TestRerankRequest_HandlerBoundary` in `rerank_normalize_test.go` — table-driven across 6 cases (string-array, object-array, empty-array, null, missing-field, invalid-shape). Each case does `sonic.Unmarshal` of a full request body into `RerankRequest` then `normalizeRerankDocuments(req.Documents)`, exercising the same sonic → json.RawMessage → normalize chain production code follows.

This was the gap that caused the 2026-06-08 bifrost incident (first patch used a unit test calling `UnmarshalJSON` via stdlib `encoding/json`, which passed; production decoded via sonic, which silently bypassed the custom unmarshaller). The handler-boundary test catches that regression class directly.

terafin:
Addressed the third coderabbit nit:

**Avoid double-decode on object-array payloads** — sniff the first non-whitespace character of the array element to dispatch to a single `sonic.Unmarshal`. `["`-leading → string array; everything else (including `[]`, object-array, and any future shape) → object-array decode. Hot-path rerank requests with large `documents` no longer pay for a failed `[]string` decode before the real `[]RerankDocument` decode.

Also added `bytes.TrimSpace` to the empty/null guard for defense against whitespace-padded `null` inputs (which sonic could surface in theory).

All 7 `TestNormalizeRerankDocuments_*` + `TestRerankRequest_HandlerBoundary` cases still pass — sniffer is a perf optimization that doesn't change behaviour for any of the existing test inputs.

terafin:
Skipping the `strings.TrimSpace` removal nit as out-of-scope for this PR.

The trim-on-validate behaviour at `inference.go:1250` + `:1261` is **pre-existing** in `prepareRerankRequest` — not introduced by this PR. This PR's scope is "accept Cohere-spec string-array documents (sonic-safe normalization)" — switching the gateway-side whitespace policy is a separate concern that affects ALL rerank validation, not just the new string-array path.

Happy to file a follow-up PR removing the trim across the validation chain (and probably similar for other handlers if the coding-guideline applies generally), but want to keep this one's diff narrow to the rerank-shape fix so review surface stays minimal.

terafin:
Sweep of the bot feedback this morning — all four items resolved on the now-pushed commit:

**greptile #1 — `null` JSON silently passes the guard**
Already handled in the current code:
```go
trimmed := bytes.TrimSpace(raw)
if len(trimmed) == 0 || bytes.Equal(trimmed, []byte("null")) {
    return nil, fmt.Errorf("documents is required")
}
```
greptile's review was posted at 10:34Z against an earlier rev; the `bytes.Equal` check landed in the squashed commit.

**greptile #2 — empty array `[]` not covered by tests**
Already covered by `TestNormalizeRerankDocuments_EmptyArray` (rerank_normalize_test.go:87) — verifies `[]` parses cleanly via the string-array branch with `len(docs) == 0`, and the caller-owned message ("documents are required for rerank") fires from `prepareRerankRequest`. Two-layer design intentional so the rerank-specific phrasing comes from the handler, not the shared helper.

**coderabbit #1 — avoid `string(raw) == "null"` allocation**
Already addressed — switched to `bytes.Equal(raw, []byte("null"))` (no allocation) in the same commit that landed the sniff-decode optimization.

**coderabbit #2 — avoid second decode on object-array payloads**
Already addressed via the sniff-first branch:
```go
if len(trimmed) >= 2 && trimmed[0] == '[' {
    inner := bytes.TrimSpace(trimmed[1:])
    if len(inner) > 0 && inner[0] == '"' {
        // string-array branch — never falls through to object-array decode
    }
}
// Otherwise: object-array decode, no prior failed attempt
```
Object-array payloads (legacy shape, the hot path) no longer pay for a failed `[]string` decode first.

**coderabbit #3 — TrimSpace on query/doc text at handler boundary**
Just removed in the pushed commit (per project convention "handlers do minimal validation, no whitespace trimming"). The two `strings.TrimSpace` calls at lines 1250 + 1261 → plain `== ""` checks. Pre-existing code from upstream, but coderabbit's cited learning explicitly covers this surface.

Greptile's 5/5 "safe to merge" verdict still stands; the bot reviews after that point were either pre-empted by the squash or just landed.

Pratham-Mishra04:
Hey @terafin I think we should handle this unmarshal type in the BifrostReRankRequest type instead

terafin:
@Pratham-Mishra04 thanks for the review — refactored as requested. Force-pushed the new commit (`77929c84d`).

## What changed

Moved the shape-flexible `Documents` normalization onto `*BifrostRerankRequest.UnmarshalJSON` in `core/schemas/rerank.go`. The schema type now owns its wire-format concerns, so any direct SDK consumer of `BifrostRerankRequest` benefits — not just the HTTP transport. The normalize logic is exposed as the package-level helper `schemas.NormalizeRerankDocuments` so the HTTP-tier `RerankRequest` can reuse it from its own `UnmarshalJSON` (which keeps `Documents []schemas.RerankDocument` as the public field type — the `json.RawMessage` is now an internal alias-shadow detail).

## Why parent-struct UnmarshalJSON, not element-level

Element-level `UnmarshalJSON` on `RerankDocument` — which was my first instinct and what I tried in the original PR — does not work under sonic. Sonic's reflective decoder rejects type mismatches at the **array-element level BEFORE** invoking element-level `UnmarshalJSON`, surfacing `Mismatch type schemas.RerankDocument with value string`. That patch passed `encoding/json` unit tests then broke in production because `transports/bifrost-http/handlers/inference.go:90` decodes via `sonic.Unmarshal`.

I empirically verified that **parent-struct** `UnmarshalJSON` works correctly under sonic before refactoring — the parent method fires for both string-array and object-array shapes:

```go
type RerankDoc struct { Text string `json:"text"` }
type MyReq struct { Documents []RerankDoc `json:"documents"` }
func (r *MyReq) UnmarshalJSON(data []byte) error { /* normalizes both shapes */ }

sonic.Unmarshal([]byte(`{"documents":["a","b"]}`), &s) // -> parent fires, Documents=[{Text:a},{Text:b}], err=nil
sonic.Unmarshal([]byte(`{"documents":[{"text":"a"}]}`), &s) // -> parent fires, err=nil
```

So the fix has to live on the parent struct (or at the field-type level via `json.RawMessage`). Putting it on the schema-tier `BifrostRerankRequest` is the cleanest of those.

## Tests

`go test ./schemas/ ./bifrost-http/handlers/` (rerank-scoped) — all green:

- `TestNormalizeRerankDocuments_*` (6 cases — string, object, with-meta, empty, empty-array, invalid) — pass
- `TestBifrostRerankRequest_UnmarshalJSON` (5 cases — string-array, object-array, null/missing decode-to-nil, invalid-shape errors) — pass
- `TestRerankRequest_HandlerBoundary` (7 cases — full sonic-decode path, the load-bearing regression guard) — pass

Two pre-existing `TestGetVirtualKeys_*` governance failures on `dev` are unrelated.

## Behavior preserved

All optimizations from prior coderabbit rounds kept verbatim:
- `bytes.Equal` null-check fast-path
- sniff-first dispatch (peek first non-whitespace element of the array) so the common case doesn't pay for a failed-decode attempt at the other shape
- `documents must be []string or [{text:string}]` error message
- caller-owned `documents are required for rerank` phrasing — null/missing `documents` decode to nil so the handler's existing `len(docs) == 0` guard surfaces the friendlier rerank-specific message rather than a raw unmarshal error

Production smoke against current bifrost (`https://bifrost.siliconspirit.net/v1/rerank`) — both shapes return HTTP 200:

```
$ curl -X POST .../v1/rerank -d '{"model":"mxbai-rerank-base-v2","query":"test","documents":["a","b"]}'
HTTP 200, results=[{index:1,score:0.78},{index:0,score:0.73}]

$ curl -X POST .../v1/rerank -d '{"model":"mxbai-rerank-base-v2","query":"test","documents":[{"text":"a"},{"text":"b"}]}'
HTTP 200, results=[{index:1,score:0.78},{index:0,score:0.73}]
```

LMK if you want any further changes.

PR #4261, opened 2026-06-10
fix: avoid duplicating inference-provider segment in backfilled huggingface model IDs

### LINKED ISSUE #4215: [Bug]: HuggingFace models show provider ID twice in /v1/models, which breaks requests
### Prerequisites

### Description

When models are selected from HuggingFace (i.e. selecting just one model, and not permitting all models), the provider name in the model ID shows up twice, such as ```huggingface/featherless-ai/featherless-ai/deepseek-ai/DeepSeek-V4-Pro``` instead of ```huggingface/featherless-ai/deepseek-ai/DeepSeek-V4-Pro```. 

Image

This can be verified by getting the model IDs manually:

```curl https://bifrost.mycompany.corp/v1/models```
`
...
{
      "id": "huggingface/featherless-ai/featherless-ai/deepseek-ai/DeepSeek-V4-Pro",
      "name": "Featherless Ai/Deepseek Ai/DeepSeek V4 Pro"
    },
...
`

In interfaces such as Open WebUI, when requests are sent to this model, bifrost returns an error message, like so:

Image

no keys found that support model: featherless-ai/featherless-ai/deepseek-ai/DeepSeek-V4-Pro

This is because Bifrost does not recognize the model with two provider names in the model ID. 

Oddly enough, permitting all models from HuggingFace does not carry this issue, and sending requests manually through curl with one provider name in the model ID instead of two works as expected. 

### Steps to reproduce

1. Configure HuggingFace as a Provider, and pick one model to invoke (DeepSeek-V4-Pro was used in this example, have replicated with Nemotron-3-Ultra and GLM-5.1)
2. Call ```curl <bifrost_url>/v1/models```
3. Examine the model ID returned, and see the provider name returned twice in any given response from HuggingFace.

### Expected behavior

The provider name shows up only once in the /v1/models response, in the format huggingface/<provider_id>/repo_name/model_name

### Actual behavior

The provider name shows up twice, in the format huggingface/<provider_id>/<provider_id>/repo_name/model_name

### Affected area(s)

Transports (HTTP)

### Version

1.4.6 Enterprise, 1.4.8 Enterprise

### Environment

```text

```

### Relevant logs/output

```shell
{
      "id": "huggingface/cohere/featherless-ai/deepseek-ai/DeepSeek-V4-Pro",
      "name": "Featherless Ai/Deepseek Ai/DeepSeek V4 Pro"
},
```

### Regression?

_No response_

### Severity

truffle-dev:
## Summary

Closes #4215.

When individual HuggingFace models are selected on a key (instead of allow-all), `/v1/models` returns IDs with the inference-provider segment duplicated, e.g. `huggingface/cohere/featherless-ai/deepseek-ai/DeepSeek-V4-Pro`. Requests against those IDs fail because `splitIntoModelProvider` parses the first segment (`cohere`) as the inference provider and the rest as the model name.

Root cause: allowlist entries selected from a previous `ListModels` response already carry an inference-provider segment (`featherless-ai/deepseek-ai/DeepSeek-V4-Pro`). The hub fetch returns model IDs without that segment, so `FilterModel` never matches the entry and `BackfillModels` emits it verbatim. The backfill re-wrap in `huggingface/models.go` then blindly prepends the current goroutine's inference provider, duplicating the segment. Because `listModelsByKey` runs one conversion per inference provider and aggregates without dedupe, the same entry was also emitted once per provider, each with a different (wrong) prefix; that is why the reproduction log shows `cohere` in front of a `featherless-ai` model. Allow-all is unaffected since the backfill path never runs without a restricted allowlist.

## Changes

- In the backfill re-wrap, check whether the entry's first segment names a known inference provider. If it matches the current pass's provider, keep the ID as-is instead of prepending again. If it names a different provider, skip the entry; that provider's own pass emits it, which also stops the one-copy-per-provider duplication.
- Entries without an inference-provider segment keep the existing behavior (current provider prepended), since we cannot know which provider serves them.
- Added `isKnownInferenceProvider` next to the re-wrap and regression tests in `models_test.go`.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [x] Core (Go)
- [ ] Transports (HTTP)
- [x] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

```sh
cd core
go test ./providers/huggingface/ -run TestToBifrostListModelsResponse -v
```

Without the fix, the new tests reproduce both reported shapes:

```
--- Expected
+++ Actual
@@ -1 +1 @@
-huggingface/featherless-ai/deepseek-ai/DeepSeek-V4-Pro
+huggingface/featherless-ai/featherless-ai/deepseek-ai/DeepSeek-V4-Pro

Should be empty, but was [{huggingface/cohere/featherless-ai/deepseek-ai/DeepSeek-V4-Pro ...}]
```

With the fix, `go test ./providers/huggingface/` and `./providers/utils/` pass; `gofmt`, `go vet`, and `golangci-lint` are clean on the touched files.

Note: `core/go.sum` on current `dev` is missing the `go.mod` hashes for `github.com/aws/aws-sdk-go-v2` (and transitively `smithy-go`), so `go test ./...` fails at setup before this change; I ran `go mod download` locally to test and left `go.sum` untouched in this PR.

## Screenshots/Recordings

N/A

## Breaking changes

- [ ] Yes
- [x] No

## Related issues

Closes #4215

## Security considerations

None.

## Checklist

- [ ] I updated documentation where needed (N/A)

### DIFF
--- a/core/providers/huggingface/models.go
+++ b/core/providers/huggingface/models.go
@@ -74,13 +74,26 @@ func (response *HuggingFaceListModelsResponse) ToBifrostListModelsResponse(provi
 		// or the "auto" policy. Prepending another segment here duplicates the
 		// provider in the compound ID and breaks request routing (#4215).
 		if first, _, found := strings.Cut(rawID, "/"); found && isKnownInferenceProviderOrPolicy(first) {
-			if !strings.EqualFold(first, string(inferenceProvider)) {
-				// The entry belongs to a different inference provider's listing
-				// (or names the "auto" policy, which no pass owns); skip it here
-				// to avoid duplicating the entry once per inference provider.
+			switch {
+			case strings.EqualFold(first, string(auto)):
+				// "auto" is owned by no inference-provider pass: the model
+				// listing loop iterates INFERENCE_PROVIDERS, which excludes it.
+				// Emit the entry exactly once, during the canonical first pass,
+				// so a routable auto-policy allowlist entry still surfaces in
+				// the listing instead of being dropped from every pass. Every
+				// other pass skips it to avoid duplicating it once per provider.
+				if len(INFERENCE_PROVIDERS) == 0 || inferenceProvider != INFERENCE_PROVIDERS[0] {
+					continue
+				}
+				m.ID = fmt.Sprintf("%s/%s", providerKey, rawID)
+			case !strings.EqualFold(first, string(inferenceProvider)):
+				// The entry belongs to a different inference provider's listing;
+				// it is emitted in that provider's pass. Skip it here to avoid
+				// duplicating the entry once per inference provider.
 				continue
+			default:
+				m.ID = fmt.Sprintf("%s/%s", providerKey, rawID)
 			}
-			m.ID = fmt.Sprintf("%s/%s", providerKey, rawID)
 		} else {
 			// Re-wrap the backfill ID to include the inferenceProvider segment
 			m.ID = fmt.Sprintf("%s/%s/%s", providerKey, inferenceProvider, rawID)
--- a/core/providers/huggingface/models.go
+++ b/core/providers/huggingface/models.go
@@ -68,13 +68,35 @@ func (response *HuggingFaceListModelsResponse) ToBifrostListModelsResponse(provi
 	// Backfill: use standard pipeline. Note that backfilled HF entries use a simplified
 	// compound ID since we don't know which inferenceProvider to assign them to.
 	for _, m := range pipeline.BackfillModels(included) {
-		// Re-wrap the backfill ID to include the inferenceProvider segment
 		rawID := strings.TrimPrefix(m.ID, string(providerKey)+"/")
-		m.ID = fmt.Sprintf("%s/%s/%s", providerKey, inferenceProvider, rawID)
+		// Allowlist entries selected from a previous ListModels response already
+		// carry an inference-provider segment (e.g. "featherless-ai/org/model").
+		// Prepending another segment here duplicates the provider in the compound
+		// ID and breaks request routing (#4215).
+		if first, _, found := strings.Cut(rawID, "/"); found && isKnownInferenceProvider(first) {
+			if !strings.EqualFold(first, string(inferenceProvider)) {
+				// The entry belongs to a different inference provider's listing;
+				// that provider's pass emits it, so skip it here to avoid
+				// duplicating the entry once per inference provider.
+				continue
+			}
+			m.ID = fmt.Sprintf("%s/%s", providerKey, rawID)
+		} else {
+			// Re-wrap the backfill ID to include the inferenceProvider segment
+			m.ID = fmt.Sprintf("%s/%s/%s", providerKey, inferenceProvider, rawID)
+		}
 		bifrostResponse.Data = append(bifrostResponse.Data, m)
 	}
 
 	return bifrostResponse
+}
+
+// isKnownInferenceProvider reports whether segment names one of the supported
+// inference providers (case-insensitive).
+func isKnownInferenceProvider(segment string) bool {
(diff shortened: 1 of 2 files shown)


TejasGhatte (dismissed):

truffle-dev:
The 06-11 review got dismissed when I pushed the follow-ups, so a quick recap of where this landed:

`2d265cf` is the core fix: the backfill path no longer prepends a second inference-provider segment to allowlist entries that already carry one — the `huggingface//featherless-ai/org/model` duplication that broke routing (#4215).

Greptile flagged the `auto` policy as a related gap, handled in two follow-ups:
- `5238499` recognizes `auto` as a known policy segment so an `auto/` entry isn't re-wrapped into a duplicate.
- `43a98e7` emits the `auto` entry once during the canonical first pass instead of skipping it in every pass. The listing loop iterates `INFERENCE_PROVIDERS`, which excludes `auto`, so before this an `auto/…`-restricted key returned an empty `/v1/models`.

Green across the matrix and mergeable. @TejasGhatte happy to walk through any of it.

core/providers/huggingface/models.go
```diff
@@ -68,15 +68,50 @@ func (response *HuggingFaceListModelsResponse) ToBifrostListModelsResponse(provi
...
+				// "auto" is owned by no inference-provider pass: the model
+				// listing loop iterates INFERENCE_PROVIDERS, which excludes it.
+				// Emit the entry exactly once, during the canonical first pass,
+				// so a routable auto-policy allowlist entry still surfaces in
+				// the listing instead of being dropped from every pass. Every
+				// other pass skips it to avoid duplicating it once per provider.
+				if len(INFERENCE_PROVIDERS) == 0 || inferenceProvider != INFERENCE_PROVIDERS[0] {
+					continue
+				}
+				m.ID = fmt.Sprintf("%s/%s", providerKey, rawID)
```
sammaji, line 88: @truffle-dev only `id` and `name` is returned by this path, we should also have `hugging_face_id` and `supported_methods`

core/providers/huggingface/models.go
```diff
@@ -68,15 +68,50 @@ func (response *HuggingFaceListModelsResponse) ToBifrostListModelsResponse(provi
...
 	// compound ID since we don't know which inferenceProvider to assign them to.
 	for _, m := range pipeline.BackfillModels(included) {
-		// Re-wrap the backfill ID to include the inferenceProvider segment
 		rawID := strings.TrimPrefix(m.ID, string(providerKey)+"/")
-		m.ID = fmt.Sprintf("%s/%s/%s", providerKey, inferenceProvider, rawID)
+		// Allowlist entries selected from a previous ListModels response already
+		// carry an inference-provider segment (e.g. "featherless-ai/org/model")
+		// or the "auto" policy. Prepending another segment here duplicates the
+		// provider in the compound ID and breaks request routing (#4215).
+		if first, _, found := strings.Cut(rawID, "/"); found && isKnownInferenceProviderOrPolicy(first) {
```
sammaji, line 76: @truffle-dev use the name from here:

```
first, modelName, found := strings.Cut(rawID, "/")
```

otherwise, the name returned does not match that one returned from the "all models" path.  
  
one allowed model:

```
{ "name": "Huggingface/Cerebras/Google/Gemma 4B It" }
```

  
all allowed models:

```
{ "name": "google/gemma-4b-it" }
```

sammaji:
* **#4261** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/4261?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/maximhq/bifrost/4261?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `dev`


<h2></h2>

PR #4296, opened 2026-06-11
feat: enforce VK tool-grant boundary on caller-provided `x-bf-mcp-include-tools` lists via `pruneMCPIncludeToolsFromContext`

Pratham-Mishra04:
## Summary

Caller-provided MCP include-tools lists (set via the `x-bf-mcp-include-tools` header) could previously reference tools that the virtual key does not grant, effectively allowing callers to expand their own tool access beyond what the key permits. This PR enforces that a caller-provided list can only narrow the virtual key's tool grant, never expand it.

## Changes

- Introduced `pruneMCPIncludeToolsFromContext`, which reads a caller-provided include-tools list from the context, intersects it with the tools the virtual key actually grants, and writes the pruned result back to the context.
- Caller wildcards (e.g. `sentry-*`) are kept verbatim only when the virtual key itself is unrestricted for that client; otherwise they are expanded to the key's specific grants to prevent the wildcard from being interpreted downstream as "all tools for this client."
- Auto-injection of the computed tool list is now skipped when a caller-provided list was present (even if it pruned to empty), replacing the previous check that only skipped injection when the context key was unset.
- Pruning is applied in both `runPreRequestRouting` and `PreRequestHook`, replacing the earlier pattern of checking for an existing context value before injecting.
- A wrong-type context value fails closed, pruning to a deny-all empty list rather than passing through.
- Added a dedicated test file covering: no caller list, disallowed tools dropped, granted tools kept, specific tools under unrestricted grants, wildcard passthrough vs. narrowing, ungranted client wildcards, empty header opt-out, deduplication, `AllowOnAllVirtualKeys` clients, explicit empty config overriding allow-all, multi-client pruning, and wrong-type fail-closed behavior.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [ ] Core (Go)
- [ ] Transports (HTTP)
- [ ] Providers/Integrations
- [x] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

```sh
go test ./plugins/governance/...
```

The new test file `prunemcpincludetools_test.go` covers all pruning scenarios. Verify that:
- A caller requesting a tool not in the virtual key's grant receives an empty (deny-all) list.
- A caller requesting a subset of granted tools receives only that subset.
- A caller wildcard is narrowed to the key's specific grants when the key is not itself unrestricted for that client.
- Auto-injection does not overwrite a caller-provided list, even when the pruned result is empty.

## Breaking changes

- [ ] Yes
- [x] No

## Security considerations

This change closes a privilege escalation path where a caller could use the `x-bf-mcp-include-tools` header to request tools beyond what their virtual key grants. The pruning logic ensures the effective tool set is always bounded by the virtual key's configured permissions, and any ambiguous or malformed input fails closed to a deny-all state.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Caller-provided MCP include-tools lists are now pruned to only the tools allowed for the active virtual key; auto-injection of tools occurs only when callers provide no list and auto-injection isn’t disabled.

* **Tests**
  * Added comprehensive tests covering pruning behavior, wildcards, deduplication, opt-out cases, multi-client scenarios, and fail-closed handling.

### DIFF
--- a/plugins/governance/main.go
+++ b/plugins/governance/main.go
@@ -428,14 +428,16 @@ func (p *GovernancePlugin) runPreRequestRouting(ctx *schemas.BifrostContext, vir
 			return modelIn, err
 		}
 
+		// A caller-provided include-tools list can only narrow the virtual key's
+		// tool grant, never expand it — prune entries the key does not allow.
+		callerProvided := p.pruneMCPIncludeToolsFromContext(ctx, virtualKey)
+
 		p.cfgMutex.RLock()
 		autoInjectDisabled := p.disableAutoToolInject != nil && *p.disableAutoToolInject
 		p.cfgMutex.RUnlock()
-		if !autoInjectDisabled {
-			if existing := ctx.Value(schemas.MCPContextKeyIncludeTools); existing == nil {
-				if tools := p.computeMCPIncludeTools(virtualKey); tools != nil {
-					ctx.SetValue(schemas.MCPContextKeyIncludeTools, tools)
-				}
+		if !callerProvided && !autoInjectDisabled {
+			if tools := p.computeMCPIncludeTools(virtualKey); tools != nil {
+				ctx.SetValue(schemas.MCPContextKeyIncludeTools, tools)
 			}
 		}
 	}
@@ -781,13 +783,19 @@ func (p *GovernancePlugin) applyRoutingRules(ctx *schemas.BifrostContext, req *s
 // directly; callers store it via ctx.SetValue(schemas.MCPContextKeyIncludeTools, ...). VK-specific
 // MCP configs take precedence over AllowOnAllVirtualKeys clients.
 func (p *GovernancePlugin) computeMCPIncludeTools(virtualKey *configstoreTables.TableVirtualKey) []string {
-	executeOnlyTools := make([]string, 0)
-
-	// Build a lookup of AllowOnAllVirtualKeys clients: clientID -> clientName
 	var allowAllVKsClients map[string]string
 	if p.inMemoryStore != nil {
 		allowAllVKsClients = p.inMemoryStore.GetMCPClientsAllowingAllVirtualKeys()
 	}
+	return p.computeMCPIncludeToolsWith(virtualKey, allowAllVKsClients)
+}
+
+// computeMCPIncludeToolsWith is the computeMCPIncludeTools variant taking a pre-fetched
+// AllowOnAllVirtualKeys map (clientID → clientName), so callers that make multiple
+// grant decisions per request can evaluate them all against one consistent snapshot.
+func (p *GovernancePlugin) computeMCPIncludeToolsWith(virtualKey *configstoreTables.TableVirtualKey, allowAllVKsClients map[string]string) []string {
+	executeOnlyTools := make([]string, 0)
+
 	if allowAllVKsClients == nil {
 		allowAllVKsClients = make(map[string]string)
 	}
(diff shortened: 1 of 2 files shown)


Pratham-Mishra04:
* **#4298** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/4298?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4297** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/4297?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4296** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/4296?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/maximhq/bifrost/4296?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `dev`


<h2></h2>

Pratham-Mishra04:
@coderabbitai full-review

Pratham-Mishra04:
@coderabbitai  
  


This suggestion is solving a problem that doesn't exist, and the sketch as written would reintroduce the leak we just fixed. Two issues:

1. There's no duplicate computation today. The two computeMCPIncludeTools calls are on mutually exclusive paths:

func (p *GovernancePlugin) pruneMCPIncludeToolsFromContext(...) bool {
existing := ctx.Value(schemas.MCPContextKeyIncludeTools)
if existing == nil {
return false        // ← returns BEFORE computing anything
}
...
vkTools := p.computeMCPIncludeTools(virtualKey)  // only on the prune path

When a caller list exists, prune computes once and the inject branch is skipped (!includeToolsProvided is false). When it doesn't, prune returns before computing and the inject branch computes once. Exactly one computation per request either way. The suggested version actually makes it worse: it computes eagerly even on the path where nothing is stamped at all (auto-inject off, no filters) — wasted work on every such request.

1. The len(vkTools) > 0 guard is a regression. computeMCPIncludeTools never returns nil — for a VK with no MCP grants it returns an empty list, and stamping that empty list is load-bearing: it's the deny-all that stops core from injecting anything when an include-clients header has opted the request into injection. With len(vkTools) > 0, that case leaves the ctx key unset → core opts in via include-clients → every tool of the named client leaks. That's precisely the bug from your curl, and TestPreRequestHookMCP_NoGrants_IncludeClients_StampsDenyAll exists to catch it — this change would fail that test. (The sketch also drops the || includeClientsPresent condition, though I assume that's just sketching against the older call site.)

If this came from a review bot, I'd resolve it with that explanation rather than apply it. If what you like about it is the signature — making the data flow explicit instead of the boolean — a corrected version would be:

includeToolsProvided, vkTools := p.pruneMCPIncludeToolsFromContext(ctx, virtualKey)
...
if !includeToolsProvided && (!autoInjectDisabled || includeClientsPresent) {
ctx.SetValue(schemas.MCPContextKeyIncludeTools, vkTools)  // unconditional — empty = deny-all
}

with prune computing vkTools up front. That's behaviorally safe but trades the lazy computation for a slightly cleaner call site. My recommendation is to leave the code as is — say the word if you'd like the corrected variant applied instead.

akshaydeo:
### Merge activity

* **Jun 11, 4:21 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/4296).
* **Jun 11, 4:22 PM UTC**: @akshaydeo merged this pull request with [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/4296).

PR #4350, opened 2026-06-12
fix: map network_config.allow_private_network into rendered Helm config

### LINKED ISSUE #4146: [Bug]: Helm chart cannot set network_config.allow_private_network (schema + config helper drop it)
### Prerequisites

### Description

v1.5.9 added `network_config.allow_private_network` (default `false`), which blocks RFC1918 IPs. maximhq/bifrost#4138 was closed with "set `allow_private_network: true` in your network config" — but on the official Helm chart (`bifrost-2.1.21`) there's no way to set it. Two separate things in the chart drop the field:

1. `values.schema.json` → `$defs/networkConfig` is `"additionalProperties": false` and doesn't list `allow_private_network`, so `helm upgrade` fails schema validation before it ever renders.
2. If you get past the schema, the `bifrost.config` helper in `templates/_helpers.tpl` (the `network_config` block) only copies a hardcoded set of keys, so the field never makes it into the rendered `config.json` anyway.

There's no `existingConfigMap` escape hatch either — `templates/deployment.yaml` hardcodes the chart-generated config volume — so you can't hand-write `config.json` without a post-renderer.

### Steps to reproduce

1. Deploy the `bifrost` Helm chart `2.1.21` with a provider whose `base_url` is an in-cluster / RFC1918 address.
2. Add `allow_private_network: true` under that provider's `network_config`.
3. `helm upgrade` fails: `additional properties 'allow_private_network' not allowed`.
4. Remove the field to upgrade successfully → provider fails at runtime with `connection to private IP <ip> is not allowed`.

### Expected behavior

The chart lets me set `allow_private_network` (in both the schema and the config helper) so I can use the same opt-in that maximhq/bifrost#4138 points to.

### Actual behavior

The schema rejects the field, and the config helper drops it, so there's no way to turn it on from Helm.

### Affected area(s)

Transports (HTTP)

### Version

Helm chart `2.1.21` (appVersion `1.5.3`); app `v1.5.9`–`v1.5.10`

### Environment

```text
- Helm chart: bifrost-2.1.21
- App image: docker.io/maximhq/bifrost:v1.5.10
- Deploy: Kubernetes (Helm)
- Provider: OpenAI-compatible, base_url = in-cluster ClusterIP (RFC1918)
```

### Relevant logs/output

```shell
# helm upgrade
values don't meet the specifications of the schema(s) in the following chart(s):
bifrost:
- at '/bifrost/providers/<provider>/network_config': additional properties 'allow_private_network' not allowed

# runtime (field omitted to pass schema)
http.error={"error":{"error":"connection to private IP 10.0.0.10 is not allowed",
"message":"failed to execute HTTP request to provider API"},
"extra_fields":{"provider":"<provider>","request_type":"chat_completion_stream"}}
```

### Regression?

Worked in app `v1.5.8`; broke in `v1.5.9`, with no Helm-chart path to the new opt-in field.

### Severity

truffle-dev:
## Summary

Fixes the Helm chart silently dropping `network_config.allow_private_network` from the rendered Bifrost config. As @gabrielcosi diagnosed in #4146, the field was rejected by the schema and dropped by `_helpers.tpl`. At current main, the schema half is already fixed — `allow_private_network` is listed under `$defs/networkConfig` since the monorepo chart import (9399212) — which leaves the worse failure mode live: the schema accepts the value, then the helper's hardcoded key-copy block discards it. `helm template` succeeds and the provider ships with the default (`false`), so requests to private base URLs like `http://10.x.x.x:11434` fail at runtime with no signal from the chart.

## Changes

- `helm-charts/bifrost/templates/_helpers.tpl`: copy `allow_private_network` into the rendered provider `network_config`, following the same `hasKey` pattern as the surrounding fields (`max_retries`, `retry_backoff_*`, `extra_headers`).
- `.github/workflows/scripts/validate-helm-config-fields.sh`: add `allow_private_network: true` to the openai provider fixture and an `assert_field_value` check so the field can't be silently dropped again.

No schema change needed — `values.schema.json` already accepts the field at main.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [ ] Core (Go)
- [ ] Transports (HTTP)
- [ ] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [ ] Docs

(Helm chart only.)

## How to test

```sh
# Field-mapping validation suite (covers the new assertion)
bash .github/workflows/scripts/validate-helm-config-fields.sh
# expect: 349/349 assertions pass, including
#   providers.openai.network_config.allow_private_network = true

# Template rendering suite (no regressions)
bash .github/workflows/scripts/validate-helm-templates.sh

# End-to-end repro from the issue
helm template bifrost helm-charts/bifrost -f values.yaml \
  --set 'providers.openai.network_config.base_url=http://10.0.12.34:11434' \
  --set 'providers.openai.network_config.allow_private_network=true'
# rendered config now contains:
#   "network_config":{"allow_private_network":true,"base_url":"http://10.0.12.34:11434"}
```

Verified the new assertion is load-bearing: with the `_helpers.tpl` change stashed, the suite fails with `got: MISSING` on exactly this field; with it applied, 349/349 pass.

## Breaking changes

- [ ] Yes
- [x] No

Charts that never set the field render identically (the copy is gated on `hasKey`).

## Related issues

Closes #4146

## Security considerations

`allow_private_network` permits a provider to target private-range base URLs; this PR only makes the chart honor the value an operator explicitly sets in `values.yaml`. The default remains unset/false.

## Checklist

### DIFF
--- a/.github/workflows/scripts/validate-helm-config-fields.sh
+++ b/.github/workflows/scripts/validate-helm-config-fields.sh
@@ -287,6 +287,7 @@ bifrost:
         max_retries: 5
         retry_backoff_initial_ms: 200
         retry_backoff_max_ms: 10000
+        allow_private_network: true
       concurrency_and_buffer_size:
         concurrency: 50
         buffer_size: 100
@@ -345,6 +346,7 @@ assert_field_value 'providers.openai.network_config.default_request_timeout_in_s
 assert_field_value 'providers.openai.network_config.max_retries' '.providers.openai.network_config.max_retries' '5'
 assert_field_value 'providers.openai.network_config.retry_backoff_initial' '.providers.openai.network_config.retry_backoff_initial' '200'
 assert_field_value 'providers.openai.network_config.retry_backoff_max' '.providers.openai.network_config.retry_backoff_max' '10000'
+assert_field_value 'providers.openai.network_config.allow_private_network' '.providers.openai.network_config.allow_private_network' 'true'
 
 # Concurrency config
 assert_field_value 'providers.openai.concurrency_and_buffer_size.concurrency' '.providers.openai.concurrency_and_buffer_size.concurrency' '50'
--- a/.github/workflows/scripts/validate-helm-config-fields.sh
+++ b/.github/workflows/scripts/validate-helm-config-fields.sh
@@ -288,6 +288,7 @@ bifrost:
         retry_backoff_initial_ms: 200
         retry_backoff_max_ms: 10000
         allow_private_network: true
+        beta_header_overrides: {}
       concurrency_and_buffer_size:
         concurrency: 50
         buffer_size: 100
@@ -347,6 +348,7 @@ assert_field_value 'providers.openai.network_config.max_retries' '.providers.ope
 assert_field_value 'providers.openai.network_config.retry_backoff_initial' '.providers.openai.network_config.retry_backoff_initial' '200'
 assert_field_value 'providers.openai.network_config.retry_backoff_max' '.providers.openai.network_config.retry_backoff_max' '10000'
 assert_field_value 'providers.openai.network_config.allow_private_network' '.providers.openai.network_config.allow_private_network' 'true'
+assert_field 'providers.openai.network_config.beta_header_overrides (explicit empty map preserved)' '.providers.openai.network_config.beta_header_overrides'
 
 # Concurrency config
 assert_field_value 'providers.openai.concurrency_and_buffer_size.concurrency' '.providers.openai.concurrency_and_buffer_size.concurrency' '50'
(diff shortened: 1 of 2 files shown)


truffle-dev:
Folded in the adjacent `beta_header_overrides` block (70d7c53). It used a truthiness guard while every other `network_config` field uses `hasKey`, so an explicitly-set empty map (`beta_header_overrides: {}`) was silently dropped — the same drop-on-empty-map class this PR fixes for `allow_private_network`. Verified by stash-bisect: empty map renders as `{}` with `hasKey`, vanishes under the old guard. Added a render-time regression assertion in the validation script for the empty-map case.

BearTS:
Hey @truffle-dev Thank you for the PR, I actually had added the support a while ago, this pr was completely missed by us.
Happy to collaborate on future PRs, please tag me in case for helm specific PR

Issue #3492, opened 2026-05-14
[Feature]: <dashscope is incompatible>


zhuangdl:
### Prerequisites

- [x] I have searched existing issues and discussions to avoid duplicates
- [x] I am using the latest version (or have tested against main/nightly)

### Description

1. When adding dashscope using Add Custom Provider with the correct base_url: https://dashscope.aliyuncs.com/compatible-mode/v1, the Chat Completion and Chat Completion Stream must have a custom path of /chat/completions; otherwise, a 404 error will be returned.

3. When using bifrost to call the dashscope model, parameters such as enable_thinking and max_tokens have no effect.

### Steps to reproduce

1. No Custom Path，return 404
<img width="1848" height="1828" alt="Image" src="https://github.com/user-attachments/assets/b9d6101b-77bb-45ce-8dc5-93da8f49fd83" />
<img width="2426" height="1612" alt="Image" src="https://github.com/user-attachments/assets/481da19d-f5e9-4c9a-a6f2-81c8a9b715c8" />
<img width="1912" height="316" alt="Image" src="https://github.com/user-attachments/assets/60d16f3a-71fe-430e-a05b-e6a89c40e745" />

2. Has custom path，return 200，But it consumed a lot of tokens,  the parameters did not take effect.
<img width="2480" height="2000" alt="Image" src="https://github.com/user-attachments/assets/4ea04d5c-7e16-4146-8c43-cb297feb10b0" />
<img width="2502" height="2288" alt="Image" src="https://github.com/user-attachments/assets/0f71a967-39c0-43f1-9792-af1ccfb7bd60" />
<img width="1680" height="946" alt="Image" src="https://github.com/user-attachments/assets/2f26aeeb-e8d6-466c-a5ca-2b3a43efe0fe" />

### Expected behavior

Can properly adapt to / work with dashscope

### Actual behavior

It seems not to be compatible / It doesn't seem to work properly

### Affected area(s)

Transports (HTTP)

### Version

v1.5.2

### Environment

```text

```

### Relevant logs/output

```shell

```

### Severity

Low (minor issue or cosmetic)

akshaydeo:
This is a feature request - we do not officially support dashscope - changing it to feature request

zhuangdl:
@akshaydeo Thank you very much. Will this take a long time?

akshaydeo:
yeah - this is not been asked a lot so would not be a priority

rlex:
[...]

Issue #3181, opened 2026-05-01
[Feature]: Support Extra Body for individial Provider


DGideas:
### Prerequisites

- [x] I have searched existing issues and discussions to avoid duplicates

### Problem to solve

Users sometimes need to specify custom request bodies to be processed by the provider. While this is currently possible using `extra_params` with the `x-bf-passthrough-extra-params=true` header, certain AI applications (such as GitHub Copilot CLI BYOK) do not support custom header/body inputs. Therefore, Bitfrost needs a built-in configuration option in the Provider settings to handle these extra body parameters automatically.

### Proposed solution

To maintain consistency with the existing Extra Headers feature, I suggest introducing a similar Extra Body configuration on the Provider config page. This would allow users to define additional body parameters in a unified manner, mirroring the current header configuration workflow.

### Area(s)

Transports (HTTP)

alshdavid:
I too would like this feature. My use case is creating preconfigured model aliases that inject configuration because clients (like vscode) cannot inject extra params.

e.g. `~/bifrost/config.json`

```json
{
  "$schema": "https://www.getbifrost.ai/schema",
  "aliases": {
    "deepseek-v4-flash-nothink": {
      "type": "static",
      "target": {
        "provider": "deepseek",
        "model": "deepseek-v4-flash"
      },
      "override_body": {
        "thinking": {
          "type": "disabled"
        }
      }
    }
  }
}
```

sammaji:
@DGideas we have a PR for this (WIP) - #4364

Issue #969, opened 2025-11-30
[Docs]: llama-server/llama-swap support


rosmur:
### Prerequisites

- [x] I have searched existing issues and docs to avoid duplicates

### Affected page(s)

bifrost/core/providers/

### What’s wrong or missing?

No information on how to add llama-swap or llama-server (llama.cpp) as a provider. It should be compatible as it serves an OpenAI API compatible endpoint

rlex:
you need to add your llama-swap address without /v1 endpoint. I just verified it and it works. If /v1 is added during configuration it appends it again, resulting in requests like /v1/v1/models.
Hasn't verified with llamacpp directly.

deardaveed:
I would love to know, also. Why is Ollama a default provider but not llama.cpp?

rosmur:
@akshaydeo bumping this for visibility

akshaydeo:
@sammcj can you check thisn

sammcj:
@akshaydeo I've spent some time testing this out (v1.4.24), and if I'm understanding this issue correctly, the request is to support llama.cpp's _native_ API (which would bring it in line with supporting Ollama's native API)?

Right now, using my test instance of Bifrost:
- llama.cpp native API: Not supported
- llama.cpp OpenAI/Anthropic compatible APIs: Bifrost does not work (does not detect any models)
- llama-swap OpenAI/Anthropic compatible APIs: Bifrost works and detects all models (but doesn't detect their context lengths (`context_length` field))

I'm assuming that Bifrost may not be correctly passing the model listing returned by llama.cpp's models endpoint (see example below)

My take is that if offering a native Ollama provider I would recommend having the same for llama.cpp's native API, especially since how popular it is, and the native API offers additional capabilities that the compatibility APIs do not that are specific to llama.cpp (a bit like Ollama's native API offers additional capabilities that are Ollama specific)

As an aside when it comes to self-hostable inference servers there's been quite a push away from Ollama in the community (and in the workplace from what I've seen) towards people running llama.cpp's server directly or through a wrapper like LM Studio, Msty or on Apple Silicon the likes of oMLX. Personally I really don't think Ollama (a wrapper around llama.cpp and MLX) should be getting preferential treatment over llama.cpp which at this point is probably the most widely deployed inference server (outside of the AI labs themselves).

Personally, I would recommend either:
A) Implement support for llama.cpp's native API
or
B) Fix model parsing with llama.cpp's OpenAI compatible API

[...]

PR #4364, opened 2026-06-13
fix: pass extra parameters for custom providers

### LINKED ISSUE #4272: [Bug]: Custom provider (openai-compatible) silently drops extra parameters
### Prerequisites

### Description

It looks like when you add openai-compatible providers via bifrost (for example opencode zen) extra params added (in my case thinking) are silently dropped, and i can't control reasoning on/off.
I verified via curl calling same providers with thinking parameter - direct call passes, thinking can be controlled, but while calling via bifrost it's ignored, and i'm always getting thinking block (since it's enabled by default for example on deepseek-v4), essentially breaking lots of 3rd party openai-compatible providers.
It seems like chat_template_kwargs are also dropped.
Looks like something like that was fixed in vllm: https://github.com/maximhq/bifrost/issues/2622

### Steps to reproduce

Add opencode-zen (for example) as provider
Call with thinking off/on - same result, deepseek-v4-flash (for example) always have thinking activated

### Expected behavior

Extra args should pass

### Actual behavior

Extra args aren't passed

### Affected area(s)

Transports (HTTP), Core (Go)

### Version

1.5.11

### Environment

```text

```

### Relevant logs/output

```shell

```

### Regression?

_No response_

### Severity

rlex:
## Summary

Fixes #4272 #3181

## Changes

This handles issue of dropping extra parameters for custom providers.
It introduces IsCustomProvider helper (might be useful in future too if another provider will decide to reinvent the wheel and introduce yet another api) and reworks merging path a bit - checking if provider is custom and enabling extra parameters passing in case it's true. Actual code change is minimal, majority of it are just tests - verifying both passing parameters in case of custom and retaining old behavior for "native" ones.

It's mostly done for openai and anthropic providers, but since it affects every custom provider, i also covered bedrock, replicate, etc, although i'm not aware of any 3rd party provider actually exposing them.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [x] Core (Go)
- [x] Transports (HTTP)
- [x] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

Tests pass, but actual change can only be tested by adding real 3rd party provider (like dashscope) and checking against bifrost if, for example, thinking can be disabled/enabled

## Breaking changes

- [ ] Yes
- [x] No

## Related issues

https://github.com/maximhq/bifrost/issues/4272 - my issue
https://github.com/maximhq/bifrost/issues/969 - since stuff like chat_template_kwargs can be passed now, it fully supports llamacpp/llamaswap
https://github.com/maximhq/bifrost/issues/3492 - fully supported now (initial issue was due to user's misconfiguration, and additional parameters should pass now.

Maybe more where people complained about they can't use options like thinking, max_tokens, etc in custom openai-compatible endpoints.

## Security considerations

Strictly theoretically if someone relied on sanitization of outgoing requests using current behavior, it will change.

## Checklist

- [ ] I read `docs/contributing/README.md` and followed the guidelines - such file do not exist, by the way.

### DIFF
--- a/core/providers/utils/utils.go
+++ b/core/providers/utils/utils.go
@@ -1272,8 +1272,7 @@ func CheckContextAndGetRequestBody(ctx context.Context, request RequestBodyGette
 // should be preserved in the outbound request.
 func ShouldPassthroughExtraParams(ctx context.Context) bool {
 	passthroughExtraParams, _ := ctx.Value(schemas.BifrostContextKeyPassthroughExtraParams).(bool)
-	isCustomProvider, _ := ctx.Value(schemas.BifrostContextKeyIsCustomProvider).(bool)
-	return passthroughExtraParams || isCustomProvider
+	return passthroughExtraParams
 }
 
 // SetExtraHeadersHTTP sets additional headers from NetworkConfig to the standard HTTP request.
--- a/core/providers/utils/utils.go
+++ b/core/providers/utils/utils.go
@@ -1272,7 +1272,8 @@ func CheckContextAndGetRequestBody(ctx context.Context, request RequestBodyGette
 // should be preserved in the outbound request.
 func ShouldPassthroughExtraParams(ctx context.Context) bool {
 	passthroughExtraParams, _ := ctx.Value(schemas.BifrostContextKeyPassthroughExtraParams).(bool)
-	return passthroughExtraParams
+	isCustomProvider, _ := ctx.Value(schemas.BifrostContextKeyIsCustomProvider).(bool)
+	return passthroughExtraParams || isCustomProvider
 }
 
 // SetExtraHeadersHTTP sets additional headers from NetworkConfig to the standard HTTP request.
--- a/core/providers/utils/utils.go
+++ b/core/providers/utils/utils.go
@@ -1251,7 +1251,7 @@ func CheckContextAndGetRequestBody(ctx context.Context, request RequestBodyGette
 			return nil, NewBifrostOperationError(schemas.ErrProviderRequestMarshal, err)
 		}
 		// Merge ExtraParams into the JSON if passthrough is enabled
-		if ctx.Value(schemas.BifrostContextKeyPassthroughExtraParams) != nil && ctx.Value(schemas.BifrostContextKeyPassthroughExtraParams) == true {
+		if ShouldPassthroughExtraParams(ctx) {
 			extraParams := convertedBody.GetExtraParams()
 			if len(extraParams) > 0 {
 				// Use order-preserving merge to avoid destroying key ordering in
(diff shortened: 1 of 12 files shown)


rlex:
Implemented greptile's suggestion.

dat911zz:
LGTM ✨

core/utils.go
```diff
@@ -351,6 +351,11 @@ func IsStandardProvider(providerKey schemas.ModelProvider) bool {
 	return ok
 }
 
+// IsCustomProvider reports whether config defines a custom provider.
+func IsCustomProvider(config *schemas.ProviderConfig) bool {
+	return config != nil && config.CustomProviderConfig != nil
+}
+
```
sammaji, line 358: @rlex why do we need this, isn't `!IsStandardProvider` good enough

core/providers/utils/utils.go
```diff
@@ -1268,6 +1268,14 @@ func CheckContextAndGetRequestBody(ctx context.Context, request RequestBodyGette
 	}
 }
 
+// ShouldPassthroughExtraParams reports whether provider-specific parameters
+// should be preserved in the outbound request.
+func ShouldPassthroughExtraParams(ctx context.Context) bool {
+	passthroughExtraParams, _ := ctx.Value(schemas.BifrostContextKeyPassthroughExtraParams).(bool)
+	isCustomProvider, _ := ctx.Value(schemas.BifrostContextKeyIsCustomProvider).(bool)
+	return passthroughExtraParams || isCustomProvider
```
sammaji, line 1276: @rlex we should set `BifrostContextKeyPassthroughExtraParams` for custom providers instead of doing this.

```
if customProvider:
  BifrostContextKeyPassthroughExtraParams = true
```

core/utils.go
rlex: Based from what i've found during testing and debugging (heavily llm-assisted) ```!IsStandardProvider``` checks the base provider type, but custom providers often use a standard provider as their base, like openai.
so for custom openai-compatible provider, ```IsStandardProvider(openai)``` is still ```true```, which means ```!IsStandardProvider(openai)``` would be false. So in the end it treats it as non-custom and drops extra params.

core/providers/utils/utils.go
rlex: @sammaji better?

sammaji (commented):
Hey @rlex, you PR looks good mostly.  
  
Ideally we want to add a UI toggle for this in the network config for each providers (including custom providers).  
  


Few things to keep in mind:  
\- Some provider like vLLM and deepseek should have this setting turned on by default.  
\- Should have a equivalent in `config.json`  
\- Should be overriden by `x-bf-passthrough-extraparams`

rlex:
Good call, can try. So we're dropping check for IsCustomProvider completely and instead set it as parameter, disabled by default?

Issue #998, opened 2025-12-04
[Feature]: OpenInference for Phoenix Arise


patrickjaja:
### Prerequisites

- [x] I have searched existing issues and discussions to avoid duplicates

### Problem to solve

 Phoenix expects OpenInference format:
  - llm.token_count.prompt
  - llm.token_count.completion
  - llm.token_count.total

 Bifrost sends GenAI extension format:
  - gen_ai.usage.prompt_tokens
  - gen_ai.usage.completion_tokens
  - gen_ai.usage.total_tokens

### Proposed solution

In bifrost OpenTelemetry UI (/workspace/observability?plugin=otel), allow changing the format to open_inference or to manually map the data.

<img width="1626" height="735" alt="Image" src="https://github.com/user-attachments/assets/628de2a8-bc97-4c24-bc16-ce609f5149bc" />

### Area(s)

Core (Go)

### Additional context

Screenshot showing that the Phoenix Span View has no data mapped correctly on Bifrost open telemetry data: 
<img width="1877" height="538" alt="Image" src="https://github.com/user-attachments/assets/0d635ed9-722f-40a7-a0dd-c0f1f73647e1" />

Screenshot showing that the data is sent to phoenix (triggered from Bifrost), but its pretty much unused:
<img width="1404" height="615" alt="Image" src="https://github.com/user-attachments/assets/1b5bbfc9-7243-431c-99bc-51ac97420a61" />

patrickjaja:
i have noticed, supporting OpenInference is more complex then that three fields i have mentioned in my usecase.
Here the docs: https://arize-ai.github.io/openinference/spec/semantic_conventions.html#reserved-attributes

And in your docs its mentioned as "comming soon" https://docs.getbifrost.ai/features/observability/otel - so maybe anyway this ticket is a ongoing placeholder.

Thanks!

PR #4374, opened 2026-06-13
Initial implementation of Arize OpenInference

rlex:
## Summary

Implements Arize OpenInference in OTEL plugin

## Changes

What i thought will be simple task turned out to be quite a nightmare... 
Anyway, this adds initial support for Arize OpenInference, allowing users to easily plug bifrost into Arize Phoenix instance or anything that supports that otel format, producing clean and useful telemetry data for debugging their prompts and sessions:

image

Since existing GenAI format already handles per-call traces with spans per each call, i filtered it out from OpenInference implementations - if users really need it, they can just enable 2nd/3rd/etc plugin with GenAI only. That leaves OpenInference implementation much lighter and cleaner - and less resource-heavy for receiving tracing stack.

This is initial implementation - i tried to keep scope as narrow as possible without messing with core, transports and etc too much.
Due to current limitations of plugin system, only certain span kinds are implemented - LLM, EMBEDDING, CHAIN, TOOL. Adding more is impossible without reworking certain parts of bifrost itself, and i'm not competent enough to properly evaluate potential damage, especially to performance.

I'll provide more screenshots and samples in coming days, and if you're OK with such partial implementation, i will continue on it - i have few more things to implement.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [ ] Core (Go)
- [x] Transports (HTTP)
- [ ] Providers/Integrations
- [x] Plugins
- [x] UI (React)
- [ ] Docs

## How to test

Needed to install https://github.com/arize-ai/phoenix - can be done with simple docker compose, create initial user, create API key in settings, and fill plugin settings:

endpoint - http://phoenix_ip/api/v1/traces
Authorization header: ```Authorization: Bearer API_KEY_FROM_PHOENIX_SETTINGS```
Transport: http

Send message and it should arrive to phoenix.

## Screenshots/Recordings

Toggle for Arize OpenInference becomes active.

## Breaking changes

- [ ] Yes
- [x] No

## Related issues

https://github.com/maximhq/bifrost/issues/998

## Security considerations

## Checklist


rlex:
I will address issues, i only tested current implementation with simple curl calls - not proper agentic workflow, and hasn't properly tested more complex cases.

rlex:
Another question for bifrost team: currently, i send model as model name only (dropping provider part), so for example if you have two upstream providers having same model (ie you route glm-5.1 via two providers at same time), in phoenix it will only appear as "GLM-5.1" regardless of provider. Theoretically i can grab provider prefix and present models as provider_name/model_name as they're represented in bifrost, which might be useful for debugging (detecting lower quantizations, evaluating different providers, etc) - would love to see some input here. Current UI doesn't allow extended settings, but it probably can be done via ENV variables if it needs to be configurable.

akshaydeo:
hey @rlex just to be sure - Arize does not use standard OpenAPI GenAI extension. So we would need to add that option to make sure users can select appropriate GenAI convertor.

also you can rename current converter as `genaiextension.go`

rlex:
> hey @rlex just to be sure - Arize does not use standard OpenAPI GenAI extension. So we would need to add that option to make sure users can select appropriate GenAI convertor.
> 
> also you can rename current converter as `genaiextension.go`

@akshaydeo not sure i follow. The old collector is still there, you can select it in dropdown menu or via config variables, i hasn't changed it much, only used it as base.

Good call on renaming, i'll do that, in case anyone will decide to implement vertex one.

rlex:
I think i'll stop here, this is already becoming too big, but some changes in MCP pipeline were required to implement this feature correctly. They can also be reused for GenAI format - which i intend to fix later too, since ATM it's broken. 
I will still need to address couple of small issues and run long agent session to test it against real-world usage, not just synthetic tests.

rlex:
Simple message:

image

Error, with extraction of contents from bifrost:

image

image

Full flow of request: 

image

Tool calling data: 

image

Tools that were available during that requests:

image

Capturing of invocation params: 

image

If you add x-bf-session-id, you can have proper flow for entire conversation (hasn't figured out how to sum them up automatically, headers seems to be the only option here)

image

rlex:
@akshaydeo ready for review. There is lingering issue i can't quite pinpoint - clients refusing to connect to MCP server sometimes - but it might be unrelated to changes - since i branched from dev branch and there might be some bug. Or it might be bug in my favourite harness (oh-my-pi) - it's the only one with connection issues.

rlex:
Couple of late targeted fixes:

Convert some provider names into OpenInference preferred ones (since spec strictly says MUST)

Add custom attributes field support - in case of phoenix, you will need openinference.project.name attached to trace to put them into project other than "default". This is strictly optional.

PR #4392, opened 2026-06-15
update write kickoff based on last activity

akshaydeo:
## Summary

Fixes a production bug where streaming LLM requests running longer than the pending log TTL had their in-memory `PendingLogData` entry evicted mid-flight by `cleanupStalePendingLogs`. This caused the final log row to be silently dropped and a per-chunk "no pending log data found" warning to be emitted for every subsequent chunk.

## Changes

- Added an atomic `LastActivity` field to `PendingLogData` that tracks the unix-nano timestamp of the most recent `PostLLMHook` call (i.e. each streaming chunk).
- `cleanupStalePendingLogs` now evicts based on idle time (time since `LastActivity`) rather than total lifetime (`CreatedAt`), so actively-streaming requests are never reaped mid-flight.
- `PreLLMHook` seeds `LastActivity` with `CreatedAt` at entry creation so the idle clock has a valid baseline even before the first chunk arrives.
- `PostLLMHook` refreshes `LastActivity` on every call, keeping long-running streams alive through the cleanup window.
- `pendingLogTTL` changed from 5 minutes to 15 minutes (idle timeout, not lifetime cap), and its comment updated to reflect the new semantics.
- `batchInterval` increased from 2s to 5s and `maxBatchBytes` from 100 MB to 300 MB.
- Added two regression tests: `TestActiveStreamSurvivesCleanup` (entry with old `CreatedAt` but recent `LastActivity` must not be evicted) and `TestIdlePendingEntryEvicted` (entry with both `CreatedAt` and `LastActivity` past the TTL must be evicted).

## Type of change

- [x] Bug fix

## Affected areas

- [x] Plugins

## How to test

```sh
go test ./plugins/logging/... -run "TestActiveStreamSurvivesCleanup|TestIdlePendingEntryEvicted|TestCleanupStalePendingMCPLogsPersistsErrorFallback"
```

Expected: all three tests pass. To validate the fix end-to-end, send a streaming chat request that takes longer than the previous 5-minute TTL and confirm the final log row is persisted and no "no pending log data found" warnings appear.

## Breaking changes

- [x] No

## Related issues

Regression fix for prod issue where long-running streaming requests lost their final log entry due to premature in-memory eviction.

## Security considerations

None. The `LastActivity` field is an in-process atomic integer with no exposure to external input.

## Checklist

- [ ] I read `docs/contributing/README.md` and followed the guidelines
- [x] I added/updated tests where appropriate
- [ ] I updated documentation where needed
[...]

PR #4329, opened 2026-06-12
adds migration catalogging

akshaydeo:
## Summary

This PR introduces a lock-free fast path for database migrations in both `configstore` and `logstore`. On PostgreSQL, before acquiring the advisory migration lock, the system now checks whether all migrations are already applied. If they are, the lock is skipped entirely, reducing startup latency in steady-state deployments where no migrations are pending.

## Changes

- Introduced a `migrationStep` struct in both `configstore` and `logstore` that pairs migration IDs with their run functions, replacing the previous flat sequence of individual `if err := migration...` calls.
- Extracted `configstoreMigrationSteps` and `logstoreMigrationSteps` as ordered slices that serve as the single source of truth for both execution and preflight checks.
- Added `runMigrationSteps`, `migrationStepIDs`, and `pendingMigrationStepIDs` helpers in both packages to iterate steps and query pending IDs without holding a lock.
- Added `PendingIDs` to the `migrator` package — a read-only helper that queries the migration table for IDs not yet recorded, returning all expected IDs if the table doesn't exist or is missing required columns.
- In `triggerMigrations` for both stores, a preflight check now runs before and after acquiring the advisory lock on PostgreSQL. If all migrations are current before the lock, the lock is skipped. If another node completes migrations while waiting for the lock, the run is skipped after acquiring it.
- Decoupled the `configstore` and `logstore` advisory lock keys so the two stores can migrate independently rather than being fully serialized against each other.
- Fixed `migrationAddCanonicalModelColumns` in `logstore` to accept a `schemas.Logger` parameter, aligning its signature with all other migration functions.
- Added unit tests for `PendingIDs` covering missing table, partial application, full application, legacy table schema, and custom options.

## Type of change

- [ ] Bug fix
- [ ] Feature
- [x] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [x] Core (Go)
- [ ] Transports (HTTP)
- [ ] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

```sh
go test ./framework/migrator/...
go test ./framework/configstore/...
go test ./framework/logstore/...
```

Start multiple instances simultaneously against a PostgreSQL database that is already fully migrated and confirm that none of them acquire the advisory migration lock (observable via log lines: `migrations already current; skipping migration lock`).

Start instances against a fresh database and confirm migrations run correctly end-to-end.

## Breaking changes

- [ ] Yes
- [x] No

## Security considerations

[...]

PR #4403, opened 2026-06-15
adds upsert based writes to avoid losing logs

akshaydeo:
## Summary

When a long-idle LLM request's in-memory pending entry is evicted by the stale-pending cleanup sweep, any response that arrives afterward was silently dropped — the log was lost entirely. This PR introduces a durability fallback: evicted entries are persisted as partial "processing" rows before being dropped from memory, and a late-arriving response reconciles its output onto that row via a batched `UPDATE` rather than losing the log.

## Changes

- **`cleanupStalePendingLogs` now persists before evicting**: before removing a stale LLM pending entry, `buildInitialLogEntry` is called to write a partial "processing" row (input preserved, `CreatedAt` reset to now, original `Timestamp` kept). An `evictedPartials` sync.Map records the eviction in-memory so a later finish can reconcile without a DB read.

- **DB-fallback reconciliation path in `PostLLMHook`**: when no in-memory pending is found on a terminal call, the hook checks `evictedPartials` first (race-free, set synchronously at eviction), then falls back to `store.IsLogEntryPresent` for rows persisted by a previous process instance. If a partial row is found, output fields are applied via `enqueueLogUpdate` rather than a full insert.

- **`BatchUpdate` added to `LogStore` interface and implementations**: `RDBLogStore.BatchUpdate` applies partial updates to multiple rows in a single transaction using GORM `Updates` semantics (non-zero fields only), silently skipping absent IDs. `HybridLogStore.BatchUpdate` delegates to the inner store.

- **`processBatch` applies inserts before updates**: the batch writer separates `log` (insert) and `logUpdate` (partial update) entries and processes inserts first, guaranteeing that a partial-row insert and its reconciling update landing in the same batch resolve correctly.

- **`applyPostHookOutputToEntry` extracted**: output, governance, cost, and status field population is factored into a shared helper used by both the normal write path and the DB-fallback reconciliation path, eliminating duplicated logic.

- **`evictedPartials` markers are reaped**: the cleanup sweep also expires stale evicted-partial markers whose request never finished, matching the processing-row Flush lifecycle.

- **Tests added** covering: persist-before-evict behavior, success and error reconciliation onto a partial row, no DB probe on non-final streaming chunks, bulk evict→reconcile integrity under a slow store with 800 concurrent requests, and `BatchUpdate` skip-missing-rows semantics.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [ ] Core (Go)
- [ ] Transports (HTTP)
- [ ] Providers/Integrations
- [x] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

```sh
go test ./framework/logstore/... ./plugins/logging/...
```

The new test suite covers the full evict→persist→reconcile cycle, including a stress test (`TestBulkEvictReconcileIntegritySQLite`) that runs 800 requests through a deliberately slow store to verify ordering guarantees under load. All 800 rows must end with `status=success`, correct token counts, and preserved input history, with zero dropped requests.

## Breaking changes

- [x] Yes
- [ ] No

`LogStore` interface gains a new `BatchUpdate(ctx context.Context, entries []*Log) error` method. Any custom `LogStore` implementation outside this repository must add this method to satisfy the interface.

## Security considerations

No auth, secrets, or PII handling changes. The fallback path uses `p.ctx` (the plugin-level context) rather than the request context when probing the DB, avoiding false negatives from already-cancelled request contexts on timeout.

## Checklist


akshaydeo:
* **#4403** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/4403?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/maximhq/bifrost/4403?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#4392** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/4392?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4327** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/4327?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#4329](https://github.com/maximhq/bifrost/pull/4329) <a href="https://app.graphite.com/github/pr/maximhq/bifrost/4329?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `dev`


<h2></h2>

PR #4410, opened 2026-06-15
fix(bedrock): preserve tool_result order to match parallel tool_use blocks

alexef:
## Summary

- `ConvertBifrostMessagesToBedrockMessages` stored pending tool results in a `map[string]*ToolResult` and iterated it directly, producing **non-deterministic ordering** of `tool_result` blocks in the Bedrock request
- When an assistant turn has N parallel `tool_use` blocks, Bedrock requires the following user turn to list `tool_result` blocks **in the same order** — otherwise it returns HTTP 400: *"toolResult blocks exceed toolUse blocks of previous turn"*
- The bug affected any conversation with 2+ parallel tool calls, causing sporadic failures (~10–15% of calls in testing)

## Root cause

`ToolCallStateManager.GetPendingResults()` returns `map[string]*ToolResult`. All three flush sites in `ConvertBifrostMessagesToBedrockMessages` iterated this map directly, with Go's randomised map iteration producing a different `tool_result` order each time.

## Fix

- Added `pendingResultIDs []string` to `ToolCallStateManager` to track registration order (parallel to the existing `pendingToolCallIDs` slice for tool calls)
- Added `GetPendingResultsOrdered() []string` which returns IDs in insertion order, filtering out already-emitted entries
- Updated all three flush sites to use the ordered IDs instead of raw map iteration
- `MarkResultsEmitted` now also purges emitted IDs from `pendingResultIDs`

## Test

Added `TestParallelToolResultOrdering` — a 200-iteration regression test that runs the full `ToBedrockResponsesRequest` conversion on a 13-message conversation mirroring the production payload that triggered this bug. Each iteration asserts that `tool_result` IDs in the Bedrock user message match `tool_use` IDs in the preceding assistant message, in the same order.

Before the fix: ~50 failures per 200 iterations.  
After the fix: 0 failures.

## Related

Discovered while investigating a production 400 error on `jira-cma-worker` at GetYourGuide (DEP-6385, PAY-2328/2333/2334/2335). The failing payload had `messages[11] = assistant: [tool_use A, tool_use B]` / `messages[12] = user: [tool_result A, tool_result B]` and was rejected intermittently.

### DIFF
--- a/core/providers/bedrock/parallel_tool_result_ordering_test.go
+++ b/core/providers/bedrock/parallel_tool_result_ordering_test.go
@@ -1,0 +1,162 @@
+package bedrock
+
+// Regression test for the non-deterministic tool_result ordering bug.
+//
+// When an assistant turn contains N parallel tool_use blocks, the corresponding
+// user turn must list the tool_result blocks in the same order.  Bedrock rejects
+// requests where the counts or ordering don't match ("toolResult blocks exceed
+// toolUse blocks of previous turn").
+//
+// Root cause: ConvertBifrostMessagesToBedrockMessages stores pending results in
+// a map[string]*ToolResult and then iterates that map directly, producing
+// non-deterministic output.  Running the conversion many times should reveal
+// the ordering instability.
+//
+// This test mirrors the production payload that triggered the bug:
+//   messages[11] = assistant: [tool_use A, tool_use B]
+//   messages[12] = user:      [tool_result A, tool_result B]
+
+import (
+	"context"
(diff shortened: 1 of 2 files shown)


akshaydeo:
@alexef please keep PRs pointed to dev 🙇

alexef:
Closing — will reopen against dev branch

akshaydeo:
no no I edited it

TejasGhatte (approved):

akshaydeo (approved):

PR #4418, opened 2026-06-15
fix: populate error message in OpenAI responses streaming error events

### LINKED ISSUE #4413: [Bug]: OpenAI Responses streaming returns empty error.message on context_length_exceeded
### Prerequisites

### Description

On the OpenAI Responses API streaming path, a context-length overflow comes back as a terminal `error` event with an empty `error.message`, so the client has no way to tell why the stream stopped. The exact same request returns the full `context_length_exceeded` error on the non-streaming Responses path and on Chat Completions streaming, so the detail is clearly available; it just gets dropped on this one path.

This looks like the same area as maximhq/bifrost#2652 / [#2681](<https://github.com/maximhq/bifrost/issues/2681>). That fix is already in my build (I'm on core v1.5.18, and the relevant code is identical on the current core v1.5.19), so it isn't a stale-version thing, it just doesn't seem to cover the overflow case.

In case it helps: the Responses error branch in `core/providers/openai/openai.go` only fills the message from `response.Message` and `response.Response.Error.Message`, and both are empty for this event. If the real message is somewhere in the raw event, `send_back_raw_response` might surface it, but the `x-bf-send-back-raw-response` header had no effect for me and I couldn't reach the admin config to set it at the provider level.

### Steps to reproduce

Send a streaming Responses request whose input is larger than the model's context window. `gpt-5.4-mini` has a 272k-token window, and the filler below is \~700k tokens:

```bash
python3 - <<'PY' > body.json
import json
print(json.dumps({
    "model": "gpt-5.4-mini",
    "stream": True,
    "max_output_tokens": 16,
    "input": "the quick brown fox jumps over the lazy dog. " * 70000,
}))
PY

curl -N https://<your-bifrost-host>/openai/v1/responses \
  --data @body.json
```

For comparison, flip `"stream"` to `false` in `body.json` and send it again: that returns an HTTP 400 with the real error.

### Expected behavior

The terminal error event carries the provider message/code, the same `context_length_exceeded` you get when streaming is off.

### Actual behavior

The error event has an empty message, so the only signal a client gets is that the stream ended without output.

### Affected area(s)

Core (Go)

### Version

v1.5.10 (transports), core v1.5.18

### Environment

```text
```

### Relevant logs/output

```shell
Streaming overflow — the whole stream is three events, and the terminal one has an empty message:


event: response.created       (response.error: null)
event: response.in_progress   (response.error: null)
data: {"type":"error","is_bifrost_error":false,"error":{"message":""},"extra_fields":{"provider":"openai","resolved_model_used":"gpt-5.4-mini","request_type":"responses_stream"}}


Same request with `"stream": false` → HTTP 400, with the full error:


{"status_code":400,"error":{"type":"invalid_request_error","code":"context_length_exceeded","message":"Your input exceeds the context window of this model. Please adjust your input and try again.","param":"input"}}


Same content sent to `/openai/v1/chat/completions` (streaming) → HTTP 400, also with the full error:


{"status_code":400,"error":{"type":"invalid_request_error","code":"context_length_exceeded","message":"Input tokens exceed the configured limit of 272000 tokens. Your messages resulted in 700007 tokens. Please reduce the length of the messages.","param":"messages"}}
```

### Regression?

*No response*

### Severity

Purvi09:
# Summary

On the OpenAI Responses API streaming path, error events like `context_length_exceeded` were returned with an empty `error.message`. The `BifrostResponsesStreamResponse` struct had no top-level `Error` field, so the nested `"error"` object sent by OpenAI was silently dropped during deserialization.

## Changes

- Added `Error *ResponsesResponseError` field to `BifrostResponsesStreamResponse` in `core/schemas/responses.go` so the nested error object is correctly deserialized
- Updated the streaming error handler in `core/providers/openai/openai.go` to read `message` and `code` from `response.Error` before falling back to `response.Response.Error`

## Type of change

- [x] Bug fix

## Affected areas

- [x] Core (Go)
- [x] Providers/Integrations

## How to test

1. Start Bifrost with an OpenAI key
2. Send a streaming Responses request whose input exceeds the model context window

**Before fix:** error.message is empty

**After fix:** error.message is "Your input exceeds the context window of this model. Please adjust your input and try again." with code "context_length_exceeded"

## Breaking changes

- [x] No

## Related issues

Closes #4413

## Security considerations

None — only affects error message deserialization in the streaming responses path.

## Checklist

- [x] I verified the fix locally with a real OpenAI streaming request


roroghost17 (approved):

Purvi09:
@roroghost17 Thanks for reviewing, can you please merge it as well, I don't have merge permissions.

akshaydeo (approved):

PR #4485, opened 2026-06-17
fix(vectorstore): KNN valkey-search without SORTBY score error

dada-engineer:
The KNN query aliased vector distance as score and issued SORTBY score. valkey-search rejects this — the alias isn't a real index field (Index field score does not exist) — so every semantic-cache lookup failed and fell back to a miss.

SORTBY is unnecessary: [KNN k …] already returns the k-nearest set, and knnLimit == searchLimit so LIMIT never drops the nearest. This removes the SORTBY clause and sorts results by similarity client-side, giving deterministic nearest-first order on both valkey-search and RediSearch (whose default output orders differ).

Changes

- redis.go: remove SORTBY score from the query; remove isSortByUnsupportedErr + the error-string-matched retry; sort filteredResults by similarity (desc, score-less last).
- redis_test.go: TestRedisStore_GetNearest_DoesNotSendSortBy — asserts one KNN query, no SORTBY, no scan fallthrough.

### DIFF
--- a/framework/vectorstore/redis.go
+++ b/framework/vectorstore/redis.go
@@ -1165,11 +1165,16 @@ func (s *RedisStore) GetNearest(ctx context.Context, namespace string, vector []
 		knnLimit = math.MaxInt32
 	}
 
+	// KNN results are returned in vector-distance order by default on both
+	// RediSearch and valkey-search ("vector results are sorted in distance order
+	// ... before the LIMIT clause is applied"), so no SORTBY is needed. Adding
+	// `SORTBY score` is redundant and valkey-search rejects it because the KNN
+	// `AS score` alias is not a real index field. Results are re-sorted by score
+	// below to stay deterministic across backends.
 	args := []interface{}{
 		"FT.SEARCH", namespace,
 		fmt.Sprintf("%s=>[KNN %d @embedding $vec AS score]", hybridQuery, knnLimit),
 		"PARAMS", "2", "vec", queryBytes,
-		"SORTBY", "score",
 	}
 
 	// Add RETURN clause - always include score for vector search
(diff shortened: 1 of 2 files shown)


akshaydeo:
this seems a really brute force fix - let me dig a bit and come back to this

framework/vectorstore/redis.go
```diff
@@ -1137,6 +1137,28 @@ func buildRedisQueryCondition(query Query, fieldTypes map[string]VectorStoreProp
...
+// rejected `SORTBY <knn-alias>` and the query should be retried without it.
+// KNN results are already distance-ordered, so dropping SORTBY is safe.
+//
+// The error wording is backend-specific:
+//   - Redis/RediSearch: "unexpected argument `SORTBY`"
+//   - valkey-search:     "Index field `score` does not exist"
+//
+// (valkey-search treats the KNN `AS score` alias as a non-existent index field
+// when it appears in SORTBY/RETURN, rather than reporting SORTBY itself.)
+func isSortByUnsupportedErr(err error) bool {
```
akshaydeo: This function seems too brute forced - are there any constants we can use from the library or something better?

dada-engineer: yeah it is, the only other option I see when using the go-redis is to not sort in the query and sort client side within bifrost, which would be alright as kNN already returns exactly k items.

dada-engineer: FT.SEARCH already sorts by distance (https://valkey.io/commands/ft.search/)

> SORTBY <field> [ASC | DESC] (Optional): If present, results are sorted according the value of the specified field and the optional sort-direction instruction. By default, vector results are sorted in distance order and non-vector results are not sorted in any particular order. Sorting is applied before the LIMIT clause is applied.

This although is not the case for redisearch... so as k is smell the client sort is IMO the best option there.

PR #4531, opened 2026-06-18
Fix Postgres password command DSN parsing

dani29:
## Summary
- quote libpq keyword DSN values when building Postgres connection strings
- preserve empty static password fields for password_command without losing dbname parsing
- add a regression test for a service-account-style username with password_command

## Verification
- go test ./postgresconn
- go test ./postgresconn ./configstore ./logstore -run '^$'
- real Postgres verification with BIFROST_POSTGRES_PASSWORD_COMMAND_TEST=1 using a service-account-style username and db bifrost


dani29:
cc @akshaydeo, another small follow-up :-(  Looks like command parsing is problematic with password command.

akshaydeo:
fixes #4564

akshaydeo (approved):

Issue #3879, opened 2026-05-29
Anthropic raw-body passthrough does not propagate betas[], causing Claude Code mid-conversation system messages to fail


qixiangyang:
# Anthropic raw-body passthrough does not propagate `betas[]`, causing Claude Code mid-conversation system messages to fail

## Summary

When using Bifrost's Anthropic-compatible endpoint with Claude Code / Claude CLI, requests can include mid-conversation system messages inside `messages[]` together with a top-level `betas` array that includes:

```json
"mid-conversation-system-2026-04-07"
```

In Bifrost's Anthropic raw-body passthrough path, the raw request body is preserved and sent downstream after some request-body mutations such as `model` replacement. However, the top-level body field `betas[]` does not appear to be propagated into the outbound `anthropic-beta` HTTP header.

As a result, Anthropic may receive a request containing:

```json
{
  "messages": [
    { "role": "system", "content": "..." }
  ]
}
```

without the required `anthropic-beta: mid-conversation-system-2026-04-07` header, and returns:

```text
400 invalid_request_error: role 'system' is not supported on this model
```

## Why this matters

Claude Code may send requests where:

- `system` exists as a top-level field;
- `messages[]` also contains a later `{ "role": "system" }` message;
- the request body includes `betas[]` with `mid-conversation-system-2026-04-07`.

This appears to be an Anthropic beta protocol feature rather than a plain invalid request body. If Bifrost takes the typed conversion path, system/developer messages can be lifted or normalized. But in raw-body passthrough, the body is preserved, so correct beta propagation becomes necessary.

## Minimal request shape

A reduced version of the incoming request body looks like this:

```json
{
  "model": "claude-sonnet-4-6",
  "max_tokens": 32000,
  "stream": true,
  "system": [
    { "type": "text", "text": "You are Claude Code..." }
  ],
  "messages": [
    {
      "role": "user",
      "content": [
        { "type": "text", "text": "hello" }
      ]
    },
    {
      "role": "system",
      "content": "The following skills are available for use with the Skill tool: ..."
    }
  ],
  "betas": [
    "claude-code-20250219",
    "interleaved-thinking-2025-05-14",
    "redact-thinking-2026-02-12",
    "thinking-token-count-2026-05-13",
    "context-management-2025-06-27",
    "prompt-caching-scope-2026-01-05",
    "mid-conversation-system-2026-04-07",
    "advisor-tool-2026-03-01",
    "effort-2025-11-24"
  ],
[...]

PR #4534, opened 2026-06-18
fix(bedrock): inline mid-conversation system messages to preserve prompt cache

### LINKED ISSUE #4068: Bedrock: mid-conversation system messages hoisted into top-level `system` block break prompt caching
## Summary

On the **Bedrock** provider, prompt-cache reads collapse intermittently for multi-turn agentic conversations (e.g. Claude Code / any client that injects `role: "system"` messages mid-conversation). Cached-read tokens drop from ~full-context down to a fixed floor (system+tools only), forcing a full re-write of the conversation prefix on the affected turn. This recurs whenever a new mid-conversation system message appears, and cascades for a turn or two after.

The root cause is that `ConvertBifrostMessagesToBedrockMessages` hoists **every** `role: system` / `role: developer` message into Bedrock's top-level `system` block, regardless of its position in the message stream.

## Affected code

`core/providers/bedrock/responses.go`, in `ConvertBifrostMessagesToBedrockMessages` (v1.5.8, around line 3445):

```go
// Convert regular message
if role == schemas.ResponsesInputMessageRoleSystem || role == schemas.ResponsesInputMessageRoleDeveloper {
    // Convert to system message
    systemMsgs := convertBifrostMessageToBedrockSystemMessages(&msg)
    systemMessages = append(systemMessages, systemMsgs...)
} else {
    ...
}
```

This unconditionally routes any system/developer message — including ones that appear *after* user/assistant/tool turns — into the top-level `system` array.

## Why it breaks caching on Bedrock specifically

Bedrock's Converse prompt cache is **prefix-based**: a `cachePoint` only matches if the entire serialized prefix in front of it is byte-identical to a prior request. The top-level `system` block sits in front of the *entire* conversation.

Clients like Claude Code legitimately emit transient `role: system` reminders **interleaved at their chronological position** in the conversation (e.g. "task tools haven't been used recently", `<new-diagnostics>` notices, skill lists, file-change notes). Each time a new one appears:

1. Bifrost appends it to the top-level `system` block.
2. The `system` block grows → the serialized prefix in front of the whole conversation changes.
3. Bedrock's prefix match fails → the read falls back to the only stable prefix (system prompt + tools), and the entire conversation is re-cached at the uncached input rate.

The **native Anthropic** provider is not sensitive to this in the same way (it tolerates system-message placement), so the bug is Bedrock-specific even though both serve Claude.

## Cost impact

Because a miss re-writes the full conversation prefix, the per-request cost is sharply bimodal — cache hits vs. full re-writes differ by ~10–30×. In a sustained agentic session the misses recur every few turns, so a meaningful fraction of requests pay the uncached price.

## Reproduction

Convert a Responses-API message list that places a `role: system` message *after* the conversation has started, and inspect the resulting Bedrock Converse request:

```go
input := []schemas.ResponsesMessage{
    sysMsg("You are a helpful assistant."), // legitimate leading system prompt
    userMsg("first user turn"),
    sysMsg("Reminder injected mid-conversation."), // transient reminder
    userMsg("second user turn"),
}
messages, systemMessages, _ := bedrock.ConvertBifrostMessagesToBedrockMessages(ctx, input)
```

**Observed:** `systemMessages` (the Bedrock top-level `system` block) contains **both** "You are a helpful assistant." *and* "Reminder injected mid-conversation." On the next turn, when another reminder is appended, the `system` block grows again and the prior turn's cached prefix no longer matches.

**Expected:** only the leading system prompt belongs in the top-level `system` block; the mid-conversation reminder should remain at its chronological position in `messages` so the conversation prefix stays stable across turns.

In a live deployment this shows up in the request logs as `cached_read_tokens` repeatedly collapsing to a fixed floor (system+tools) on turns where a new mid-conversation system message appeared, despite the conversation being well within the cache TTL.

## Suggested fix

Only hoist the **leading run** of `role: system` / `role: developer` messages (the genuine system prompt, before the first non-system message) into the top-level `system` block. Inline any later system/developer message at its chronological position as a `user` message (Bedrock has no message-level system role) — optionally wrapped in `<system-reminder>…</system-reminder>` to preserve provenance for the model.

This keeps the serialized prefix stable across turns, so Bedrock's prefix cache keeps hitting. We've validated this approach against production Claude Code traffic (misses eliminated, p90 request cost dropped ~5×) and it leaves the Anthropic-direct and non-Bedrock paths untouched. Happy to open a PR with the change + regression tests if useful.

## Environment

- Bifrost transport **v1.5.8**
- Provider: **bedrock** (Converse API), model `anthropic.claude-*` (observed on Opus 4.x)
- Client: Claude Code (any client that injects mid-conversation `role: system` messages)

mickgvirtu:
## Summary

Fixes #4068. On the Bedrock provider, `ConvertBifrostMessagesToBedrockMessages` hoists **every** `role:system`/`role:developer` message into Bedrock's top-level `system` block, regardless of position. Because Bedrock's Converse prompt cache is prefix-based, a `role:system` message injected mid-conversation (e.g. the reminders Claude Code emits) grows the `system` prefix in front of the cached conversation and collapses cache reads to the system+tools floor — recurring on every such turn.

## Changes

- When the model is in the Anthropic family, keep only the **leading run** of system/developer messages in `system`; messages appearing after the conversation starts are inlined in place. Non-Anthropic models keep the historical hoist-everything behavior.
- This mirrors the native Anthropic provider's existing `SupportsMidConversationSystem` handling. Bedrock has no message-level `system` role, so an inlined message is rendered as a `user` turn (wrapped in `<system-reminder>…</system-reminder>`, matching the convention clients already use for pre-wrapped reminders).
- Gating is an `inlineSystemReminders bool` computed by the caller via `IsAnthropicModelFamily(ctx, model)` (alias-aware, consistent with the other Anthropic gates in the file).
- `cache_control` on tool calls/results is preserved as a `CachePoint` carrying the requested TTL.

## Type of change

- [x] Bug fix

## Affected areas

- [x] Core (Go)
- [x] Providers/Integrations

## How to test

```sh
go test ./core/providers/bedrock/
```

Adds `TestMidConversationSystemReminderStaysInline`, `…HoistedForNonAnthropic`, `TestToolCacheControlBecomesCachePointWithTTL` (positive TTL assertion), a lone-system early-return test, and a no-leading-system-block gate test.

Issue #4068 has the full root-cause plus a real cache-read trace (cached tokens dropping to the system/tools floor and recovering after the prefix re-warms). Related native-side work: #4276, #3879.

---

Notes for reviewers: this re-adds a parameter the converter previously dropped — happy to thread it differently (e.g. derive from a typed context) if you prefer. The `<system-reminder>` wrapping follows the client convention; open to gating it if you'd rather it not be implicit.

### DIFF
--- a/core/providers/bedrock/bedrock_test.go
+++ b/core/providers/bedrock/bedrock_test.go
@@ -2958,7 +2958,7 @@ func TestToolResultJSONParsingResponsesAPI(t *testing.T) {
 				},
 			}
 
-			messages, _, err := bedrock.ConvertBifrostMessagesToBedrockMessages(context.Background(), input)
+			messages, _, err := bedrock.ConvertBifrostMessagesToBedrockMessages(context.Background(), input, false)
 			require.NoError(t, err)
 			require.Len(t, messages, 1)
 
@@ -5104,7 +5104,7 @@ func TestToolResultImageContentResponsesAPI(t *testing.T) {
 			},
 		}
 
-		messages, _, err := bedrock.ConvertBifrostMessagesToBedrockMessages(context.Background(), input)
+		messages, _, err := bedrock.ConvertBifrostMessagesToBedrockMessages(context.Background(), input, false)
 		require.NoError(t, err)
 		require.Len(t, messages, 1)
 
@@ -5148,7 +5148,7 @@ func TestToolResultImageContentResponsesAPI(t *testing.T) {
 			},
 		}
 
-		messages, _, err := bedrock.ConvertBifrostMessagesToBedrockMessages(context.Background(), input)
+		messages, _, err := bedrock.ConvertBifrostMessagesToBedrockMessages(context.Background(), input, false)
 		require.NoError(t, err)
 		require.Len(t, messages, 1)
 
@@ -5181,7 +5181,7 @@ func TestToolResultImageContentResponsesAPI(t *testing.T) {
 			},
 		}
 
-		messages, _, err := bedrock.ConvertBifrostMessagesToBedrockMessages(context.Background(), input)
+		messages, _, err := bedrock.ConvertBifrostMessagesToBedrockMessages(context.Background(), input, false)
 		require.NoError(t, err)
 		require.Len(t, messages, 1)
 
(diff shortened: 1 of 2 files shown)


mickgvirtu:
Thanks — both addressed:

- **TTL test exercised the pre-existing end-of-sequence flush, not the new flush-before-message path** (greptile): `TestToolCacheControlBecomesCachePointWithTTL` is now table-driven over two shapes — `end of sequence` (no following message) and `followed by a message` (`[user, FunctionCall(+cache), FunctionCallOutput(+cache), user]`), which reaches the new CachePoint code inside `case ResponsesMessageTypeMessage`. Both assert the `1h` TTL survives.
- **Lone `developer` branch** (coderabbit): `TestLoneSystemMessageReturnsUserMessage` now runs over both `system` and `developer` roles.

mickgvirtu:
Rebased onto `transports/v1.5.16` to keep current with `dev`. The upstream changes to `core/providers/bedrock/responses.go` and `bedrock_test.go` in this release did not conflict with the inline-system-reminder logic; `go test ./core/providers/bedrock/` passes. Still mergeable into `dev`.

mickgvirtu:
## Known limitation: non-Anthropic Bedrock families (Nova, imported models) still hoist

Flagging the boundary of this fix. The inline behavior is gated on `IsAnthropicModelFamily(model)` (ultimately `strings.Contains(model, "anthropic.") || strings.Contains(model, "claude")`), so it only applies to Claude-on-Bedrock. Other Bedrock families keep the historical hoist-everything behavior.

That leaves a real gap: **Amazon Nova also prefix-caches** -- `BedrockModelSupportsCachePoints` is `IsAnthropicModel(model) || IsNovaModel(model)` -- so a Nova request with Claude Code's mid-conversation `role:system` reminders gets the same prefix-cache breakage this PR fixes for Claude. Custom/imported models on Bedrock (e.g. a GLM imported via Custom Model Import) are in the same boat.

This narrow gating is deliberate, not an oversight: the Bedrock inline transform is **lossy** -- because Converse has no message-level system role, the reminder is rewritten into a `user` turn wrapped in the Claude-Code-specific `<system-reminder>\n...\n</system-reminder>\n` envelope. Applying that envelope blindly to Nova/Llama/imported models could change how those models interpret the turn. Generalizing safely needs a family-appropriate envelope (or a per-model opt-in), not just widening the gate.

(For contrast, the Anthropic-compatible custom-provider counterpart -- #4592 / #4593 -- is a **no-op transform**: the wire format has a native inline `role:system`, so "inline" just means "don't move it." That is why it can safely generalize to all custom anthropic-base providers, whereas the Bedrock side cannot generalize without an envelope-per-family.)

Filing this as a known limitation / follow-up rather than expanding this PR's scope.

TejasGhatte (approved):

akshaydeo (approved):

PR #4536, opened 2026-06-18
feat(config): -logs-dir flag to relocate logs.db out of the config dir

mickgvirtu:
## Summary

`logs.db` is created as a sibling of `config.db` in the app dir, forcing the config and the large, churning request-log DB onto the same directory/volume. This adds a `-logs-dir` flag (default from `BIFROST_LOGS_DIR`, mirroring the `-host`/`BIFROST_HOST` precedence) so the two can live on independent mounts — useful for putting logs on separate or ephemeral storage.

## Changes

- New `-logs-dir` flag threaded `main.go → Server.LogsDir → LoadConfig(ctx, configDir, logsDir)`.
- Applies to the default SQLite logs store. When the logs store is explicitly configured in `config.json` or is non-SQLite, the override is **ignored with a warning** (not silently), so the operator gets feedback.

## Type of change

- [x] Feature

## Affected areas

- [x] Transports (HTTP)

## How to test

```sh
go test ./transports/bifrost-http/lib/ -run TestLoadConfig_LogsDirRelocatesLogsDB
```

Covers relocation, the empty-default case, and explicit-config-wins.


mickgvirtu:
Thanks for the review. Addressed in the latest push:

- **Startup crash when `-logs-dir` points at a non-existent path** (greptile): `LoadConfig` now `os.MkdirAll`s the logs dir before opening `logs.db` (SQLite won't create the parent). Added a `logs dir created when it does not exist` subtest.
- **Over-aggressive override of an explicitly-configured stored SQLite path** (coderabbit): the relocation now only applies when the stored path is the implicit default (next to `config.db`, or empty). An operator-configured stored path is left unchanged with a warning, so `-logs-dir` never silently switches an explicitly-chosen logs DB.
- **Explicit-config precedence test**: covered for SQLite (`explicit logs store config wins over logs dir override`). I left out the non-SQLite explicit-config case because initializing a non-SQLite (e.g. Postgres) logstore in a unit test needs a live DB connection — the non-SQLite path is already guarded by the `LogStoreType != SQLite` warn branch, which is exercised at runtime.

`go test ./transports/bifrost-http/lib/ -run TestLoadConfig_LogsDirRelocatesLogsDB` passes (relocation, empty-default, explicit-config-wins, auto-created dir).

mickgvirtu:
Rebased onto `transports/v1.5.16` to keep current with `dev`. Resolved the line-offset shifts in `config.go`/`server.go` from this release; the logs-dir relocation tests (incl. `TestLoadConfig_LogsDirRelocatesLogsDB`) pass.

mickgvirtu:
Rebased onto the latest `dev` (post-v1.6.0). Updated the test for the v1.6.0 `LoadConfig` signature (now takes the `logsDir` arg); the logs-dir relocation tests pass.

mickgvirtu:
Rebased onto the latest `dev` (post-v1.6.6). Updated the test for the v1.6.6 `LoadConfig` signature (now takes the `logsDir` arg) and added a stub for the new `ConfigStore.UpdateBudgetOverride` interface method; the logs-dir relocation tests pass.

Issue #2695, opened 2026-04-14
/metrics endpoint requires authentication, breaks Prometheus scraping


bilby91:
When basic auth or admin key authentication is enabled, the `/metrics` Prometheus endpoint returns 401 Unauthorized. This prevents Prometheus from scraping bifrost metrics without embedding credentials in the scrape config.

The `/health` endpoint is already whitelisted from auth in `APIMiddleware()` (`transports/bifrost-http/handlers/middlewares.go`, line 556). `/metrics` should receive the same treatment — it's a standard observability endpoint that's typically unauthenticated on internal networks.

**Current behavior:**
```
$ curl http://bifrost:8080/metrics
HTTP/1.1 401 Unauthorized
```

**Expected behavior:**
`/metrics` should respond without authentication, same as `/health`.

**Suggested fix:**

Add `/metrics` to the `whitelistedRoutes` in `APIMiddleware()`:

```go
whitelistedRoutes := []string{
    "/api/session/is-auth-enabled",
    "/api/session/login",
    "/api/oauth/callback",
    "/health",
    "/metrics",  // <-- add this
}
```

Alternatively, a config option like `metrics.public: true` (defaulting to `true`) would give operators control.

**Version:** transports/v1.4.19

dennypradipta:
You can configure it in the Prometheus config though?

```yaml
        - job_name: "bifrost"
          scrape_interval: 15s
          static_configs:
            - targets: ["bifrost:8080"]
          basic_auth:
            username: "dennypradipta"
            password: "somethingsomething"
```

akshaydeo:
yes and we have added whitelisting of paths in prerelease cut; where you can whitelist /metrics endpoint to avoid going through auth

PR #4537, opened 2026-06-18
feat(auth): serve /metrics without auth by default (opt-out via metrics_require_auth)

mickgvirtu:
## Summary

Follow-up to maximhq/bifrost#2695. That issue was resolved by letting operators add `/metrics` to the path whitelist — which works. This proposes making the conventional behavior the **default**: `/metrics` (like `/health`) bypasses auth out of the box, since Prometheus scrapers typically can't carry admin credentials, with an explicit **opt-out** for operators who consider the metric labels sensitive.

## Changes

* `/metrics` bypasses `APIMiddleware` auth by default, via an exact-match gate (no prefix widening: `/metricsX`, `/metrics/foo` stay authenticated).
* New `ClientConfig.metrics_require_auth` (default `false`); set `true` to keep `/metrics` behind auth — for deployments where labels (provider/model/virtual-key/team/customer/cost) are sensitive.
* Removed a duplicate `/health` entry in the whitelist.

## Type of change

- [X] Feature

## Affected areas

- [X] Transports (HTTP)
- [X] Core (Go)

## How to test

```sh
go test ./transports/bifrost-http/handlers/ -run TestAuthMiddleware_Metrics
```

Covers public-by-default, gated-when-`metrics_require_auth=true`, and the exact-match negatives.

## Follow-up: UI toggle (happy to send as a companion PR)

This adds a `metrics_require_auth` client-config field, which the dashboard should surface alongside the existing auth controls. The natural home is the **Security** settings view (`ui/app/workspace/config/views/securityView.tsx`) — next to the "Enable authentication" switch and the "Whitelisted routes" editor — plus the `ClientConfig` type in `ui/lib/types/config.ts` (beside `whitelisted_routes`): a single switch ("Require auth for /metrics", default off) wired through the existing `setLocalConfig` flow. Kept this PR backend-only to stay reviewable; glad to open the matching UI PR following your securityView patterns if you'd like it.

---

Deferential note: maximhq/bifrost#2695 was closed with the position that the existing path-whitelist is the intended mechanism. We're proposing a default change on the principle that the conventional unauthenticated endpoint should require config to *secure*, not to *expose* — but we fully understand if you prefer to keep it whitelist-only. Happy to close this or rework it into docs for the whitelist path instead.


mickgvirtu:
Thanks — addressed:

- **`metrics_require_auth` missing from `transports/config.schema.json`** (greptile P1 / coderabbit): added the field to the schema (boolean, default false) next to `whitelisted_routes`, with a description noting it gates the public-by-default `/metrics`. `go test ./transports/schema_test/` passes, and the schema↔struct sync check no longer reports `metrics_require_auth` as drift.

On the design itself (per the PR description): this remains a *proposal* to flip the default to public — happy to defer to the whitelist-only approach from #2695 if you'd prefer. I also offered a companion UI toggle for the Security settings view; glad to send it if this lands.

mickgvirtu:
Rebased onto `transports/v1.5.16` to keep current with `dev`. Re-applied the `metrics_require_auth` entry against this release's restructured `config.schema.json`; `TestConfigSchemaSync` and `TestAuthMiddleware_Metrics` pass. Still mergeable into `dev`.

mickgvirtu:
Rebased onto the latest `dev` (post-v1.6.0). Updated the test for the v1.6.0 `EnvVar`→`SecretVar` rename (`schemas.NewSecretVar`); `TestAuthMiddleware_Metrics` and the schema-sync check pass.

mickgvirtu:
Rebased onto the latest `dev` (post-v1.6.6). Updated the test for the v1.6.6 `EnvVar`→`SecretVar` rename (`schemas.NewSecretVar`) and merged `MetricsRequireAuth` into the rewritten `ClientConfig` struct; `TestAuthMiddleware_Metrics` and the schema-sync check pass.

akshaydeo:
~~we can't do this as this was marked in our VAPT report. Auth on every endpoint is required - you can opt out from settings~~ - my bad - let me confirm the changes first 🙇

Issue #4530, opened 2026-06-18
[Bug]: /v1/responses strips OpenRouter server tools (openrouter:web_search) via filterUnsupportedTools


abdenasseraroukhsiss:
### Summary

On the `/v1/responses` path with `provider: openrouter`, Bifrost silently removes OpenRouter server tools such as `openrouter:web_search` from the request. The upstream OpenRouter call ends up with `tools: []`, so the tool never runs. With `tool_choice: "required"`, the upstream provider then returns `400` (a required tool was declared but no tool is present).

### Environment
- Provider: `openrouter`
- Endpoint: `/v1/responses`
- Tool: `openrouter:web_search` (an OpenRouter-native server tool)

### Reproduction

Calling OpenRouter directly with the server tool works (the tool runs, citations are returned):

```bash
curl https://openrouter.ai/api/v1/responses \
  -H "Authorization: Bearer $OPENROUTER_API_KEY" -H "Content-Type: application/json" \
  -d '{"model":"anthropic/claude-haiku-4.5",
       "input":[{"role":"user","content":"Latest news today? Give source URLs."}],
       "tools":[{"type":"openrouter:web_search"}],
       "tool_choice":"required"}'
# -> message contains annotations[].url_citation
```

The same request routed through Bifrost loses the tool (no search, no citations):

```bash
curl http://localhost:8080/v1/responses \
  -H "x-bf-vk: $VK" -H "Content-Type: application/json" \
  -d '{"provider":"openrouter","model":"openrouter/anthropic/claude-haiku-4.5",
       "input":[{"role":"user","content":"Latest news today? Give source URLs."}],
       "tools":[{"type":"openrouter:web_search"}],
       "tool_choice":"required"}'
# -> tools:[] reaches OpenRouter; no annotations / no search
```

### Root cause

`filterUnsupportedTools` in `core/providers/openai/responses.go` filters the request's tools against a whitelist of OpenAI-native `ResponsesToolType` values. OpenRouter server tools (`openrouter:*`) are not in that whitelist, so they are dropped — even when the provider is OpenRouter, which natively supports them:

```go
supportedTypes := map[schemas.ResponsesToolType]bool{
    // function, file_search, web_search, web_fetch, mcp, code_interpreter, ...
    // openrouter:web_search is not present
}
...
for _, tool := range resp.Tools {
    if supportedTypes[tool.Type] { ... } // openrouter:web_search -> false -> dropped
}
```

[...]

PR #4543, opened 2026-06-18
fix(schemas/responses): forward server-tool parameters on ResponsesTool

### LINKED ISSUE #4541: [Bug]: /v1/responses drops server-tool parameters (ResponsesTool has no parameters field)
### Summary

On `/v1/responses`, server tools that carry a `parameters` object lose it. For example:

```json
"tools": [{ "type": "openrouter:web_search", "parameters": { "engine": "exa" } }]
```

reaches the provider as `{"type":"openrouter:web_search"}` — the `parameters` are dropped. Server tools that need configuration (engine, max_results, …) therefore can't be controlled, and tools that rely on a non-default engine effectively don't work.

### Root cause

`ResponsesTool` in `core/schemas/responses.go` models typed tool variants (function, web_search, etc.) but has no generic `parameters` passthrough. Neither `UnmarshalJSON` nor `MarshalJSON` carries a `parameters` field, so it is silently discarded during (de)serialization.

### Reproduction

Send a Responses request with a server tool that has parameters:

```json
{
  "model": "...",
  "input": [{ "role": "user", "content": "..." }],
  "tools": [{ "type": "openrouter:web_search", "parameters": { "engine": "exa" } }]
}
```

The request forwarded to the provider contains the tool `type` but no `parameters`.

### Expected

Forward the `parameters` object verbatim for server tools (at least the `openrouter:` namespace), so engine/config selection works.

### Related

#4530 — same area (that one is the tool-type whitelist in `filterUnsupportedTools`).

abdenasseraroukhsiss:
### What

`ResponsesTool` models typed tool variants (function, web_search, etc.) but has no generic `parameters` passthrough. As a result, configuration for provider/server tools — e.g. OpenRouter's `openrouter:web_search` with `{"engine":"exa","max_results":5}` — is silently dropped during (de)serialization on `/v1/responses`.

Fixes #4541.

### Change

Add a `Parameters map[string]interface{}` field to `ResponsesTool`, captured in `UnmarshalJSON` and emitted (deterministically, via `MarshalSorted`) in `MarshalJSON`, so server-tool parameters are forwarded verbatim.

### Test

`TestResponsesTool_ParametersRoundTrip` — unmarshal a tool with `parameters`, assert it is captured, marshal it back, assert the parameters survive on the wire.

### Note

Independent of #4530 / #4532 (the `filterUnsupportedTools` whitelist). Both are needed for OpenRouter server tools to fully work on `/v1/responses`: the type must pass the whitelist **and** its parameters must be forwarded.

### DIFF
--- a/core/changelog.md
+++ b/core/changelog.md
@@ -1,0 +1,1 @@
+- fix: deterministic MCP tool ordering for prompt cache stability (closes #2347)
--- a/core/mcp/toolmanager.go
+++ b/core/mcp/toolmanager.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"slices"
 	"strings"
 	"sync/atomic"
 	"time"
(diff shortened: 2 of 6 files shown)


abdenasseraroukhsiss:
Hi @akshaydeo  👋 
Rebased on latest dev (diff back to the 2 intended files) and addressed all review feedback:
- Parameters now scoped to the openrouter: namespace (marshal + unmarshal) — no longer captures a function tool's typed parameters.
- Round-trip test asserts engine and max_results=5 nested under parameters (no substring check), plus a negative test for function tools.

PR #4548, opened 2026-06-18
feat: auto-detect ChatGPT subscription tokens and route to chatgpt.com

### LINKED ISSUE #4459: [Feature]: Auto-detected ChatGPT/Codex OAuth passthrough (forward the official client's existing token, no toggle)
### Prerequisites

*Related: #2316 (original request, framed as device-code flow), and the prior implementation attempts #2715 / #2775 (both closed). This reframes the problem more narrowly.*

*Context: raising this on behalf of envelio, where we're evaluating Bifrost for company-wide rollout and the enterprise tier. Routing the official Codex CLI through the gateway is load-bearing for that decision, so I'd like it upstream rather than carried as a fork patch.*

### Problem to solve

This isn't "add OAuth to Bifrost". It's forwarding the official client's existing OAuth correctly. Codex already holds a valid "Sign in with ChatGPT" token. Pointing its `base_url` at Bifrost only breaks because Bifrost forwards to `api.openai.com/v1/responses`, while a subscription token is accepted only at `chatgpt.com/backend-api/codex/responses`. Bifrost stores/forges nothing and pools nothing. No new OAuth flow is added.

### Proposed solution

Auto-detect, don't toggle. Like the Anthropic Claude-Code passthrough: when a `/responses` request carries a bearer whose JWT has the `https://api.openai.com/auth` -> `chatgpt_account_id` claim, route to the ChatGPT backend, applying:

- `/v1` stripped from the path,
- `chatgpt-account-id` (already in the token's JWT claim) echoed as a header,
- `OpenAI-Beta: responses=experimental`,
- body normalized (`store=false`, `stream=true`).

No config switch, no UI toggle, no separate provider to declare. The config toggle is what made #2715 / #2775 over-scoped, not the logic itself.

### Alternatives considered

**Out-of-tree plugin, and I'd like your read on whether it's viable.** My understanding is that the `/v1` strip and body normalization happen on the *outbound* call inside the OpenAI provider (derived from the request type), while plugin hooks only see the *inbound* request, with no outbound URL/body rewrite hook, so this can't be a plugin and has to live in core. I'm not certain I've read the plugin surface correctly, and I'd rather not put logic in core if there's a clean out-of-tree path. Is there a hook or pattern I'm missing that would let a plugin rewrite the outbound path/body? If not, I'll do it in core. Either way, a recommendation on the preferred shape would help. The actual logic is small: one helper file plus a few touch-points.

### Area(s)

Core (Go), Transports (HTTP)

### Additional context

**On whether it's allowed** (relevant for fleet-wide use): the prohibited pattern is credential sharing / account pooling, which this isn't, each user is on their own seat. The other hook is circumventing rate limits, which a passive forward-only gateway doesn't do. OpenAI's terms are otherwise silent on routing the official client through a corporate gateway. No clause is violated and no protective measure circumvented, as long as the impl stays strictly forward-only (no rewriting limit responses, no pooling, no faking a client). Orgs at scale should still confirm with their OpenAI account owner.

I have a narrow implementation (auto-detect, no toggle, reusing the path/header/body helpers) and will open a PR referencing this issue.

Purvi09:
## Summary

When a user logs into Codex CLI using their ChatGPT subscription (`codex login` with ChatGPT OAuth), the resulting bearer token is a JWT containing a `chatgpt_account_id` claim. Sending this token to `api.openai.com/v1/responses` returns 401 because it is a subscription token, not an API key.

This PR auto-detects that token in Bifrost's `/v1/responses` pre-hook and transparently reroutes the request to `chatgpt.com/backend-api/codex/responses` — the endpoint that actually accepts it. No config change or URL change is required from the user.

## Changes

- **`core/providers/openai/chatgpt_passthrough.go`** — new file with `ParseChatGPTJWT` (decodes the JWT payload and checks for the `chatgpt_account_id` claim under `https://api.openai.com/auth`) and `IsChatGPTPassthrough` (reads the detection flag from context)
- **`core/schemas/bifrost.go`** — adds `BifrostContextKeyChatGPTPassthrough` context key (bool) set by the pre-hook and read by the provider
- **`transports/bifrost-http/integrations/openai.go`** — pre-hook for `POST /v1/responses` that calls `ParseChatGPTJWT` on the incoming bearer token; if detected, sets the URL override to `chatgpt.com/backend-api/codex/responses`, skips Bifrost key selection, and forwards the original bearer token
- **`core/providers/openai/openai.go`** — `Responses()` and `ResponsesStream()` force `store: false` when `IsChatGPTPassthrough` is true (required by the chatgpt.com backend)
- **`core/providers/openai/chatgpt_passthrough_test.go`** — unit tests for `ParseChatGPTJWT` covering valid JWT, missing claim, malformed tokens, non-JWT API keys, and edge cases

> Design decision: JWT signature is intentionally not verified — the token is used only as a routing hint. The upstream chatgpt.com endpoint validates it for real.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [x] Core (Go)
- [x] Transports (HTTP)
- [x] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

**Unit tests:**

```sh
cd core
go test ./providers/openai/ -run TestParseChatGPTJWT -v
```

## Breaking changes

- [ ] Yes
- [x] No

## Related issues

Closes #4459

## Security considerations

- The JWT is decoded (base64) but **not signature-verified** — it is used only as a routing hint. The chatgpt.com backend performs real authentication with the token.
- No credentials are stored or logged; the bearer token passes through in-memory only.
- `BifrostContextKeySkipKeySelection` is set so Bifrost does not attempt to inject a stored API key over the user's token.

## Checklist

### DIFF
--- a/core/providers/openai/chatgpt_passthrough.go
+++ b/core/providers/openai/chatgpt_passthrough.go
@@ -1,0 +1,57 @@
+package openai
+
+import (
+	"encoding/base64"
+	"strings"
+
+	"github.com/bytedance/sonic"
+	"github.com/maximhq/bifrost/core/schemas"
+)
+
+const (
+	chatGPTAccountIDKey = "chatgpt_account_id"
+	openAIAuthClaim     = "https://api.openai.com/auth"
+
+	// ChatGPTCodexURL is the full upstream URL for ChatGPT subscription token requests.
+	ChatGPTCodexURL = "https://chatgpt.com/backend-api/codex/responses"
+)
+
+// ParseChatGPTJWT parses a raw bearer token, checks for the ChatGPT subscription
+// JWT claim, and returns the chatgpt_account_id. No signature verification is
+// Returns ("", false) for any non-ChatGPT or malformed token.
+func ParseChatGPTJWT(token string) (accountID string, ok bool) {
+	parts := strings.Split(token, ".")
+	if len(parts) != 3 {
+		return "", false
+	}
+
+	payload, err := base64.RawURLEncoding.DecodeString(parts[1])
+	if err != nil {
+		return "", false
+	}
+
+	// Extract the nested claim: {"https://api.openai.com/auth": {"chatgpt_account_id": "..."}}
+	var claims map[string]interface{}
+	if err := sonic.Unmarshal(payload, &claims); err != nil {
+		return "", false
+	}
+
+	authClaim, ok := claims[openAIAuthClaim].(map[string]interface{})
+	if !ok {
+		return "", false
+	}
+
+	accountID, ok = authClaim[chatGPTAccountIDKey].(string)
+	if !ok || accountID == "" {
+		return "", false
+	}
+
+	return accountID, true
+}
+
+// IsChatGPTPassthrough reports whether the current request was auto-detected
+// as a ChatGPT subscription token and should be routed to chatgpt.com.
+func IsChatGPTPassthrough(ctx *schemas.BifrostContext) bool {
+	v, _ := ctx.Value(schemas.BifrostContextKeyChatGPTPassthrough).(bool)
+	return v
+}
--- a/core/providers/openai/chatgpt_passthrough.go
+++ b/core/providers/openai/chatgpt_passthrough.go
@@ -18,6 +18,7 @@ const (
 
 // ParseChatGPTJWT parses a raw bearer token, checks for the ChatGPT subscription
 // JWT claim, and returns the chatgpt_account_id. No signature verification is
+// performed — the token is used only as a routing hint.
 // Returns ("", false) for any non-ChatGPT or malformed token.
 func ParseChatGPTJWT(token string) (accountID string, ok bool) {
 	parts := strings.Split(token, ".")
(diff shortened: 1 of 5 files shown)


Purvi09:
@coderabbitai review

Pratham-Mishra04:
Hey @Purvi09, thanks for the PR! I don't think this is the right approach for solving this problem. Could you take a look at how the auto-detection for max accounts is implemented in the Anthropic integration? I think we can use a similar pattern here by leveraging `BifrostContextKeyUseRawRequestBody`, `BifrostContextKeyExtraHeaders`, `BifrostContextKeyURLPath`, and `BifrostContextKeySkipKeySelection`.

Purvi09:
> Hey @Purvi09, thanks for the PR! I don't think this is the right approach for solving this problem. Could you take a look at how the auto-detection for max accounts is implemented in the Anthropic integration? I think we can use a similar pattern here by leveraging `BifrostContextKeyUseRawRequestBody`, `BifrostContextKeyExtraHeaders`, `BifrostContextKeyURLPath`, and `BifrostContextKeySkipKeySelection`.

Hey @Pratham-Mishra04, thanks for the guidance. I've reworked it to follow the Anthropic max-account pattern.

PR #4559, opened 2026-06-19
fix: double message start event in anthropic stream closes #4556

### LINKED ISSUE #4556: [Bug]: duplicate message_start SSE frame on Anthropic passthrough (responses_stream)
### Prerequisites

### Description

On the Anthropic-compatible streaming path (`POST /anthropic/v1/messages?beta=true`, served as `responses_stream`), Bifrost emits **two `message_start` SSE frames for a single response**, both carrying the **same `message.id`**. The [Anthropic Messages streaming spec](https://platform.claude.com/docs/en/build-with-claude/streaming) allows exactly one `message_start` per response.

The two frames differ: the **first** lacks `stop_reason`/`service_tier`/`inference_geo`; the **second** includes them. This is consistent with the first being a Bifrost-synthesized frame (from the internal Responses-API `response.created` lifecycle event) and the second being the genuine upstream Anthropic `message_start` forwarded through — i.e. the response is round-tripped Anthropic → internal Responses → Anthropic, and the synthesized opening frame is not de-duplicated against the forwarded one.

The reference client (Claude Code) tolerates this — its accumulator keys on `content_block_delta` + `message_delta`/`message_stop` — so inference still works. But it is a protocol violation that (a) breaks any strict SSE consumer that resets state on `message_start`, and (b) is a **usage-accounting hazard for downstream consumers**: the two frames carry different usage payloads, so a client that accumulates usage on each `message_start` can **double-count in its own telemetry**. To be clear, this does **not** double actual token spend or billing — Anthropic generates and bills a single response (confirmed by the direct control, and by Bifrost's own single-`usage`/`cost` log entry); the risk is purely client-side double-counting of the duplicated frame.

### Steps to reproduce

1. Front an Anthropic provider in **passthrough** (OAuth bearer passthrough, no provider key) so requests resolve to the `responses_stream` Anthropic path.
2. Send a streaming `claude-opus-4-8` request to `POST /anthropic/v1/messages?beta=true` (Claude Code's exact shape: `anthropic-beta: claude-code-...,oauth-2025-04-20,...`, `Authorization: Bearer <oauth>`).
3. Capture the raw SSE and count event frames.
4. As a control, send the **same** request directly to `api.anthropic.com`.

A self-contained repro script (`sse-framecheck.py`, stdlib only) is in the collapsed section below. It drives a real `claude -p` through a tiny local capture, tallies SSE frames, and flags the violation. `--direct` runs the control against `api.anthropic.com`.

Measured (live):

```
via Bifrost gateway        ->  message_start: 2   (same id repeated)   VIOLATION
direct to api.anthropic.com ->  message_start: 1                        OK  (control, confirmed twice)
```

### Expected behavior

Exactly one `message_start` SSE frame per response (matching a direct Anthropic stream and the published spec).

### Actual behavior

Two `message_start` frames with the same `message.id`; the second carries `stop_reason`/`service_tier`/`inference_geo` that the first omits.

### Affected area(s)

- Core (Go)

### Version

`transports/v1.5.15` (latest at time of filing; OSS, Helm chart `maximhq/bifrost`). The repro capture was taken against the gateway as the deployment was being bumped from `v1.5.11` to `v1.5.15` — config now targets `v1.5.15`. The `v1.5.12`–`v1.5.15` HTTP-transport changelogs contain no duplicate-`message_start` / Anthropic-stream-framing fix (streaming-related entries cover `stop_reason`, `max_output_tokens`, and cache diagnostics), so the behaviour is expected to persist on latest. Can re-confirm on a clean `v1.5.15` instance on request.

### Environment

```
- Server: Bifrost OSS v1.5.11 (Helm), storage postgres
- Path: POST /anthropic/v1/messages?beta=true  (object: responses_stream)
- Provider: anthropic in OAuth-bearer passthrough (no provider key configured)
- Model: claude-opus-4-8
- Client: claude-cli/2.1.179
- Repro harness: macOS, Python 3 stdlib
```

### Relevant logs/output

```
=== SSE frame counts (via Bifrost) ===
  message_start          2
  content_block_start    1
  content_block_delta    2
  content_block_stop     1
  message_delta          1
  message_stop           1
  ping                   1

=== event order ===
  message_start -> content_block_start -> ping -> content_block_delta -> content_block_stop -> message_delta -> message_stop

=== message_start ids ===
  #1  msg_01TLCs...        (no stop_reason / service_tier / inference_geo)
  #2  msg_01TLCs...        (includes stop_reason, service_tier, inference_geo)   <- same id

=== control: direct to api.anthropic.com ===
  message_start          1     OK
```

### Likely origin (code pointers, for maintainer confirmation)

In `core/providers/anthropic/responses.go`:

- **Inbound** `ToBifrostResponsesStream` (~L254-319): one Anthropic `message_start` is expanded into **both** `response.created` and `response.in_progress` ("emit response.created and response.in_progress (OpenAI-style lifecycle)").
- **Outbound** `ToAnthropicResponsesStreamResponse` (~L1426-1467): `response.created` -> `message_start` (~L1427); `response.in_progress` -> `return nil` (~L1466, comment: "Skip ... to avoid duplicate message_start events").

The outbound guard correctly prevents `in_progress` from becoming a second `message_start`, so the duplicate is **not** from that pair. The remaining un-deduplicated case is the **synthesized** `created`->`message_start` frame coexisting with the **forwarded native** Anthropic `message_start` on the passthrough path (the field delta between the two frames matches this: the synthesized frame at ~L1432 never sets `stop_reason`/`service_tier`/`inference_geo`, which is exactly what frame #1 is missing). Flagging as likely origin for maintainers to confirm against the two captured frames.

### Severity

High (major functionality broken) — non-conformant streaming on the Anthropic passthrough path. Note: the Claude Code reference client tolerates it, so user-visible inference still works; re-triage to Medium if that weighting is preferred.

<details>
<summary>Repro script: <code>sse-framecheck.py</code> (stdlib only; <code>--direct</code> for the control)</summary>

```python
#!/usr/bin/env python3
"""Count SSE frames in a real Claude Code response and flag protocol violations.

Drives a real `claude -p` through a local capture that records the raw SSE the
upstream sends, then tallies each event type. The Anthropic Messages streaming
spec says exactly ONE message_start and ONE message_stop per response; the
envelio gateway (Bifrost) has been observed emitting message_start TWICE for the
same message id. This script proves whether that duplication is present.

Compare two upstreams:
  python3 sse-framecheck.py                # default: through the gateway
  python3 sse-framecheck.py --direct       # straight to api.anthropic.com
The --direct run is the control: Anthropic itself emits exactly one
message_start, so a difference between the two runs localizes the bug to the gateway.

On VPN for the gateway run. Stdlib only. No secrets printed.
"""
import collections
import gzip
import json
import os
import socket
import subprocess
import sys
import threading
import zlib
import http.client
from http.server import BaseHTTPRequestHandler, ThreadingHTTPServer
from urllib.parse import urlsplit

DIRECT = "--direct" in sys.argv
UPSTREAM = "https://api.anthropic.com" if DIRECT else "https://ai-gateway.envelio.org"
# Path Claude uses; against api.anthropic.com the /anthropic prefix must be dropped.
PATH_PREFIX = "" if DIRECT else "/anthropic"
HOP = {"connection", "keep-alive", "transfer-encoding", "content-length",
       "te", "trailer", "upgrade", "proxy-authorization", "proxy-authenticate", "host"}
cap = {}


def free_port():
    s = socket.socket(); s.bind(("127.0.0.1", 0)); p = s.getsockname()[1]; s.close()
    return p


def analyze_sse(raw):
    """Tally SSE event types and the message ids seen on each message_start."""
    counts = collections.Counter()
    order = []
    start_ids = []
    for line in raw.splitlines():
        if line.startswith("event:"):
            ev = line.split(":", 1)[1].strip()
            counts[ev] += 1
            # record a compact order signature (collapse repeated deltas)
            if not order or order[-1] != ev:
                order.append(ev)
        elif line.startswith("data:"):
            payload = line[5:].strip()
            if payload.startswith("{") and '"message_start"' in payload:
                try:
                    start_ids.append(json.loads(payload)["message"].get("id", "?"))
                except Exception:
                    pass
    return counts, order, start_ids


class Cap(BaseHTTPRequestHandler):
    protocol_version = "HTTP/1.1"

    def _relay(self):
        n = int(self.headers.get("content-length") or 0)
        body = self.rfile.read(n) if n else b""
        up = urlsplit(UPSTREAM)
        fwd = {k: v for k, v in self.headers.items() if k.lower() not in HOP}
        fwd["Host"] = up.netloc
        # Ask the upstream NOT to compress, so we can read the SSE text directly.
        # (Claude sends Accept-Encoding: gzip,...; Anthropic honors it and the raw
        # bytes would be unparseable without this.)
        fwd["Accept-Encoding"] = "identity"
        # rewrite path: capture is mounted at /anthropic locally; map to upstream prefix
        path = self.path
        if path.startswith("/anthropic") and PATH_PREFIX == "":
            path = path[len("/anthropic"):] or "/"
        conn = http.client.HTTPSConnection(up.hostname, 443, timeout=120)
        try:
            conn.request(self.command, path, body=body or None, headers=fwd)
            r = conn.getresponse()
            data = r.read()
        except Exception as e:
            self.send_error(502, str(e)); return
        if "messages" in self.path and "raw" not in cap:
            cap["status"] = r.status
            cap["ctype"] = r.getheader("content-type", "")
            # Decompress defensively if the upstream ignored identity.
            enc = (r.getheader("content-encoding", "") or "").lower()
            decoded = data
            try:
                if "gzip" in enc:
                    decoded = gzip.decompress(data)
                elif "deflate" in enc:
                    decoded = zlib.decompress(data)
            except Exception:
                decoded = data
            cap["enc"] = enc or "identity"
            cap["raw"] = decoded.decode("utf-8", "replace")
        self.send_response_only(r.status, r.reason)
        for k, v in r.getheaders():
            if k.lower() in HOP:
                continue
            self.send_header(k, v)
        self.send_header("content-length", str(len(data)))
        self.end_headers()
        try:
            self.wfile.write(data)
        except Exception:
            pass
        self.close_connection = True

    do_GET = do_POST = do_PUT = do_DELETE = do_PATCH = _relay

    def log_message(self, *a):
        pass


def main():
    port = free_port()
    srv = ThreadingHTTPServer(("127.0.0.1", port), Cap)
    threading.Thread(target=srv.serve_forever, daemon=True).start()
    base = "http://127.0.0.1:%d/anthropic" % port
    print("upstream: %s  (%s)" % (UPSTREAM, "DIRECT to Anthropic" if DIRECT else "via gateway"))
    print("driving claude -p through capture ...")
    try:
        subprocess.run(
            ["claude", "-p", "Reply with exactly: OK", "--model", "claude-opus-4-8",
             "--settings", json.dumps({"env": {"ANTHROPIC_BASE_URL": base}})],
            capture_output=True, text=True, timeout=90, stdin=subprocess.DEVNULL,
        )
    except FileNotFoundError:
        sys.exit("claude CLI not found on PATH")
    except subprocess.TimeoutExpired:
        pass

    if "raw" not in cap:
        sys.exit("no messages response captured (check VPN / base url override)")

    counts, order, start_ids = analyze_sse(cap["raw"])
    print("\n=== response ===")
    print("status:", cap["status"], "| content-type:", cap["ctype"], "| content-encoding:", cap.get("enc", "?"))
    if "text/event-stream" not in cap["ctype"]:
        print("not an SSE stream — body:", cap["raw"][:300]); return

    print("\n=== SSE frame counts ===")
    for ev in ("message_start", "content_block_start", "content_block_delta",
               "content_block_stop", "message_delta", "message_stop", "ping", "error"):
        if counts.get(ev):
            print("  %-22s %d" % (ev, counts[ev]))

    print("\n=== event order (deltas collapsed) ===")
    print("  " + " -> ".join(order))

    print("\n=== message_start ids ===")
    for i, mid in enumerate(start_ids, 1):
        print("  #%d  %s" % (i, mid))

    print("\n=== verdict ===")
    ms = counts.get("message_start", 0)
    mstop = counts.get("message_stop", 0)
    ok = True
    if ms == 1:
        print("  message_start: 1  OK (spec-conformant)")
    else:
        ok = False
        dup = len(set(start_ids)) < len(start_ids)
        print("  message_start: %d  VIOLATION (spec allows exactly 1)%s"
              % (ms, "  [same id repeated]" if dup else ""))
    if mstop != 1:
        ok = False
        print("  message_stop: %d  (spec allows exactly 1; Anthropic messages API"
              " may omit [DONE] but sends one message_stop)" % mstop)
    if ok:
        print("  -> stream is spec-conformant.")
    else:
        print("  -> NON-CONFORMANT stream. If the --direct run is clean and this"
              " one isn't, the duplication originates at the gateway (Bifrost).")


if __name__ == "__main__":
    main()
```

</details>

TejasGhatte:
## Summary

Fixes a duplicate `message_start` event emitted on the Anthropic passthrough responses stream path. When a single upstream `message_start` is received, it expands into two bifrost responses: `response.created` and `response.in_progress`. The raw upstream frame was previously attached to the last chunk (`in_progress`), which caused the transport to emit two `message_start` frames — a lossy synthesized one from `created` and the raw-forwarded one from `in_progress`. The fix attaches the raw frame to `response.created` instead, ensuring exactly one `message_start` is forwarded with all upstream fields intact (including `service_tier` and `inference_geo`).

## Changes

- When attaching the upstream raw response to a bifrost response chunk, the default "last chunk" rule is now overridden for `message_start` expansions: the raw is attached to `response.created` rather than `response.in_progress`.
- Added `TestResponsesStream_MessageStart_NoDuplicateOnPassthrough` to explicitly guard this fix, verifying that the correct placement emits exactly one verbatim `message_start` and that the old placement produces the duplicate behavior.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [x] Core (Go)
- [ ] Transports (HTTP)
- [x] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

```sh
go test ./core/providers/anthropic/... -run TestResponsesStream_MessageStart_NoDuplicateOnPassthrough -v
```

Expected output: both subtests pass — `raw on created emits one full message_start` confirms a single verbatim frame, and `raw on in_progress (old behavior) duplicates message_start` confirms the old rule produced two frames.

## Screenshots/Recordings

N/A

## Breaking changes

- [ ] Yes
- [x] No

## Security considerations

No security implications.

## Checklist

### DIFF
--- a/core/providers/anthropic/advisor_test.go
+++ b/core/providers/anthropic/advisor_test.go
@@ -358,3 +358,105 @@ func TestAdvisorStream_RoundTrip(t *testing.T) {
 		t.Error("text block dropped")
 	}
 }
+
+// countPassthroughMessageStarts mirrors the transport passthrough converter
+// (transports/bifrost-http/integrations/anthropic.go): for each bifrost stream
+// response it forwards the raw upstream frame verbatim when present (except
+// ContentPartAdded), otherwise falls back to the normalized converter. Returns
+// how many message_start frames are emitted and the raw bytes of the forwarded one.
+func countPassthroughMessageStarts(ctx *schemas.BifrostContext, responses []*schemas.BifrostResponsesStreamResponse) (int, string) {
+	count := 0
+	forwarded := ""
+	for _, r := range responses {
+		if r.ExtraFields.RawResponse != nil && r.Type != schemas.ResponsesStreamResponseTypeContentPartAdded {
+			raw, _ := r.ExtraFields.RawResponse.(string)
+			if gjson.Get(raw, "type").String() == "message_start" {
+				count++
+				forwarded = raw
+			}
+			continue
+		}
+		for _, ev := range ToAnthropicResponsesStreamResponse(ctx, r) {
+			if ev.Type == AnthropicStreamEventTypeMessageStart {
+				count++
+			}
+		}
+	}
+	return count, forwarded
+}
+
+// TestResponsesStream_MessageStart_NoDuplicateOnPassthrough guards the fix for the
+// duplicate message_start emitted on the Anthropic passthrough responses_stream
+// path. A single upstream message_start expands inbound into [response.created,
+// response.in_progress]. The raw upstream frame must ride on response.created
+// (which maps back to message_start) so it is forwarded once with all upstream
+// fields intact; response.in_progress must carry no raw and convert to nil.
+// Attaching the raw to in_progress instead (the old "last chunk" rule) produced
+// two message_start frames: a lossy synthesized one from created plus the
+// raw-forwarded one from in_progress.
+func TestResponsesStream_MessageStart_NoDuplicateOnPassthrough(t *testing.T) {
+	ctx := schemas.NewBifrostContext(nil, time.Time{})
+
+	const rawMessageStart = `{"type":"message_start","message":{"id":"msg_01","type":"message","role":"assistant","model":"claude-opus-4-8","content":[],"stop_reason":null,"usage":{"input_tokens":3450,"output_tokens":4,"service_tier":"standard","inference_geo":"not_available"}}}`
+
+	var chunk AnthropicStreamEvent
+	if err := sonic.Unmarshal([]byte(rawMessageStart), &chunk); err != nil {
+		t.Fatalf("unmarshal message_start: %v", err)
+	}
+
+	state := acquireAnthropicResponsesStreamState()
+	defer releaseAnthropicResponsesStreamState(state)
+
+	responses, bErr, _ := chunk.ToBifrostResponsesStream(ctx, 0, state)
+	if bErr != nil {
+		t.Fatalf("ToBifrostResponsesStream error: %v", bErr)
+	}
+
+	// Inbound must expand the single message_start into [created, in_progress].
+	if len(responses) != 2 {
+		t.Fatalf("expected 2 bifrost responses (created, in_progress), got %d", len(responses))
+	}
+	createdIdx := -1
+	for i, r := range responses {
+		if r.Type == schemas.ResponsesStreamResponseTypeCreated {
+			createdIdx = i
+		}
+	}
+	if createdIdx == -1 {
+		t.Fatal("expected a response.created in the message_start expansion")
+	}
+	// The fix attaches the raw to response.created; the old rule attached it to the
+	// last response (in_progress). Confirm they differ so the cases below exercise it.
+	if createdIdx == len(responses)-1 {
+		t.Fatal("response.created is last; raw attachment would not exercise the fix")
+	}
+
+	t.Run("raw on created emits one full message_start", func(t *testing.T) {
+		for _, r := range responses {
+			r.ExtraFields.RawResponse = nil
+		}
+		responses[createdIdx].ExtraFields.RawResponse = rawMessageStart
+
+		count, forwarded := countPassthroughMessageStarts(ctx, responses)
+		if count != 1 {
+			t.Fatalf("expected exactly 1 message_start, got %d", count)
+		}
+		// The surviving frame must be the verbatim upstream bytes, preserving
+		// service_tier/inference_geo that the synthesized frame drops.
+		if forwarded != rawMessageStart {
+			t.Errorf("forwarded message_start is not the verbatim upstream frame: %s", forwarded)
+		}
+	})
+
+	t.Run("raw on in_progress (old behavior) duplicates message_start", func(t *testing.T) {
+		for _, r := range responses {
+			r.ExtraFields.RawResponse = nil
+		}
+		responses[len(responses)-1].ExtraFields.RawResponse = rawMessageStart
+
+		count, _ := countPassthroughMessageStarts(ctx, responses)
+		if count != 2 {
+			t.Fatalf("expected the old last-chunk rule to duplicate message_start (2), got %d", count)
+		}
+	})
+}
--- a/core/providers/anthropic/anthropic.go
+++ b/core/providers/anthropic/anthropic.go
@@ -1330,6 +1330,16 @@ func HandleAnthropicResponsesStream(
 					responseChan, postHookSpanFinalizer)
 				continue
 			}
+			// Attach the upstream raw to exactly one bifrost response. Default to the last,
+			// but for the message_start expansion ([created, in_progress]) attach it to
+			// response.created
+			rawIdx := len(responses) - 1
+			for j, r := range responses {
+				if r != nil && r.Type == schemas.ResponsesStreamResponseTypeCreated {
+					rawIdx = j
+					break
+				}
+			}
 			// Handle each response in the slice
 			for i, response := range responses {
 				if response != nil {
@@ -1347,8 +1357,7 @@ func HandleAnthropicResponsesStream(
 					lastChunkTime = time.Now()
 					chunkIndex++
 
-					// Only add raw response to the last chunk of the incoming event
-					if providerUtils.ShouldSendBackRawResponse(ctx, sendBackRawResponse) && i == len(responses)-1 {
+					if providerUtils.ShouldSendBackRawResponse(ctx, sendBackRawResponse) && i == rawIdx {
 						response.ExtraFields.RawResponse = eventData
 					}
 


TejasGhatte:
* **#4559** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/4559?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/maximhq/bifrost/4559?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `dev`


<h2></h2>

akshaydeo:
### Merge activity

* **Jun 19, 10:30 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/4559).
* **Jun 19, 10:30 AM UTC**: @akshaydeo merged this pull request with [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/4559).

PR #4563, opened 2026-06-19
fix: cache compiled regexps in CORS wildcard origin matching

matiasinsaurralde:
`matchesWildcardPattern` was calling `regexp.Compile` on every HTTP request for each wildcard pattern in `AllowedOrigins`. Add a `sync.Map` cache keyed by the raw pattern string so each regexp is compiled once and reused.

Benchmarks show ~36x speedup sequential, ~350x under concurrency, with zero allocations on the hot path.

Also adds unit tests for `matchesWildcardPattern` and `IsOriginAllowed`.

## Summary

`matchesWildcardPattern` compiled a new `*regexp.Regexp` on every call — which happens on every HTTP request through the CORS middleware, plus every WebSocket upgrade. Since the wildcard patterns come from static config (`AllowedOrigins`), the same patterns were recompiled thousands of times per second for no reason.

This adds a `sync.Map` cache so each pattern is compiled once on first encounter and reused thereafter. This follows the same pattern already used by `celMapKeyRegexCache` in `plugins/governance/routing.go` for CEL map-key regex caching.

## Changes

- Added `wildcardRegexpCache sync.Map` in `handlers/utils.go` with cache-first lookup in `matchesWildcardPattern`
- Added `"sync"` import
- Added `handlers/wildcard_test.go` with 26 test cases covering `matchesWildcardPattern` and `IsOriginAllowed`

### Benchmark (benchstat, 8 rounds, main vs fix)

| Benchmark | main | fix | delta |
|-----------|-----:|----:|------:|
| `MatchesWildcardPattern` | 11,595 ns | 320 ns | **-97.24%** (36x) |
| `MatchesWildcardPattern_Parallel` | 3,501 ns | 10 ns | **-99.71%** (350x) |
| `IsOriginAllowed_WithWildcards` | 15,278 ns | 527 ns | **-96.55%** (29x) |
| `IsOriginAllowed_WithWildcards_Parallel` | 4,080 ns | 23 ns | **-99.44%** (178x) |
| `IsOriginAllowed_ExactOnly` | 24 ns | 24 ns | ~ (no regression) |

Memory drops to **0 B/op, 0 allocs/op** across all wildcard benchmarks (from 27-36 KiB and 340-422 allocs/op on main).

Benchmark:
[wildcard_bench_test.txt](https://github.com/user-attachments/files/29145796/wildcard_bench_test.txt)

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [ ] Core (Go)
- [x] Transports (HTTP)
- [ ] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

```sh
cd transports/bifrost-http
go test ./handlers/... -run='TestMatchesWildcard|TestIsOriginAllowed' -v


akshaydeo:
@matiasinsaurralde ❤️ for a kickass PR - could you please rebase it with dev?

matiasinsaurralde:
> @matiasinsaurralde ❤️ for a kickass PR - could you please rebase it with dev?

Just rebased, should be good now

akshaydeo (approved):

PR #4582, opened 2026-06-21
feat: add per-provider disable_model_discovery to skip live model discovery

### LINKED ISSUE #4581: Live model discovery on /v1/models leaks connections (Azure) with no opt-out for built-in providers
## Summary

`GET /v1/models` triggers live model discovery against **every** configured provider. For the built-in **Azure** provider with multiple regional keys, those per-key list-models calls open upstream connections that accumulate and don't drain promptly, so repeated `/v1/models` requests climb the gateway's open-FD count until the endpoint hangs — and can starve inference too. Critically, there is **no configuration option to disable live model discovery for a built-in provider.**

## Environment

- Bifrost **v1.5.15**, HTTP transport (`maximhq/bifrost:v1.5.15`)
- Providers: `azure` (9 regional keys, each with a per-key `azure_key_config` endpoint), `openrouter`, and a custom `cohere` provider (base type `cohere`)
- ConfigStore: SQLite (enabled); governance plugin enabled, `is_vk_mandatory: true`

## Observed behaviour

With a valid virtual key:

- A single `GET /v1/models` returns `200` but with **highly variable latency** — ~0.5s sometimes, 11–31s other times — and under repeated calls it times out entirely. Even **spaced** calls (one every ~2.5s) degrade: first one or two are fast, then they all time out.
- Bifrost's own logs show `request_duration_ms` reaching **23000–31000** for `/v1/models` while still returning 200.
- The gateway's **open file descriptors climb from ~19 to ~155 after 8 `/v1/models` calls.**
- During a slow/hung call, the gateway holds **~120 ESTABLISHED sockets to Azure IPs** (`20.x`, `51.12.x`) spread across the 9 regional endpoints; they don't drain quickly.
- Each provider endpoint is fast **individually** (OpenRouter `/v1/models` ~0.1s; the Azure endpoints answer in <0.6s), so the problem is the aggregation / per-key connection handling, not a slow upstream.
- Inference (`/v1/chat/completions`) is always fast and reliable — it never enters this path.

## Root cause (from reading the source)

- On startup and on each `/v1/models`, the server fans out `RefreshLiveModelsForProvider` → `FetchAndStoreLiveForKey` for every provider/key (`transports/bifrost-http/server/server.go`). `FetchAndStoreLiveForKey` issues a filtered + unfiltered `ListModelsRequest` per `(provider, key)`. With 9 Azure keys this is a wide fan-out and the connections pile up.
- There **is** a skip path: `FetchAndStoreLiveForKey` early-returns when
  `pc.CustomProviderConfig != nil && !pc.CustomProviderConfig.IsOperationAllowed(ListModelsRequest)`
  — i.e. a provider whose `custom_provider_config.allowed_requests` omits `list_models`. This is exactly why our custom `cohere` provider never participates and never leaks.
- **But that skip is only reachable for providers that can carry a `custom_provider_config`**, which requires `base_provider_type ∈ SupportedBaseProviders = [Anthropic, Bedrock, Cohere, Gemini, OpenAI, HuggingFace, Replicate]` (`core/schemas/bifrost.go`). **Azure and OpenRouter are not in that set**, so they can't be wrapped, and there is no `allowed_requests`/list-models opt-out for built-in providers — they always do live discovery.
- The model catalog is initialized unconditionally (`modelcatalog.Init` in `transports/bifrost-http/lib/config.go`) with no enable/disable flag; turning off `config_store` is the only thing that changes catalog behaviour, but it also disables auth, so it's not a viable opt-out.

## Impact

- `/v1/models` is unreliable for any setup with several Azure regional keys.
- Because the sockets accumulate rather than drain, an OpenAI-compatible client that polls `/v1/models` (many do, on startup/refresh) can exhaust the gateway's FDs and **stall inference for everyone**, not just the model-list endpoint.

## Suggested fixes (any one would resolve it)

1. A global opt-out — config field or env var (e.g. `BIFROST_DISABLE_LIVE_MODEL_DISCOVERY=true`) that skips `FetchAndStoreLiveForKey` and serves `/v1/models` from the statically configured models.
2. Allow a list-models opt-out for **built-in** providers too (e.g. a provider-level `disable_model_discovery: true`, or honoring `allowed_requests` on standard providers), so Azure/OpenRouter can opt out the way custom providers can.
3. Fix the leak in the per-key live list-models path so connections are closed/drained promptly even when a provider has many keys.

## Current workaround

For affected built-in providers there is no config workaround (you can't attach a `custom_provider_config`). The only mitigations are to avoid `/v1/models` entirely (point clients at a static model list) or patch the source to early-return in `FetchAndStoreLiveForKey`.

ragokan:
## Summary

`GET /v1/models` triggers live model discovery (list-models) against **every** configured provider. For built-in providers like Azure with many regional keys this fans out per key, and the connections can pile up until `/v1/models` hangs (and can starve inference). Built-in providers can't opt out via `custom_provider_config.allowed_requests` because they aren't in `SupportedBaseProviders`, so there was previously **no way to disable discovery for them**. Full investigation in #4581.

This adds a per-provider `disable_model_discovery` flag. When set, `FetchAndStoreLiveForKey` early-returns, so `/v1/models` serves only the statically configured models for that provider and the per-key list-models fan-out is skipped. It composes with the existing `allowed_requests` skip used by custom providers.

This implements option 2 from #4581 (a per-provider opt-out for built-in providers), not the global env var. Default behaviour is unchanged — discovery stays on unless explicitly disabled.

## Changes

- **`core/schemas/provider.go`** — add `ProviderConfig.DisableModelDiscovery` (`disable_model_discovery`, default `false`).
- **`transports/bifrost-http/server/server.go`** — `FetchAndStoreLiveForKey` skips the live fetch when the flag is set, alongside the existing custom-provider `allowed_requests` skip.
- **`framework/configstore`** — persist the field end-to-end: the configstore `ProviderConfig`, the `providers` table column, `GenerateConfigHash`, the load/save conversions, and an `add_disable_model_discovery_column` migration. Because the field defaults to `false` (which writes nothing to the config hash), existing rows keep a valid hash and the migration only adds the column — **no backfill needed**.
- **`transports/config.schema.json`** — document the field on every provider block.
- **`transports/schema_test/config_schema_test.go`** — assert the schema documents it as a boolean on every provider block.
- **changelogs** — `core/`, `framework/`, `transports/`.

## Type of change

- [x] Feature

## Affected areas

- [x] Core (Go)
- [x] Transports (HTTP)
- [x] Providers/Integrations

## How to test

```sh
go version   # 1.26
# builds (go.work spans the modules)
go build ./core/... ./framework/... ./transports/...
# schema documents the field on every provider block
go test ./transports/schema_test/ -run TestSchemaDisableModelDiscovery
# configstore hash + migrations still green
go test ./framework/configstore/ -run 'Hash|Migration'
# schema <-> Go validators
bash .github/workflows/scripts/validate-schema-sync.sh
bash .github/workflows/scripts/validate-go-config-fields.sh
```

Runtime: set `disable_model_discovery: true` on a provider in `config.json`. `GET /v1/models` then returns only that provider's statically configured models, and the gateway no longer opens list-models connections for it on startup or per request.

## Breaking changes

- [x] No

Defaults to `false`; existing configs and stored `config_hash` values are unaffected.

## Related issues

Closes #4581

## Security considerations

None. The flag only suppresses outbound list-models calls; it doesn't touch auth, secrets, or request handling (and can reduce outbound connection pressure).

## Checklist

- [x] I read the contributing guidelines and followed them (`docs/contributing/raising-a-pr.mdx`)
- [x] I updated documentation where needed (config schema + changelogs)
- [x] I verified builds succeed (Go; UI unaffected)
- [x] I verified the schema-sync, go-config-fields validators and configstore tests pass locally


akshaydeo:
Hey not the right solution again - we will take over it

akshaydeo (changes requested):
This is not the right way - we just need to use cached table and trigger a background sync to make sure they are in the right state. 

This does not require these many changes

PR #4593, opened 2026-06-21
fix(anthropic): keep mid-conversation system inline for custom Anthropic-base providers (GLM, Kimi, ...) to preserve prompt cache

### LINKED ISSUE #4592: Custom Anthropic-base providers (sglang/vLLM: GLM, Kimi, ...) hoist mid-conversation system messages, breaking prompt cache
## Summary

On **custom Anthropic-base providers** (`base_provider_type: anthropic` pointed at a self-hosted Anthropic-compatible engine such as sglang/SGLang, vLLM, TGI, or llama.cpp), prompt-cache reads collapse for multi-turn agentic conversations (Claude Code / any client that emits `role: "system"` messages mid-conversation).

This is the **custom-provider sibling of #4068** (which is Bedrock-specific, fixed by #4534). It applies to **any self-hosted model** behind a custom Anthropic-base provider -- e.g. GLM (`glm-5` on sglang) and Kimi (`Kimi-K2` on vLLM) -- not one model in particular.

## Root cause

`SupportsMidConversationSystem(provider, model)` (`core/providers/anthropic/utils.go`) gates whether a mid-conversation `role:system` message stays **inline** at its position or is **hoisted** into the top-level `system` block via `appendToSystemContent`. Both the chat and responses inbound converters call this gate.

```go
func SupportsMidConversationSystem(provider schemas.ModelProvider, model string) bool {
	if provider != schemas.Anthropic {   // (1) a custom provider key fails here
		return false
	}
	m := strings.ToLower(model)
	if IsFableFamily(m) {
		return true
	}
	return strings.Contains(m, "opus") &&   // (2) and non-opus-4.8 models fail here
		(strings.Contains(m, "4-8") || strings.Contains(m, "4.8"))
}
```

For a custom provider, `GetProviderName` returns the **operator-chosen custom key** (e.g. `amd_qre_001`, `virtuaireason`), not `schemas.Anthropic`, so guard (1) returns `false` immediately. (Even setting that aside, the served model -- `glm-5`, `Kimi-K2` -- is neither opus-4.8 nor Fable, so guard (2) also fails.) Result: **every** mid-conversation `role:system` message is hoisted to the front of the prompt.

## Why it breaks caching

Self-hosted engines use a **prefix/radix KV cache** (sglang RadixAttention, vLLM automatic prefix caching, ...): a hit is the longest common prefix from position 0. The top-level `system` block sits in front of the entire conversation + tools + history.

Claude Code emits transient `role:system` reminders at their chronological position (`<system-reminder>`, `<new-diagnostics>`, "task tools haven't been used recently", skill lists, file-change notes). Each new one, once hoisted:

1. grows the top-level `system` block,
2. changes the serialized prefix in front of the whole conversation,
3. forks the radix tree near position 0 -> the prefix match fails and the entire cacheable tail (tools + all history) is stranded and re-prefilled.

These engines **do** render `role:system` inline at any position (their chat templates emit a system turn mid-stream), so keeping the reminder inline is both render-legal and cache-preserving. An operator who set `base_provider_type: anthropic` is asserting the endpoint speaks the Anthropic Messages format, which natively includes mid-conversation system messages.

## Evidence (natural experiment)

Same Claude Code traffic, same Bifrost, only the **provider transform** differs. Placement of the "task tools haven't been used" reminder:

| reminder lands in... | custom anthropic-base provider (glm-5, unpatched) | Bedrock (after #4534) |
|---|---|---|
| top-level `system` block (cache-breaking hoist) | **98%** (n=17,913) | **0%** |
| inline as a `user` turn (cache-safe) | 1% | **100%** (n=22,350) |

The provider that keeps the reminder inline sustains ~98% cached-read; the one that hoists it sits around ~89% with a 5x excess of partial-hit requests.

## Fix direction

A custom provider key (not one of Bifrost's built-in providers) reaching the Anthropic converter exists only because the operator chose an Anthropic-compatible base for a self-hosted engine. Treat that as mid-conversation-system supported by default. A PR (#4593) does this: a non-standard provider key keeps mid-conv `role:system` inline; built-in providers (Bedrock/Vertex/Azure/standard SGL/OpenAI) keep their historical model-based gating. Open to a per-custom-provider `supports_mid_conversation_system` config flag instead if maintainers prefer an explicit operator opt-out for an engine whose template cannot render mid-stream system.

Note: this is the inverse policy of #4276, which hoists mid-conv system for "unsupported" models. Hoisting is correct only for engines that **cannot** render `role:system` inline; for engines that can (sglang/vLLM/TGI Anthropic-compatible), hoisting is what breaks the cache. The `SupportsMidConversationSystem` gate is exactly where that decision belongs.

## Environment
- Bifrost: 1.5.16 line
- Providers: custom, `base_provider_type: anthropic`, pointed at sglang (`glm-5`) and vLLM (`Kimi-K2`) engines
- Client: Claude Code

mickgvirtu:
## Summary

Fixes #4592. Keeps mid-conversation `role:system` messages **inline** for **any custom Anthropic-base provider**, instead of hoisting them into the top-level `system` block.

This is the custom-provider sibling of #4534 (Bedrock). A custom provider whose key is not one of Bifrost's built-in providers reaches the Anthropic converter only because the operator set `base_provider_type: anthropic` and pointed it at a self-hosted Anthropic-compatible engine (sglang, vLLM, TGI, llama.cpp). The gate `SupportsMidConversationSystem` returned `false` for these (the custom key is not `schemas.Anthropic`, and the served model -- `glm-5`, `Kimi-K2`, ... -- is neither opus-4.8 nor Fable), so Claude Code's per-turn reminders were hoisted every turn, forking the engine's prefix/radix cache and stranding the cacheable tail (tools + history).

This covers GLM, Kimi, and any other self-hosted Anthropic-compatible model in one rule rather than per-model special-casing.

## Changes

- `SupportsMidConversationSystem`: a non-standard (custom) provider key reaching the Anthropic converter now returns `true` -- the operator chose an Anthropic-compatible base, whose engines render `role:system` inline. Built-in providers (Bedrock, Vertex, Azure, standard SGL/OpenAI) fall through to the existing model-based opus-4.8/Fable gate, so their behavior is unchanged.
- New `schemas.IsStandardProvider` (built from `StandardProviders`) so the gate can distinguish a custom provider key from a built-in one without an import cycle.

## How to test

`go test ./core/providers/anthropic/ -run "TestSupportsMidConversationSystem|MidConversation"` -- `TestSupportsMidConversationSystem` adds: custom keys (GLM under `amd_qre_001`, Kimi under `virtuaireason`, a generic custom key) return `true` regardless of model; built-in `schemas.SGL`/`schemas.OpenAI` with a self-hosted model name still return `false`; existing Anthropic/opus-4.8/Fable/Bedrock/Vertex cases unchanged.

## Evidence

#4592 has the full root-cause and the natural-experiment data: on identical Claude Code traffic, the "task tools haven't been used" reminder lands in the top-level `system` block 98% of the time on the unpatched custom-provider path vs 0% on Bedrock after #4534 (where it lands inline as a `user` turn 100% of the time, sustaining ~98% cached-read).

## Notes for reviewers

- The signal is "provider key is not a built-in provider" -- i.e. a custom provider configured with an Anthropic base. Happy to rework this into an explicit per-custom-provider `supports_mid_conversation_system` config flag (default on) if you'd prefer an operator-visible opt-out for a backend whose chat template cannot render mid-stream system.
- Inverse policy to #4276: hoisting is correct only for engines that cannot render `role:system` inline; this gate keeps it inline for those that can. Glad to coordinate the two so they don't regress each other.

### DIFF
--- a/core/changelog.md
+++ b/core/changelog.md
@@ -1,0 +1,1 @@
+- fix: deterministic MCP tool ordering for prompt cache stability (closes #2347)
--- a/core/mcp/toolmanager.go
+++ b/core/mcp/toolmanager.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"slices"
 	"strings"
 	"sync/atomic"
 	"time"
@@ -214,7 +215,15 @@ func (m *ToolsManager) GetAvailableTools(ctx *schemas.BifrostContext) []schemas.
 	// Track tool names to prevent duplicates
 	seenToolNames := make(map[string]bool)
 
-	for clientName, clientTools := range availableToolsPerClient {
+	// Sort client names for deterministic tool ordering
+	sortedClients := make([]string, 0, len(availableToolsPerClient))
+	for clientName := range availableToolsPerClient {
+		sortedClients = append(sortedClients, clientName)
+	}
+	slices.Sort(sortedClients)
+
+	for _, clientName := range sortedClients {
+		clientTools := availableToolsPerClient[clientName]
 		client := m.clientManager.GetClientByName(clientName)
 		if client == nil {
 			m.logger.Warn("%s Client %s not found, skipping", MCPLogPrefix, clientName)
--- a/core/mcp/utils.go
+++ b/core/mcp/utils.go
@@ -71,10 +71,19 @@ func (m *MCPManager) GetToolPerClient(ctx context.Context) map[string][]schemas.
 
 	m.logger.Debug("%s GetToolPerClient: Total clients in manager: %d, Filter: %v", MCPLogPrefix, len(m.clientMap), includeClients)
 
-	tools := make(map[string][]schemas.ChatTool)
+	// Collect and sort client names for deterministic tool ordering
+	clientNames := make([]string, 0, len(m.clientMap))
+	clientsByName := make(map[string]*schemas.MCPClientState, len(m.clientMap))
 	for _, client := range m.clientMap {
-		// Use client name as the key (not ID)
-		clientName := client.ExecutionConfig.Name
+		name := client.ExecutionConfig.Name
+		clientNames = append(clientNames, name)
+		clientsByName[name] = client
+	}
+	slices.Sort(clientNames)
+
+	tools := make(map[string][]schemas.ChatTool)
+	for _, clientName := range clientNames {
+		client := clientsByName[clientName]
 		clientID := client.ExecutionConfig.ID
 
 		m.logger.Debug("%s Evaluating client %s (ID: %s) for tools", MCPLogPrefix, clientName, clientID)
@@ -91,12 +100,19 @@ func (m *MCPManager) GetToolPerClient(ctx context.Context) map[string][]schemas.
 			continue
 		}
 
-		// Add all tools from this client
+		// Collect and sort tool names for deterministic ordering
+		toolNames := make([]string, 0, len(client.ToolMap))
+		for toolName := range client.ToolMap {
+			toolNames = append(toolNames, toolName)
+		}
+		slices.Sort(toolNames)
+
 		// FILTERING HIERARCHY (restrictive, not permissive):
 		// 1. Client-level configuration (ToolsToExecute) - Global allow-list, most restrictive
 		// 2. Request context (MCPContextKeyIncludeTools) - Can only further narrow, not expand
 		// Context filtering CANNOT override client configuration - it can only be more restrictive.
-		for toolName, tool := range client.ToolMap {
+		for _, toolName := range toolNames {
+			tool := client.ToolMap[toolName]
 			// First check: Client configuration is the global allow-list
 			// If client config blocks a tool, it CANNOT be overridden by context
 			if shouldSkipToolForConfig(toolName, client.ExecutionConfig) {
--- a/core/providers/anthropic/utils.go
+++ b/core/providers/anthropic/utils.go
@@ -840,13 +840,31 @@ func appendToSystemContent(existing *AnthropicContent, newContent AnthropicConte
 
 // SupportsMidConversationSystem returns true if the provider+model combination
 // supports role:"system" entries inside the messages array (mid-conversation
-// system messages). Available on the Anthropic API only — not on Bedrock or
-// Vertex. Supported on Claude Opus 4.8+ and the Claude Fable/Mythos family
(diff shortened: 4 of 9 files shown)


mickgvirtu:
## Evidence + simulation results

Adding the data behind this, since the change is a behavior shift on a hot path.

### Natural experiment (proves the hoist is Bifrost-side, not the client)

Same Claude Code traffic, same Bifrost, only the provider transform differs. Placement of the "task tools haven't been used" reminder in the post-Bifrost request:

| reminder lands in... | GLM (custom anthropic-base, unpatched) | Bedrock/Claude (analogous inline fix applied, #4534) |
|---|---|---|
| top-level `system` block (cache-breaking hoist) | **99.0%** (n=17,913) | **0.5%** |
| inline as a `user` turn (cache-safe) | 1.0% | **99.5%** (n=22,350) |

Identical inbound, opposite placement, split purely by the provider path.

### Simulation (faithful token-exact sglang RadixCache model)

We replayed real captured traffic through a transcription of sglang's RadixCache (edge-keyed patricia, page_size=1, LRU by last_access, L1 cap = real `max_total_num_tokens`). It reproduces the engine's real ~82-89% hit at real capacity (validation anchor). The fix is modeled as relocating the hoisted system-block reminders back inline:

| workload | base (hoisted) | fix (inline) | delta |
|---|---|---|---|
| churning session (Bifrost-hoisted reminders) | 88.2% | **94.0%** | +5.8pt |
| session w/ reminders already inline | 97.5% | 97.4% | no regression |
| combined (char-weighted) | 96.6% | 96.6% | - |

The post-fix ceiling on the churning session (~94%) is bounded by genuinely-new content (new user/assistant text + large tool-output JSON), not by the cache.

### Distributional signature

GLM shows **21.8%** of requests in the 5-90% partial-hit band vs Bedrock's **4.4%** -- a 5x excess of partial-hit requests, which is exactly the gap. ~91% of stranded tokens are system-region breaks.

Methodology: faithful tokenizer (real GLM tokenizer + `chat_template.jinja`, rendered token count matches the engine's logged `input_tokens` to within 0.1% median; rendered head matches logged `input_ids` token-for-token) feeding a Rust radix model. cache_control/cachePoint breakpoints were ruled out as a lever (they only choose where to STORE prefix checkpoints; KV causality forbids off-prefix reuse), as were cache size (infinite-cache ceiling ~91-92%, LRU within 0.6%) and DP routing (sticky, 0/87 sessions bounced).

PR #4594, opened 2026-06-21
fix(codemode): stop duplicating MCP tool raw response into model-visible logs

### LINKED ISSUE #4434: [Bug]: Code mode echoes every tool's full raw response into executeToolCode output, duplicating the returned value (~2x token usage)
### Prerequisites

### Description

When an MCP client is configured with `is_code_mode_client: true`, the code-mode executor (`executeToolCode`) writes the **full raw response of every proxied tool call** into the script's captured output (the `Print output` block returned to the caller), *in addition to* the script's structured `Return value`.

The script never requests this — the gateway injects it automatically after each tool call. The compiled format strings are:

```
[TOOL] %s.%s raw response: %s
[TOOL] %s.%s error result: %s
[TOOL] %s.%s error: %v
```

Effect: every byte of every tool's raw payload is sent to the model **twice** — once in `Print output` (the auto-injected `[TOOL] ... raw response:` line) and once in `Return value` (whatever the script assigns to `result`).

This defeats the core value proposition of code mode. The point of running a script in the sandbox is to *reduce* tool output (filter / count / extract) so the model only receives the distilled `result`. But the raw responses are echoed verbatim regardless, so a script that deliberately returns a 50-token summary still ships the model the entire multi-kilobyte raw payload of each underlying call. On multi-call scripts this roughly doubles (or worse) the token cost of the result.

It is also not gated by log level — it appears at `-log-level info`, because it is written to the script's captured output sink rather than to the leveled logger.

Related but distinct: maximhq/bifrost#2603 ("Allow disabling code mode tools injection") concerns the meta-tools' *schema* being injected into the tool list. This issue is about runtime duplication of tool *outputs* in the `executeToolCode` result, and applies even when code-mode tools are the intended path.

* Every executeToolCode result also appends a fixed block, on success as well as on error:

```
Environment:
Available server keys: <every connected server name>
Note: This is a Starlark (Python subset) environment.
```

* This is appended to every request and should be optional.

### Steps to reproduce

1. Configure an MCP client with `is_code_mode_client: true`, fronting an upstream server whose tools return sizable payloads.
2. Call `executeToolCode` with a script that calls one or more tools and returns only a reduced value:

   ```python
   r = SomeServer.some_tool(param="x")
   result = {"count": len(r["items"])}   # intentionally tiny
   ```
3. Inspect the `executeToolCode` result.

### Expected behavior

The caller/model-visible output should contain only what the script explicitly prints plus the `Return value`. The raw tool response should not be unconditionally dumped into the returned output. Reasonable options:

* Route the `[TOOL] ... raw response: ...` trace to the **server-side logger** only (gated by log level or a debug flag), not to the script's captured stdout that is returned to the model.
* Or make it opt-in via a config flag / request header.
* Or skip the echo for calls whose result is already captured/returned by the script.

### Actual behavior

For every tool call, the `Print output` block contains a `[TOOL] <server>.<tool> raw response: <ENTIRE raw response>` line, and the same data also appears in `Return value`. The full raw payload reaches the model even when the script returns only a small derived value — roughly doubling token usage on the tool result.

### Affected area(s)

Core (Go) — MCP gateway / code mode (`executeToolCode`)

### Version

v1.5.13 (HTTP, as reported by `GET /api/version`). Self-hosted Docker image built 2026-06-12 with code mode enabled.

### Environment

* Bifrost: HTTP v1.5.13 (self-hosted Docker)
* Deployment: Docker on Linux
* MCP gateway with `is_code_mode_client: true`; multiple stdio MCP servers behind the gateway

### Relevant logs/output

Real captured example (a `gh issue list` call made through code mode). The raw response appears in `Print output` and again, verbatim, in `Return value`:

```
Print output:
[TOOL] GitHub.gh_issue_list raw response: "2603  OPEN  [Feature]: Allow disabling code mode tools injection ... <full result> ..."

Execution completed successfully.
Return value: "2603  OPEN  [Feature]: Allow disabling code mode tools injection ... <same full result, verbatim> ..."
```

### Severity

Medium (functionality works, but it ~doubles token usage on every code-mode tool result and negates code mode's data-reduction benefit).

vjsai:
## Summary

Fixes #4434. In the Starlark code-mode executor, the raw MCP tool response was sent to the model **twice** — once as the tool's return value and again inside the model-visible execution logs — roughly doubling token cost and defeating code mode's data-reduction purpose.

In `core/mcp/codemode/starlark/executecode.go`, `callMCPTool` did:

```go
resultStr := formatResultForLog(rawResult)
logToolName := strings.ReplaceAll(effectiveToolName, "-", "_")
appendLog(fmt.Sprintf("[TOOL] %s.%s raw response: %s", clientName, logToolName, resultStr))
```

`appendLog` feeds `ExecutionResult.Logs`, which `handleExecuteToolCode` renders into the model-visible `Print output:` block. The same `rawResult` is already returned as the tool's `Return value` (`createToolResponseMessage` immediately below), so the full payload reached the model in both places.

## Changes

- Route the `[TOOL] ... raw response` trace to the server-side debug logger (`s.logger.Debug`) instead of `appendLog`. The data still reaches the model once, as the tool's return value; the raw dump remains available for server-side debugging.

## Type of change

- [x] Bug fix

## Affected areas

- [x] Core (Go)

## How to test

Added `TestCallMCPToolDoesNotEchoRawResponseIntoLogs` in `core/mcp/codemode/starlark/starlark_test.go`. It stands up an in-process MCP server whose tool echoes a known payload, drives `callMCPTool`, and asserts the payload appears in the **return value** but not in the appended logs.

```
cd core && go test ./mcp/codemode/starlark/ -run TestCallMCPToolDoesNotEchoRawResponseIntoLogs -v
```

- **Before:** fails — `raw tool response leaked into execution logs: [TOOL] testserver.echo raw response: "ECHO_PAYLOAD_b3f19a27"`
- **After:** passes. Full `go test ./mcp/codemode/starlark/` and `go vet` pass.

---
*Prepared with AI assistance (Claude Code); the change was reviewed, built, and tested by a human before submitting.*


Pratham-Mishra04:
Hey @vjsai can you rebase the pr once?

Pratham-Mishra04:
Hey @vjsai can you pls rebase this pr once

akshaydeo:
Hi @vjsai — thanks for the contribution! This PR is currently blocked because our CLA bot shows the Contributor License Agreement as not yet signed. Could you sign it here so we can move this forward: https://cla-assistant.io/maximhq/bifrost?pullRequest=4594

PR #4612, opened 2026-06-22
fix(responses): deep-copy author, recipient, encrypted_content in message copies

### LINKED ISSUE #4608: ResponsesMessage drops author, recipient, encrypted_content fields — breaks Codex multi_agent_v2 subagent spawning
### Summary

Bifrost silently drops three fields from Responses API `input` items during deserialization/re-serialization. This completely breaks Codex 0.141.0+ when `multi_agent_v2 = true` is enabled, causing every request after the first `collab_tool_call` to fail with a 400.

### Affected fields

| Field | Location in request | Error |
|---|---|---|
| `author` | `input[N]` (top-level on `collab_tool_call` items) | `Missing required parameter: 'input[N].author'` |
| `recipient` | `input[N]` (top-level on `collab_tool_call` items) | `Missing required parameter: 'input[N].recipient'` |
| `encrypted_content` | `input[N].content[0]` (reasoning content blocks) | `Missing required parameter: 'input[N].content[0].encrypted_content'` |

### Root cause

`ResponsesMessage` in `core/schemas/responses.go` has no `Author` or `Recipient` struct fields. `ResponsesMessageContentBlock` has no `EncryptedContent` field. Go's JSON decoder silently discards unknown fields, so these values are stripped on inbound decode and never re-serialized to the upstream provider.

`send_back_raw_request: true` does **not** fix this — it only echoes the raw request in error responses, it does not bypass re-serialization.

### How to reproduce

**Setup:**
- Bifrost deployed as a proxy in front of any OpenAI-compatible Responses API endpoint
- Codex CLI 0.141.0 configured with `multi_agent_v2 = true` routing via bifrost

**Steps:**
```bash
codex exec \
  -c model_provider=bifrost \
  -c model=gpt-5.5 \
  --enable multi_agent_v2 \
  --dangerously-bypass-approvals-and-sandbox \
  "write hello to hello.txt then spawn a subagent to verify it exists"
```

**Expected:** Subagent spawns and returns result, session completes.

**Actual (without fix):**
```
ERROR: {"status_code":400,"error":{"message":"Missing required parameter: 'input[9].author'."}}
ERROR: {"status_code":400,"error":{"message":"Missing required parameter: 'input[10].recipient'."}}
ERROR: {"status_code":400,"error":{"message":"Missing required parameter: 'input[4].content[0].encrypted_content'."}}
```

**Actual (with fix):** Session completes successfully. Subagent spawns, verifies file, reports back.

### Fix

Add the three missing fields to `core/schemas/responses.go`:

```go
// In ResponsesMessage struct:
Author    json.RawMessage `json:"author,omitempty"`
Recipient json.RawMessage `json:"recipient,omitempty"`

// In ResponsesMessageContentBlock struct:
EncryptedContent *string `json:"encrypted_content,omitempty"`
```

`json.RawMessage` is used for `author`/`recipient` to preserve the objects verbatim without coupling bifrost to OpenAI's evolving multi-agent schema shape.

### What Codex sends (collab_tool_call item)

```json
{
  "type": "collab_tool_call",
  "id": "ctc_abc123",
  "author": { "role": "assistant", "name": "orchestrator" },
  "recipient": { "role": "assistant", "name": "subagent-1" },
  "action": { "type": "spawn_agent", ... }
}
```

After bifrost round-trip (without fix): `author` and `recipient` are gone → OpenAI 400.

### Environment

- Bifrost: `maximhq/bifrost:latest`
- Codex: 0.141.0
- Feature flag: `multi_agent_v2 = true`
- Upstream: OpenAI Responses API

Shaik-Sirajuddin:
Fixes #4608

### Follow-up to #4609

#4609 added `Author`/`Recipient` to `ResponsesMessage` and `EncryptedContent` to `ResponsesMessageContentBlock` so Codex `multi_agent_v2` `collab_tool_call` fields survive bifrost's JSON decode/re-encode path. That fix covered the request decode/encode path but **not the deep-copy helpers**.

### Problem

Both deep-copy implementations copy `ResponsesMessage` / `ResponsesMessageContentBlock` field-by-field and were not updated for the new fields, so they get silently stripped whenever a message is deep-copied:

- `framework/streaming/responses.go` — `deepCopyResponsesMessage` + `deepCopyResponsesMessageContentBlock` run for **every streamed output item**. Any plugin that inspects or re-serializes the accumulated streaming response sees `Author`/`Recipient`/`EncryptedContent` absent. As the multi-agent protocol evolves to return `collab_tool_call` output items, the same stripping bug reappears on the response side.
- `core/schemas/utils.go` — `DeepCopyResponsesMessage` + `deepCopyResponsesMessageContentBlock` have the identical gap on the request-side copy.

### Fix

Mirror the new schema fields in both deep-copy implementations:

| Field | Type | Copy method |
|---|---|---|
| `Author` | `json.RawMessage` (`[]byte`) | `append(json.RawMessage(nil), src...)` (nil-guarded, fresh backing array) |
| `Recipient` | `json.RawMessage` (`[]byte`) | same |
| `EncryptedContent` | `*string` | value-copy via deref |
| `Signature` | `*string` | value-copy via deref (sibling reasoning-replay field with the same latent gap) |

Also added the missing `Phase` (`*string`) copy in the `core/schemas/utils.go` helper for parity with the streaming helper.

### Testing

- `go build ./schemas/`, `go vet ./schemas/`, and `go test ./schemas/` pass.
- `go vet ./streaming/` passes.


core/schemas/utils.go
```diff
@@ -1223,11 +1223,26 @@ func DeepCopyResponsesMessage(original ResponsesMessage) ResponsesMessage {
 		copy.Status = &copyStatus
 	}
 
+	if original.Phase != nil {
+		copyPhase := *original.Phase
+		copy.Phase = &copyPhase
```
akshaydeo: just use new(*original.Phase)

Shaik-Sirajuddin: applied

akshaydeo: new please - Ptr is deprecated

Issue #4648, opened 2026-06-23
[Bug]: PUT /api/providers/{provider} silently discards payload.Keys (blocked-models edits revert)


aeciolevy:
### Prerequisites

- [x] I have searched existing issues and discussions to avoid duplicates (none found across 6 search variants for `blacklisted_models`, `updateProvider`, `"blocked models"`, and `"keys are managed separately"`).
- [x] I am using the latest version (verified bug is present on `transports/v1.5.16`, `main`, and `dev` by reading the handler at each ref).

### Description

`PUT /api/providers/{provider}` (`updateProvider` in `transports/bifrost-http/handlers/providers.go`) accepts a `keys` array in its JSON request body but never assigns it to `config.Keys` — instead always assigning `oldConfigRaw.Keys`. Any client that sends key-level fields (e.g. `blacklisted_models`, `weight`, `enabled`, `name`, `models`) through this endpoint gets HTTP 200 but their changes are silently discarded.

Worse: the call **also overwrites the in-memory cache** with `oldConfigRaw.Keys`, so subsequent reads from the list endpoint (`GET /api/providers/{provider}/keys`) return stale data that diverges from the DB.

This was introduced intentionally in #2160 (per the comment `keys are managed separately via /keys endpoints`), but the contract is silently violated — the field is accepted at JSON decode and the response is 200, with no error/warning indicating the field was ignored.

### Steps to reproduce

Setup: Bifrost HTTP transport `v1.5.15` on ECS Fargate, single task, Postgres `config_store` + `logs_store` enabled, governance admin auth enabled. OpenAI provider added via the dashboard with a single API key.

```bash
ADMIN_USER='medeloop-admin'
ADMIN_PWD='<admin-password>'
KEY_ID='<openai-key-uuid>'
HOST='https://your-bifrost-host'

# 1. Read current key state
curl -sS -u "${ADMIN_USER}:${ADMIN_PWD}" \
  "${HOST}/api/providers/openai/keys/${KEY_ID}" | jq '{blacklisted_models, weight}'
# Output: {"blacklisted_models": ["gpt-3.5-turbo"], "weight": 0.8}

# 2. Send a provider-level PUT with NEW key payload
curl -sS -u "${ADMIN_USER}:${ADMIN_PWD}" \
  -X PUT -H 'Content-Type: application/json' \
  -d '{
    "network_config": {
      "default_request_timeout_in_seconds": 30,
[...]

Issue #4640, opened 2026-06-23
[Bug]: provider config key_ids:["*"] not synced to allow_all_keys for existing virtual keys


acarpe:
### Prerequisites

- [x] I have searched existing issues and discussions to avoid duplicates
- [x] I am using the latest version (or have tested against main/nightly)

### Description

With `source_of_truth: "config.json"` and a Postgres `config_store`, editing a virtual key's `provider_configs[].key_ids` to `["*"]` does not flip `allow_all_keys` to `true` in `governance_virtual_key_provider_configs` for a  **virtual key that already exists** in the DB. The column stays `false`.

Because `allow_all_keys = false` with an empty key set means deny-all-keys (deny-by-default), the affected provider config ends up with no usable keys, so inference through that VK/provider fails even though the file declares all keys allowed.

### Root cause (two spots):

1.  `reconcileVirtualKeyAssociations` (`transports/bifrost-http/lib/config.go`) copies `weight`, `allowed_models`, `rate_limit_id`, and `keys` from the file onto the stored provider-config row, but never copies `allow_all_keys` (or `blacklisted_models`). `.Save()`　then writes the stale DB value back. New VKs work (create path uses the file struct directly); only existing VKs updated via reconcile are affected.
2. `GenerateVirtualKeyHash` (`framework/configstore/clientconfig.go`) omits `allow_all_keys` and `blacklisted_models`, so in split mode the change does not even register as a hash mismatch. Under  `source_of_truth=config.json` the forced sync runs but still drops the field via (1).

### Steps to reproduce

1. Run Bifrost with a Postgres `config_store` and `"source_of_truth": "config.json"`.
2. In `config.json`, define a virtual key whose provider config references explicit keys (so the stored `allow_all_keys` is `false`). Start Bifrost so the VK is persisted.
3. Edit `config.json`: change that provider config to `"key_ids": ["*"]`.
4. Restart/redeploy Bifrost.
5. Inspect the DB:
```
  SELECT vk.name, pc.provider, pc.allow_all_keys
  FROM governance_virtual_key_provider_configs pc
  JOIN governance_virtual_keys vk ON vk.id = pc.virtual_key_id;
```

### Expected behavior

 `allow_all_keys = true` for the provider config (key_ids:["*"] = allow all keys), matching config.json.

### Actual behavior

`allow_all_keys` stays `false`; the provider config remains deny-all-keys.

### Affected area(s)

Framework, Transports (HTTP)

### Version

[...]

Issue #4627, opened 2026-06-23
[Bug]: Gemini video reference fields are sent under parameters


vojthor:
### Prerequisites

- [x] I have searched existing issues and discussions to avoid duplicates
- [x] I am using the latest version (or have tested against main/nightly)

### Description

Gemini/Veo video generation requests that include `referenceImages` or `lastFrame` are converted into a request body where those fields are placed under top-level `parameters`.

For Vertex/Gemini video REST generation, these fields need to be on the first instance instead:

```json
{
  "instances": [
    {
      "prompt": "use the provided reference image",
      "referenceImages": [
        {
          "image": {
            "bytesBase64Encoded": "...",
            "mimeType": "image/jpeg"
          },
          "referenceType": "asset"
        }
      ],
      "lastFrame": {
        "bytesBase64Encoded": "...",
        "mimeType": "image/jpeg"
      }
    }
  ],
  "parameters": {
    "durationSeconds": 8,
    "aspectRatio": "16:9"
  }
}
```

When Bifrost sends `referenceImages` under `parameters`, Vertex appears to ignore them as subject/reference inputs. The generated video does not preserve the referenced subject, while the same request shape sent through direct Vertex/native passthrough with `instances[0].referenceImages` does use the reference image.

### Steps to reproduce

1. Configure Bifrost with a Gemini or Vertex provider that can run a Veo 3.1 video model.
2. Send a video generation request through Bifrost with a prompt and `extra_params.referenceImages`, for example:

   ```json
   {
     "model": "veo-3.1-generate-001",
     "input": {
       "prompt": "use the provided reference image"
     },
     "params": {
       "seconds": "8",
       "extra_params": {
         "referenceImages": [
           {
             "image": {
               "bytesBase64Encoded": "<base64 image>",
               "mimeType": "image/jpeg"
             },
             "referenceType": "asset"
           }
         ]
       }
     }
   }
   ```

3. Inspect the outbound Gemini/Vertex video REST payload generated by `ToGeminiVideoGenerationRequest`.
4. Observe that `referenceImages` is serialized under `parameters.referenceImages` instead of `instances[0].referenceImages`.
5. Repeat with `extra_params.lastFrame` and observe the same placement issue for `lastFrame`.

### Expected behavior

`referenceImages` and `lastFrame` should be serialized on `instances[0]` for Gemini/Veo video generation requests.

### Actual behavior

[...]

PR #4692, opened 2026-06-25
changelogs for 1.6.0

### LINKED ISSUE #2347: [Bug]: MCP tool ordering is non-deterministic, breaking prefix-based prompt caching
### Prerequisites

### Description

MCP tools backed by Go maps have non-deterministic iteration order, causing the tool list and property keys to shuffle across requests. This breaks prefix-based prompt caching (Anthropic, OpenAI) since tool definitions form the first level of the cache prefix.

## Changes

- Sort `GetAvailableTools()` result by function name (`core/mcp/toolmanager.go`)
- Sort MCP tool property keys via `OrderedMapFromMap` default sorting (`core/mcp/utils.go`)
- Make `OrderedMapFromMap` sort keys lexicographically by default, with an optional `sortFn` override for custom ordering (`core/schemas/orderedmap.go`)

### Steps to reproduce

1. Configure bifrost with an MCP server that exposes tools with multiple properties (5+ per tool is enough to trigger Go's map randomization reliably).

2. Send identical chat completion requests through bifrost with MCP tool injection enabled

3. Observe the cache hit (read/write) tokens across 4-5 consecutive identical requests. They stay relative stable, usually around ~4k tokens (where the tool call section starts depending on the provider). 

### Expected behavior

Identical chat completion requests with MCP tool injection should produce identical tool JSON across requests, enabling prompt cache hits after the first request

### Actual behavior

Every request pays full cache creation cost because the tool JSON differs across requests due to non-deterministic map iteration order

### Affected area(s)

Core (Go)

### Version

1.4.17

### Environment

```text

```

### Relevant logs/output

```shell

```

### Regression?

_No response_

### Severity

### LINKED ISSUE #3443: [Bug]: Anthropic→OpenAI streaming tool_call deltas violate OpenAI spec on continuation chunks (breaks strict clients)

### LINKED ISSUE #4068: Bedrock: mid-conversation system messages hoisted into top-level `system` block break prompt caching

### LINKED ISSUE #4413: [Bug]: OpenAI Responses streaming returns empty error.message on context_length_exceeded

### LINKED ISSUE #4460: GLM-5.2 reasoning_effort "max" silently downgraded to "high" (same normalization gap as #4320 for DeepSeek V4)
## Problem

`reasoning_effort: "max"` is silently downgraded to `"high"` for **GLM-5.2** ([Z.ai](<http://Z.ai>)). GLM-5.2 natively supports both `high` and `max` effort levels ([Z.ai docs](<https://docs.z.ai/devpack/latest-model>)), but `normalizeOpenAIReasoningEffort()` collapses `max → high` because GLM-5.2 is not in `supportsMaxReasoningEffort()`.

This is the **exact same pattern** as issue maximhq/bifrost#4320 (DeepSeek V4), which was fixed in PR maximhq/bifrost#4330 by adding `deepseek-v4` to the supported list. GLM-5.2 needs the same treatment.

## Root Cause

`core/providers/openai/utils.go`, `normalizeOpenAIReasoningEffort()`:

```go
case "max":
    if supportsMaxReasoningEffort(model) {   // ← only "deepseek-v4"
        return effort
    }
    if supportsOpenAIXHighReasoningEffort(model) {  // ← only gpt-5.x
        return "xhigh"
    }
    return "high"   // ← GLM-5.2 lands here: max silently becomes high
```

`supportsMaxReasoningEffort()` (line 134) only matches `deepseek-v4`:

```go
func supportsMaxReasoningEffort(model string) bool {
    _, parsedModel := schemas.ParseModelString(model, schemas.OpenAI)
    if parsedModel != "" {
        model = parsedModel
    }
    return strings.HasPrefix(strings.ToLower(model), "deepseek-v4")  // ← GLM-5.2 missing
}
```

## Evidence

Direct probe through a Bifrost gateway (v1.5.13) routing to [Z.ai](<http://Z.ai>) GLM-5.2. The `extra_fields.raw_request` shows what Bifrost sent upstream:

**Sent to Bifrost:**

```json
{ "reasoning_effort": "max" }
```

**Bifrost sent to [Z.ai](<http://Z.ai>) (from** `extra_fields.raw_request`**):**

```json
{ "reasoning_effort": "high" }   // ← downgraded
```

The passthrough header (`x-bf-passthrough-extra-params`) does **not** bypass this normalization — `reasoning_effort` is a known field parsed into `ChatParameters.Reasoning.Effort` and normalized regardless.

## Suggested Fix

Add GLM-5.2 to `supportsMaxReasoningEffort()` (one-line change):

```diff
 func supportsMaxReasoningEffort(model string) bool {
     _, parsedModel := schemas.ParseModelString(model, schemas.OpenAI)
     if parsedModel != "" {
         model = parsedModel
     }
-    return strings.HasPrefix(strings.ToLower(model), "deepseek-v4")
+    modelLower := strings.ToLower(model)
+    return strings.HasPrefix(modelLower, "deepseek-v4") ||
+        strings.HasPrefix(modelLower, "glm-5.2")
 }
```

## Impact

* GLM-5.2 is a major coding model; [Z.ai](<http://Z.ai>) recommends `max` effort for coding tasks
* Affects all users routing GLM-5.2 through Bifrost with `reasoning_effort: "max"`
* The fix is scoped (only GLM-5.2 models), so no risk to other providers

## Environment

* Bifrost version: v1.5.13 (also confirmed absent in latest v1.5.20 / `main` branch)
* Upstream: [Z.ai](<http://Z.ai>) GLM-5.2 (`https://api.z.ai/api/coding/paas/v4`)

Related: maximhq/bifrost#4320, maximhq/bifrost#4330, [#3369](<https://github.com/maximhq/bifrost/issues/3369>)

### LINKED ISSUE #4496: [Bug]: Frequent intermittent broken pipe and closed connection errors with vllm provider
### Prerequisites

### Description

A certain percentage of chat streaming requests fail after a couple of ms with either 
`
failed to execute HTTP request to provider API
Details
the server closed connection before returning the first response byte. Make sure the server returns 'Connection: close' response header before closing the connection
`

or

`failed to execute HTTP request to provider API
Details
write tcp x.x.x.x:xxxxx->x.x.x.x:8000: write: broken pipe`

We assume this is due to idle connections timing out after 5 seconds on the vllm provider by default and bifrost reusing them with FIFO for up to 30s while only checking liveness after 10s. As per our understanding of the code, the same request needs to hit dead connections twice in a row to surface this way. As fasthttp is configured to use FIFO, once the oldest connection on the pool is found to be stale, it will try the next newest, which is also likely to be stale. On the streaming path, setting max_retries > 0 in the network_config does not alleviate this - requests that fail in this way surface as an 500 error, which is not retried.

### Steps to reproduce

1) Connect vllm provider
2) Direct normal traffic to bifrost
3) Experience intermittent failing requests

### Expected behavior

Requests process fine with status 200

### Actual behavior

Streaming requests fail with error 500, non-retriable. Non-streaming requests fail with 502

### Affected area(s)

Core (Go), Transports (HTTP)

### Version

1.5.7, 1.5.15

### Environment

```text

```

### Relevant logs/output

```shell

```

### Regression?

_No response_

### Severity

### LINKED ISSUE #4544: [Bug]: Cerebras + /anthropic endpoint fails after first turn with 400 provider API error
### Prerequisites

### Description

Hi! Firstly, super cool project! Thanks for making it and OSS'ing it.

I am facing a bug that happens particularly after I've setup a Cerebras API provider to use `cerebras/gpt-oss-120b`. I'm on a MacOS and have the latest version etc.

I've narrowed down the bug a bit.
When I use `ANTHROPIC_BASE_URL='http://127.0.0.1:8080/anthropic' ANTHROPIC_AUTH_TOKEN='<local_token>' ANTHROPIC_MODEL='cerebras/gpt-oss-120b' claude`, the first message successfully returns 100% of the time. But on the second message, 100% of the time, I get `400 provider API error (status 400)`:

Image

Strangely, when I try the same model / provider on codex, it works fine up for as many turns as I want:
Image

I think it may be related to the local anthropic endpoint on + 1 turns. I only have cerebras as a provider, so not sure if other providers are also affected, but it reliably fails 100% of the time for me on the second time when I launch Claude using /anthropic with Cerebras.

I also have tried to look into the Bifrost logs a bit, but sadly they only show the same 400 as far as I can tell:

Image

Lastly, I DO also see the failed 400 API request hitting Cerebras when I login to my Cerebras console -- so it's getting rejected by their API for something, hence why I think it may be specific to Cerebras. If you could help me figure out what's going wrong, I would really appreciate it!

Thanks
Ellek

### Steps to reproduce

1. Add Cerebras api provider
2. Launch `ANTHROPIC_BASE_URL='http://127.0.0.1:8080/anthropic' ANTHROPIC_AUTH_TOKEN='<local_token>' ANTHROPIC_MODEL='cerebras/gpt-oss-120b' claude`
3. Send 2 messages
4. Second message returns 400

### Expected behavior

Non failure

### Actual behavior

400

### Affected area(s)

Framework

### Version

latest using npx

### Environment

```text
MacOs Tahoe
Node v20 and v22 (tested both)
```

### Relevant logs/output

```shell

```

### Regression?

_No response_

### Severity

### LINKED ISSUE #4608: ResponsesMessage drops author, recipient, encrypted_content fields — breaks Codex multi_agent_v2 subagent spawning

### LINKED ISSUE #4617: [Bug]: idle-timeout timer goroutine can panic in closeBodyStream and crash the process
### Description

`NewIdleTimeoutReader` (`core/providers/utils/utils.go`) registers a
`time.AfterFunc` that, on idle timeout, calls `closeBodyStream(r.bodyStream, …)`.
If that timer fires **after** the stream's connection has already been released
to / reused from the fasthttp pool — an orphaned timer on an already-completed
stream — `CloseWithError` nil-dereferences inside fasthttp's
`(*HostClient).CloseConn` and panics.

Because the panic happens in the **timer goroutine**, it is unrecoverable by
any caller (the request that created the reader is long gone) and crashes the
entire process.

PR #3677 ("fix idle timeout panic") added a `recover()` to the
`idleTimeoutReader.Read()` path, but the `AfterFunc`'s own `closeBodyStream`
call has no equivalent guard, so this crash path remains open on current `dev`.

### Steps to reproduce

The race is timing-dependent in production, but the crashing code path is
deterministic to reproduce in a unit test. A body whose `CloseWithError`
panics (mimicking the fasthttp `CloseConn` nil-deref), wired through
`NewIdleTimeoutReader` with a short timeout:

```go
type timerPanicCloser struct{}

func (timerPanicCloser) Read([]byte) (int, error)   { return 0, io.EOF }
func (timerPanicCloser) CloseWithError(error) error { panic("simulated fasthttp CloseConn nil-deref") }

func TestIdleTimerPanicCrashes(t *testing.T) {
    _, cleanup := NewIdleTimeoutReader(&someReader{}, timerPanicCloser{}, 10*time.Millisecond, nil)
    time.Sleep(50 * time.Millisecond) // let the idle timer fire
    cleanup()
}
```

On current `dev` this **crashes the test process** (`panic: … goroutine [running]`).

### Expected behavior

A stale/orphaned idle timer firing on an already-released connection must not
crash the process. The teardown panic should be contained to the timer
goroutine, consistent with the `Read()`-path guard from #3677.

### Actual behavior

The panic propagates out of the `time.AfterFunc` goroutine and terminates the
process. Observed in production as a gateway built on Bifrost crashing under
sustained streaming load, with the crash firing minutes after the originating
stream completed (when the stale idle timer finally elapses), making it hard to
correlate with any single request.

### Affected area(s)

Core (Go) — `core/providers/utils` (streaming idle-timeout reader).

### Version

Reproduced on `core` v1.5.22 and on current `dev` (the `AfterFunc` has no
`recover()` around its `closeBodyStream` call).

### Regression?

Partial. #3677 (merged 2025-05-21) fixed the `Read()`-path panic but did not
cover the timer-callback path, which still crashes.

### Severity

High — uncontained process crash in a long-running gateway under normal
streaming load.

### Proposed fix

PR #4616 adds the companion `recover()` inside the `AfterFunc`, plus a
regression test that crashes without the fix and passes with it.

### LINKED ISSUE #4622: [Bug]: Bedrock Converse document blocks with format xlsx/xls/doc/docx are silently rewritten to format=pdf, causing ValidationException
## Prerequisites

## Description

When a request reaches the Bifrost HTTP transport via the `/bedrock/model/{modelId}/converse` endpoint and contains a `document` content block with `format: "xlsx"` (or `"xls"`, `"doc"`, `"docx"`), Bifrost internally converts it to `format: "pdf"` before forwarding to Bedrock. The document bytes (which are valid xlsx/xls/doc/docx content) are then sent to Bedrock labeled as PDF, causing Bedrock to reject the request with a `ValidationException`.

**Affected code:** `core/providers/bedrock/responses.go` — two functions:

1. **`convertSingleBedrockMessageToBifrostMessages`** (~line 4041): The switch that maps `block.Document.Format` to a MIME type is missing cases for `xlsx`, `xls`, `doc`, and `docx`. These fall through to `default: fileType = "application/pdf"`.

```go
// current (incomplete)
switch block.Document.Format {
case "pdf":
    fileType = "application/pdf"
case "txt", "md", "html", "csv":
    fileType = "text/plain"
default:
    fileType = "application/pdf"  // xlsx, xls, doc, docx all hit this
}
```

2. **`convertBifrostResponsesMessageContentBlocksToBedrockContentBlocks`** (~line 4287): The reverse mapping from MIME type back to Bedrock `doc.Format` also only handles `text/*` and `pdf`. Non-PDF binary formats not covered here default to `doc.Format = "pdf"`.

These two gaps together mean the round-trip `Bedrock format → Bifrost internal → Bedrock format` corrupts the document format for any Office document type.

Note: `chat.go` lines 162–185 (`ToBifrostChatResponse`) already has the complete and correct mapping for all formats including `xlsx`, `xls`, `doc`, `docx`. The responses path just needs to match it.

## Steps to reproduce

Send a `POST /bedrock/model/{modelId}/converse` request to the Bifrost HTTP transport with a document content block:

```json
{
  "messages": [
    {
      "role": "user",
      "content": [
        { "text": "Parse this spreadsheet" },
        {
          "document": {
            "name": "data",
            "format": "xlsx",
            "source": {
              "bytes": "<base64-encoded xlsx bytes>"
            }
          }
        }
      ]
    }
  ]
}
```

Observe that Bedrock returns:
```
ValidationException: The model returned the following errors:
messages.0.content.1.image.source.bytes: The PDF specified was not valid.
```

The error path `image.source.bytes` is Bedrock's internal error path for document content validation failures — the block is still a `document` block on the wire, but Bedrock rejects it because the bytes are not valid PDF.

**Reproducer confirming the underlying problem** (direct Bedrock call):

```python
import boto3

bedrock = boto3.client("bedrock-runtime", region_name="us-east-1")
fake_xlsx = b'PK\x03\x04' + b'A' * 100  # xlsx magic bytes

# Sending xlsx bytes labeled as "pdf" → error (what Bifrost sends after the bug)
bedrock.converse(
    modelId="us.anthropic.claude-haiku-4-5-20251001-v1:0",
    messages=[{
        "role": "user",
        "content": [
            {"text": "test"},
            {"document": {"name": "test", "format": "pdf", "source": {"bytes": fake_xlsx}}}
        ]
    }],
    inferenceConfig={"maxTokens": 10}
)
# => ValidationException: messages.0.content.1.image.source.bytes: The PDF specified was not valid.

# Sending the same bytes labeled as "xlsx" → Bedrock accepts the request
bedrock.converse(
    modelId="us.anthropic.claude-haiku-4-5-20251001-v1:0",
    messages=[{
        "role": "user",
        "content": [
            {"text": "test"},
            {"document": {"name": "test", "format": "xlsx", "source": {"bytes": fake_xlsx}}}
        ]
    }],
    inferenceConfig={"maxTokens": 10}
)
# => succeeds (model error about unreadable content, but no ValidationException)
```

## Expected behavior

A `document` block with `format: "xlsx"` passes through Bifrost and reaches Bedrock with `format: "xlsx"` preserved.

## Actual behavior

The `format` field is rewritten to `"pdf"` and Bedrock rejects the request with `ValidationException: The PDF specified was not valid.`

## Affected area(s)

- Core (Go)
- Transports (HTTP)

## Version

`transports/v1.5.7` — confirmed present in latest `main` as of HEAD `ff8876d0`

## Environment

- Go (core provider code, no specific OS dependency)

## Relevant logs/output

```
ValidationException: An error occurred (ValidationException) when calling the Converse operation:
The model returned the following errors:
messages.0.content.1.image.source.bytes: The PDF specified was not valid.
```

## Severity

## Suggested fix

In `core/providers/bedrock/responses.go`, extend both switch statements to cover the missing Office document formats. The correct MIME types are already used in `chat.go` (`ToBifrostChatResponse`, lines 162–185):

**In `convertSingleBedrockMessageToBifrostMessages`** (the inbound format→MIME mapping):
```go
case "xlsx":
    fileType = "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet"
case "xls":
    fileType = "application/vnd.ms-excel"
case "docx":
    fileType = "application/vnd.openxmlformats-officedocument.wordprocessingml.document"
case "doc":
    fileType = "application/msword"
```

**In `convertBifrostResponsesMessageContentBlocksToBedrockContentBlocks`** (the outbound MIME→format mapping):
```go
case strings.Contains(fileType, "spreadsheetml") || fileType == "xlsx":
    doc.Format = "xlsx"
case fileType == "application/vnd.ms-excel" || fileType == "xls":
    doc.Format = "xls"
case strings.Contains(fileType, "wordprocessingml") || fileType == "docx":
    doc.Format = "docx"
case fileType == "application/msword" || fileType == "doc":
    doc.Format = "doc"
```

Per the [AWS Bedrock Converse API docs](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_DocumentBlock.html), valid `format` values are: `pdf | csv | doc | docx | xls | xlsx | html | txt | md`.

akshaydeo:
## Summary

This PR bumps core to v1.6.0, framework to v1.4.0, and transports to v1.6.0, shipping a batch of new provider integrations, a typed secret-reference system, streaming reliability improvements, and a broad set of bug fixes across Bedrock, Gemini, Anthropic, and the Responses API.

## Changes

- **Runware & Runway** — Added Runware provider with image and video generation operations, and Runway image generation operations
- **Customer Attribution** — Added `x-bf-customer-id` and `x-bf-customer-name` header support for per-customer attribution (governance plugin)
- **Enriched Model Listing** — `list models` now returns `ContextLength`, `MaxInputTokens`, `MaxOutputTokens`, `Architecture`, and `WebSearch` pricing from pricing entries; Fireworks models enriched from config
- **Streaming Pause/Resume** — Added pause/resume flows for streaming calls and reduced streaming memory usage
- **SecretVar** — Introduced typed `SecretVar` env/vault reference support (`env.*`, `vault.*`) with store/resolve/remove hooks, replacing `EnvVar` across core, framework, and plugins. Note: `TableVirtualKey.Value` was rolled back to a plain string; virtual key values do not support env/vault references in this release
- **GORM Vault Callbacks** — Added global GORM vault callbacks with `VaultPathKeyer` interface and `map[string]EnvVar` support, replacing per-model `BeforeSave`/`AfterDelete` hooks
- **Circuit Breaker UI** — Added circuit breaker config schema and UI per provider
- **SSO** — Added SailPoint and generic SSO providers; refactored SCIM attribute-mapping schemas into shared `$defs`
- **OTEL Plugin** — Added `group_traces_by_session` support, root-span content logging disable toggle, and typed `SecretVar` support
- **Mantle** — Added Mantle list models endpoint and OpenAI-compatible endpoint support
- **OpenCode Gateways** — Added Go and Zen gateway providers
- **Advisor Tool** — Added advisor tool compatibility for Claude
- **Bedrock** — Corrected GLM-5 stop sequences, removed erroneous web search error, broadened document file extension support (xlsx/xls/doc/docx), inlined mid-conversation system reminders to preserve prompt cache, preserved `tool_result` order to match parallel `tool_use` blocks, and set TTL in cache points
- **Gemini** — Fixed video reference fields mapping to instances; preserved request conversion for Gemini/Vertex batch requests
- **Responses API** — Preserved `author`, `recipient`, and `encrypted_content` in multi-agent items and message copies; accepted object-valued tool-call arguments (e.g. `tool_search_call`) on the streaming path; populated error message in streaming error events
- **Anthropic** — Fixed tool call handling; surfaced prompt-cache diagnostics via new `CacheDiagnostics`/`CacheMissReason` schema
- **Reliability** — Added retry on stale connections; fixed ctx panic in fasthttp; recovered from `closeBodyStream` panic in idle-timeout timer goroutine; dropped assistant reasoning for Cerebras
- **Provider Config Sync** — `allow_all_keys` and `blacklisted_models` now sync from `config.json` source of truth; `PUT /api/providers/{provider}` no longer silently discards `keys`/blocked-model edits
- **MCP** — Deterministic MCP tool ordering for prompt cache stability
- **Key Pool** — Added key pool filter; committed routing-pinned key ID to reserved `BifrostContextKeyAPIKeyID` after `PreRequestHook` unblock
- **Misc** — Preserved `extra_content` on `ChatAssistantMessageToolCall` for lossless round-tripping; preserved structured errors for streaming plugin blocks; configurable server logs; raw SQL column-add migrations; `responses_input_history` fallback for hybrid logstore logs UI

## Type of change

- [x] Bug fix
- [x] Feature
- [x] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [x] Core (Go)
- [x] Transports (HTTP)
- [x] Providers/Integrations
- [x] Plugins
- [x] UI (React)
- [ ] Docs

## How to test

```sh
# Core/Transports
go version
go test ./...

# UI
cd ui
pnpm i || npm i
pnpm test || npm test
pnpm build || npm run build
```

- Verify Runware and Runway image generation endpoints respond correctly with valid credentials
- Confirm `x-bf-customer-id` / `x-bf-customer-name` headers are forwarded and attributed in logs
- Confirm `list models` returns enriched fields (`ContextLength`, `MaxInputTokens`, etc.)
- Test `env.*` and `vault.*` references in provider config fields resolve correctly; confirm virtual key values remain plain strings
- Validate streaming pause/resume behavior under load and confirm reduced memory footprint
- Confirm Bedrock document uploads with xlsx/xls/doc/docx formats are no longer rewritten to pdf

## Breaking changes

- [x] Yes
- [ ] No

`EnvVar` has been renamed to `SecretVar` across core, framework, and several plugins (jsonparser, maxim, mocker, semanticcache, telemetry). Any code or config referencing `EnvVar` directly will need to be updated to `SecretVar`. Virtual key values (`TableVirtualKey.Value`) do not support env/vault references in this release; that capability was reverted.

## Related issues

- Closes [#2347](https://github.com/maximhq/bifrost/issues/2347) — Non-deterministic MCP tool ordering breaking prefix-based prompt caching
- Closes [#3443](https://github.com/maximhq/bifrost/issues/3443) — Anthropic→OpenAI streaming tool_call deltas violate OpenAI spec on continuation chunks
- Closes [#4068](https://github.com/maximhq/bifrost/issues/4068) — Bedrock mid-conversation system messages break prompt caching
- Closes [#4413](https://github.com/maximhq/bifrost/issues/4413) — OpenAI Responses streaming returns empty `error.message` on context_length_exceeded
- Closes [#4460](https://github.com/maximhq/bifrost/issues/4460) — GLM-5.2 `reasoning_effort` "max" silently downgraded to "high"
- Closes [#4496](https://github.com/maximhq/bifrost/issues/4496) — Frequent broken pipe / closed connection errors with vllm provider
- Closes [#4544](https://github.com/maximhq/bifrost/issues/4544) — Cerebras + `/anthropic` endpoint fails after first turn with 400 error
- Closes [#4608](https://github.com/maximhq/bifrost/issues/4608) — ResponsesMessage drops author/recipient/encrypted_content, breaking Codex multi_agent_v2 subagent spawning
- Closes [#4617](https://github.com/maximhq/bifrost/issues/4617) — idle-timeout timer goroutine can panic in `closeBodyStream` and crash the process
- Closes [#4622](https://github.com/maximhq/bifrost/issues/4622) — Bedrock Converse document blocks with xlsx/xls/doc/docx silently rewritten to pdf
- Closes [#4627](https://github.com/maximhq/bifrost/issues/4627) — Gemini video reference fields sent under `parameters`
- Closes [#4640](https://github.com/maximhq/bifrost/issues/4640) — Provider config `key_ids:["*"]` not synced to `allow_all_keys` for existing virtual keys
- Closes [#4648](https://github.com/maximhq/bifrost/issues/4648) — `PUT /api/providers/{provider}` silently discards `payload.Keys`

## Security considerations

- `SecretVar` introduces typed env/vault references for secret fields; ensure vault paths and env variable names do not leak sensitive identifiers in logs or error responses
- Internal error details (stack traces, SQL) are now sanitized before being sent to clients, reducing information disclosure risk
- Virtual key values remain plain strings in this release; vault/env reference support for that field was explicitly reverted to avoid unintended exposure

## Checklist


akshaydeo:
* **#4692** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/4692?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/maximhq/bifrost/4692?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `dev`


<h2></h2>

PR #4726, opened 2026-06-26
feat: add ElevenLabs sound effects (text-to-sound) support

### LINKED ISSUE #4712: [Feature]: ElevenLabs sound effects (/v1/sound-generation)
## Summary

ElevenLabs exposes a dedicated **text-to-sound-effects** API at `POST /v1/sound-generation` (model `eleven_text_to_sound_v2`). Bifrost’s ElevenLabs provider currently covers TTS, STT, list models, and ConvAI WebSocket realtime, but **does not expose sound generation** in any form.

Official reference: https://elevenlabs.io/docs/api-reference/text-to-sound-effects/convert

## Motivation

Sound generation is a first-class ElevenLabs product surface (short SFX clips, optional seamless loops) and is **orthogonal to TTS**:

| | TTS (`/v1/text-to-speech/{voice_id}`) | Sound effects (`/v1/sound-generation`) |
|---|----------------------------------------|----------------------------------------|
| Requires `voice_id` | Yes | No |
| Primary input | Spoken text | Sound description |
| Typical duration | Variable speech | 0.5–30s (or auto) |
| Loop support | No | Yes (`loop`, v2 model) |
| Extra params | `voice_settings`, pronunciation dicts | `duration_seconds`, `prompt_influence`, `loop` |

Users who route all inference through Bifrost today cannot call this endpoint without bypassing the gateway (direct `api.elevenlabs.io` + raw API key).

## Current behavior

In `core/providers/elevenlabs/`, the only upstream paths implemented are:

- `GET /v1/models`
- `POST /v1/text-to-speech/{voice_id}` (+ `/stream`, `/with-timestamps`)
- `POST /v1/speech-to-text`
- `wss://.../v1/convai/conversation` (realtime)

There is no reference to `sound-generation`, `sound_effects`, or `eleven_text_to_sound_v2` in the codebase. `Passthrough` for ElevenLabs returns `UnsupportedOperationError`.

## Proposed API contract (upstream)

```
POST https://api.elevenlabs.io/v1/sound-generation?output_format=mp3_22050_32
Headers: xi-api-key, Content-Type: application/json

Body:
{
  "text": "<prompt>",
  "model_id": "eleven_text_to_sound_v2",   // default per docs
  "duration_seconds": 2.0,                 // optional, 0.5–30; omit for auto length
  "loop": false,                           // optional, v2 only
  "prompt_influence": 0.3                    // optional, 0–1
}

Response: binary audio (application/octet-stream)
```

No streaming variant exists for this endpoint.

### Option A — ElevenLabs passthrough (smallest diff)

- Implement `ElevenlabsProvider.Passthrough` (mirror OpenAI/Anthropic pattern, `xi-api-key` auth).
- Register `/elevenlabs_passthrough` router (as with `/openai_passthrough`).
- Enables `/v1/sound-generation` and also `/v1/music` without separate feature work.

**Pros:** ~200 LOC, unlocks other uncovered REST endpoints.  
**Cons:** Weak governance/model semantics; billing may not attribute SFX correctly.

### Option B — Extend existing speech path

- In `ElevenlabsProvider.Speech()`, branch when `model == eleven_text_to_sound_v2` (or explicit flag) to call `/v1/sound-generation`.
- Reuse `POST /v1/audio/speech` gateway; map `input` → `text`, `extra_params` → `loop` / `duration_seconds` / `prompt_influence`.
- Skip `voice` requirement for sound model; reject streaming for that model.

**Pros:** Virtual keys, routing, logging, tracing work out of the box.  
**Cons:** Semantic mismatch with OpenAI TTS schema; music (`/v1/music`) still needs separate handling.

### Option C — New `RequestType` (not recommended)

Full new operation across provider interface, HTTP handlers, OpenAPI, UI — disproportionate for a single provider endpoint.

## Acceptance criteria

- [ ] Callers can generate sound effects through Bifrost using configured ElevenLabs keys (no direct `api.elevenlabs.io` required).
- [ ] Request supports at minimum: `text`, optional `duration_seconds`, `loop`, `prompt_influence`, `output_format` query param.
- [ ] Response returns binary audio with correct `Content-Type`.
- [ ] Errors from ElevenLabs are surfaced via existing `parseElevenlabsError` patterns.
- [ ] Documentation updated in `docs/providers/supported-providers/elevenlabs.mdx` and provider overview matrix.
- [ ] Unit/integration tests (mock or opt-in with `ELEVENLABS_API_KEY`).

## Related gaps (out of scope for this issue, but worth tracking)

- `/v1/music` (music generation) — separate endpoint, similar passthrough or speech-extension pattern.
- Broader `output_format` mapping for TTS/SFX (e.g. `mp3_22050_32` alias in `utils.go`).
- Accurate cost attribution for duration-based SFX billing when `duration_seconds` is set.

## References

- ElevenLabs sound effects API: https://elevenlabs.io/docs/api-reference/text-to-sound-effects/convert
- Existing ElevenLabs provider: `core/providers/elevenlabs/`
- Provider capability matrix: `docs/providers/supported-providers/overview.mdx`

SecretSun:
## Summary

Adds ElevenLabs **sound effects** (text-to-sound) support, closing the gap described in #4712. ElevenLabs' `POST /v1/sound-generation` (model `eleven_text_to_sound_v2`) is orthogonal to TTS (no voice, prompt-described SFX, optional looping, 0.5–30s duration). Until now it could not be called through the gateway.

This implements **Option B-ish (minimal, governed)**: sound models reuse the existing **speech** request type and are dispatched inside the ElevenLabs provider based on the model id, so they ride the full virtual-key governance pipeline (provider/model allowlists, budgets, rate limits) with **zero changes to existing flows**.

## What changed

- **`core/providers/elevenlabs/sound_generation.go`** (new): `IsElevenlabsSoundModel`, request mapping (`ToElevenlabsSoundGenerationRequest`) for `duration_seconds` / `loop` / `prompt_influence` with range clamping, and `soundGeneration()` calling `POST /v1/sound-generation` (reuses the speech URL builder for `output_format`).
- **`core/providers/elevenlabs/elevenlabs.go`**: `Speech()` dispatches to `soundGeneration()` when the model is a sound model; the existing TTS path is unchanged.
- **`transports/bifrost-http/handlers/inference.go`**: new `POST /v1/audio/sound-effects` alias (also works via `POST /v1/audio/speech`); the voice requirement is relaxed **only** for sound models; shared `serveSpeech` sets the right attachment filename.
- **`core/schemas/speech.go`**: additive `SpeechUsage.AudioSeconds` (forward-compatible) for future per-second billing.
- **`core/providers/elevenlabs/sound_generation_test.go`** (new): unit tests for model detection, request mapping, and clamping.
- **`docs/.../elevenlabs.mdx`**: documents the feature, parameters, and billing caveats.

## Design notes

- **No new RequestType / Provider interface method** — avoids touching ~20 providers. Dispatch is by model id inside `Speech()`, so all non-streaming speech entry points (native speech, the new sound-effects alias, async speech, OpenAI-compat speech) route correctly.
- **Backward compatible**: existing TTS, streaming, transcription, list-models, and governance are untouched (diff is additive or guarded behind a sound-model check).
- **No streaming**: the upstream sound-generation API has no streaming variant.

## Billing

The default pricing datasheet does not yet include `eleven_text_to_sound_v2`, so without a pricing entry the request records `cost = 0`. The response now carries `audio_seconds`; wiring it into per-second cost is left as a follow-up. Until then, a pricing override (works with any config store, incl. SQLite) can bill it.

## Testing

- Unit tests pass (`go test ./providers/elevenlabs/`), `gofmt`/`go vet` clean.
- End-to-end verified against the live ElevenLabs API through a locally built gateway:
  - `POST /v1/audio/sound-effects` and `POST /v1/audio/speech` both return valid MP3 (HTTP 200, `audio/mpeg`); `duration_seconds` affects output length.
  - With a SQLite config store + a pricing override, governance is active and a non-zero cost is recorded — no Postgres required.

Closes #4712

### DIFF
--- a/core/providers/elevenlabs/soundgeneration.go
+++ b/core/providers/elevenlabs/soundgeneration.go
@@ -1,0 +1,148 @@
+package elevenlabs
+
+import (
+	"math"
+	"net/http"
+
+	providerUtils "github.com/maximhq/bifrost/core/providers/utils"
+	schemas "github.com/maximhq/bifrost/core/schemas"
+	"github.com/valyala/fasthttp"
+)
+
+// ToElevenlabsSoundGenerationRequest maps a Bifrost speech request onto the
+// sound-generation body. SFX-specific fields arrive via ExtraParams (they are not
+// known fields of the speech schema) and are clamped to the upstream-allowed ranges.
+// model_id uses the canonical model resolved from virtual-key aliases so upstream
+// receives a real ElevenLabs identifier, not the user-facing alias string.
+func ToElevenlabsSoundGenerationRequest(ctx *schemas.BifrostContext, bifrostReq *schemas.BifrostSpeechRequest) *ElevenlabsSoundGenerationRequest {
+	if bifrostReq == nil || bifrostReq.Input == nil {
+		return nil
+	}
+
+	out := &ElevenlabsSoundGenerationRequest{
+		Text:    bifrostReq.Input.Input,
+		ModelID: schemas.ResolveCanonicalModel(ctx, bifrostReq.Model),
+	}
+
+	if bifrostReq.Params != nil && bifrostReq.Params.ExtraParams != nil {
+		// Copy the caller's map before consuming keys: the original
+		// BifrostSpeechRequest may be reused (e.g. fallback / multi-key retries),
+		// so mutating it in place would silently drop these params on retry.
+		extra := make(map[string]interface{}, len(bifrostReq.Params.ExtraParams))
+		for k, v := range bifrostReq.Params.ExtraParams {
+			extra[k] = v
+		}
+		out.ExtraParams = extra
+
+		if v, ok := schemas.SafeExtractFloat64Pointer(extra["duration_seconds"]); ok {
+			delete(extra, "duration_seconds")
+			out.DurationSeconds = clampFloat64Ptr(v, 0.5, 30)
+		}
+		if v, ok := schemas.SafeExtractBoolPointer(extra["loop"]); ok {
+			delete(extra, "loop")
+			out.Loop = v
+		}
+		if v, ok := schemas.SafeExtractFloat64Pointer(extra["prompt_influence"]); ok {
+			delete(extra, "prompt_influence")
+			out.PromptInfluence = clampFloat64Ptr(v, 0, 1)
+		}
+	}
+
+	return out
+}
+
+// clampFloat64Ptr returns a pointer to v clamped to [min, max].
+func clampFloat64Ptr(v *float64, min, max float64) *float64 {
+	if v == nil {
+		return nil
+	}
+	c := *v
+	if c < min {
+		c = min
+	}
+	if c > max {
+		c = max
+	}
+	return &c
+}
+
+// soundGeneration performs a text-to-sound-effects request against
+// POST /v1/sound-generation. It mirrors Speech() but targets a different endpoint
+// and has no voice / timestamps concept. Returns audio bytes in BifrostSpeechResponse.
+func (provider *ElevenlabsProvider) soundGeneration(ctx *schemas.BifrostContext, key schemas.Key, request *schemas.BifrostSpeechRequest) (*schemas.BifrostSpeechResponse, *schemas.BifrostError) {
+	req := fasthttp.AcquireRequest()
+	resp := fasthttp.AcquireResponse()
+	defer fasthttp.ReleaseRequest(req)
+	defer fasthttp.ReleaseResponse(resp)
+
+	providerUtils.SetExtraHeaders(ctx, req, provider.networkConfig.ExtraHeaders, nil)
+
+	// Reuses the speech URL builder for base URL + output_format query handling;
+	// only the path differs. enable_logging / optimize_streaming_latency are not
+	// set for sound requests, so no TTS-only query params leak in.
+	requestURL := provider.buildBaseSpeechRequestURL(ctx, "/v1/sound-generation", schemas.SpeechRequest, request)
+	req.SetRequestURI(requestURL)
+
+	req.Header.SetMethod(http.MethodPost)
+	req.Header.SetContentType("application/json")
+	if key.Value.GetValue() != "" {
+		req.Header.Set("xi-api-key", key.Value.GetValue())
+	}
+
+	// Build once so the resolved duration can also feed per-second billing below.
+	soundReq := ToElevenlabsSoundGenerationRequest(ctx, request)
+
+	jsonData, bifrostErr := providerUtils.CheckContextAndGetRequestBody(
+		ctx,
+		request,
+		func() (providerUtils.RequestBodyWithExtraParams, error) {
+			return soundReq, nil
+		})
+	if bifrostErr != nil {
+		return nil, bifrostErr
+	}
+
+	if !providerUtils.ApplyLargePayloadRequestBody(ctx, req) {
+		req.SetBody(jsonData)
+	}
+
+	latency, bifrostErr, wait := providerUtils.MakeRequestWithContext(ctx, provider.client, req, resp)
+	defer wait()
+	if bifrostErr != nil {
+		return nil, providerUtils.EnrichError(ctx, bifrostErr, jsonData, nil, provider.sendBackRawRequest, provider.sendBackRawResponse)
+	}
+	ctx.SetValue(schemas.BifrostContextKeyProviderResponseHeaders, providerUtils.ExtractProviderResponseHeaders(resp))
+
+	if resp.StatusCode() != fasthttp.StatusOK {
+		return nil, providerUtils.EnrichError(ctx, parseElevenlabsError(resp), jsonData, nil, provider.sendBackRawRequest, provider.sendBackRawResponse)
+	}
+
+	body, err := providerUtils.CheckAndDecodeBody(resp)
+	if err != nil {
+		return nil, providerUtils.EnrichError(ctx, providerUtils.NewBifrostOperationError(schemas.ErrProviderResponseDecode, err), jsonData, nil, provider.sendBackRawRequest, provider.sendBackRawResponse)
+	}
+
+	bifrostResponse := &schemas.BifrostSpeechResponse{
+		Audio: body,
+		ExtraFields: schemas.BifrostResponseExtraFields{
+			Latency:                 latency.Milliseconds(),
+			ProviderResponseHeaders: providerUtils.ExtractProviderResponseHeaders(resp),
+		},
+	}
+
+	// ElevenLabs bills sound effects per generated second when a duration is
+	// specified. Surface it on usage so the model-catalog per-second rate
+	// (OutputCostPerSecond) can bill accurately; auto-duration requests carry no
+	// known length and fall back to whatever token/char pricing is configured.
+	if soundReq != nil && soundReq.DurationSeconds != nil {
+		bifrostResponse.Usage = &schemas.SpeechUsage{
+			AudioSeconds: int(math.Round(*soundReq.DurationSeconds)),
+		}
+	}
+
+	if providerUtils.ShouldSendBackRawRequest(ctx, provider.sendBackRawRequest) {
+		providerUtils.ParseAndSetRawRequest(&bifrostResponse.ExtraFields, jsonData)
+	}
+
+	return bifrostResponse, nil
+}
--- a/transports/bifrost-http/handlers/inference.go
+++ b/transports/bifrost-http/handlers/inference.go
@@ -736,6 +736,9 @@ func (h *CompletionHandler) RegisterRoutes(r *router.Router, middlewares ...sche
 	r.POST("/v1/embeddings", lib.ChainMiddlewares(h.embeddings, baseMiddlewares...))
 	r.POST("/v1/rerank", lib.ChainMiddlewares(h.rerank, baseMiddlewares...))
 	r.POST("/v1/ocr", lib.ChainMiddlewares(h.ocr, baseMiddlewares...))
+	// ElevenLabs sound-effect models also flow through /v1/audio/speech; the
+	// provider routes them to /v1/sound-generation by model id, keeping SDK and
+	// transport APIs at parity (no separate sound-effects endpoint).
 	r.POST("/v1/audio/speech", lib.ChainMiddlewares(h.speech, baseMiddlewares...))
 	r.POST("/v1/audio/transcriptions", lib.ChainMiddlewares(h.transcription, baseMiddlewares...))
 	r.POST("/v1/images/generations", lib.ChainMiddlewares(h.imageGeneration, baseMiddlewares...))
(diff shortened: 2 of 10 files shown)


SecretSun:
Thanks for the reviews! Addressed the feedback in `7ac7c59`:

- **[P1, greptile] `ToElevenlabsSoundGenerationRequest` mutated the caller's `ExtraParams`** — good catch. It now copies the map before consuming `duration_seconds` / `loop` / `prompt_influence`, so the original `BifrostSpeechRequest` stays intact across fallback / multi-key retries. Added a regression test asserting the caller's map is untouched.

- **[P2, greptile + coderabbit] duplicated sound-model detection** — moved `ElevenlabsSoundModelPrefix` and `IsElevenlabsSoundModel` into the shared `core/providers/utils` package (which both the provider and the transport already import). The handler now calls `providerUtils.IsElevenlabsSoundModel(...)` instead of a hardcoded `strings.HasPrefix`, so there's a single source of truth and no layering coupling to the elevenlabs provider.

- **[coderabbit] billing docs were self-contradictory** — the example now leads with `input_cost_per_character` (what actually bills today) and clearly marks `output_cost_per_second` as the field that applies once the `audio_seconds` wiring lands.

`go build`, `gofmt`, `go vet`, and the elevenlabs + utils unit tests all pass.

Also re-pushed the commit under a GitHub-linked email to resolve the CLA association.

core/providers/elevenlabs/soundgeneration.go
```diff
@@ -0,0 +1,163 @@
+package elevenlabs
+
```
akshaydeo, line 2: we are trying to maintain file naming without _

can you rename this to soundgeneration.go

akshaydeo: test file can have soundgeneration_test.go

akshaydeo (changes requested):
overall PR looks good - lets do this change so we can merge and do the internal tests

SecretSun:
Thanks @akshaydeo! Renamed the files to drop the underscore in `06b219a`:

- `sound_generation.go` → `soundgeneration.go`
- `sound_generation_test.go` → `soundgeneration_test.go`

No other changes; `go build`, `gofmt`, `go vet`, and the elevenlabs unit tests still pass. Ready for another look whenever you get a chance.

core/providers/elevenlabs/soundgeneration.go
SecretSun: OK

core/providers/elevenlabs/soundgeneration.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+
+	// ElevenLabs bills sound effects per generated second when a duration is
+	// specified. Surface it on usage so the model-catalog per-second rate
+	// (OutputCostPerSecond) can bill accurately; auto-duration requests carry no
+	// known length and fall back to whatever token/char pricing is configured.
+	if soundReq != nil && soundReq.DurationSeconds != nil {
+		bifrostResponse.Usage = &schemas.SpeechUsage{
+			AudioSeconds: int(math.Round(*soundReq.DurationSeconds)),
+		}
```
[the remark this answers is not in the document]
SecretSun: `InputChars` is actually populated for this path — it's just done centrally rather than here. After `Speech()` returns, `core/bifrost.go` (the `schemas.SpeechRequest` case) calls `speechResponse.BackfillParams(req.BifrostRequest.SpeechRequest)`, and `BifrostSpeechResponse.BackfillParams` preserves any existing `Usage` and sets `Usage.InputChars = utf8.RuneCountInString(request.Input.Input)`. That happens before the post-hooks / billing run, so:

- duration given → `Usage = {AudioSeconds, InputChars}`
- auto duration → `Usage = {InputChars}`

So prompt usage is never zero, and the existing `input_cost_per_character` / `InputChars` contract is honored exactly like the normal TTS path (ElevenLabs TTS returns no token usage either, so `InputChars` is all the speech path populates). I set only `AudioSeconds` here precisely so backfill can add `InputChars` on top without me duplicating that logic.

Happy to set `InputChars` explicitly here too if you'd prefer the provider be self-contained, but it'd be redundant with the central backfill. Skipping for now to keep the change minimal — let me know.

core/providers/elevenlabs/soundgeneration.go
```diff
@@ -0,0 +1,163 @@
...
+	"net/http"
+
+	providerUtils "github.com/maximhq/bifrost/core/providers/utils"
+	schemas "github.com/maximhq/bifrost/core/schemas"
+	"github.com/valyala/fasthttp"
+)
+
+// ElevenlabsSoundGenerationRequest is the request body for POST /v1/sound-generation.
+// output_format is sent as a query parameter (see buildBaseSpeechRequestURL), not here.
+type ElevenlabsSoundGenerationRequest struct {
```
Pratham-Mishra04: can you pls shift this to types.go?

core/providers/utils/audio.go
```diff
@@ -5,8 +5,23 @@ package utils
...
+// (e.g. "eleven_text_to_sound_v2"). These route to /v1/sound-generation, a
+// different upstream API than text-to-speech, so they must not use a voice.
+// It lives here (not in the elevenlabs provider) so both the provider and the
+// transport layer can share one source of truth and avoid drift.
+const ElevenlabsSoundModelPrefix = "eleven_text_to_sound"
+
+// IsElevenlabsSoundModel reports whether a model id targets the ElevenLabs
+// sound-generation API rather than text-to-speech. Used by the provider to
+// dispatch and by the HTTP transport to relax the voice requirement.
+func IsElevenlabsSoundModel(model string) bool {
```
Pratham-Mishra04: lets try to replicate the methods in `schemas/account.go` like `IsElevenlabSoundModelFamily()` which also takes aliases into account along with their canonical model names

Pratham-Mishra04 (changes requested):
Hey @SecretSun I have added a few comments, rest looks good!

transports/bifrost-http/handlers/inference.go
```diff
@@ -729,6 +730,10 @@ func (h *CompletionHandler) RegisterRoutes(r *router.Router, middlewares ...sche
 	r.POST("/v1/rerank", lib.ChainMiddlewares(h.rerank, baseMiddlewares...))
 	r.POST("/v1/ocr", lib.ChainMiddlewares(h.ocr, baseMiddlewares...))
 	r.POST("/v1/audio/speech", lib.ChainMiddlewares(h.speech, baseMiddlewares...))
+	// Sound effects reuse the speech request/handler; the provider routes to the
+	// sound-generation API based on the model id. This is a semantic alias, not a
+	// new request type, so virtual-key governance and billing are unchanged.
+	r.POST("/v1/audio/sound-effects", lib.ChainMiddlewares(h.soundEffects, baseMiddlewares...))
```
Pratham-Mishra04: Also lets just reuse /speech endpoint which would now also serve sound models, to maintain sdk-transports api parity

transports/bifrost-http/handlers/inference.go
```diff
@@ -1327,8 +1332,19 @@ func prepareSpeechRequest(ctx *fasthttp.RequestCtx, config *lib.Config) (*Speech
...
 		return nil, nil, fmt.Errorf("input is required for speech completion")
 	}
-	if req.SpeechParameters == nil || req.VoiceConfig == nil || (req.VoiceConfig.Voice == nil && len(req.VoiceConfig.MultiVoiceConfig) == 0) {
-		return nil, nil, fmt.Errorf("voice is required for speech completion")
+	// ElevenLabs sound-generation models (e.g. "eleven_text_to_sound_v2") are not
+	// tied to a voice; they map to a separate upstream endpoint. Only enforce the
+	// voice requirement for regular text-to-speech models. The provider still
+	// rejects voiceless TTS requests, so this only relaxes the sound-effects path.
+	// Detection is shared with the provider via providerUtils to avoid drift.
+	isSoundModel := providerUtils.IsElevenlabsSoundModel(base.ModelName)
```
Pratham-Mishra04: this would break for aliases, like `best-model: eleven_text_to_sound-pro` - lets add the validation while making the sound gen request where we have resolved the alias

Pratham-Mishra04 (changes requested):

core/providers/elevenlabs/soundgeneration.go
SecretSun: OK

core/providers/utils/audio.go
SecretSun: OK

core/providers/elevenlabs/soundgeneration.go
SecretSun: Done — moved `ElevenlabsSoundGenerationRequest` (and its `GetExtraParams`) into `types.go` under a new "SOUND EFFECTS TYPES" section. `soundgeneration.go` now only has the mapping/logic.

core/providers/utils/audio.go
SecretSun: Done — removed the helper from `providers/utils` and replicated the `account.go` pattern instead: added `IsElevenlabsSoundModel(model)` in `schemas/utils.go` (alongside `IsAnthropicModel` etc.) and an alias-aware `IsElevenlabsSoundModelFamily(ctx, model)` in `schemas/account.go` that resolves the canonical model via `ResolveCanonicalModel` first. The provider dispatch now uses the alias-aware family helper.

transports/bifrost-http/handlers/inference.go
SecretSun: Done — dropped the dedicated `/v1/audio/sound-effects` route, handler, and `PathToTypeMapping` entry. Sound models now go through `POST /v1/audio/speech` (provider routes by model id), keeping SDK/transport parity. Collapsed the temporary `serveSpeech` helper back into `speech`. Docs updated accordingly.

transports/bifrost-http/handlers/inference.go
SecretSun: Good catch on the alias break. I removed the transport-layer voice check entirely (it couldn't see the resolved model/provider — aliases like `best-model -> eleven_text_to_sound-pro` were misclassified, and the resolved provider isn't known at this layer for VK-routed requests). Voice validation now lives only in the provider after alias resolution: the ElevenLabs TTS path already returns `voice parameter is required` when no voice is set, and sound models skip it. Heads up: this also means OpenAI speech no longer gets a transport-side voice 400 — it defers to the provider/upstream. Happy to add a provider-side check for the OpenAI speech path too if you'd like that behavior preserved.

core/providers/elevenlabs/soundgeneration.go
```diff
@@ -0,0 +1,146 @@
...
+// sound-generation body. SFX-specific fields arrive via ExtraParams (they are not
+// known fields of the speech schema) and are clamped to the upstream-allowed ranges.
+func ToElevenlabsSoundGenerationRequest(bifrostReq *schemas.BifrostSpeechRequest) *ElevenlabsSoundGenerationRequest {
+	if bifrostReq == nil || bifrostReq.Input == nil {
+		return nil
+	}
+
+	out := &ElevenlabsSoundGenerationRequest{
+		Text:    bifrostReq.Input.Input,
+		ModelID: bifrostReq.Model,
```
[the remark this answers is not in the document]
SecretSun: Fixed in `c8fc97b2d`. `ToElevenlabsSoundGenerationRequest` now takes `ctx` and sets `ModelID` via `schemas.ResolveCanonicalModel(ctx, bifrostReq.Model)` so virtual-key aliases whose `ModelName` is a sound model send the canonical ElevenLabs id upstream. Added `TestToElevenlabsSoundGenerationRequest_ResolvesAliasModelName` covering the `best-sfx -> eleven_text_to_sound_v2` case.

Also rebased onto latest `dev` (resolved conflicts in `account.go` / `utils.go`, keeping upstream's new `IsOpenAIModel*` helpers alongside ours).

SecretSun:
Addressed the latest CodeRabbit feedback in `c8fc97b2d`:

- **Canonical model_id for aliases**: sound-generation body now uses `ResolveCanonicalModel(ctx, bifrostReq.Model)` instead of the raw alias string.
- **Rebased onto latest `dev`**: resolved conflicts in `schemas/account.go` and `schemas/utils.go` (kept upstream `IsOpenAIModel` / `IsOpenAIModelFamily` + our `IsElevenlabsSoundModel*` helpers). PR should no longer be `DIRTY`.

Note: latest `dev` currently has a duplicate `ResponsesMessage.UnmarshalJSON` in `responses.go` (upstream #4121 area) that blocks local `go build ./...` — unrelated to this PR.

SecretSun:
**Rebased onto `dev` and cleaned up.** The branch is now a single commit touching just the 9 feature files (`+513/-9`) instead of the earlier 124-file diff, and it's back to `MERGEABLE` — no conflicts. `core` + `transports` build, `gofmt`/`go vet` are clean, and the provider unit tests pass.

All outstanding CodeRabbit items are now addressed:

- **Canonical model id in the sound-generation body** — resolved earlier.
- **Prompt usage alongside `AudioSeconds`** — resolved (`core/bifrost.go` backfills `InputChars`).
- **Voice requirement scope** — the `voice` check is back to the original handler-level `400` for all text-to-speech; only sound-effect models (`IsElevenlabsSoundModel`) bypass it, so normal TTS behavior is unchanged. Alias→sound is additionally enforced at the provider layer (`ElevenlabsProvider.Speech`), where the resolved model is known.
- **Attachment filename** — added `speechAttachmentFilename(...)`; the download filename now follows the requested `response_format`/`output_format` (mp3/opus/aac/flac/wav/pcm/ulaw/alaw) instead of hard-coding `speech.mp3`.

Two design questions from #4712 I'd love a maintainer's call on before finalizing:

1. **Approach** — the `Speech()` branch on `eleven_text_to_sound_v2` (this PR) vs a dedicated endpoint vs Passthrough. The Speech branch keeps VK model allowlists, routing, and `RequestType=speech` observability/billing intact, which is why I went this way.
2. **Billing** — is extending `SpeechUsage` with `AudioSeconds` for duration-based SFX billing acceptable? Happy to adjust if you'd prefer a different shape.

Would appreciate a review when you get a chance. 🙏

akshaydeo (approved):

PR #4742, opened 2026-06-27
fix: make provider prefix matching case-insensitive in ParseModelString

### LINKED ISSUE #4395: [Bug]: provider/model prefix silently ignored on OpenAI-compatible routes for providers with uppercase/space in the name
### Prerequisites

### Description

On /v1/chat/completions, the provider/model prefix fails to pin a configured provider whose name contains uppercase letters or spaces; the request silently falls back to the model-catalog resolver (alphabetical-first provider).

Root cause looks like a case mismatch: the config layer lowercases provider names when registering them in the known-provider set (transports/bifrost-http/lib/config.go ~L1173/1249/1288 strings.ToLower(providerName) → RegisterKnownProvider), but the provider config is stored under the original name, while ParseModelString → IsKnownProvider (core/schemas/utils.go) is case-sensitive.

Capitalized prefix B/… → IsKnownProvider("B") false → provider empty → resolver picks alphabetical-first. Prefix ignored.
Lowercase prefix b/… → IsKnownProvider("b") true → provider a, but no config under a → 400 unsupported_operation.

### Possible fix
Lowercase parts[0] before the IsKnownProvider check in ParseModelString, and/or store/register the provider config under the same lowercased key as the registry so the two agree.

### Steps to reproduce

1. Add a custom OpenAI-compatible provider serving gpt-5.5; add another (alphabetically-earlier) provider also serving gpt-5.5.
2. POST /v1/chat/completions {"model":"B/gpt-5.5"} → routed to the other provider (prefix ignored).
3. POST /v1/chat/completions {"model":"b/gpt-5.5"} → 400 {"code":"unsupported_operation"}.

### Expected behavior

B/gpt-5.5 routes to the B provider.

### Actual behavior

capitalized prefix ignored; lowercase prefix → unsupported_operation.

### Affected area(s)

Transports (HTTP)

### Version

Bifrost HTTP v1.5.13 / core v1.5.19

### Environment

```text

```

### Relevant logs/output

```shell

```

### Regression?

_No response_

### Severity

Critical (blocks releases or production)

Purvi09:
## Summary

ParseModelString did a case-sensitive lookup against knownProviders. Built-in providers are stored lowercase (openai, anthropic), so uppercase prefixes like "OpenAI/gpt-4" or "Anthropic/claude-3" were silently ignored — the prefix was not stripped, and the full string was passed as the model name to the backend.

Custom providers, however, are registered in their exact configured case (e.g. NVIDIA). So the fix can't simply lowercase the prefix: lowercasing would both fail to match a display-case custom provider and return a provider value that no longer matches the canonical name used in config, governance, and routing (which silently drops such providers' models from GET /v1/models). Instead, the lookup is made case-insensitive while always returning the canonical registered case.

## Changes

In core/schemas/utils.go:
1)Added knownProvidersLower, a lowercased name → canonical registered case map kept in sync with knownProviders in RegisterKnownProvider / UnregisterKnownProvider.
2)Added ResolveKnownProvider, which matches a prefix case-insensitively (exact case first, then lowercase fallback) and returns the canonical ModelProvider.
3)ParseModelString now resolves the prefix via ResolveKnownProvider, so both built-in providers given in mixed case and custom providers registered in display case are recognized and returned in their canonical case.
4)Added tests in core/schemas/utils_test.go covering mixed-case built-ins, display-case custom providers (NVIDIA/… and nvidia/…), namespace preservation (meta-llama/…), and register/unregister behavior.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [x] Core (Go)
- [ ] Transports (HTTP)
- [ ] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

Start Bifrost with a custom provider configured (e.g. `MyProvider` with `base_provider_type: openai`), then send a request using an uppercase provider prefix:

```sh
curl -s -X POST http://localhost:8080/v1/chat/completions \
  -d '{"model": "MyProvider/gpt-4", "messages": [{"role": "user", "content": "hi"}]}'
```

**Without fix:** `"resolved_model_used":"MyProvider/gpt-4"` — prefix not stripped, full string sent as model name.

**With fix:** `"resolved_model_used":"gpt-4"` — prefix correctly stripped, only model name sent to backend.

## Screenshots/Recordings

N/A

## Breaking changes

- [ ] Yes
- [x] No

## Related issues

Closes #4395

## Security considerations

None.

## Checklist


Xytronix:
@Purvi09 Your PR introduces a regression for custom providers registered in non-lowercase display case which silently drops them from v1/models and makes the models unroutable by prefixed id.


### Issues
1. `IsKnownProvider` is backed by `knownProviders`, a set keyed by the **exact registered case** (`RegisterKnownProvider(provider)` stores `string(provider)` verbatim; custom providers are commonly registered in display case). Lowercasing the prefix before the lookup means `IsKnownProvider("nvidia")` is **false** for a provider registered as `NVIDIA` → the `/`-split is skipped → `ParseModelString("NVIDIA/meta/llama-3.1-8b-instruct", "")` returns `("", "NVIDIA/meta/llama-3.1-8b-instruct")`.
2. Even when it does match, it returns the **lowercased** `ModelProvider`, which no longer equals the provider key used everywhere else (config `config_providers.name`, governance `provider_configs.provider`, routing), so downstream equality checks fail.

Downstream impact — `governance.filterModelsForVirtualKey` (`plugins/governance`, run from `PostLLMHook` on `ListModelsRequest`) parses each `model.ID` with `ParseModelString`, then keeps a model only if some `pc.Provider == string(parsedProvider)`. With `parsedProvider == ""`, **every** model for a display-case custom provider is dropped, so `GET /v1/models` returns 0 for it. Providers whose canonical key is already lowercase (`openai`, `anthropic`, and lowercase-named custom providers) are unaffected, which masks the regression.

### Steps to reproduce

1. Register a custom OpenAI-compatible provider whose name has uppercase letters or a space, e.g. `NVIDIA` (base `https://integrate.api.nvidia.com`, `list_models` enabled) with a valid key.
2. Grant it on a virtual key with `allowed_models: ["*"]`.
3. Observe **0 models** for `NVIDIA`, while `GET /api/models?provider=NVIDIA&unfiltered=true` (admin, ungoverned) shows the full list and the response `key_statuses` is `success`.

### Alternative fix
I propose the following fix instead:


```go
// lower(name) -> canonical registered name
var knownProvidersLower = /* built from StandardProviders; updated in Register/UnregisterKnownProvider */

func ResolveKnownProvider(p string) (ModelProvider, bool) {
    if knownProviders[p] { return ModelProvider(p), true }
    if c, ok := knownProvidersLower[strings.ToLower(p)]; ok { return ModelProvider(c), true }
    return "", false
}

func ParseModelString(model string, def ModelProvider) (ModelProvider, string) {
    if strings.Contains(model, "/") {
        parts := strings.SplitN(model, "/", 2)
        if len(parts) == 2 {
            if canonical, ok := ResolveKnownProvider(parts[0]); ok {
                return canonical, parts[1]
            }
        }
    }
    return def, model
}
```

### Relevant logs

```text
# admin, ungoverned — provider HAS models:
GET /api/models?provider=NVIDIA&unfiltered=true      -> total 121
# inference via virtual key on a build with PR #4742 — 0, key success:
GET /v1/models?provider=NVIDIA (Authorization: Bearer sk-bf-…)
{ "data": [], "key_statuses": [{"status":"success","provider":"NVIDIA"}] }
```

PR #4749, opened 2026-06-28
fix: preserve tool search responses API round-trip for Codex

### LINKED ISSUE #4713: Responses streaming drops tool_search_call items because arguments is typed *string (object args fail to parse)
# Responses streaming drops `tool_search_call` items because `arguments` is typed `*string` (object args fail to parse)

## Summary

On the OpenAI `/v1/responses` **streaming** path, Bifrost types every tool-call `arguments` field as a string (`ResponsesMessage.Arguments *string`). A `tool_search_call` output item carries `arguments` as a **JSON object** (per the Responses API — only `function_call.arguments` is a string). The strict SSE parser therefore fails to unmarshal the event and **silently drops it**, which also drops the `response.completed` that follows — so the client sees the stream end without completion.

## Reproduction

Drive any `/v1/responses` stream through Bifrost where the model emits a `tool_search_call` (deferred-tool discovery). The upstream model/backend sends:

```json
{"type":"tool_search_call","status":"completed","call_id":"call_…","arguments":{"query":"…","limit":20}}
```

## Actual behavior

Bifrost's streaming parse fails on the object and `continue`s past the event (per-event drop is only a `warn`):

```
WARN Failed to parse stream response: Mismatch type string with value object
```

Because the `output_item.added` / `output_item.done` for the tool_search_call AND the trailing `response.completed` are dropped, strict clients (OpenAI Codex CLI, `wire_api = "responses"`) see:

```
stream closed before response.completed
```

…and reconnect/fail. The request is logged as a success at the request level (the drop is only a per-event warn), so it is invisible in the request log — only visible in the Bifrost process log.

## Expected behavior

`tool_search_call.arguments` is relayed unchanged as a JSON **object**; the item and the following `response.completed` are forwarded to the client.

## Root cause

`ResponsesMessage.Arguments` is typed `*string` (`core/schemas/responses.go`), and the strict streaming unmarshal (sonic) rejects an object value for that field. The wire-format distinction is correct for `function_call` (string) but wrong for `tool_search_call` (object), so any turn that uses tool search is dropped.

## Reference

OpenAI Python SDK: `ResponseToolSearchCall.arguments: object` vs `ResponseFunctionToolCall.arguments: str`.

## Environment

- Bifrost: `core` Responses provider (OpenAI-compatible passthrough), streaming SSE
- Client: OpenAI Codex CLI `0.142.2`, `wire_api = "responses"`

Shaik-Sirajuddin:
Codex's tool-search Responses API calls broke on round-trip because Bifrost re-marshaled `tool_search_call`/`tool_search_output` fields incorrectly or dropped them in deep-copy paths.

Fixes #4713

## Details

- Re-emits `tool_search_call.arguments` as a JSON object instead of a quoted string.
- Preserves and always emits `execution` for `tool_search_call`/`tool_search_output`.
- Preserves `tool_search_output.tools[].type` on request round-trip.
- Preserves tool-search fields in shared response deep-copy paths used by streaming and schema helpers.

## Checklist

- [x] Added tests
- [ ] Reproduced against live OpenAI API


Shaik-Sirajuddin:
multi_agent v1 tools are made available via deferred tool_search tool 
Pr addresses multi agent usability

PR #4772, opened 2026-06-29
fix(mcp): bound per-call ephemeral connect so a dead per-user MCP fails fast

terafin:
## Problem

The per-call ephemeral connect used by per-user-auth MCP clients (`AcquireClientConn`) reuses `MCPClientConnectionEstablishTimeout` — the **30s shared-connection STARTUP** bound. A per-user client opens a fresh upstream transport on **every** tool call, so a single dead/unreachable per-user upstream adds ~30s to every invocation, stacking to 90–120s per-turn stalls when an agent turn invokes several dead per-user MCPs (observed with OpenWebUI fronting multiple per-user MCPs).

This is the per-*call* connect path specifically — not the shared-startup connect (which the 30s bound is correct for), and not auth-required (which already fast-fails with a submit/auth URL).

## Fix

- **`MCPPerCallConnectTimeout` (5s):** the per-call `Initialize` now fails fast instead of inheriting the 30s startup bound.
- **Short-cooldown circuit-breaker (`MCPPerCallConnectBreakerCooldown`, 30s):** after a connect-acquisition failure for a client, subsequent per-call connects for that client fail instantly, bounding per-turn cost to ~one timeout regardless of how many dead-tool calls the agent makes. Auto-recovers after cooldown.
- **Auth-required does NOT trip the breaker** — missing/stale per-user creds are a fast, actionable result (submit/auth URL), not a connect failure.

## Test

`core/mcp/peruser_connect_timeout_test.go` (new): a blackhole endpoint exercises the per-call connect and asserts the 1st call is bounded (`<10s`, was ~30s) and the breaker makes the 2nd call instant (`<1s`). Fail-first verified: 30.001s → 5.001s (timeout) / ~6µs (breaker).

`go build ./mcp/`, `go test ./mcp/` (full package), and `gofmt` all clean on current `dev`.


R-droid101 (commented):
Hey, can you resolve the AI code comments from greptile?

terafin:
Thanks for the review @R-droid101 — resolved the greptile comments in ff353f8:

**Fixed in code**
- **Plugin Denials Trip Breaker (P1):** a `connectAttempted` flag now gates the breaker — it flips true only once `client.Start`/`Initialize` is actually about to run. A `PreConnectionHook` policy denial, a credential-resolution error, or a transport/TLS build error all fail *before* the upstream is contacted and no longer arm the breaker, so a denial can't fast-fail later valid per-user calls against a healthy server for the whole cooldown. Added `TestPerUserConnectBreakerNotTrippedByPreConnectFailure` (RED before / GREEN after).
- **Blackhole Connections Stay Open (P2):** the test now closes every accepted `net.Conn` at teardown (previously only the listener was closed → FD leak on repeated runs), and `held` is mutex-guarded (also clears a `-race` finding).

**Intentional trade-offs of the per-call model — documented rather than changed (happy to turn either into code here or a follow-up, your call):**
- **Slow Initialize Poisons Client (P2):** tripping on an init-timeout is the PR's core purpose — the target is exactly "TCP open succeeds but JSON-RPC `initialize` never returns" (a hung/sleepy upstream), which is indistinguishable at connect time from a legitimately-slow server. So we fail-fast and auto-recover after the 30s cooldown. For a known-slow upstream, a per-client connect-timeout override is the clean follow-up.
- **Parallel Calls Miss Breaker (P1):** each per-user call opens its *own* ephemeral connection, so concurrent connects can't be coalesced into a shared result the way a persistent client could. An initial concurrent batch to a dead upstream each hits the ~5s timeout in parallel (bounded wall-clock ≈ one timeout, not serial), after which the breaker fast-fails every subsequent call for the cooldown. A singleflight-style connect coalescer would trim that transient batch cost — glad to add it here or as a separate PR if you'd like.

All `core/mcp` tests pass under `-race`.

PR #4829, opened 2026-07-01
fix: Anthropic - default unrecognized Claude models to adaptive thinking

### LINKED ISSUE #4828: New Claude models need a Bifrost patch, and then you need to upgrade to get it
Anthropic's newer models, Opus 4.7 and up, Fable, Sonnet 5, only run on adaptive thinking. No `budget_tokens`, no `temperature`/`top_p`/`top_k`. Bifrost has to know that per model, so it keeps an allowlist in `core/providers/anthropic/utils.go`. When a new model ships and isn't on that list yet, Bifrost downgrades it to `budget_tokens` and the request 400s.

This has happened four times now: #2773 (Opus 4.7), #4224 (Fable/Mythos), #3868 (Opus 4.8), #4812 (Sonnet 5). Same fix each time, a new entry in the allowlist.

The annoying part isn't waiting for the PR. It's that once the PR lands, you still have to upgrade your Bifrost server to get it. So trying a new model isn't "point it at the new model id", it's "wait for the PR, then upgrade your deployment, then try the model." That's a lot of friction just to try something new, and Anthropic's direction is adaptive-only becoming the default going forward, so this keeps happening.

Fix: flip the default. Assume a Claude model supports adaptive thinking unless it's on a small, closed list of known legacy models (Haiku, Claude 2.x/3.x, Opus/Sonnet 4.x below .6). New models then work without a patch or an upgrade, and everything already on the allowlist keeps behaving exactly as it does today. I have this ready, PR linked below.

adrianlyjak:
## Summary

`SupportsAdaptiveThinking` / `IsAdaptiveOnlyThinkingModel` were allowlists — an unrecognized Claude model id fell through to `false` and got downgraded to `budget_tokens` extended thinking. Every Claude model since Opus 4.7 (including Sonnet 5) rejects `budget_tokens` and sampling params with a 400, so a new/unlisted model would 400 on first use instead of working.

## Changes

- Reversed the default: an unrecognized Claude model is now assumed to support adaptive thinking instead of being downgraded.
- Added a small closed legacy denylist (Haiku 4.5, Claude 2.x/3.x, Opus/Sonnet 4.x below the `.6` minor) via a new `claudeOpusSonnet4Minor` parser, so every currently-known model keeps its exact prior behavior — only genuinely new/unrecognized Claude ids flip to adaptive.
- Bedrock reuses the same helpers, so it inherits the fix with no separate change needed.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [x] Core (Go)
- [ ] Transports (HTTP)
- [x] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

```sh
cd core
go build ./...
go test ./providers/anthropic/... ./providers/bedrock/...
go vet ./providers/anthropic/... ./providers/bedrock/...
```

## Breaking changes

- [ ] Yes
- [x] No

All currently-known model ids (Haiku 4.5, Claude 2.x/3.x, Opus/Sonnet 4.0–4.5, dual-mode 4.6, adaptive-only 4.7+/5+/Fable/Mythos) keep identical behavior. Only genuinely new/unrecognized Claude model ids change (adaptive instead of a 400).

## Related issues

Closes #4828

## Security considerations

None.

## Checklist

- [x] I read `docs/contributing/raising-a-pr.mdx` and followed the guidelines
- [x] I updated `core/changelog.md`
- [ ] UI build/tests (not applicable — no UI changes)

### DIFF
--- a/core/providers/anthropic/utils_test.go
+++ b/core/providers/anthropic/utils_test.go
@@ -7,6 +7,7 @@ import (
 	"reflect"
 	"slices"
 	"sort"
+	"strings"
 	"testing"
 	"time"
 
@@ -2291,6 +2292,24 @@ func TestSupportsAdaptiveThinking(t *testing.T) {
 		{"claude-haiku-4-6-20250514", false}, // haiku does not support adaptive
 		{"claude-haiku-4-7-20260401", false}, // haiku, not opus
 		{"claude-haiku-4-8-20260601", false}, // haiku, not opus
+		{"claude-haiku-4-9-20270101", false}, // future haiku still legacy
+		// Fail-open: new/unrecognized Claude models default to adaptive.
+		{"claude-opus-5", true},
+		{"claude-opus-5-20270101", true},
+		{"claude-sonnet-6", true},
+		{"claude-opus-4-9", true},                  // future Opus 4.x minor >= 6
+		{"claude-sonnet-7-20270101", true},
+		{"global.anthropic.claude-opus-5", true},
+		// Legacy stays budget_tokens-only (must NOT flip to adaptive).
+		{"claude-sonnet-4-20250514", false}, // bare Sonnet 4.0 (dated)
+		{"claude-opus-4-20250514", false},   // bare Opus 4.0 (dated)
+		{"claude-opus-4-1", false},
+		{"claude-opus-4-0", false},
+		{"anthropic.claude-opus-4-v1", false}, // bare Opus 4.0 (bedrock)
+		{"claude-3-7-sonnet", false},
+		{"claude-3-5-sonnet-20241022", false},
+		{"claude-3-opus", false},
+		{"claude-2.1", false},
 		{"", false},
 	}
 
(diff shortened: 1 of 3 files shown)


R-droid101 (commented):
Love the inversion in this PR where adaptive thinking is the default and only legacy models are excluded. It feels much more scalable than adding support model by model.

This also aligns well with the direction we're planning for supporting adaptive thinking going forward:
- Use our internal model parameter datasheet to determine support.
- Fall back to this logic if the datasheet doesn't have the information.

Could you address the Greptile comment before we approve?

core/providers/anthropic/utils_test.go
```diff
@@ -2288,9 +2289,35 @@ func TestSupportsAdaptiveThinking(t *testing.T) {
...
-		{"claude-haiku-4-7-20260401", false}, // haiku, not opus
-		{"claude-haiku-4-8-20260601", false}, // haiku, not opus
+		// Haiku 4.5 (the latest Haiku) and 3.x Haiku stay budget_tokens-only.
+		{"claude-haiku-4-5", false},
+		{"claude-haiku-4-5-20251001", false},
+		{"anthropic.claude-haiku-4-5-20251001-v1:0", false},
+		{"claude-3-5-haiku-20241022", false},
+		{"claude-3-haiku", false},
+		// A future/adaptive Haiku fails open like every other new model — the
+		// blanket "haiku" denylist would have wrongly downgraded these.
```
R-droid101: I would suggest not adding tests for model versions that aren't out yet. We don't know for sure if these model versions would definitely support adaptive thinking. Similarly, could you remove test cases for other future model versions added here?

R-droid101 (approved):

R-droid101:
Hey! After some further internal discussion, we've decided to re-architect this as part of a broader datasheet framework we're introducing, so we're going to hold off on merging this PR for now.

@TejasGhatte will be working on this, and the work you've done here will definitely help inform that implementation.

Thanks a lot for taking the time to work on this and put together the PR!

PR #4836, opened 2026-07-02
[feat]: virtual-keys endpoint: add user_id filter and remove from_memory in-memory path

### LINKED ISSUE #4793: [Feature]: Virtual keys endpoint: add user-scoped or search filter to avoid full team key fetch
### Prerequisites

## Problem

  The `/api/governance/virtual-keys` endpoint returns all keys for a team in a single
  response. We are currently fetching every key for a team (10,000+) and filtering
  client-side to find the key for a specific user. This is slow and wastes bandwidth.

  **Current call:**
  GET /api/governance/virtual-keys?limit=100&offset=0&team_id=&from_memory=true
  Returns 10,537 virtual keys. We then scan the list for a key where the `name` field
  contains `USERID:<target_user_id>`.

  

## Request

  Add one or both of the following:

  **Option A — filter by user**
  GET /api/governance/virtual-keys?team_id=&user_id=
  Returns only the virtual key(s) for that user on the given team.

  **Option B — search/filter parameter**
  GET /api/governance/virtual-keys?team_id=&search=USERID:
  Filters results by a substring or pattern match on the key `name` field.

  ## Why

  Fetching 10k+ keys per request to find one key is not viable at scale. A server-side
  filter would cut response size from ~1MB+ down to a single record.

  

## Alternatives tried

  **Pagination** -- we tried paginating through the results but the team key count
  (10,537+) made this impractical. Iterating pages to find a single user key caused
  request timeouts before we completed the scan.

  **`from_memory=true`** -- this flag helped with response time on smaller teams but
  still times out on large teams where the in-memory dataset itself is too big to
  serialize and return in one shot.

  **Caching the full list client-side** -- we cache the response, but the initial
  load still times out for large teams and the cache invalidation adds complexity we
  would rather not carry.

  None of these are reliable at the team sizes we are seeing in production.

### Area(s)

Core (Go)

## Context

  We are building per-user virtual key lookups as part of a portfolio token resolution
  flow. Each user has one key per team, identified by `USERID:<uuid>` in the key name.
  Right now we pull the entire team key list on every request.

jimseiwert:
## Summary

- Remove the `from_memory` GovernanceManager block from `GET /api/governance/virtual-keys`, which returned all virtual keys from in-memory state with no filter or pagination support. The param is now a no-op and falls through to the regular DB-backed paths.
- Add a new `user_id` query parameter that filters results to virtual keys whose `created_by_user_id` matches the given value (Option A from the issue).
- Fix 2 pre-existing failing tests that already encoded the expected DB-backed behavior for the `from_memory` path.

## Type of change

- [x] Feature
- [ ] Bug fix
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [x] Core (Go)
- [x] Transports (HTTP)
- [ ] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

```sh
# Run affected tests
go test ./transports/bifrost-http/handlers/ -run TestGetVirtualKeys -v
go test ./framework/configstore/ -run TestGetVirtualKeysPaginated -v
```

Filter by user ID (new):
```
GET /api/governance/virtual-keys?user_id=<user-uuid>
```

Filter by name substring combined with team (existing, now also works when from_memory param is present):
```
GET /api/governance/virtual-keys?team_id=<id>&search=USERID:<user-uuid>
```

## Breaking changes

- [x] No

`from_memory=true` continues to return the same logical dataset via the DB store. Response shape is identical.

## Related issues

Closes #4793

## Security considerations

No auth, secrets, or PII changes. The `user_id` filter is an exact match on `created_by_user_id`, a column already exposed in list responses.

## Checklist

### DIFF
--- a/framework/configstore/rdb.go
+++ b/framework/configstore/rdb.go
@@ -3134,6 +3134,9 @@ func (s *RDBConfigStore) GetVirtualKeysPaginated(ctx context.Context, params Vir
 		search := "%" + strings.ToLower(params.Search) + "%"
 		baseQuery = baseQuery.Where("LOWER(name) LIKE ?", search)
 	}
+	if params.UserID != "" {
+		baseQuery = baseQuery.Where("created_by_user_id = ?", params.UserID)
+	}
 
 	// Get total count before pagination
 	var totalCount int64
--- a/framework/changelog.md
+++ b/framework/changelog.md
@@ -1,1 +1,2 @@
+[feat]: add UserID filter to VirtualKeyQueryParams and GetVirtualKeysPaginated for created_by_user_id lookups [@jamesseiwert](https://github.com/jamesseiwert)
 [fix]: rebuild token usage from denormalized columns in hybrid log list [@G-XD](https://github.com/G-XD)
--- a/framework/configstore/rdb_test.go
+++ b/framework/configstore/rdb_test.go
@@ -1872,6 +1872,61 @@ func TestGetVirtualKeysUsesInternalPagination(t *testing.T) {
 	require.Equal(t, fmt.Sprintf("vk-page-%04d", totalVirtualKeys-1), virtualKeys[len(virtualKeys)-1].ID)
 }
 
+// TestGetVirtualKeysPaginatedFiltersByUserID verifies that the UserID filter
+// returns only virtual keys whose created_by_user_id matches.
+func TestGetVirtualKeysPaginatedFiltersByUserID(t *testing.T) {
+	store := setupRDBTestStore(t)
+	ctx := context.Background()
+
+	userA := "user-aaa"
+	userB := "user-bbb"
+
+	vk1 := &tables.TableVirtualKey{
+		ID:              "vk-user-a-1",
+		Name:            "Key for User A",
+		Value:           "vk-val-a-1",
+		IsActive:        schemas.Ptr(true),
+		CreatedByUserID: &userA,
+	}
+	vk2 := &tables.TableVirtualKey{
+		ID:              "vk-user-a-2",
+		Name:            "Another Key for User A",
+		Value:           "vk-val-a-2",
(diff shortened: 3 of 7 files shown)


framework/configstore/rdb.go
```diff
@@ -3134,6 +3134,9 @@ func (s *RDBConfigStore) GetVirtualKeysPaginated(ctx context.Context, params Vir
 		search := "%" + strings.ToLower(params.Search) + "%"
 		baseQuery = baseQuery.Where("LOWER(name) LIKE ?", search)
 	}
+	if params.UserID != "" {
+		baseQuery = baseQuery.Where("created_by_user_id = ?", params.UserID)
```
akshaydeo, line 3138: the aim here is to get virtual keys created by user or attached to user - cause both concepts dont exist in OSS and we track them in parallel in our enterprise version.

so this flow would mess up with what we have in enterprise (as same base is used there). 

Im thinking how can we accomodate both

jimseiwert: I could drop that part and just do the name filtering? In enterprise even though I link a key to a user and a team, the call to get virtual keys for a user returns an empty array (unless I am doing something wrong calling get keys for user by email)?

akshaydeo:
Hi @jimseiwert — thanks for the contribution! This PR is currently blocked because our CLA bot shows the Contributor License Agreement as not yet signed. Could you sign it here so we can move this forward: https://cla-assistant.io/maximhq/bifrost?pullRequest=4836

jimseiwert:
@akshaydeo - I did sign the first time, repeated attempts I just see this. Please let me know if you need anything from me to proceed.

Screenshot 2026-07-31 at 3 38 35 PM

PR #4862, opened 2026-07-02
retry on query cache fix

akshaydeo:
## Summary

During rolling upgrades, a migration running on one instance can invalidate prepared-statement caches on other instances, causing Postgres to return `SQLSTATE 0A000` ("cached plan must not change result type"). This PR introduces a retrying connection pool wrapper that transparently re-executes statements that fail with this error, so application code does not need to handle it explicitly.

## Changes

- Introduced `retryConnPool`, a `gorm.ConnPool` implementation that wraps `*sql.DB` and retries `ExecContext` and `QueryContext` calls that fail with the cached-plan invalidation error, up to `maxCachedPlanRetries` (3) times. Each failed attempt causes pgx to drop the stale cache entry, so retries re-prepare against the current schema and converge quickly.
- `QueryRowContext` is not retried because `*sql.Row` surfaces its error only at `Scan` time. `PrepareContext` is also not retried. Statements inside explicit transactions bypass the retry intentionally — after an error the transaction is aborted and must be retried as a whole by the caller.
- Unified the `Open` code path so both the password-command and no-password-command cases go through `stdlib.OpenDB` and `openGormFromSQLDB`, eliminating the separate early-return branch.
- Added `isCachedPlanError` which matches both the `0A000` SQLSTATE code and the specific message to avoid false positives from other `feature_not_supported` errors.
- Added `retrypool_test.go` covering `isCachedPlanError`, retry behavior for `ExecContext` and `QueryContext` across success, partial failure, and exhausted-retry scenarios, non-retry behavior for `PrepareContext` and `QueryRowContext`, and `GetDBConn` resolution.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [x] Core (Go)
- [ ] Transports (HTTP)
- [ ] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

```sh
go test ./framework/postgresconn/...
```

To validate end-to-end, run a rolling upgrade scenario where a migration adds a column while the application is serving traffic and confirm that queries no longer fail with "cached plan must not change result type".

## Breaking changes

- [x] No

## Security considerations

The retry logic re-executes the original query and arguments unchanged. No credentials, PII, or secrets are introduced or logged. Password-command behaviour is unaffected.

## Checklist

### DIFF
--- a/framework/postgresconn/postgresconn.go
+++ b/framework/postgresconn/postgresconn.go
@@ -99,30 +99,28 @@ func BuildDSN(config *Config) string {
 
 // Open opens a *gorm.DB against the configured Postgres instance.
 func Open(dsn string, config *Config, logger gormlogger.Interface) (*gorm.DB, error) {
-	if config.PasswordCommand == nil {
-		return gorm.Open(postgres.New(postgres.Config{DSN: dsn}), &gorm.Config{
-			Logger: logger,
-		})
-	}
-
 	pgxConfig, err := pgx.ParseConfig(dsn)
 	if err != nil {
 		return nil, err
 	}
-	sqlDB := stdlib.OpenDB(*pgxConfig, stdlib.OptionBeforeConnect(func(ctx context.Context, connConfig *pgx.ConnConfig) error {
-		password, err := RunPasswordCommand(ctx, config.PasswordCommand)
-		if err != nil {
-			return err
-		}
-		connConfig.Password = password
-		return nil
-	}))
-	return openGormFromSQLDB(sqlDB, logger)
+	var opts []stdlib.OptionOpenDB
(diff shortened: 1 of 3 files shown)


akshaydeo:
* **#4862** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/4862?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/maximhq/bifrost/4862?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `dev`


<h2></h2>

PR #4869, opened 2026-07-03
[fix]: sweep orphaned deferred spans in trace store TTL cleanup

### LINKED ISSUE #4868: Memory leak: orphaned deferred spans in TraceStore are never TTL-swept
## Describe the bug

`TraceStore.deferredSpans` (framework/tracing/store.go) entries are only removed by `CompleteTrace` or `ClearDeferredSpan`. Both depend on a streaming request reaching its normal terminal path:

- `completeDeferredSpan` (core/providers/utils/utils.go) runs in the provider stream goroutine on the final chunk / error, and
- `CompleteTrace` runs from the transport's trace completer.

If neither runs, the entry leaks for the process lifetime. The TTL sweeper `cleanupOldTraces` iterates **only** `s.traces` — it reaps the trace itself after the TTL but never touches `deferredSpans`.

Concrete orphan paths:

1. `ProcessAndSendResponse` (core/providers/utils/utils.go) returns early when the final chunk's `GateSendChunk` fails while the context is **not** cancelled — `completeDeferredSpan` is skipped. The cancellation defers only recover this when `ctx.Err() != nil`.
2. The HTTP transport's SSE producer goroutine invokes the trace completer in a defer, but if the goroutine is never spawned (handler error after `StoreDeferredSpan`) or `traceCompleter` is nil, completion never happens.
3. A provider stream goroutine that dies without emitting a terminal chunk.

Each leaked `DeferredSpanInfo` can pin the **full accumulated streaming response** via `AccumulatedResponse` (set by `SetAccumulatedResponse` at final-chunk time), so on long-running deployments with streaming traffic and client disconnects this accumulates as a slow, unbounded RSS climb.

## Environment

- Bifrost v1.6.2 (also present on current `dev`), self-hosted via `maximhq/bifrost` Docker image on ECS Fargate
- Postgres config/logs stores, governance enabled
- Workload: streaming chat completions proxied to Bedrock, with intermittent client disconnects and client-side retry loops

## Expected behavior

Orphaned deferred spans should be reaped by the same TTL cleanup that reaps orphaned traces.

## Proposed fix

Sweep `deferredSpans` in `cleanupOldTraces` using `DeferredSpanInfo.StartTime` against the same cutoff. Deleting an entry early is safe: `GetDeferredSpanHandle` returns nil and `completeDeferredSpan` becomes a no-op — the same semantics as a TTL-reaped trace today. PR incoming.

citrocat:
## Summary

`TraceStore.deferredSpans` entries are only removed by `CompleteTrace` or `ClearDeferredSpan`, both of which require a streaming request to reach its normal terminal path. When a stream never gets there — the final chunk's `GateSendChunk` fails without a cancelled context (`ProcessAndSendResponse` returns before `completeDeferredSpan`), the SSE producer goroutine is never spawned, or the provider goroutine dies before emitting a terminal chunk — the entry leaks for the process lifetime. `cleanupOldTraces` reaps the orphaned **trace** after the TTL but never touched `deferredSpans`, and each leaked `DeferredSpanInfo` can pin the full accumulated streaming response via `AccumulatedResponse`. On long-running deployments with streaming traffic and client disconnects this shows up as a slow, unbounded RSS climb.

## Changes

- `framework/tracing/store.go`: `cleanupOldTraces` now also sweeps `deferredSpans`, deleting entries whose `StartTime` exceeds the same TTL cutoff used for traces. The cleanup log line reports both counts.
- `framework/tracing/store_test.go`: new test `TestCleanupOldTraces_RemovesOrphanedDeferredSpans` — an expired deferred span (and its trace) is reaped while a fresh deferred span survives.
- `framework/changelog.md`: changelog entry.

Deleting an entry early is safe: `GetDeferredSpanHandle` returns nil and `completeDeferredSpan` becomes a no-op (nil-handle guard at the top), which matches the existing semantics of a TTL-reaped trace.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Affected areas

- [ ] Core
- [ ] Transports
- [ ] Providers
- [ ] Plugins
- [ ] UI
- [ ] Docs
- [x] Other: framework/tracing

## How to test

```bash
cd framework
go test ./tracing/ -run TestCleanupOldTraces_RemovesOrphanedDeferredSpans -v
go test -race ./tracing/
```

Full `go test -race ./tracing/` passes; `gofmt` clean on the touched files. (The pre-existing `go vet` warning in `tracer_test.go:31` about copying a lock is unrelated to this change and present on `dev`.)

## Breaking changes

None. The sweep only removes entries older than the store TTL (default 60 min in the HTTP transport), long past any legitimate stream completion, and early deletion degrades to the existing no-op semantics.

## Related issues

Closes #4868

## Security considerations

None — reduces retention of accumulated response payloads in memory.

## Checklist

- [x] Tests added for the change
- [x] Changelog updated (`framework/changelog.md`)
- [x] Commit message follows `[type]: description` format
- [x] PR kept small and focused (+49/−3)


citrocat:
Re the CodeRabbit note on long-running streams: this is intentional and consistent with existing behavior — the trace itself is already reaped from `s.traces` at the same TTL cutoff today, so a stream outliving the TTL already loses its trace before this change. The deferred-span sweep only removes the entry that used to outlive that reaping. Operationally the HTTP transport constructs the store with a 60-minute TTL, which comfortably exceeds provider stream timeouts (the core stream idle timeout is 120s). Happy to add a doc comment on `NewTraceStore` noting that `ttl` should exceed the longest expected stream duration if maintainers would like.

framework/tracing/store.go
```diff
@@ -436,8 +436,24 @@ func (s *TraceStore) cleanupOldTraces() {
...
 	})
 
-	if count > 0 && s.logger != nil {
-		s.logger.Debug("tracing: cleaned up %d orphaned traces", count)
+	// Deferred spans are normally removed by CompleteTrace or ClearDeferredSpan.
+	// A streaming request that never reaches either (e.g. the final chunk send
+	// fails without a cancelled context, or the producer goroutine dies before
+	// the trace completer runs) would otherwise leave its entry — and the
+	// accumulated response it pins — in the map forever.
+	deferredCount := 0
```
akshaydeo, line 444: this is a good fallback - but we should ideally handle that streaming case here right? Making sure if we detect failure - we should cancel and complete the loop here cc @Pratham-Mishra04

framework/tracing/store_test.go
```diff
@@ -279,3 +279,32 @@ func TestSetTraceAttribute(t *testing.T) {
...
+	if store.GetDeferredSpan(orphanTraceID) != nil {
+		t.Error("orphaned deferred span should be removed by TTL cleanup")
+	}
+	if store.GetTrace(orphanTraceID) != nil {
+		t.Error("orphaned trace should be removed by TTL cleanup")
+	}
+	if store.GetDeferredSpan(freshTraceID) == nil {
+		t.Error("fresh deferred span should survive TTL cleanup")
+	}
+}
```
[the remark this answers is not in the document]
citrocat: Applied in 99e4e8c0c — the test now also asserts the fresh trace survives the sweep.

framework/tracing/store.go
citrocat: Agreed — this sweep is a backstop, not a substitute for completing the span at failure time. The concrete source-level gap I found while tracing this: in `ProcessAndSendResponse` (core/providers/utils/utils.go), when the **final** chunk's `GateSendChunk` fails while the context is *not* cancelled, the function returns before `completeDeferredSpan` runs — the cancellation defers only recover the case where `ctx.Err() != nil`. That's the main in-tree path that strands a deferred span (plus any transport path where the producer goroutine never runs its completer).

Happy to do the source-level fix as a follow-up PR — detect the failed final-chunk send and synchronously complete/clear the deferred span (and the stream accumulator) right there. I kept it out of this PR since it touches the streaming hot path and deserves its own review, while the sweep also covers failure modes we can't enumerate (e.g. a producer goroutine dying before its defer). Want me to open that follow-up?

akshaydeo (approved):

akshaydeo (approved):

PR #4871, opened 2026-07-03
[fix]: zero pooled ChannelMessage references on release

### LINKED ISSUE #4870: Pooled ChannelMessage retains request body, context, and undelivered response while idle in pool
## Describe the bug

`releaseChannelMessage` (core/bifrost.go) returns `ChannelMessage` objects to `channelMessagePool` with:

1. the embedded `schemas.BifrostRequest` still set — its `ChatRequest` (and sibling request pointers) reference the fully parsed request body;
2. `Context` still set — a `*schemas.BifrostContext` carrying per-request user values (raw bodies, large-response previews, plugin state);
3. the response/error channels put back into their pools while they can still hold an undelivered `*BifrostResponse` / `BifrostError` (they are drained on **Get**, not before **Put**).

An idle pooled message therefore pins request-body-sized and response-sized allocations until the pool entry is next reused or dropped by GC's victim cache. With large request bodies (multi-MB chat histories) and a pool high-water mark in the hundreds-to-thousands (matching provider concurrency), this retains a significant amount of heap indefinitely and inflates the steady-state RSS of long-running deployments.

AGENTS.md's own pool-hygiene rule says every pooled object must have all fields zeroed before `pool.Put()` — `releaseChannelMessage` currently only nils the three channel fields.

## Expected behavior

Releasing a `ChannelMessage` should drop all request-scoped references so pooled objects hold no payload memory while idle.

## Impact analysis

Zeroing on release is behaviorally inert for reuse: `getChannelMessage` overwrites the embedded request on acquire, callers set `Context` immediately after acquire, and the acquire-side channel drains already exist. All `releaseChannelMessage` call sites construct their error returns from local `req`/`ctx` variables, never from `msg` after release.

PR incoming.

citrocat:
## Summary

`releaseChannelMessage` returned `ChannelMessage` objects to the pool with the embedded `BifrostRequest` (pointing at the fully parsed request body) and `Context` (per-request user values) still set, and put the response/error channels back while they could still hold an undelivered `*BifrostResponse` / `BifrostError` (drained on Get, not before Put). Idle pooled objects therefore pinned request- and response-sized allocations until next reuse — with multi-MB request bodies and a pool high-water mark matching provider concurrency, this retains significant heap indefinitely on long-running deployments.

This also aligns the function with the pool-hygiene rule in AGENTS.md ("Every pooled object must have all fields zeroed before pool.Put()").

## Changes

- `core/bifrost.go` — `releaseChannelMessage` now (1) drains `msg.Response` / `msg.Err` before putting the channels back (acquire-side drains kept as defense in depth), and (2) zeroes the embedded `BifrostRequest` and clears `Context` before pooling the message. Comment updated to clarify that `msg.BifrostRequest` is a value copy, so zeroing it does not affect the separately-pooled `*BifrostRequest` used for fallbacks.
- `core/bifrost_test.go` — new test `TestReleaseChannelMessage_ClearsPooledReferences`.
- `core/changelog.md` — changelog entry.

## Why it's safe

Behavior on reuse is unchanged: `getChannelMessage` overwrites `msg.BifrostRequest` on acquire, every caller sets `msg.Context` immediately after acquire, and channels were already drained on acquire. I audited all `releaseChannelMessage` call sites — each constructs its error return from local `req`/`ctx` variables and never reads `msg` after release.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Affected areas

- [x] Core
- [ ] Transports
- [ ] Providers
- [ ] Plugins
- [ ] UI
- [ ] Docs

## How to test

```bash
cd core
go test -race -run TestReleaseChannelMessage -v .
go test -race -count=1 .
```

Full `go test -race` on the core package passes; `gofmt` clean on touched files.

## Breaking changes

None.

## Related issues

Closes #4870

## Security considerations

None — reduces retention of request/response payloads in pooled objects.

## Checklist

- [x] Tests added for the change
- [x] Changelog updated (`core/changelog.md`)
- [x] Commit message follows `[type]: description` format
- [x] PR kept small and focused (+76/−2)


core/bifrost_test.go
```diff
@@ -2920,3 +2920,56 @@ func TestClearAnthropicPassthroughForNonNativeProvider(t *testing.T) {
...
+	case <-respCh:
+		t.Error("pooled response channel should be drained before Put")
+	default:
+	}
+	select {
+	case <-errCh:
+		t.Error("pooled error channel should be drained before Put")
+	default:
+	}
+}
```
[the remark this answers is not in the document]
citrocat: Good catch — added `TestReleaseChannelMessage_ClearsPooledReferences_Streaming` in aa404928b, which pre-populates `msg.ResponseStream` with an undelivered stream handoff and asserts it is drained and the reference cleared on release. `go test -race -run TestReleaseChannelMessage` passes for both variants.

core/bifrost.go
```diff
@@ -7678,9 +7690,17 @@ func (bifrost *Bifrost) releaseChannelMessage(msg *ChannelMessage) {
...
+	// as it is required for fallbacks; msg.BifrostRequest is a value copy, so
+	// zeroing it here only drops this message's references.
 
-	// Clear references and return to pool
+	// Clear all references before returning to the pool. The embedded
+	// BifrostRequest holds pointers to the fully parsed request body and
+	// Context holds per-request user values; leaving them set pins those
+	// allocations for as long as the message sits idle in the pool.
+	// getChannelMessage overwrites both on acquire.
+	msg.BifrostRequest = schemas.BifrostRequest{}
```
akshaydeo, line 7989: This is like allocating a new request again - and defeating pooling right ?

citrocat: No allocation happens here — `ChannelMessage` embeds `schemas.BifrostRequest` **by value**, so `msg.BifrostRequest = schemas.BifrostRequest{}` just zeroes the embedded struct's fields in place (a memset over memory the message already owns). Nothing new goes on the heap.

Pooling is also unaffected on the acquire side: `getChannelMessage` already does `msg.BifrostRequest = req` on every Get, which overwrites the same memory whether or not it was zeroed at release. What the pool saves — reusing the `ChannelMessage` object and its channels — is untouched.

The separately pooled `*schemas.BifrostRequest` (from `bifrostRequestPool`, released in the handle methods for fallbacks) is not touched by this line either; only the message's value copy is cleared, so its pointers (`ChatRequest` → full parsed body, etc.) stop pinning that memory while the message sits idle in the pool.

Net cost: one struct-sized zeroing per release. Net benefit: an idle pooled message no longer retains the last request's multi-MB body and context.

akshaydeo (dismissed):

PR #4878, opened 2026-07-03
feat(vertex): add Vertex AI as a custom-provider base type

### LINKED ISSUE #3502: [Feature]: Allow multiple instances of the same standard provider (e.g. Vertex AI) in the UI
### Prerequisites

### Problem to solve

The UI currently prevents adding a second instance of a standard provider like Google Vertex AI. Once you've added one Vertex provider, the option is greyed out / unavailable when trying to add another.

**Use case:** I need two separate Vertex AI configurations pointing to different GCP projects (e.g., one for production workloads, one for experimentation), each with their own ADC credentials, regions, and model allowlists. The UI only allows one `vertex` entry.

The "Custom Providers" feature ([#249](<https://github.com/maximhq/bifrost/issues/249>)) was implemented to address similar needs, but **Vertex is not in the** `SupportedBaseProviders` **list** (`core/schemas/bifrost.go:64-72`), which only includes: `Anthropic, Bedrock, Cohere, Gemini, OpenAI, HuggingFace, Replicate`. So there's no workaround via Custom Providers for Vertex today either.

### Proposed solution

Either (or both):

1. **Add Vertex AI to** `SupportedBaseProviders` so users can create custom providers backed by Vertex (similar to how issue maximhq/bifrost#3131 proposes adding SGLang).
2. **Allow multiple instances of standard providers in the UI** — let users add a second `vertex` entry with a unique name and independent configuration.

### Additional context

* My organization does not use service account API keys — we rely entirely on ADC / Workload Identity, so the `vertex_key_config` with just `project_id` and `region` (no `auth_credentials`) is the only auth path available to us.
* The `config.json` provider setup already supports multiple keys within a single provider for load balancing, but that doesn't cover the case of completely separate GCP projects with different auth contexts.
* Related closed issues: [#243](<https://github.com/maximhq/bifrost/issues/243>), [#249](<https://github.com/maximhq/bifrost/issues/249>)

Shaik-Sirajuddin:
Users can have multiple instances of Vertex AI configured with Bifrost through Custom Providers, each with independent auth mechanisms.

Adds Google Vertex AI to `SupportedBaseProviders`, enabling multiple independently-named Vertex instances (each with its own project/region/credentials) via the existing Custom Providers mechanism.

Closes #3502

## Checklist

- [x] Added tests
- [x] Reproduced against live GCP Vertex AI — 3 real projects, real OAuth2 service-account credentials, multi-key weighted routing across separate projects/regions, per-key error isolation, and confirmed no cross-instance config bleed between a standard `vertex` provider and a custom one


Shaik-Sirajuddin:
Screenshot: "Add Custom Provider" flow with Vertex AI selected.

Vertex custom provider modal

Confirms:
- Base URL field correctly hidden (Vertex ignores `NetworkConfig.BaseURL`)
- Keyless toggle correctly hidden (Vertex can't be keyless)
- Path-override settings icons correctly absent on every toggle
- `Allowed Request Types` correctly reflects Vertex's real operation support — Text Completion/Speech greyed out, Chat Completion/Responses/Embedding/Batch*/File* enabled

Shaik-Sirajuddin:
Tested on load balancing works with native vertex , custom provider with vertex

@danpiths Could you review when you get a chance

KamilDziemba (changes requested):
## Code Review: Vertex as a custom-provider base type

Review run via `/code-review max` (finder panel completed; consolidated). The root pattern across these findings: **this PR lets Vertex instances exist under a custom name, but provider-specific branches compare against the literal `schemas.Vertex` — and now silently miss custom-named instances.** The key-selection path was verified correct (both call sites resolve the base type before `selectKeyFromProviderForModelWithPool`); the misses are in body conversion, header filtering, and transport shaping.

1. **Structured-outputs conversion gated on `bifrostReq.Provider == schemas.Vertex`** (`core/providers/anthropic/chat.go:415`). A custom-named Vertex provider (`my-vertex`, base_provider_type=vertex) sending `response_format` (json_schema) takes the native `output_config.format` branch that the comment says Vertex rejects — upstream 400, while the identical request to the standard vertex provider succeeds (converted to a forced tool). The re-strip in `requestbuilder.go:545` doesn't save it: `stripUnsupportedAnthropicFields` never strips `OutputConfig.Format`.

2. **`isVertex := provider == schemas.Vertex` false for custom Vertex** (`core/providers/gemini/responses.go:3075, 3314`). Multimodal function-response media parts get built Gemini-API-style (inline) instead of Vertex's `$ref` form — the exact divergence the comment documents between the two APIs.

3. **Streaming beta-header filtering disabled for custom Vertex** (`core/providers/vertex/vertex.go:826, 905, 1428`). `ChatCompletionStream`/`ResponsesStream` pass `GetProviderKey()` — now the custom name — as `providerName` to the shared Anthropic streaming handlers; `FilterBetaHeadersForProvider` (`core/providers/anthropic/utils.go:1639`) returns headers **unfiltered** for unknown providers. Non-streaming paths filter correctly via `cfg.Provider=schemas.Vertex`, so stream vs non-stream silently diverge, and unsupported `anthropic-beta` tokens reach Vertex on streams.

4. **Batch response ID base64url encoding keyed on `ExtraFields.Provider == schemas.Vertex`** (`transports/bifrost-http/integrations/openai.go:1583`). The worker stamps the custom name, so the encoding is skipped — full resource names with slashes returned where the RawURLEncoding exists to make IDs path-safe; the decode case is skipped too, so the pair is inconsistent for the same job.

5. **Native Vertex batch raw-body passthrough gated on `createReq.Provider == schemas.Vertex`** (`transports/bifrost-http/integrations/genai.go:725`). Custom Vertex forces the lossy typed conversion for BigQuery output / non-JSONL formats — the path the comment says exists precisely because only raw passthrough round-trips those.

6. **UI: invalid base_url typed under another base format stays after switching to Vertex** (`addNewCustomProviderSheet.tsx:228`) — the hidden field keeps its `.url()` error, the superRefine doesn't cover it, and submit is a dead button with no rendered message.

**Suggested root fix** (covers 1-5): thread the *resolved base provider type* — `customProviderConfig.BaseProviderType` when present — into every `== schemas.Vertex` gate, or centralize a `providerBaseType(p)` resolver next to `GetProviderKey()` and use it at these sites. The key-selection path already does exactly this and is the in-repo precedent.

Also noted (non-blocking): the `CheckOperationAllowed` per-method prelude ×31 follows repo convention (bedrock/gemini/anthropic do the same) — fine as-is; a `schemas.NewNoopLogger()` exists and would replace the 6th hand-rolled test logger copy.

PR #4899, opened 2026-07-04
fix: persist Responses stream usage when providers omit or reuse sequence numbers

### LINKED ISSUE #4846: [Bug]: Responses Stream usage is present in response.completed but not persisted in LLM Logs
### Prerequisites

### Description

Bifrost v1.6.2
Provider: Unsloth
Type: Responses Stream

The final SSE event contains:
response.usage.input_tokens = 9
response.usage.output_tokens = 12
response.usage.total_tokens = 21

But Observability > LLM Logs shows:
Tokens in/out = 0 / 0
Total tokens = 0
Raw JSON = No raw JSON available

Expected:
Bifrost should persist response.usage from response.completed for responses_stream logs.

### Steps to reproduce

Codex CLI managed "Responses Stream", in that case the token count does not work.

### Expected behavior

Bifrost should persist response.usage from response.completed for responses_stream logs.

### Actual behavior

Responses Stream usage is present in response.completed but not persisted in LLM Logs

### Affected area(s)

UI (React)

### Version

v1.6.2

### Environment

```text

```

### Relevant logs/output

```shell

```

### Regression?

_No response_

### Severity

Low (minor issue or cosmetic)

eyeveil:
## Problem

For Responses API streaming requests, the final `response.completed` SSE event contains `response.usage` (e.g. 9/12/21 tokens), but Observability → LLM Logs shows Tokens in/out 0/0, total 0, and "No raw JSON available". Reported with Codex CLI driving a Responses Stream through an OpenAI-compatible provider (Unsloth).

## Root cause

The responses stream accumulator (`framework/streaming`) dedupes chunks by `ChunkIndex` and extracts stream metadata (usage, raw response) from the chunk with the highest index. Some OpenAI-compatible providers omit or reuse `sequence_number`, so the terminal `response.completed` event arrives with `ChunkIndex == 0` (or another already-seen index) — it gets deduped away or never selected as the metadata chunk, and logging persists zero usage.

## Fix

When a terminal responses-stream event arrives with a chunk index that is not greater than the current max, remap it to a reserved stable trailing index (`TerminalResponseChunkIndex`, initialized to -1 like the existing `TerminalErrorChunkIndex` pattern). The terminal event carrying `response.usage` survives dedup and is selected for persistence; duplicate final processing stays idempotent (the reserved index is reused, so the terminal chunk is only stored once).

## Testing

Added `TestProcessStreamingResponsePreservesTerminalResponsesUsageWhenChunkIndexIsReused`:
- prior stream chunks with indexes 0 and 7 (including a duplicate index 7);
- terminal `response.completed` with usage and reused `ChunkIndex == 0`;
- asserts accumulated usage 31/17/48, and that reprocessing the same final event does not append a duplicate terminal chunk.

With the repo go.work workspace (`core` + `framework` + `transports`):
- `go build ./...` (framework) — pass
- `go vet ./streaming/` — pass
- `go test -count=1 ./streaming/` — pass

Note: `go vet ./...` on framework has one pre-existing unrelated finding in `framework/tracing/tracer_test.go` (copies a `schemas.Trace` containing a mutex), present on clean `dev`.

Fixes #4846

### DIFF
--- a/framework/streaming/accumulator.go
+++ b/framework/streaming/accumulator.go
@@ -153,6 +153,7 @@ func (a *Accumulator) createStreamAccumulator(requestID string) *StreamAccumulat
 		MaxTranscriptionChunkIndex: -1,
 		MaxAudioChunkIndex:         -1,
 		TerminalErrorChunkIndex:    -1,
+		TerminalResponseChunkIndex: -1,
 		IsComplete:                 false,
 		mu:                         sync.Mutex{},
 		Timestamp:                  now,
@@ -192,6 +193,7 @@ func (a *Accumulator) getOrCreateStreamAccumulator(requestID string) *StreamAccu
 		MaxTranscriptionChunkIndex: -1,
 		MaxAudioChunkIndex:         -1,
 		TerminalErrorChunkIndex:    -1,
+		TerminalResponseChunkIndex: -1,
 		IsComplete:                 false,
 		mu:                         sync.Mutex{},
 		Timestamp:                  now,
@@ -321,6 +323,15 @@ func (a *Accumulator) addResponsesStreamChunk(requestID string, chunk *Responses
 	// Track first chunk timestamp for TTFT calculation
 	if accumulator.FirstChunkTimestamp.IsZero() {
 		accumulator.FirstChunkTimestamp = chunk.Timestamp
+	}
+	if isFinalChunk && chunk.StreamResponse != nil && chunk.ChunkIndex <= accumulator.MaxResponsesChunkIndex {
+		if accumulator.TerminalResponseChunkIndex >= 0 {
+			chunk.ChunkIndex = accumulator.TerminalResponseChunkIndex
+		} else {
+			accumulator.MaxResponsesChunkIndex++
+			chunk.ChunkIndex = accumulator.MaxResponsesChunkIndex
+			accumulator.TerminalResponseChunkIndex = chunk.ChunkIndex
+		}
 	}
 	if _, seen := accumulator.ResponsesChunksSeen[chunk.ChunkIndex]; !seen {
 		accumulator.ResponsesChunksSeen[chunk.ChunkIndex] = struct{}{}
--- a/framework/streaming/accumulator.go
+++ b/framework/streaming/accumulator.go
@@ -311,6 +311,25 @@ func (a *Accumulator) addAudioStreamChunk(requestID string, chunk *AudioStreamCh
 	return nil
 }
 
+// reserveTerminalChunkIndex returns the stable chunk index reserved for a terminal
+// chunk, seeding the reservation on first call: an index that is already unique
+// (greater than the current max) is reserved as-is, while a reused/lower one gets a
+// fresh trailing index. Duplicate plugin deliveries of the same terminal chunk then
+// remap to the reservation and are dropped by the seen-index dedup. Keep separate
+// reservation fields per terminal kind (error vs response) — merging them would
+// change dedup behavior if a stream ever produced both. Caller must hold mu.
+func (sa *StreamAccumulator) reserveTerminalChunkIndex(field *int, chunkIndex int) int {
+	if *field >= 0 {
+		return *field
+	}
+	if chunkIndex <= sa.MaxResponsesChunkIndex {
+		sa.MaxResponsesChunkIndex++
+		chunkIndex = sa.MaxResponsesChunkIndex
+	}
+	*field = chunkIndex
+	return chunkIndex
+}
+
 // addResponsesStreamChunk adds a responses stream chunk to the stream accumulator
 func (a *Accumulator) addResponsesStreamChunk(requestID string, chunk *ResponsesStreamChunk, isFinalChunk bool) error {
 	accumulator := a.getOrCreateStreamAccumulator(requestID)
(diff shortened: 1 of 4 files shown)


framework/streaming/accumulator.go
```diff
@@ -322,6 +324,15 @@ func (a *Accumulator) addResponsesStreamChunk(requestID string, chunk *Responses
 	if accumulator.FirstChunkTimestamp.IsZero() {
 		accumulator.FirstChunkTimestamp = chunk.Timestamp
 	}
+	if isFinalChunk && chunk.StreamResponse != nil && chunk.ChunkIndex <= accumulator.MaxResponsesChunkIndex {
```
akshaydeo: **CONFIRMED regression (blocking): duplicate delivery of a normal final chunk is double-appended.**

On a monotonic stream the terminal chunk arrives with the highest index N while `MaxResponsesChunkIndex == N-1`, so `chunk.ChunkIndex <= accumulator.MaxResponsesChunkIndex` is **false** on the first delivery and `TerminalResponseChunkIndex` is never seeded. When a second plugin delivers the same terminal chunk (logging and maxim both call `tracer.ProcessStreamingChunk` with the same final result: `plugins/logging/main.go:1082`, `plugins/maxim/main.go:587`), the condition `N <= N` is now true, `TerminalResponseChunkIndex` is still -1, so the else branch mints N+1 and the terminal chunk is stored a **second** time. Its `RawResponse` is then written twice into the persisted raw log (`responses.go:907-913`). Before this PR the seen-set deduped that second delivery.

Fix: reserve the terminal index **unconditionally** on the first `isFinalChunk` (exactly like the error path at `responses.go:948-957`), and only remap when it collides:

```go
if isFinalChunk && chunk.StreamResponse != nil {
	if accumulator.TerminalResponseChunkIndex >= 0 {
		chunk.ChunkIndex = accumulator.TerminalResponseChunkIndex
	} else {
		if chunk.ChunkIndex <= accumulator.MaxResponsesChunkIndex {
			accumulator.MaxResponsesChunkIndex++
			chunk.ChunkIndex = accumulator.MaxResponsesChunkIndex
		}
		accumulator.TerminalResponseChunkIndex = chunk.ChunkIndex
	}
}
```

Please also add a test for the monotonic case (terminal chunk has the highest index, delivered twice); the current test only covers the colliding-index case.

framework/streaming/types.go
```diff
@@ -140,6 +140,8 @@ type StreamAccumulator struct {
 
 	// TerminalErrorChunkIndex holds the reserved chunk index for the terminal error (-1 = unset); reused across plugin calls for correct dedup.
 	TerminalErrorChunkIndex int
+	// TerminalResponseChunkIndex holds the reserved index for a terminal response event when providers omit or reuse sequence numbers.
+	TerminalResponseChunkIndex int
```
akshaydeo, line 144: **Cleanup (non-blocking):** this reservation pattern is now duplicated near-verbatim with `TerminalErrorChunkIndex` (`responses.go:948-957`), and at a different layer: `addResponsesStreamChunk` becomes the only `add*` helper that mutates `chunk.ChunkIndex` (the chat/audio/transcription siblings do not). Consider extracting a shared helper on `StreamAccumulator`, e.g.

```go
// caller holds mu
func (sa *StreamAccumulator) reserveTerminalChunkIndex(field *int) int {
	if *field >= 0 {
		return *field
	}
	sa.MaxResponsesChunkIndex++
	*field = sa.MaxResponsesChunkIndex
	return *field
}
```

and calling it from both terminal paths. Keep the two fields separate: merging them would change dedup behavior if a stream ever produced both an error terminal and a completed terminal.

akshaydeo (changes requested):
## Review summary

The diagnosis is correct and the collision-case test is good, but the reservation guard introduces a confirmed regression in the normal monotonic-index case, and the root cause lives one layer below this fix.

### Findings

| 1 | High | `framework/streaming/accumulator.go` (new block) | Duplicate delivery of a normal (highest-index) final chunk is double-appended instead of deduped (see inline comment) | CONFIRMED |
| 2 | Medium | `core/providers/openai/openai.go:1293`, `:1947` | Root cause is at ingestion: `ExtraFields.ChunkIndex = response.SequenceNumber` verbatim; `SequenceNumber` is a non-pointer `int` so an omitted value becomes 0 for every chunk. Non-terminal delta collisions still silently drop distinct content after this fix (this PR's own test treats the second, distinct index-7 delta as an expected drop). The chat path in the same file already mints a bifrost-side monotonic counter (`chunkIndex++` at ~1384-1386). | CONFIRMED |
| 3 | Low | `framework/streaming/types.go` / `accumulator.go` | Reservation logic is a near-verbatim copy of the `TerminalErrorChunkIndex` block at `responses.go:948-957`, at a different layer (see inline comment) | PLAUSIBLE |

### Recommendation

**Request changes** for finding 1 only; it triggers in the common 2+ plugin configuration (e.g. logging + maxim) and the fix is small. Findings 2 and 3 can land as follow-ups.

### Suggested followups (separate PR)

1. Assign a bifrost-side monotonic counter to `ExtraFields.ChunkIndex` in the Responses stream handlers (`openai.go:1293`, `:1947`), mirroring the chat path, keeping provider `sequence_number` as metadata only. Nothing outside `framework/streaming` derives behavior from `ExtraFields.ChunkIndex` (consumers are pure field copies in `core/schemas/mux.go` / `chatcompletions.go`), and the client-facing value is the separate top-level `SequenceNumber` field, so this is safe and fixes the delta-loss class entirely.
2. Extract a shared `reserveTerminalChunkIndex(field *int)` helper on `StreamAccumulator` used by both the error path and this new path (keep the two fields separate; merging them would change dedup behavior if both terminals ever fire).

### Checked and cleared (for the record)

- Out-of-order arrival defeating the remap: refuted; every provider assigns the terminal chunk the global-max index and max-index selection is arrival-order independent.
- Terminal-error vs terminal-response index race masking errors: refuted; persisted status/`ErrorDetails` come directly from `respErr`, not from the max-index chunk.
- The `StreamResponse != nil` guard dropping usage-only terminals: refuted; `chunk.TokenUsage` is only set inside the branch that guarantees non-nil `StreamResponse` (`responses.go:964-968`).

akshaydeo (approved):

PR #4901, opened 2026-07-04
fix: attribute passthrough virtual keys sent via Azure api-key header

### LINKED ISSUE #4477: [Bug]: Passthrough calls using a Virtual Key log as actual key
### Prerequisites

### Description

Calling the API using a virtual key via passthrough registers the call as if it was done using the original key.

### Steps to reproduce

1. Create a virtual key
2. Call the API at passthrough, using the virtual key

Image

### Expected behavior

Log registers as using the virtual key

### Actual behavior

Log registers as if call was done using the base key

### Affected area(s)

UI (React), Core (Go)

### Version

v1.5.13

### Environment

```text
Using the Azure Open AI API for Foundry.
```

### Relevant logs/output

```shell

```

### Regression?

_No response_

### Severity

eyeveil:
## Problem

Provider-passthrough calls authenticated with a Bifrost Virtual Key are logged/attributed as if made with the underlying base provider key: the VK column in LLM Logs is empty and governance sees the base key. Reported with Azure OpenAI (Foundry). Severity high — VK-level attribution/quotas silently don't apply to passthrough traffic.

## Root cause

`ConvertToBifrostContext` (`transports/bifrost-http/lib/ctx.go`) extracts a virtual key from `x-bf-vk`, `Authorization: Bearer`, `x-api-key`, and `x-goog-api-key` — but not from `api-key`, Azure OpenAI's native auth header. Azure SDKs send the VK as `api-key: sk-bf-...` on passthrough, so `BifrostContextKeyVirtualKey` is never populated, governance never resolves the VK, and logging attributes the call to the selected base key.

## Fix

Recognize the `api-key` header as a virtual-key carrier in `ConvertToBifrostContext`, gated on the `sk-bf-` prefix — exactly mirroring the existing `x-api-key`/`x-goog-api-key` handling:

- real Azure keys (no `sk-bf-` prefix) are untouched;
- the passthrough handler (`handlePassthrough`) already excludes `api-key` from headers forwarded upstream, so the VK value still never leaks to the provider — this change only affects attribution into the governance/logging context.

## Testing

- Added `TestConvertToBifrostContext_VirtualKeyFromAzureAPIKeyHeader` (VK via `api-key` is captured into context) and `TestConvertToBifrostContext_APIKeyHeaderNonVirtualKeyIgnored` (non-VK `api-key` value is not misread as a VK).
- `go vet ./bifrost-http/lib/`, `go test -count=1 -run TestConvertToBifrostContext ./bifrost-http/lib/` (24 tests incl. 2 new) — pass.

Fixes #4477


eyeveil:
updated to the convention mentioned in 4897 — added `api-key` (Azure OpenAI style) to the allowed virtual-key headers list in `docs/features/governance/virtual-keys.mdx`, matching the code change.

plugins/governance/utils.go
```diff
@@ -20,7 +20,7 @@ import (
 //   - *string: The virtual key if found, nil otherwise
 func ParseVirtualKeyFromFastHTTPRequest(req *fasthttp.RequestCtx) *string {
 	vkHeader := string(req.Request.Header.Peek("x-bf-vk"))
-	if vkHeader != "" && strings.HasPrefix(strings.ToLower(vkHeader), VirtualKeyPrefix) {
+	if vkHeader != "" {
 		return bifrost.Ptr(vkHeader)
 	}
```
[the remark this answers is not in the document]
eyeveil: Intentional — it aligns `ParseVirtualKeyFromFastHTTPRequest` with the HTTP transport extractor (`ctx.go` accepts any `x-bf-vk` value) and the documented behavior in `docs/features/governance/virtual-keys.mdx`: \"Old virtual keys (without `sk-bf-*` prefix) are only supported by `x-bf-vk` header.\" The prefix guard here silently broke legacy VKs on this path while the transport accepted them. Only the shared auth headers (Authorization, x-api-key, x-goog-api-key, api-key) keep the prefix gate. Added `TestParseVirtualKeyFromFastHTTPRequest_XBfVkAcceptsLegacyUnprefixedKeys` to pin it.

akshaydeo (dismissed):

plugins/governance/utils.go
```diff
@@ -20,7 +20,7 @@ import (
 //   - *string: The virtual key if found, nil otherwise
 func ParseVirtualKeyFromFastHTTPRequest(req *fasthttp.RequestCtx) *string {
 	vkHeader := string(req.Request.Header.Peek("x-bf-vk"))
-	if vkHeader != "" && strings.HasPrefix(strings.ToLower(vkHeader), VirtualKeyPrefix) {
+	if vkHeader != "" {
```
TejasGhatte: @nnNyx can you revert the change here, this guard is needed, we don't support old vk formats

eyeveil: ctx.go already accepts x-bf-vk with no prefix check (transports/bifrost-http/lib/ctx.go:357, unrelated to this PR), and the doc note right above this says old VKs are x-bf-vk only. reverting just here would make the plugin reject what the transport already accepts for the same header. if we're dropping legacy VK support for real, that guard needs to come out of ctx.go too and the doc note needs to go, that's bigger than this PR's scope.

eyeveil: I've reverted it, the prefix guard is back. For context, it was aligned with the note in docs/features/governance/virtual-keys.mdx; "Old virtual keys (without sk-bf-* prefix) are only supported by x-bf-vk header". Since old formats aren't supported anymore I'm down to drop that note in this PR as well if you want.

rodigu:
this is awesome, thank you mate

akshaydeo (approved):

Issue #3278, opened 2026-05-06
[Bug]: Model Aliases Must Be Double Specified


ccreighton-apptio:
### Prerequisites

- [x] I have searched existing issues and discussions to avoid duplicates
- [x] I am using the latest version (or have tested against main/nightly)

### Description

When available models are restricted, aliases must be double-specified. If they are not, the request fails to route.

### Steps to reproduce

1. Define a provider with restricted models and model aliases, say:
```
  "providers": {
    "bedrock": {
      "keys": [
        {
          "name": "iam",
          "region": "env.AWS_REGION",
          "role_arn": "env.AWS_ROLE_NAME",
          "session_name": "env.AWS_SESSION_NAME",
          "models": [
            "us.anthropic.claude-sonnet-4-6"
          ],
          "aliases": {
            "anthropic.claude-sonnet-4-6": "us.anthropic.claude-sonnet-4-6",
            "foo": "us.anthropic.claude-sonnet-4-6"
          },
          "weight": 1.0
        }
      ]
    }
  }
```
2. Make an inference call using the model alias, `anthropic.claude-sonnet-4-6`

Expected:

Works

Actual:

"no keys found that support model: anthropic.claude-sonnet-4-6"
"provider is required in model field (format: provider/model) — no providers found for model \"foo\" in model catalog to auto-resolve"

---

If the aliases are also added to the model list, this works. This shouldn't be necessary.

If all models are enabled (`*`), this works. This shouldn't be necessary.

### Expected behavior

Model aliases are defined in `aliases` and become available for use

### Actual behavior

Model aliases must be double-specified before they work

### Affected area(s)

Core (Go)

### Version

v1.5.0-prerelease8

### Environment

```text
Docker
```

### Relevant logs/output

```shell
{"is_bifrost_error":false,"status_code":400,"error":{"message":"provider is required in model field (format: provider/model) — no providers found for model \"foo\" in model catalog to auto-resolve"},"extra_fields":{}}
{"is_bifrost_error":false,"error":{"error":"no keys found that support model: anthropic.claude-sonnet-4-6","message":"no keys found that support model: anthropic.claude-sonnet-4-6"},"extra_fields":{"provider":"bedrock","original_model_requested":"anthropic.claude-sonnet-4-6","resolved_model_used":"anthropic.claude-sonnet-4-6","request_type":"chat_completion_stream"}}
```

### Regression?

Dunno

### Severity

Medium (some functionality impaired)

[...]

PR #4905, opened 2026-07-04
fix: always surface configured model aliases in ListModels regardless of key allowlist

### LINKED ISSUE #4170: [Bug]: Configuring provider model aliases breaks model listing
### Prerequisites

### Description

Setting provider model aliases (the new ones, not through the routing rules) breaks model listing
`curl "${BIFROST_URL}/v1/models" -H "Authorization: Bearer ${BIFROST_TOKEN}"`. After aliases are configured it only returns them for the provider

Providers without aliases are not affected.
The ui models tab, is also not affected.

### Steps to reproduce

1. List models with `curl "${BIFROST_URL}/v1/models" -H "Authorization: Bearer ${BIFROST_TOKEN}` to make sure it works. All models are listed here.
2. Create a provider model alias.
3. List models with `curl "${BIFROST_URL}/v1/models" -H "Authorization: Bearer ${BIFROST_TOKEN}` to make sure it does not work. Only aliases are listed for this provider.

### Expected behavior

Provider model aliases being added to the list of available models of the provider in the list models response.

### Actual behavior

Provider model aliases overwriting the list of available models of the provider in the list models response.

### Affected area(s)

Core (Go)

### Version

1.5.7

### Environment

```text

```

### Relevant logs/output

```shell

```

### Regression?

_No response_

### Severity

eyeveil:
## Problem

Configuring provider model aliases breaks `GET /v1/models`: once aliases are configured, the listing stops being the provider's model list with aliases *added* and instead drops models and/or aliases depending on the key's `models` allowlist. Related: #3278 (aliases had to be double-specified in `models` to surface).

## Root cause

The shared ListModels pipeline (`core/providers/utils/models.go`) gated explicitly-configured aliases behind the key's `models` allowlist:

- `FilterModel` filtered *every* candidate — including alias-derived ones — against `AllowedModels`, so an alias whose name wasn't in the allowlist was dropped.
- `BackfillModels` only backfilled aliases in the wildcard (`["*"]`) case, so restricted or empty allowlists never surfaced them.

The allowlist is meant to restrict the provider's own models; it was also hiding the user's explicitly-configured aliases.

## Fix

- `FilterModel`: the allowlist check now applies only to raw provider model IDs. Alias-derived candidates bypass the allowlist (blacklist still wins), so a restricted allowlist never hides configured aliases; alias entries keep the provider's metadata when the target is in the API response.
- `BackfillModels`: two additive passes — restricted-allowlist entries as before (now marking `included` to prevent duplicates), then an unconditional pass backfilling every configured alias not yet surfaced (blacklist still wins, deterministic ordering). Unfiltered listings still skip backfill.

Aliases are now additive to the listing; allowlist filtering of the provider's own models, blacklist precedence, and default `["*"]` behavior are unchanged. Exported signatures unchanged.

## Testing

New offline unit tests (`models_test.go`) cover: wildcard+alias, restricted-allowlist keeps alias (the reported regression), empty-allowlist surfaces alias, empty-allowlist+no-alias stays empty, alias backfill when the target isn't in the API response, blacklist blocks alias, no duplicate when the alias name is also allowlisted, and unfiltered listings.

`go build ./...`, `go vet ./providers/utils/`, `go test -count=1 ./providers/utils/ ./providers/openai/ ./providers/gemini/` — all pass.

Fixes #4170

### DIFF
--- a/core/providers/utils/models.go
+++ b/core/providers/utils/models.go
@@ -246,8 +246,10 @@ func (p *ListModelsPipeline) resolveModelID(modelID string) []aliasMatch {
 //  1. Resolve name — check alias VALUES for a match (uses MatchFns).
 //     If matched: resolvedName = alias KEY, aliasValue = provider ID.
 //     If not matched: resolvedName = original modelID, aliasValue = "".
-//  2. Allowlist check (only when allowlist is restricted, i.e. not wildcard):
-//     Skip if resolvedName is not in AllowedModels.
+//  2. Allowlist check (only for raw provider IDs under a restricted allowlist):
+//     Skip a raw candidate (aliasValue == "") if resolvedName is not in AllowedModels.
+//     Alias candidates (aliasValue != "") are always allowed — the allowlist filters
+//     the provider's own models, never the user's explicitly-configured aliases.
 //  3. Blacklist check (always):
 //     Skip if resolvedName is blacklisted. Blacklist takes precedence over everything.
 //  4. Return one FilterResult per passing candidate.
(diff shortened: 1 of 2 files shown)


eyeveil:
@coderabbitai review

Pratham-Mishra04:
Hey @nnNyx sorry but this issue is already picked up internally so we'll be closing this PR. Plus the correct way here would also require checking model+alias in key selection gating and not just surfacing model names in list models

eyeveil:
Makes sense — thanks for the heads-up, and agreed on the key-selection gating point. Looking forward to the internal fix.

PR #4907, opened 2026-07-04
fix: Vertex list-models quota project, GenAI schema-mismatch errors, and dropped message.images

### LINKED ISSUE #2367: [Feature]: Support for message.images array from OpenRouter image generation models (Gemini Nano Banana / gemini-*-image)
### Prerequisites

### Problem to solve

When using image generation models from OpenRouter (e.g. `google/gemini-2.5-flash-image`, `google/gemini-3-pro-image-preview`, or any "Nano Banana" variant) via Bifrost, the model successfully generates images (visible in usage stats: high `completion_tokens` + `image_tokens` ~1290), but the final response returned by Bifrost contains **only text** in `message.content`.

The actual generated image, which OpenRouter returns in the standard field:
`"message": {
  "role": "assistant",
  "content": "Here's your cat!",
  "images": [
    {
      "type": "image_url",
      "image_url": {
        "url": "data:image/png;base64,..."
      }
    }
  ]
}`
is missing / stripped during response normalization in Bifrost.

### Expected behavior
Bifrost should properly pass through or convert the message.images array (with base64 data URLs) into the unified OpenAI-compatible response, similar to how other multimodal output is handled.
### Current behavior
Only the textual part reaches the client. The image is lost even though it was generated upstream.

### Alternatives considered

_No response_

### Area(s)

Core (Go)

### Additional context

_No response_

### LINKED ISSUE #3468: [Bug]: GenAI streamGenerateContent returns 500 "Invalid JSON" when sonic.Unmarshal fails on GeminiGenerationRequest
### Prerequisites

### Description

Bifrost HTTP GenAI route accepts `POST …/genai/v1beta/models/{model}:streamGenerateContent?alt=sse`. For normal JSON bodies, `integrations/router.go` unmarshals the raw body with `sonic.Unmarshal` into the route’s request type (for chat this is `*gemini.GeminiGenerationRequest`).

When the client is **Vercel AI SDK `@ai-sdk/google`** behind Bifrost (OpenCode / similar), the outbound request can include **large `tools` / `functionDeclarations` / `toolConfig`** payloads that are valid JSON but **`sonic.Unmarshal` into `GeminiGenerationRequest` fails**. The handler then returns **HTTP 500** with a JSON body whose `message` is **`Invalid JSON`**, which is misleading: the payload is usually syntactically valid JSON; the failure is **schema/type mismatch during unmarshal**, not a random malformed body.

The error is surfaced to clients as `AI_APICallError` with `statusCode: 500` and:

`{"error":{"code":500,"message":"Invalid JSON","status":"","details":null}}`

### Steps to reproduce

1. Run Bifrost HTTP with GenAI integration (path prefix e.g. `/genai`, route `POST /genai/v1beta/models/{model:*}`).
2. From a client using **`@ai-sdk/google`** with `baseURL` set to `http(s)://<bifrost>/genai/v1beta` and model id including provider scope, e.g. `models/google/gemini-3-flash-preview`, call **`streamGenerateContent`** with the **same tool list** a full OpenCode / agent stack sends (many function declarations with nested JSON Schema).
3. Observe response **500** and body above.

(Optional minimal path: capture one failing request body from Bifrost access logs and replay with `curl` against the same URL.)

### Expected behavior

- Either the request unmarshals successfully and is forwarded to the upstream Gemini-compatible provider, **or** - Bifrost returns **4xx** with a **clear** error (e.g. “request body incompatible with GenAI schema: &lt;sonic error detail&gt;”) so gateways and SDKs can distinguish **unmarshal** failures from true invalid JSON.

### Actual behavior

- **HTTP 500** with `message: "Invalid JSON"` and no unmarshaling detail in the default error envelope. - Client retries treat it as retryable server error.

### Affected area(s)

Core (Go)

### Version

v1.5.2

### Environment

```text

```

### Relevant logs/output

```shell

```

### Regression?

_No response_

### Severity

Low (minor issue or cosmetic)

--- OpenCode / AI SDK side (error object tail) ---

service=llm
providerID=google
modelID=google/gemini-3-flash-preview
sessionID=<redacted>

url: http://<bifrost-host>:8080/genai/v1beta/models/google/gemini-3-flash-preview:streamGenerateContent?alt=sse

error.name: AI_APICallError
statusCode: 500

responseHeaders (subset):
  server: fasthttp
  content-type: application/json

responseBody (verbatim):
{"error":{"code":500,"message":"Invalid JSON","status":"","details":null}}

--- Notes ---

Request URL shape matches Bifrost GenAI v1beta route; failure is not HTTP 405.
Upstream hypothesis: sonic.Unmarshal into GeminiGenerationRequest failed; error text is generic "Invalid JSON".

### LINKED ISSUE #4106: Vertex AI: model listing fails with ADC user credentials (listModelsByKey omits X-Goog-User-Project / quota project)
**Bifrost version:** HTTP transport `v1.5.8` (Docker `maximhq/bifrost:latest`)

## Summary

With a Vertex AI provider authenticated via **Application Default Credentials** (the UI's *Service Account (Attached)* auth mode), every model-catalog / `ListModels` refresh fails with:

```
failed to list models for provider vertex: Your application is authenticating by using local
Application Default Credentials. The aiplatform.googleapis.com API requires a quota project,
which is not set by default.
```

The provider then shows a permanent red error (`list_models_failed`) in the UI. **Inference is unaffected** — chat/image generation through the same provider returns `200`s. So it's a misleading "broken" badge on a provider that actually works, and the Model-Garden catalog never populates for ADC users. Still reproduces on `v1.5.8` (current latest).

## Root cause — `core/providers/vertex/vertex.go` (`main`)

`listModelsByKey` calls the **project-less** Model Garden endpoint and sets only `Authorization`:

```go
// L245
requestURL := fmt.Sprintf("https://%s/v1beta1/publishers/%s/models?pageSize=%d", host, publisher, MaxPageSize)
// L258
req.Header.Set("Authorization", "Bearer "+token.AccessToken)   // <- no X-Goog-User-Project
```

Because the URL has no `projects/{id}/...` segment, Google can't infer a quota/billing project, and **ADC user credentials require one to be supplied explicitly** (via `x-goog-user-project`). So this call 403s for user creds.

By contrast, within the same file:

- The **inference** path embeds the project in the URL (e.g. L2733 `.../projects/%s/locations/%s/publishers/...`), so it works without the header.
- The **rerank** path already sets the header for its own project-less call (L1656: `req.Header.Set("X-Goog-User-Project", projectID)`) — so the project ID is available in the key config and this is the established pattern.
- `listModelsByKey` itself already reads the key config (L200: `key.VertexKeyConfig.Region.GetValue()`), so the project is in scope.

Note: setting `GOOGLE_CLOUD_QUOTA_PROJECT` (env) or `quota_project_id` in the ADC file (via `gcloud auth application-default set-quota-project`) does **not** help — neither appears to be read by the provider.

## Steps to reproduce

1. Run Bifrost in Docker on a non-GCP host (local mac/linux — no metadata server).
2. Authenticate to Vertex with ADC user creds, and set a quota project:
   ```
   gcloud auth application-default login
   gcloud auth application-default set-quota-project <PROJECT>
   ```
   Mount the ADC file into the container and point `GOOGLE_APPLICATION_CREDENTIALS` at it.
3. Add a Vertex provider with **Service Account (Attached)** auth, Project ID `<PROJECT>`, region `global`.
4. Observe: the provider shows the red error; logs show `failed to list models ... requires a quota project`. Meanwhile a real inference call (e.g. `POST /v1/images/generations` with `vertex/gemini-3.1-flash-image-preview`) returns `200` with `extra_fields.provider=vertex`.

## Logs

```
{"level":"error","message":"failed to list models for provider vertex: ... The aiplatform.googleapis.com API requires a quota project, which is not set by default. ..."}
{"level":"info","message":"models added to catalog"}   // catalog still partially builds; inference unaffected
```

## Expected

`listModelsByKey` should set `X-Goog-User-Project: <key's Vertex project id>` on the `publishers.models.list` request — mirroring the rerank path (L1656) — and/or honor `GOOGLE_CLOUD_QUOTA_PROJECT` / the ADC `quota_project_id`. Then model listing succeeds for ADC user credentials, matching the inference path's behavior.

## Environment

- Bifrost `v1.5.8` (Docker, `maximhq/bifrost:latest`)
- Vertex provider, auth = *Service Account (Attached)* / ADC **user** credentials (`authorized_user` type), region `global`
- Host: local macOS (Docker Desktop), non-GCP (no metadata server)

Shaik-Sirajuddin:
Closes #4106
Closes #3468
Closes #2367

## Summary

- **#4106** — Vertex `listModelsByKey` now sets `X-Goog-User-Project`, fixing model listing for ADC credentials.
- **#3468** — GenAI routes now return a proper schema-mismatch error instead of generic "Invalid JSON" for any decode failure.
- **#2367** — `message.images` from OpenAI-compatible image-gen models (e.g. OpenRouter) is now preserved instead of dropped.


core/schemas/chatcompletions.go
```diff
@@ -1375,6 +1376,16 @@ type ChatAssistantMessage struct {
...
 	Annotations      []ChatAssistantMessageAnnotation `json:"annotations,omitempty"`
 	ToolCalls        []ChatAssistantMessageToolCall   `json:"tool_calls,omitempty"`
+	// Images carries generated images returned alongside (or instead of) text content,
+	// e.g. from OpenRouter image-generation models (Gemini "Nano Banana" / gemini-*-image).
+	Images []ChatMessageImage `json:"images,omitempty"`
+}
+
+// ChatMessageImage represents a single generated image entry in an assistant message's
+// "images" array (OpenAI-compatible image-generation output shape).
+type ChatMessageImage struct {
```
Pratham-Mishra04: lets call it ChatAssistantMessageImage to follow the conventions

Pratham-Mishra04 (changes requested):

Pratham-Mishra04:
Hey @Shaik-Sirajuddin thanks for the PR! I have added a small comment rest LGTM

Shaik-Sirajuddin:
Applied the suggestion!

PR #3584, opened 2026-05-19
fix: fixes forwarding of reasoning content while conversion of Responses to Chat

roroghost17:
## Summary

Reasoning content (e.g. from DeepSeek thinking mode or Anthropic extended thinking) was being silently dropped when converting between the Responses and Chat message formats. This caused multi-turn flows that route through the Responses→Chat fallback path to 400 on providers that require reasoning content to be echoed back. This PR fixes the bidirectional conversion so reasoning is preserved across both directions.

## Changes

- **`ToResponsesMessages`**: Reasoning content on an assistant `ChatMessage` is now emitted as a `reasoning`-typed `ResponsesMessage` *before* any tool calls or text content, matching the order providers expect and allowing clients to echo it back correctly.
- **`ToChatMessages`**: Reasoning messages are no longer silently skipped. Instead, they are buffered and attached to the next assistant turn (whether that turn carries text content or tool calls), populating both `Reasoning` and `ReasoningDetails` fields including signatures, summaries, and encrypted content.
- Null-safety guards were added in the UI log detail view and column helpers to prevent crashes when `input_history` contains `null`/`undefined` entries (which can occur when reasoning messages are injected into history).

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [x] Core (Go)
- [ ] Transports (HTTP)
- [ ] Providers/Integrations
- [ ] Plugins
- [x] UI (React)
- [ ] Docs

## How to test

```sh
# Core/Transports
go version
go test ./...
```

Specific test cases added:
- `TestToChatMessages_AttachesReasoningToNextAssistantMessage` — verifies reasoning is buffered and attached to the following assistant text message, including signature preservation.
- `TestToChatMessages_AttachesReasoningToToolCallAssistantMessage` — verifies reasoning is attached to tool-call assistant messages.
- `TestToResponsesMessages_EmitsReasoningMessageBeforeToolCalls` — verifies reasoning is emitted first when an assistant message has both reasoning and tool calls.
- `TestToResponsesMessages_EmitsReasoningMessageBeforeTextContent` — verifies reasoning is emitted first when an assistant message has both reasoning and text content.

## Breaking changes

- [ ] Yes
- [x] No

## Related issues

Fixes multi-turn DeepSeek thinking mode and Anthropic extended thinking flows that were returning 400 errors due to missing reasoning content in echoed history.

## Security considerations

[...]

PR #4908, opened 2026-07-04
fix: address 3 Anthropic-specific schema-compatibility bugs

### LINKED ISSUE #3233: [Bug]: function.strict is not being dropped for Anthropic
### Prerequisites

### Description

I was trying to use the Responses API with Anthropic models hosted via Vertex AI, and encountered the error message:
```
tools.0.custom.strict: Extra inputs are not permitted
```

As per [the docs](https://docs.getbifrost.ai/providers/supported-providers/anthropic#tool-conversion), the `function.strict` fields are supposed to be dropped. I both manually searched the code and let Codex scan the code in search for the implementation of that, and did not find it anywhere (on the `main` branch).

After adding a workaround to our code that forcefully omits the `function.strict` field, the error message is gone.

I don't run the Bifrost instance, so I don't know the version, but it's likely the latest one or one released in the last few weeks.

### Steps to reproduce

1. Send request to Responses API endpoint with tool that contains `"function.strict": false` (or true), routed to Anthropic model
2. See error

### Expected behavior

The LLM provider answers normally

### Actual behavior

Error response

### Affected area(s)

Core (Go)

### Version

?, likely latest

### Environment

```text

```

### Relevant logs/output

```shell
tools.0.custom.strict: Extra inputs are not permitted
```

### Regression?

_No response_

### Severity

### LINKED ISSUE #3802: [Bug]: reasoning_content dropped on assistant tool-call turns when /anthropic (extended thinking) requests are converted to the Responses API for Kimi (v1.5.4–v1.5.5; regression of #2093/#2284)
### Prerequisites

- [x] I am using the latest version (tested on v1.5.5)

### Problem to solve

Multi-turn, tool-using conversations from an **Anthropic Messages** client (Claude Code, `POST /anthropic/v1/messages`) with **extended thinking enabled**, routed to a thinking-enabled provider (**Kimi For Coding**: custom provider, `base_provider_type: anthropic`, `base_url: https://api.kimi.com/coding/`, model `kimi-for-coding`), intermittently fail with an upstream HTTP 400:

```
thinking is enabled but reasoning_content is missing in assistant tool call message at index N
```

(`is_bifrost_error: false` — the 400 is the upstream's, triggered by the payload Bifrost emits.)

This is the **same class of bug as #2093** (closed, fixed via #2284), but on a **different code path**. Here the inbound request is **Anthropic Messages with native `thinking` blocks**, and Bifrost converts it to the **OpenAI Responses API** to call the Kimi upstream (the request is logged as `object_type: responses_stream`, with `thinking` mapped to `reasoning: {effort, summary}`). During that conversion, Bifrost **drops `reasoning_content` from historical assistant tool-call turns**, which Kimi strictly requires when thinking is active.

Kimi documents this as a hard requirement:
> "reasoning_content field is missing from the tool call message. This is a Kimi Code-specific field required when thinking mode is active."
— https://www.kimi.com/code/docs/en/kimi-code/error-reference.html

It only surfaces on longer histories (observed at index 6, 7, 10, 11, 17), not on short ones — consistent with reasoning content being dropped from earlier assistant tool-call turns as the conversation grows.

### Steps to reproduce

1. Configure a custom provider `kimi` (`base_provider_type: anthropic`, `base_url: https://api.kimi.com/coding/`), model `kimi-for-coding`.
2. From an Anthropic Messages client (e.g. Claude Code) with **extended thinking enabled**, run a multi-turn session that makes several tool calls.
3. After a few assistant tool-call turns, requests fail with the 400 above.

Reproducible directly against the Responses API that Bifrost emits upstream: a multi-turn `input` with assistant `function_call` items that lack preceding `reasoning` items, plus `reasoning: {effort: "medium"}`, returns the 400; the identical payload with reasoning disabled returns 200.

### Expected behavior

When converting Anthropic (extended thinking) requests to the Responses API for an `anthropic`-type provider, Bifrost should **preserve/carry `reasoning_content` (reasoning items) on every assistant tool-call message** (as #2284 did for the path it covered), or pass through natively as Anthropic Messages so the thinking blocks are retained.

### Actual behavior

`reasoning_content` is absent from historical assistant tool-call turns in the converted upstream request → upstream 400.

### Additional findings

- The **chat-completions** upstream tolerates the missing `reasoning_content` (HTTP 200); the **Responses** upstream enforces it (HTTP 400) for the same conversation.
- Setting `responses: false` / `responses_stream: false` in the provider's `allowed_requests` does **not** make Bifrost fall back to chat-completions — it hard-fails with `500 "responses_stream is not supported by kimi provider"`. So there is no provider-config workaround.

### Affected area(s)

Core (Go) — Anthropic ↔ Responses conversion / reasoning handling.

### Version

v1.5.5 (also reproduced on v1.5.4).

### Environment

```text
Self-hosted Bifrost HTTP gateway: npx @maximhq/bifrost --transport-version v1.5.5
Config/logs store: PostgreSQL
Client: Claude Code via /anthropic/v1/messages with extended thinking enabled
Provider: Kimi For Coding (base_provider_type: anthropic, base_url: https://api.kimi.com/coding/)
```

### Regression?

Likely — the same error as #2093 (fixed via #2284, closed 2026-03-25) reappears on v1.5.4–v1.5.5 via the Anthropic-inbound → Responses-upstream conversion path. Related: #1977 (Chat→Responses mux not creating reasoning items), #2301 (thinking-response normalization).

### Severity

Medium (breaks multi-turn tool use with thinking enabled on Kimi; intermittent, surfaces on longer sessions).

### LINKED ISSUE #4780: [Bug]: Anthropic server-side tool_search results are dropped on /v1/responses (advertised as supported, but not wired up)
## TL;DR

The Anthropic provider advertises `tool_search` as supported, but server-side `tool_search` results are never forwarded on the `/v1/responses` streaming path. The model's turn ends cleanly with no tool call and no answer. Only happens through Bifrost — going straight to Anthropic works.

## What happens

`ToolSearch: true` is set for Claude models in `core/providers/anthropic/types.go`, so Bifrost reports `tool_search` as supported. The request side maps the tool correctly and Claude does run a server-side search. But the result block coming back is dropped, so the discovered tools never reach the model's next turn — the stream just ends with a clean stop, no tool call, no output.

## Root cause

In `core/providers/anthropic/responses.go`, the streaming converter has a handler for every server-tool result **except** `tool_search`:

| Server tool | Result block | Streaming handler in `responses.go` |
|---|---|---|
| Web search | `web_search_tool_result` | ✅ |
| Web fetch | `web_fetch_tool_result` | ✅ |
| Code execution | `code_execution_tool_result` | ✅ |
| Advisor | advisor result | ✅ |
| **Tool search** | **`tool_search_tool_result`** | ❌ **missing** |

The pieces are half-wired:

- `AnthropicContentBlockTypeToolSearchToolResult` (the block type) and `ToolReferences` (the field meant to carry the result) are **defined** in `types.go`...
- ...but are **never read or written anywhere** else in the provider.

So the `tool_search_tool_result` block is silently discarded. It reads like a feature that was started — types and feature flag in place, request side done — but the response/streaming side was never finished.

## Steps to reproduce

1. Send a Claude request (e.g. Sonnet) through `/v1/responses` with a `tool_search` tool (`tool_search_tool_regex_20251119`) plus some `defer_loading` tools.
2. Ask something that needs one of the deferred tools.
3. Claude emits the server-side tool_search step and Bifrost receives a `tool_search_tool_result` — but nothing is forwarded. The turn ends with no tool call and no answer.

## Expected

`tool_search_tool_result` (with its `tool_references`) should be forwarded the same way `web_search_tool_result` is, so the model can continue, see the discovered tool, call it, and answer.

## Affected versions

Reproduces from `v1.5.x` through `v1.6.0` and current `main` — the handler has never been present.

## Not a duplicate

This is distinct from the OpenAI/Codex `tool_search` fixes (#4711, #4749, #4644, #4121, …), which are all on the `/v1/responses` **client-dispatch** path. This one is the **Anthropic provider's server-side** path.

Shaik-Sirajuddin:
Closes #4780
Closes #3233
Closes #3802

## Summary

- **#4780** — `tool_search_tool_result` was dropped on `/v1/responses` (streaming, non-streaming, and Anthropic egress/replay).
- **#3233** — `tools[].strict` now dropped only where Anthropic-family providers don't support it (Vertex), kept for native Anthropic.
- **#3802** — regression test confirming `reasoning_content` survives tool-call turns with extended thinking (already fixed by #3584).
- **`web_search_call`** — same class as #4780: missing from `response.completed`, plus a concurrency bug dropping all but the last call in a multi-search turn.

## Test plan

- [x] Regression tests in `core/providers/anthropic`, full suite passing (one pre-existing unrelated failure confirmed on baseline: `TestResponsesMessageToolCallArguments/real_tool_search_call_frames_from_openai`)
- [x] Live verification against a real Anthropic key: streaming + non-streaming `tool_search`, concurrent `web_search` calls, `strict` forwarding


Shaik-Sirajuddin:
Performed a round trip validation againt anthropic platform api 

@TejasGhatte Could you take a look when you get a chance

TejasGhatte:
Hey @Shaik-Sirajuddin is this compatible with both openai and anthropic tool search tool?
I mean it should have correct response structure when calling from both openai and anthropic sdks.

Shaik-Sirajuddin:
Box | Config | Result
-- | -- | --
#1 | Anthropic-endpoint (/v1/responses) → Anthropic backend, bm25 | ✅ PASS
#1 | Anthropic-endpoint (/v1/responses) → Anthropic backend, regex | ✅ PASS
#1 (native) | /anthropic/v1/messages via real Anthropic SDK, bm25 | ✅ PASS — re-verified on latest binary
#3 | Same session, switched → OpenAI backend, bm25-origin | ✅ PASS
#3 | Same session, switched → OpenAI backend, regex-origin | ✅ PASS
#2 (ingest) | OpenAI-endpoint namespace bridge → Anthropic backend | ✅ PASS
#2 (egress) | Caller-facing namespace collapse on response | ✅ PASS — wired
#2 (egress→switch) | Same session, namespace-disguised pair replayed → OpenAI backend, bm25 | ✅ PASS — wired
#2 (egress→switch) | Same session, namespace-disguised pair replayed → OpenAI backend, regex | ✅ PASS — wired
#4 | OpenAI-endpoint native → OpenAI backend | ✅ PASS
L3 | OpenAI-native-only decl → Anthropic backend | ✅ PASS — deterministic bm25 default
L3→#3 | Same session, switched → OpenAI backend | ✅ PASS

Shaik-Sirajuddin:
> Hey @Shaik-Sirajuddin is this compatible with both openai and anthropic tool search tool? I mean it should have correct response structure when calling from both openai and anthropic sdks.

Addressed issues that was missing handling for OpenAi v1/responses handling with anthropic as provider

Anthropic publishes two types of tool_search_tool_regex_* , tool_search_tool_bm25_* , 

Open AI has one default tool_search_tool , support in its semantics of v1/respones 
To Adjust for this anthropics defaults to using regex serach as backend , 

_So Drop In Replacement works for both v1/respones with anthropic or openai as backends_ 

--- 

Additionally , users might want to leverage anthropics both tools ,
To allow for such case , by passing a tools from the client , 
Exact schema 

```
{
  "type": "namespace",
  "name": "bifrost_tool_search_bridge",
  "description": "Search hidden/deferred tools by keyword (bm25) or regex pattern.",
  "tools": [
    {
      "type": "function",
      "name": "tool_search_bm25",
      "description": "Keyword search over deferred tools."
    },
    {
      "type": "function",
      "name": "tool_search_regex",
      "description": "Regex pattern search over deferred tools."
    }
  ]
}
```

This internally handles using both tools when provider supports ( anthropic in this case ) 
@TejasGhatte 
Awaiting your view on the namespace type overrides for specific users , or an another aspect to rethink in this

ws4charlie:
Thanks for the fix! We opened #4780 and the initial fix #4786, and it's great to see the response-side handling carried further here. 🙏

While verifying #4908 against the OpenAI /v1/responses path (the native /anthropic/v1/messages already works), we found a complementary request-side gap: `normalizeResponsesToolType` (core/schemas/responses.go) doesn't preserve the `tool_search_tool_regex_20251119` type, so on /v1/responses Anthropic never server-runs the search and the forwarding has nothing to forward.

Filed as #5279, with a small standalone fix in #5280 (one case in `normalizeResponsesToolType`, ~9 lines, mirroring the existing memory/advisor cases; base `dev`). Verified end-to-end on top of #4908: /v1/responses now returns the full `server_tool_use` + `tool_search_tool_result` + discovered `tool_use`, and the tool loop continues to a final answer.

Happy to land it separately or fold it into #4908 — whatever's easiest for you. Thanks again for the work here!

PR #4911, opened 2026-07-04
fix: clear the stream close claim between attempts so streaming retries and fallbacks work

### LINKED ISSUE #4788: [Bug]: DeepSeek Anthropic-compatible provider causes "stream closed" error in v1.6.0 (regression from v1.5.16)
### Prerequisites

### Description

When configuring DeepSeek as a custom Anthropic-type provider in Bifrost (using https://api.deepseek.com/anthropic as the base URL), streaming requests fail with the following error visible in the LLM logs:
`Error reading DeepSeekClaude stream: stream closed`

### Steps to reproduce

1. Add DeepSeek as a provider with type: anthropic and base_url: https://api.deepseek.com/anthropic
2. Configure it as a fallback for an OpenAI-compatible primary provider
3. Trigger a request that falls back to DeepSeek
4. Observe Error reading DeepSeekClaude stream: stream closed in LLM logs

### Expected behavior

Streaming response from DeepSeek's Anthropic-compatible endpoint completes successfully, as it does on v1.5.16.

### Actual behavior

Stream is closed prematurely by Bifrost's Anthropic stream reader. The request never completes.

### Affected area(s)

Core (Go), Transports (HTTP)

### Version

1.6.0

### Environment

```text
- Bifrost version: v1.6.0 (regression), confirmed working on v1.5.16
- Installed via: npm install -g @maximhq/bifrost
- Provider: DeepSeek configured as anthropic type, base_url: https://api.deepseek.com/anthropic
- OS: Linux (Debian/glibc)
- Usage: fallback routing from OpenAI primary → DeepSeek Anthropic fallback
```

### Relevant logs/output

```shell

```

### Regression?

1.5.16

### Severity

fus3r:
## Summary

Fixes #4788. Since v1.6.0, any streaming retry or fallback that follows a provider error embedded in an HTTP 200 SSE stream was dead on arrival: the new attempt's stream failed every read with `stream closed` before delivering a single chunk, and the request ultimately failed with the primary error. The reporter hit this with an OpenAI-compatible primary falling back to DeepSeek's Anthropic-compatible endpoint, but the bug is not specific to either provider.

When a stream dies through an SSE-embedded error, `CheckFirstStreamChunkForError` requalifies it as an attempt failure so retries and fallbacks can run. By then the dead stream's teardown has already run `ReleaseStreamingResponse`, which since #4678 claims the `connection_closed` flag on the request's shared context (an atomic guard against double-releasing the fasthttp body stream). That claim is scoped to the response it released, but the flag stayed set on the context. Nothing cleared it between attempts, so the idle-timeout reader of the next attempt's stream treated its own fresh stream as already closed and every read returned `stream closed`. On v1.5.16 the guard was read-only, which is why the same setup worked there.

## Changes

- `executeRequestWithRetries` clears `BifrostContextKeyConnectionClosed` when it requalifies a first-chunk error, right after the dead stream's teardown is known to be complete (`<-drainDone`). This covers both the in-loop retry and the fallback path, which re-enters `executeRequestWithRetries` per attempt. Clearing is race-free at that point: the drain only completes after the provider goroutine's deferred `ReleaseStreamingResponse`/`CloseStream` have run, and the cancellation watcher and idle timer are stopped before that.
- Regression tests in `core/streamfallback_test.go` cover both paths against local `httptest` servers: an OpenAI-compatible primary that returns HTTP 200 with an SSE error event, with an Anthropic fallback that streams normally, and a single-provider retry where the first attempt fails the same way. Both fail before the fix (the fallback test fails with the primary error after the fallback stream dies, the retry test fails with `Error reading stream: stream closed`) and pass with it.
- Changelog entry.

Cancellation semantics are unchanged: a genuinely cancelled request surfaces `RequestCancelled`, which neither retries nor falls back, and non-streaming requests never take this branch.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [x] Core (Go)
- [ ] Transports (HTTP)
- [ ] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

```sh
cd core
go test . -run 'TestStreamFallbackAfterFirstChunkError|TestStreamRetryAfterFirstChunkError' -v
```

Both tests fail on `dev` without this change and pass with it. `go build ./...` passes and the package tests (`go test .`) are green, including under the race detector. The pre-existing `TestResponsesMessageToolCallArguments/real_tool_search_call_frames_from_openai` failure in `core/schemas` fails on `dev` without this change as well.

## Screenshots/Recordings

N/A

## Breaking changes

- [ ] Yes
- [x] No

## Related issues

Closes #4788

## Security considerations

None. The flag only coordinates stream teardown ownership within a single request.

## Checklist

### DIFF
--- a/core/bifrost.go
+++ b/core/bifrost.go
@@ -6065,6 +6065,12 @@ func executeRequestWithRetries[T any](
 				checkedStream, drainDone, firstChunkErr := providerUtils.CheckFirstStreamChunkForError(ctx, streamChan)
 				if firstChunkErr != nil {
 					<-drainDone
+					// The dead stream's teardown (ReleaseStreamingResponse) claimed the
+					// connection_closed flag on the shared context. That claim is scoped
+					// to the response it released; clear it so the retry or fallback
+					// attempt that follows doesn't see its own fresh stream as already
+					// closed and fail every read with ErrStreamClosed.
+					ctx.ClearValue(schemas.BifrostContextKeyConnectionClosed)
 					bifrostError = firstChunkErr
 				} else {
 					result = any(checkedStream).(T)
--- a/core/changelog.md
+++ b/core/changelog.md
@@ -1,3 +1,4 @@
+- fix: clear the per-attempt stream close claim so streaming retries and fallbacks are not dead on arrival after an SSE-embedded provider error (closes #4788) [@fus3r](https://github.com/fus3r)
 - fix: emit reads-only `cached_tokens` in usage per the OpenAI spec so cache writes are not priced as cache reads (closes #4816) [@fus3r](https://github.com/fus3r)
 - fix: preserve Gemini file upload MIME types for GenAI file URI completions
 - fix: Gemini video reference fields map to instances [@vojthor](https://github.com/vojthor)
--- a/core/streamfallback_test.go
+++ b/core/streamfallback_test.go
@@ -1,0 +1,194 @@
+package bifrost
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	schemas "github.com/maximhq/bifrost/core/schemas"
+)
+
+// Regression tests for https://github.com/maximhq/bifrost/issues/4788.
+//
+// When a streaming attempt fails through an error embedded in an HTTP 200 SSE
+// stream (e.g. rate limits sent as SSE events), the provider goroutine exits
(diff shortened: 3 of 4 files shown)


TejasGhatte:
Hey the fix is at correct place can you also add it in function `clearCtxForFallback`, then its good to merge

fus3r:
Done, added the clear to clearCtxForFallback as well.

akshaydeo (approved):

PR #4923, opened 2026-07-05
fix: emit contentBlockStop events on the Bedrock ConverseStream egress

### LINKED ISSUE #4262: [Bug]: Bedrock ConverseStream egress never emits contentBlockStop (breaks strands streaming)
## Summary

When Bifrost serves a streaming response on its **Bedrock `ConverseStream` egress** (Converse-shaped output), it **never emits a `contentBlockStop` event**. Each content block is opened (`contentBlockStart`) and filled (`contentBlockDelta`), but is never closed before `messageStop`.

This violates the AWS Bedrock `ConverseStream` contract, in which every content block is terminated by a `contentBlockStop` event. SDKs and frameworks that assemble the final message on content-block boundaries (i.e. they finalize a block when they receive `contentBlockStop`) end up with an **empty assembled message** even though all the text/tool deltas did stream — a silent functional break: tokens flow over the wire, but `result.message.content` comes back empty.

Notably, Bifrost's **other** Bedrock streaming egress path — the legacy `InvokeModelWithResponseStream` / Anthropic-Messages shape — *does* correctly emit a block terminator (`content_block_stop`). Only the Converse path is missing it.

Tested against `main` (version `1.5.11`, commit `726422e6d462daef2120b7eb3304a96c4b964029`); the relevant code is unchanged across recent versions.

## Affected code

All references are in `core/providers/bedrock/`.

**1. `ToBedrockConverseStreamResponse` skips the block-done events** (`responses.go`, the `ToBedrockConverseStreamResponse` function near line 1524). The cases that correspond to "a content block finished" return `nil` with a comment asserting Bedrock has no such event:

```go
case schemas.ResponsesStreamResponseTypeOutputTextDone,
    schemas.ResponsesStreamResponseTypeContentPartDone,
    schemas.ResponsesStreamResponseTypeReasoningSummaryTextDone:
    // Content block done - Bedrock doesn't have explicit done events, so we skip them
    return nil, nil

case schemas.ResponsesStreamResponseTypeOutputItemDone:
    // Item done - Bedrock doesn't have explicit done events, so we skip them
    return nil, nil
```

The comment is incorrect for the Converse API (see "AWS contract" below).

**2. `BedrockStreamEvent.ToEncodedEvents` can only encode 6 event types** (`responses.go`, the `ToEncodedEvents` method near line 1776). It emits `chunk`, `messageStart`, `contentBlockStart`, `contentBlockDelta`, `messageStop`, `metadata` — there is no branch that can ever produce `contentBlockStop`.

**3. The `BedrockStreamEvent` struct has no field to represent a stop event** (`types.go`, near line 753) — it carries `Role`, `Start`, `Delta`, `StopReason`, `Usage`, etc., but nothing that maps to a `contentBlockStop`. So the Converse path is structurally incapable of emitting one.

**Contrast — the invoke/messages path does it correctly** (`invoke.go`, `ToBedrockInvokeMessagesStreamResponse`, near line 1376):

```go
case schemas.ResponsesStreamResponseTypeOutputItemDone:
    idx := 0
    if resp.ContentIndex != nil {
        idx = *resp.ContentIndex
    }
    event = map[string]interface{}{
        "type":  "content_block_stop",
        "index": idx,
    }
```

So `OutputItemDone` is mapped to a block terminator on the invoke path but dropped (`return nil`) on the Converse path.

## AWS contract (evidence)

The AWS Bedrock Runtime API reference defines `contentBlockStop` as a first-class member of the ConverseStream output event stream:

- **`ConverseStreamOutput`** (the event-stream union returned by `ConverseStream`) lists these event members: `messageStart`, `contentBlockStart`, `contentBlockDelta`, **`contentBlockStop`**, `messageStop`, `metadata`, plus the exception types.
  https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_ConverseStreamOutput.html
- **`ContentBlockStopEvent`** — *"A content block stop event."* — with a required `contentBlockIndex` field.
  https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_ContentBlockStopEvent.html

A real Bedrock `ConverseStream` response closes each content block with a `contentBlockStop` (carrying the block's `contentBlockIndex`) before the next block starts or before `messageStop`. The AWS SDKs (e.g. Java V2, Python) model the stream as a union that includes `ContentBlockStopEvent`, and Bedrock-targeting frameworks rely on it to finalize a block.

## Expected behavior

The ConverseStream egress should emit a `contentBlockStop` event (with the correct `contentBlockIndex`) at the end of each content block — matching the AWS contract and the behavior already present on the invoke/messages path.

Expected event order, e.g.:

```
text-only: messageStart → contentBlockStart → contentBlockDelta(×N) → contentBlockStop → messageStop → metadata
tool-use : messageStart → contentBlockStart(toolUse) → contentBlockDelta(×N) → contentBlockStop → messageStop → metadata
```

## Actual behavior

```
text-only: messageStart → contentBlockDelta(×N) → messageStop → metadata          (no contentBlockStop, and no contentBlockStart for the text block)
tool-use : messageStart → contentBlockStart(toolUse) → contentBlockDelta(×N) → messageStop → metadata   (no contentBlockStop)
```

`contentBlockStop` count is **0** in both cases.

## Impact

- Any consumer of Bifrost's `ConverseStream` egress that assembles the final message on `contentBlockStop` boundaries receives an **empty** final message, even though all deltas streamed correctly. Because tokens do stream (callbacks fire), the failure is silent — callers that read the live delta callback see output, but callers that read the assembled result get nothing.
- Tool-use streaming is also suspect, since the closing `contentBlockStop` for the toolUse block is likewise never sent.
- Consumers that read only `contentBlockDelta` (raw delta readers) are unaffected, which is why this can go unnoticed.

## Steps to reproduce

1. Route a streaming Converse request through Bifrost's Bedrock ConverseStream egress (`.../converse-stream`) to any Claude model.
2. Capture the raw AWS event-stream frames (or use an SDK/framework that builds the final message from content-block boundaries).
3. Observe: `contentBlockDelta` frames arrive and `messageStop` arrives, but **no `contentBlockStop` frame** is ever emitted; an assembling consumer's final message content is empty.

## Suggested fix

In the Bedrock Converse stream translation:

1. Add a representation for a stop event to `BedrockStreamEvent` (e.g. a `Stop *BedrockContentBlockStop` field or equivalent flag, carrying `contentBlockIndex`).
2. In `ToBedrockConverseStreamResponse`, map the block-done event (`OutputItemDone`, mirroring the invoke path) to that stop representation instead of `return nil`.
3. In `ToEncodedEvents`, emit a `contentBlockStop` event for it.
4. Also emit the `contentBlockStart` for the text block (today only the toolUse block gets a `contentBlockStart`), to fully match the AWS contract.

fus3r:
## Summary

Fixes #4262. The Bedrock ConverseStream egress (`/bedrock/model/{modelId}/converse-stream`) never emitted `contentBlockStop` events: content blocks were opened and filled with deltas, but never closed before `messageStop`. The AWS contract terminates every content block with a `contentBlockStop` event carrying the block's `contentBlockIndex` (a required field), and consumers that assemble the final message on block boundaries (the AWS SDK stream unions, frameworks like strands) only finalize a block when they receive it. Through Bifrost they received all the deltas and still ended up with an empty assembled message, so the break is silent: live delta readers see output, assembling consumers get nothing.

The invoke egress (`invoke-with-response-stream`, Anthropic Messages shape) already maps `OutputItemDone` to `content_block_stop`. The Converse path dropped the same event with a comment saying Bedrock has no done events, which holds for the part-level done events but not at the block level.

## Changes

- `ToBedrockConverseStreamResponse` now maps `OutputItemDone` to a `contentBlockStop` event carrying the same content block index the block's start and delta events use, mirroring the invoke path. `OutputTextDone`, `ContentPartDone` and `ReasoningSummaryTextDone` stay skipped, so each block is closed exactly once, on its item's done event.
- `BedrockStreamEvent` gets an explicit `ContentBlockStop` marker: the wire payload of `contentBlockStop` is just `{"contentBlockIndex": n}`, so the flat event union had no way to represent it. The marker is never serialized; `ToEncodedEvents` builds the payload from `ContentBlockIndex`.
- `ToEncodedEvents` encodes the new event after deltas and before `messageStop`, so a text stream now encodes as `messageStart`, `contentBlockDelta` (xN), `contentBlockStop`, `messageStop`, `metadata`.
- Regression tests replay the Bifrost stream lifecycle for a text block, a tool use block and a reasoning block through the converter and the encoder, asserting the exact wire event sequences and the stop payloads.

Two notes on scope:

- The issue also suggests emitting `contentBlockStart` for text blocks. Real ConverseStream responses do not do that: the `ContentBlockStart` union only has `image`, `toolResult` and `toolUse` members, so text blocks start implicitly with their first delta. The current behavior already matches the contract there, so this PR leaves the start events unchanged.
- Nova models on `invoke-with-response-stream` delegate to the same Converse converter, so that path picks up the block terminator as well, which also matches Nova's native stream shape.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [x] Core (Go)
- [ ] Transports (HTTP)
- [ ] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

```sh
cd core
go test ./providers/bedrock/ -run 'ConverseStream.*ContentBlockStop' -v
```

All three tests fail on `dev` without this change, with the exact event sequence from the issue report (`messageStart, contentBlockDelta, contentBlockDelta, messageStop, metadata` and no `contentBlockStop`), and pass with it. `go build ./...` and the full `./providers/bedrock/` package suite are green.

## Screenshots/Recordings

N/A

## Breaking changes

- [ ] Yes
- [x] No

Consumers that only read `contentBlockDelta` are unaffected. The new event is part of the documented `ConverseStreamOutput` union, so Bedrock clients already handle it.

## Related issues

Closes #4262

## Security considerations

None. This only adds a lifecycle event to the outgoing stream encoding.

## Checklist

### DIFF
--- a/core/changelog.md
+++ b/core/changelog.md
@@ -1,3 +1,4 @@
+- fix: emit `contentBlockStop` events on the Bedrock ConverseStream egress so consumers that assemble messages on block boundaries get complete content (closes #4262) [@fus3r](https://github.com/fus3r)
 - fix: set the raw-storage log flag on standalone MCP tool executions so logging hooks see an explicit value
 - fix: clear the per-attempt stream close claim so streaming retries and fallbacks are not dead on arrival after an SSE-embedded provider error (closes #4788) [@fus3r](https://github.com/fus3r)
 - fix: emit reads-only `cached_tokens` in usage per the OpenAI spec so cache writes are not priced as cache reads (closes #4816) [@fus3r](https://github.com/fus3r)
--- a/core/providers/bedrock/conversestreamstop_test.go
+++ b/core/providers/bedrock/conversestreamstop_test.go
@@ -1,0 +1,201 @@
+package bedrock
+
+import (
+	"reflect"
+	"testing"
+
+	"github.com/bytedance/sonic"
+	"github.com/maximhq/bifrost/core/schemas"
+)
+
+// encodeConverseStream runs each Bifrost stream event through the Converse stream
+// converter and flattens the results into wire-ready encoded events, in order,
+// the same way the HTTP transport does before AWS Event Stream encoding.
+func encodeConverseStream(t *testing.T, chunks []*schemas.BifrostResponsesStreamResponse) []BedrockEncodedEvent {
+	t.Helper()
+	var events []BedrockEncodedEvent
+	for i, chunk := range chunks {
+		bedrockEvent, err := ToBedrockConverseStreamResponse(chunk)
+		if err != nil {
+			t.Fatalf("chunk %d (%s): unexpected error: %v", i, chunk.Type, err)
+		}
+		if bedrockEvent == nil {
+			continue
+		}
+		events = append(events, bedrockEvent.ToEncodedEvents()...)
+	}
+	return events
+}
+
+func encodedEventTypes(events []BedrockEncodedEvent) []string {
+	types := make([]string, 0, len(events))
+	for _, event := range events {
+		types = append(types, event.EventType)
+	}
+	return types
+}
+
+func contentBlockStopPayloads(t *testing.T, events []BedrockEncodedEvent) []string {
+	t.Helper()
+	var payloads []string
+	for _, event := range events {
+		if event.EventType != "contentBlockStop" {
+			continue
+		}
+		data, err := sonic.Marshal(event.Payload)
+		if err != nil {
+			t.Fatalf("marshal contentBlockStop payload: %v", err)
+		}
+		payloads = append(payloads, string(data))
+	}
+	return payloads
+}
+
+// TestConverseStreamTextBlockEmitsContentBlockStop replays the stream lifecycle of a
+// plain text response and asserts the Converse egress closes the text block with a
+// contentBlockStop event before messageStop, as the Bedrock ConverseStream contract
+// requires (issue #4262). Text blocks have no contentBlockStart (the ContentBlockStart
+// union has no text member), so the block is opened implicitly by its first delta.
+func TestConverseStreamTextBlockEmitsContentBlockStop(t *testing.T) {
+	messageType := schemas.ResponsesMessageTypeMessage
+	chunks := []*schemas.BifrostResponsesStreamResponse{
+		{Type: schemas.ResponsesStreamResponseTypeCreated},
+		{Type: schemas.ResponsesStreamResponseTypeInProgress},
+		{
+			Type:        schemas.ResponsesStreamResponseTypeOutputItemAdded,
+			OutputIndex: schemas.Ptr(0),
+			Item:        &schemas.ResponsesMessage{Type: &messageType},
+		},
+		{
+			Type:         schemas.ResponsesStreamResponseTypeOutputTextDelta,
+			ContentIndex: schemas.Ptr(0),
+			Delta:        schemas.Ptr("Hello"),
+		},
+		{
+			Type:         schemas.ResponsesStreamResponseTypeOutputTextDelta,
+			ContentIndex: schemas.Ptr(0),
+			Delta:        schemas.Ptr(" world"),
+		},
+		{
+			Type:         schemas.ResponsesStreamResponseTypeOutputTextDone,
+			ContentIndex: schemas.Ptr(0),
+			Text:         schemas.Ptr("Hello world"),
+		},
+		{
+			Type:         schemas.ResponsesStreamResponseTypeContentPartDone,
+			ContentIndex: schemas.Ptr(0),
+		},
+		{
+			Type:         schemas.ResponsesStreamResponseTypeOutputItemDone,
+			OutputIndex:  schemas.Ptr(0),
+			ContentIndex: schemas.Ptr(0),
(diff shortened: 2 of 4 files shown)


akshaydeo (approved):

TejasGhatte (approved):

akshaydeo:
### Merge activity

* **Jul 5, 9:59 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/4923).
* **Jul 5, 9:59 PM UTC**: [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/4923) couldn't merge this PR because it failed for an unknown reason (Fast-forward merges are not supported for forked repositories. Please create a branch in the target repository in order to merge).

PR #4927, opened 2026-07-05
fix: skip plaintext fallback lookup for secret-based virtual key values

BearTS:
## Summary

Prevents plaintext database fallback lookups when a virtual key value appears to be a secret reference (e.g., a vault or environment variable reference). Without this guard, a value like `vault.my-secret` or `env.MY_VAR` would be passed as a raw string into a `WHERE value = ?` query, which could never match a real row but still leaks information about the lookup path and wastes a database round-trip.

## Changes

- In `GetVirtualKeyByValue`, when a hash-based lookup returns no record, the value is parsed as a `SecretVar`. If it originates from a secret reference, the function returns `ErrNotFound` immediately instead of falling back to a plaintext query.
- In `GetVirtualKeyQuotaByValue`, the same early-exit behavior is applied by checking for `vault.` and `env.` prefixes directly, consistent with the rationale in `GetVirtualKeyByValue`.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [x] Core (Go)
- [ ] Transports (HTTP)
- [ ] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

```sh
go test ./framework/configstore/...
```

- Submit a request using a virtual key value formatted as `vault.<secret-name>` or `env.<VAR_NAME>`.
- Confirm the response returns a not-found error without triggering a secondary plaintext database query.
- Confirm that normal virtual key values (non-secret-reference strings) still fall back to plaintext lookup as expected.

## Breaking changes

- [ ] Yes
- [x] No

## Security considerations

Secret references (vault/env-style values) should never be stored or queried as plaintext in the database. Allowing them to reach a `WHERE value = ?` clause could expose the reference string in query logs or slow query logs. This change ensures such values are short-circuited before any plaintext lookup occurs.

## Checklist

### DIFF
--- a/core/schemas/secretvar.go
+++ b/core/schemas/secretvar.go
@@ -39,13 +39,15 @@ func inferSecretType(ref string) SecretType {
 	return SecretTypePlainText
 }
 
-// NewSecretVar creates a new SecretVar from a string.
-func NewSecretVar(value string) *SecretVar {
+// parseSecretRef classifies value and returns a *SecretVar with SecretType and ref
+// populated but Val unresolved — no vault HTTP calls, no env lookups.
+// For JSON-encoded SecretVars the raw "value" field is preserved in Val unchanged.
+// Adding a new secret type only requires updating this function.
+func parseSecretRef(value string) *SecretVar {
 	val := value
 	if unquoted, err := strconv.Unquote(value); err == nil {
 		val = unquoted
 	}
-	// If it's a valid JSON object following the SecretVar schema, unmarshal it
 	if sonic.Valid([]byte(value)) {
 		valueNode, _ := sonic.Get([]byte(val), "value")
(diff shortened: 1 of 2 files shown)


BearTS:
* **#4927** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/4927?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/maximhq/bifrost/4927?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `dev`


<h2></h2>

akshaydeo (approved):

PR #4328, opened 2026-06-12
feat: add `canonical_model_name` and `alias_model_family` columns to logs

Pratham-Mishra04:
## Summary

When a request's model is resolved through an alias mapping, the alias config can define a canonical model name and a model family. Previously these values were not captured in logs. This PR persists both fields (`canonical_model_name` and `alias_model_family`) through the full stack — from the database schema to the log entry writer to the UI detail view.

## Changes

- Added `canonical_model_name` and `alias_model_family` columns to the `logs` table via a new `migrationAddCanonicalModelColumns` migration, with both forward and rollback support.
- Added the two new fields to the `Log` struct and included them in the `listSelectColumns` query.
- Added `applyResolvedAliasInfo` in the logging plugin writer, which copies `ModelName` and `ModelFamily` from the resolved key alias onto the log entry. This is called for both error and success paths in `PostLLMHook`.
- Exposed `canonical_model_name` and `alias_model_family` on the `LogEntry` TypeScript type and rendered them in the log detail view when present.
- Removed the `go-git` dependency and its transitive dependencies from `transports/go.mod`, cleaning up unused packages.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [x] Core (Go)
- [ ] Transports (HTTP)
- [ ] Providers/Integrations
- [x] Plugins
- [x] UI (React)
- [ ] Docs

## How to test

1. Configure a model alias that sets a `canonical_model_name` and/or `model_family`.
2. Send a request using the alias name as the model.
3. Inspect the resulting log entry — both fields should be populated in the database and visible in the log detail view under "Canonical Model" and "Model Family".
4. Verify that log entries for requests not resolved via alias leave both fields as `null`.

```sh
go test ./framework/logstore/...
go test ./plugins/logging/...

cd ui
pnpm i
pnpm build
```

## Screenshots/Recordings

The log detail view will display "Canonical Model" and "Model Family" rows beneath the "Alias" row when the resolved alias defines those values.

## Breaking changes

- [ ] Yes
- [x] No

## Security considerations

No new secrets or PII surface area. The canonical model name and model family are operator-configured strings copied from alias config, consistent with how the existing `alias` field is handled.

## Checklist

- [ ] I read `docs/contributing/README.md` and followed the guidelines
- [ ] I added/updated tests where appropriate
- [ ] I updated documentation where needed
- [ ] I verified builds succeed (Go and UI)
[...]

Issue #4071, opened 2026-06-05
`aliases` filter ignored on the materialized-view stats/observability path


honeyhaarin:
## Summary

The `aliases` log filter is silently ignored on every materialized-view aggregation path (dashboard stats, counts, histograms, per-model breakdowns). The raw-table path applies it correctly, so the **Logs list filters by alias, but the Observability dashboard / `/api/logs/stats` does not**.

## Environment

- Bifrost transports `v1.5.7` (also present on `main` at the time of writing)
- Postgres logstore with hourly materialized-view aggregation enabled

## Reproduction

Use a provider that populates the `alias` column — e.g. an AWS Bedrock application-inference-profile, where the stored `model` is the profile **resource-id** and `alias` is the human-readable name the caller used.

```
GET /api/logs?aliases=<alias>         → correctly filtered
GET /api/logs/stats?aliases=<alias>   → returns ALL rows (filter ignored)
GET /api/logs/stats?models=<model>    → correctly filtered (for contrast)
```

The Models filter works on both paths; the Aliases filter works only on the raw Logs path, not on the matview-backed stats/observability path.

## Root cause

`framework/logstore/matviews.go` → `applyMatViewFiltersOnly()` builds WHERE predicates for `providers / models / status / object_type / selected_key_ids / virtual_key_ids / routing_rule_ids / team_ids / customer_ids / user_ids / business_unit_ids` but **omits `aliases`**.

More fundamentally, the `mv_logs_hourly` materialized view has **no `alias` dimension** — it is absent from the DDL `SELECT` / `GROUP BY`, the unique index, and `mvLogsHourlyRequiredColumns`. So adding only the predicate would not be enough; the column has to exist in the view first.

For contrast, the raw path (`rdb.go`) already applies it:

```go
if len(filters.Aliases) > 0 {
    baseQuery = baseQuery.Where("alias IN ?", filters.Aliases)
}
```

`applyMatViewFilters` is shared by ~6 aggregation call sites (count, stats, histograms, per-model breakdown), so the gap affects all matview-backed aggregates, not just stats.

## Proposed fix

[...]

PR #4019, opened 2026-06-03
feat: replace flat aliases table with rich deployments table supporting per-deployment model family, canonical name, and provider-specific overrides

Pratham-Mishra04:
## Summary

Replaces the flat `Record<string, string>` aliases map with a rich `DeploymentsTable` component backed by a structured `AliasConfig` type. Users can now expand each deployment row to set a canonical model name, model family, description, and provider-specific overrides (Azure endpoint/API version, Vertex project/region, Bedrock inference profile ARN, Replicate deployments endpoint toggle) — all of which feed cost/pricing logs and family-based routing.

## Changes

- Introduced `AliasConfig` TypeScript interface and `ModelFamily` enum in `config.ts`, mirroring the Go `schemas.AliasConfig` struct with flattened embedded provider sub-configs.
- Added `aliasConfigSchema` and `modelFamilySchema` Zod schemas in `schemas.ts` to validate the new shape.
- Replaced the `HeadersTable`-based aliases editor in `apiKeysFormFragment.tsx` with the new `DeploymentsTable` component. Removed the now-unnecessary `normalizeAliasesValue` helper.
- Created `deploymentsTable.tsx`: a collapsible table with a draft row for adding new entries, per-row expand panels for advanced config, and provider-specific override sections (Azure, Vertex, Bedrock, Replicate).
- Updated `isValidAliases` in `validation.ts` to validate the new object shape — every entry must have a non-empty deployment name and `model_id`; legacy string handling removed.
- Updated `KeySchema` in `providerForm.ts` to use `aliasConfigSchema` and updated the validation error message accordingly.
- Bumped several indirect Go dependencies in `plugins/governance` and `plugins/modelcatalogresolver` (`bytedance/sonic`, `cloud.google.com/go/auth`, `andybalholm/brotli`, `envoyproxy`, `googleapis`, `grpc`, `protobuf`, `opentelemetry`, etc.).

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [ ] Core (Go)
- [ ] Transports (HTTP)
- [ ] Providers/Integrations
- [x] Plugins
- [x] UI (React)
- [ ] Docs

## How to test

```sh
# UI
cd ui
pnpm i
pnpm build

# Navigate to a provider key form and verify:
# 1. The "Deployments" table renders with Deployment name / Model ID columns.
# 2. Adding a row with both fields filled commits it to the list.
[...]

PR #4941, opened 2026-07-06
feat: show canonical model names in dashboard model rankings

### LINKED ISSUE #4940: [Feature]: Show canonical model names instead of Bedrock inference-profile IDs on the dashboard Model Rankings
### Prerequisites

### Problem to solve

When requests go through AWS Bedrock application inference profiles, the logged `model` is the profile's opaque resource ID (e.g. `4xg7dq2mkz9v`). The Observability dashboard's Model Rankings tab (Top Models chart, legend, and rankings table) displays these raw IDs, so anyone reading the dashboard has to keep a mental mapping from profile IDs to actual models. With a handful of profiles the tab becomes effectively unreadable:

Model Rankings showing raw inference-profile IDs

The data to fix this already exists: since #4328 the logs table stores `canonical_model_name` per row (populated when a deployment/key alias sets `model_name`), and the deployments table (#4019) carries per-deployment canonical names. The log detail view shows it — but no aggregation path (rankings, histograms) reads the column, so the dashboard never benefits.

This affects anyone using Bedrock inference profiles, Azure deployment IDs, or similar provider-side deployment identifiers as the wire model.

### Proposed solution

Surface the canonical name in the model rankings aggregation and render it in the UI, keeping the raw model as the grouping/filter key:

- `GetModelRankings` (raw and `mv_logs_hourly` paths) additionally selects `MAX(NULLIF(canonical_model_name, ''))` per `(model, provider)` group and returns it as an optional `canonical_model_name` field on each ranking entry. Grouping stays keyed by the raw model, so two profiles pointing at the same underlying model remain distinct rows.
- `canonical_model_name` becomes a dimension of `mv_logs_hourly` (same approach as the `alias` dimension added for #4071; `repairMatViewShapes` rebuilds old-shape views automatically).
- The rankings table shows the canonical name with the raw profile ID as muted secondary text, so the deployment is still identifiable. The Top Models legend/tooltip resolve labels through a shared helper with fallback to the raw model. CSV export gains a "Canonical Model" column.

I have this implemented and tested (SQLite, Postgres raw + matview, ClickHouse via the logstore parity suite) and will open a PR right after filing this.

### Alternatives considered

- **Grouping rankings by canonical name instead of raw model** — merges deployments of the same model into one row. Arguably nicer for "which model is used most", but it changes existing semantics and breaks the Models filter round-trip (filter values are raw models). Keeping the raw key + display label is non-breaking; merging could be a later opt-in.
- **Resolving profile IDs via the Bedrock `GetInferenceProfile` API at write time** — would populate `canonical_model_name` even without a configured alias `model_name`. Complementary follow-up rather than an alternative; the display change is needed either way.
- **Only using the existing Aliases filter/column** — aliases are the caller-facing names, populated only when the caller used an alias, and the dashboard rankings don't read them either.

### Area(s)

Framework, UI (React)

### Additional context

Out of scope for the first PR (happy to follow up): the Models filter sidebar also lists raw profile IDs (`GetDistinctModels` returns plain strings; changing its shape touches the LogStore interface, hybrid store, plugin manager, and filterdata cache), and the Overview tab's model usage chart has the same raw-ID display.

satyamkrishna:
## Summary

The dashboard's Model Rankings tab displays raw wire model values, which for AWS Bedrock application inference profiles are opaque resource IDs (e.g. `4xg7dq2mkz9v`). The logs table already stores `canonical_model_name` per row (#4328, #4019), but no aggregation path surfaces it. This PR propagates the canonical name through the model rankings queries and renders it in the UI, with the raw model kept as the grouping/filter key.

## Changes

- **`GetModelRankings` (raw + matview paths)**: additionally selects `MAX(NULLIF(canonical_model_name, ''))` per `(model, provider)` group, returned as an optional `canonical_model_name` field on `ModelRankingEntry`. Grouping stays keyed by the raw model, so two profiles backed by the same underlying model remain distinct rows. On the raw path, the previous-period trend query keeps the canonical-free select clause since it never reads the column.
- **`mv_logs_hourly`**: `canonical_model_name` added as a dimension (DDL, unique index, `mvLogsHourlyRequiredColumns`) — the same approach as the `alias` dimension added for #4071. `repairMatViewShapes` detects old-shape views and rebuilds them on startup; no manual migration. Cardinality impact is minimal since canonical name is effectively functionally dependent on `model`.
- **UI**: the rankings table shows the canonical name with the raw profile ID as muted secondary text; the Top Models legend and tooltip resolve labels through a shared `displayModelLabel` helper in `chartUtils.ts` (fallback to raw model). The label map is built from the rankings response — single source of truth, no new histogram fields. CSV export gains a "Canonical Model" column.
- **Tests**: `canonicalmodel_test.go` covers SQLite, Postgres raw, and Postgres matview paths with shared seed/assert helpers (including the mixed NULL/non-NULL canonical case and the "no canonical stays unset" case). The three-backend parity suite (`logstoreparity_test.go`) gets canonical fixtures, which verifies ClickHouse agrees with SQLite/Postgres on the new aggregate.

Design notes:

- `NULLIF(..., '')` keeps empty strings out of `MAX`, and matview rows store `COALESCE(canonical_model_name, '')` like other dimensions, so mixed populated/NULL rows for one model resolve to the populated name on both paths.
- ClickHouse needs no schema work: its `logs` table derives columns from the `Log` struct, so `canonical_model_name` already exists; the store reuses the raw-path query.

Deliberately out of scope (follow-ups, noted on the issue): canonical labels in the Models filter sidebar (`GetDistinctModels` shape change ripples through the LogStore interface/hybrid store/filterdata cache), the Overview tab's model usage chart, and write-time resolution of profile IDs via `GetInferenceProfile`.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [ ] Core (Go)
- [ ] Transports (HTTP)
- [ ] Providers/Integrations
- [ ] Plugins
- [x] UI (React)
- [ ] Docs

(Framework — `framework/logstore` — is the main surface; there's no checkbox for it above.)

## How to test

```sh
# Backend (Postgres + ClickHouse via framework/docker-compose.yml; tests skip if unavailable)
cd framework
docker compose up -d postgres clickhouse
go test ./logstore/ -run 'TestCanonicalModelRankings' -count=1 -v
go test ./logstore/ -run 'TestLogStoreParity' -count=1

# UI
cd ui
npm i
npm run build
```

Manual: run `make dev`, send traffic through a Bedrock key whose deployment/alias sets `model_name` (or seed `logs.canonical_model_name` directly), open Dashboard → Model Rankings. Rankings rows and the Top Models legend show the canonical name; the raw profile ID appears as secondary text in the table. `GET /api/logs/rankings` entries carry `canonical_model_name` only for models that have one.

## Screenshots/Recordings

Before:

before

After:

after

## Breaking changes

- [ ] Yes
- [x] No

`canonical_model_name` is additive and `omitempty` on the rankings response; the matview reshape is handled by the existing startup repair path (same rollout as #4071's alias dimension).

## Related issues

Closes #4940

## Security considerations

None — read-side display of a column that's already stored and shown on the log detail view.

## Checklist

- [ ] I updated documentation where needed (no doc page covers ranking response fields; happy to add if pointed at one)


satyamkrishna:
@codeant-ai: review

satyamkrishna:
Re: the CodeRabbit note on mv_logs_hourly cardinality — intentional, and now documented in the DDL comment (85a0b18d8). canonical_model_name is effectively functionally dependent on model, so steady-state group count is unchanged; buckets only split transiently while a model's rows mix empty and populated canonical values (e.g. an alias gains model_name mid-hour), and both readers re-aggregate per raw model via SUM / MAX(NULLIF(...)). Same rollout mechanics as the alias dimension added for #4071.

Pratham-Mishra04 (approved):

impoiler (approved):

PR #4943, opened 2026-07-06
fix: round-trip anthropic redacted_thinking blocks on the chat completions surface

### LINKED ISSUE #4942: [Bug]: redacted_thinking blocks are dropped on chat completions, breaking tool-use replay with extended thinking
### Prerequisites

### Description

When Anthropic flags part of Claude's reasoning, the API returns it as a `redacted_thinking` content block (an encrypted `data` payload) instead of a `thinking` block. Anthropic requires these blocks to be passed back unmodified in multi-turn tool use; their docs state that dropping them is the most common cause of the 400 error `` `thinking` or `redacted_thinking` blocks in the latest assistant message cannot be modified ``, and the payload is decrypted server side so the model keeps its reasoning context.

On the chat completions surface (`/v1/chat/completions` with an Anthropic provider, or any custom provider with `base_provider_type: anthropic`), Bifrost loses these blocks in all three directions:

1. Non-streaming responses: the content switch in `ToBifrostChatResponse` (`core/providers/anthropic/chat.go`) has cases for `text`, `tool_use` and `thinking` only. A `redacted_thinking` block falls through and produces no `reasoning_details` entry, so the client never sees it.
2. Streaming: the `content_block_start` handler only emits for `tool_use`. A `redacted_thinking` start (which carries the complete `data` payload; the official SDKs model it as a `content_block_start` union member with no deltas) is silently discarded.
3. Requests: `ToAnthropicChatRequest` converts every `reasoning_details` entry to a `thinking` block. A `reasoning.encrypted` detail (which carries `data` and no text or signature) becomes an empty `thinking` block, which `StripEmptyThinkingBlocks` then removes from the outgoing body.

So even a client that faithfully echoes the assistant message back gets a rewritten history: the redacted block is gone. When all of the turn's reasoning was redacted, the replayed assistant message has no thinking block at all in a thinking-enabled tool-use conversation.

The schema already has the right representation (`reasoning_details` type `reasoning.encrypted` with a `data` field), the Responses surface of the same provider already round-trips redacted thinking through `ResponsesReasoning.EncryptedContent`, and the chat/responses mux maps `data` to `EncryptedContent`. Only the chat converters never got the mapping.

### Steps to reproduce

Deterministic, no API key: convert an Anthropic message containing a `redacted_thinking` block with `ToBifrostChatResponse` and observe that no `reasoning.encrypted` detail is produced; convert an assistant message carrying a `reasoning.encrypted` detail with `ToAnthropicChatRequest` and observe the outgoing body contains no `redacted_thinking` block.

Against the live API: enable thinking plus a tool on `/v1/chat/completions` with an Anthropic key and send the documented test string that forces redaction (`ANTHROPIC_MAGIC_STRING_TRIGGER_REDACTED_THINKING_46C9A13E193C177646C7398A98432ECCCE4C1253D5E2D82641AC0E52CC2876CB`) as the user prompt, then answer the tool call and send the follow-up turn.

### Expected behavior

`redacted_thinking` blocks survive the chat completions round trip: responses (streaming and not) expose them as `reasoning_details` entries of type `reasoning.encrypted` with the `data` payload, and request conversion re-materializes them as `redacted_thinking` blocks, so the follow-up turn replays the assistant message unmodified.

### Actual behavior

The block is dropped from responses and from replayed requests. Multi-turn tool use with extended thinking breaks with an upstream 400 when redaction occurs, and the model loses the redacted reasoning context.

### Affected area(s)

Core (Go)

### Version

Current dev (reproduced at 6731d0b79, still present at e0c541da8); the chat converters never handled `redacted_thinking`.

### Severity

fus3r:
## Summary

Fixes #4942. When Anthropic's safety systems flag part of Claude's reasoning, the API returns it as a `redacted_thinking` block carrying an encrypted `data` payload, and requires the block to be replayed unmodified on later turns; their docs call out dropping these blocks as the most common cause of the 400 `` `thinking` or `redacted_thinking` blocks in the latest assistant message cannot be modified ``. The chat completions converters dropped them everywhere: the response switch had no `redacted_thinking` case, the streaming `content_block_start` handler only emitted for `tool_use`, and request conversion turned `reasoning.encrypted` details into empty `thinking` blocks that `StripEmptyThinkingBlocks` then removed from the outgoing body. Tool-use conversations with extended thinking broke as soon as redaction occurred, and the model lost that reasoning context even when it didn't hard fail.

The schema already models this (`reasoning_details` type `reasoning.encrypted` with a `data` field), the Responses surface of the same provider already round-trips redacted thinking via `ResponsesReasoning.EncryptedContent`, and the chat/responses mux maps between the two. This PR gives the chat converters the same mapping.

## Changes

- `ToBifrostChatResponse` preserves a `redacted_thinking` block as a `reasoning_details` entry of type `reasoning.encrypted` carrying the `data` payload, keeping block order relative to visible thinking blocks.
- `ToBifrostChatCompletionStream` emits the same detail from a `redacted_thinking` `content_block_start` (the block arrives complete in the start event; the official SDKs model it as a start-event union member with no deltas).
- Streaming reasoning details now carry a per-block index tracked in the stream state instead of a hardcoded 0. Consumers group reasoning deltas by index (Bifrost's own logging accumulator does too), so in a stream mixing redacted and visible thinking a shared index would merge both into one text-typed detail and the encrypted payload would be lost again when that message is replayed. Single-block reasoning streams keep index 0 exactly as before.
- `ToAnthropicChatRequest` re-materializes `reasoning.encrypted` details that carry a non-empty payload as `redacted_thinking` blocks. Encrypted details without data (for example Gemini thought signatures, which only carry a signature) and empty payloads keep the existing thinking-block mapping, pinned by regression tests.
- Regression tests cover all three directions plus the data-less guards; the streaming accumulator already deep-copies and folds `reasoning_details` data, so logged messages pick the payload up with no framework change.
- Changelog entry.

The conversions are gated on the `redacted_thinking` type and a non-empty payload, so messages without redacted thinking keep their exact shape, with one deliberate exception: a stream carrying several reasoning blocks now gives each block its own `reasoning_details` index (previously every delta carried index 0), which index-keyed consumers need to keep the blocks apart.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [x] Core (Go)
- [ ] Transports (HTTP)
- [ ] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

```sh
cd core
go test ./providers/anthropic/ -run 'RedactedThinking|EncryptedDetail|MixedReasoning' -v
```

The four behavior tests fail on `dev` without this change (the response conversion produces no `reasoning.encrypted` detail, the request conversion emits a `thinking` block instead of `redacted_thinking` with no data, the streaming start produces no chunk, and mixed reasoning streams have no per-block detail indices) and pass with it. The guard tests pass before and after; they pin the data-less and empty-payload cases. `go build ./...` and the full `./providers/anthropic/` suite are green, including under the race detector.

Against the live API, Anthropic documents a test string that deterministically triggers redacted thinking (`ANTHROPIC_MAGIC_STRING_TRIGGER_REDACTED_THINKING_46C9A13E193C177646C7398A98432ECCCE4C1253D5E2D82641AC0E52CC2876CB`); with thinking and a tool enabled, the follow-up turn after the tool result round-trips through Bifrost with this change.

## Screenshots/Recordings

N/A

## Breaking changes

- [ ] Yes
- [x] No

Clients that ignore `reasoning_details` are unaffected; clients that echo the assistant message back now replay exactly what the model produced.

## Related issues

Fixes #4942

## Security considerations

None. The payload is opaque ciphertext produced and decrypted by Anthropic; Bifrost only carries it through.

## Checklist


core/providers/anthropic/chat.go
```diff
@@ -1264,6 +1309,33 @@ func (chunk *AnthropicStreamEvent) ToBifrostChatCompletionStream(ctx *schemas.Bi
 			return streamResponse, nil, false
 		}
 
+		// Redacted thinking blocks arrive complete in content_block_start (no
+		// deltas follow). Surface the encrypted payload as a reasoning.encrypted
+		// detail so clients can replay it on the next turn; Anthropic rejects
+		// tool-use follow-ups whose latest assistant message dropped it.
+		if chunk.Index != nil && chunk.ContentBlock != nil && chunk.ContentBlock.Type == AnthropicContentBlockTypeRedactedThinking &&
```
R-droid101: I think this can be handled more cleanly, similar to how we handle types in other functions within this provider.

Instead of checking each type inline, we can first guard on `chunk.Index != nil && chunk.ContentBlock != nil`, and then use a `switch` on `chunk.ContentBlock.Type` inside that block. Each supported type can have its own case, with a `default` that returns `nil, nil, false`. This should make the logic easier to read and extend.

Let me know what you think.

core/providers/anthropic/chat.go
```diff
@@ -1349,7 +1421,7 @@ func (chunk *AnthropicStreamEvent) ToBifrostChatCompletionStream(ctx *schemas.Bi
 										Reasoning: schemas.Ptr(thinkingText),
 										ReasoningDetails: []schemas.ChatReasoningDetails{
 											{
-												Index: 0,
+												Index: state.reasoningDetailIndex(*chunk.Index),
```
R-droid101, line 1430: Hey, what's the purpose for changing this?

core/providers/anthropic/redactedthinking_test.go
```diff
@@ -0,0 +1,285 @@
...
+	"github.com/maximhq/bifrost/core/schemas"
+)
+
+// Anthropic returns redacted_thinking blocks when reasoning is flagged by its
+// safety systems, and requires thinking and redacted_thinking blocks to be
+// replayed unmodified on the next turn during tool use. These tests pin the
+// chat-completions round trip of that payload: response conversion preserves
+// it as a reasoning.encrypted detail, request conversion re-materializes the
+// block, and streaming surfaces the payload from content_block_start.
+
```
R-droid101, line 17: It would be great if you could document what each of these test cases test for. A couple of comment lines just describing the test cases will help along with the function names you already have.

core/providers/anthropic/chat.go
fus3r: Agreed, a switch reads better here. Done in 1210fbd3d, tool_use and redacted_thinking each have their own case now with a silent default. The structured output early return lives inside the tool_use case.

core/providers/anthropic/chat.go
fus3r: reasoning_details get grouped by index downstream, the streaming accumulator does it and OpenRouter style clients do the same. These deltas hardcoded index 0, which was fine while thinking was the only reasoning emitter, but a redacted block in the same stream would now share index 0 with the visible thinking. The two get merged into a single detail typed reasoning.text that also carries data, and since request conversion only turns reasoning.encrypted details back into redacted_thinking blocks, the payload would be dropped again on replay. Per block indices keep the entries separate. Single reasoning block streams still emit index 0 like before, and TestToBifrostChatCompletionStream_MixedReasoningBlocksKeepDistinctIndices pins the mixed case.

core/providers/anthropic/redactedthinking_test.go
fus3r: Done in 1210fbd3d, each test has a short description now.

TejasGhatte (approved):

Pratham-Mishra04 (approved):

PR #4957, opened 2026-07-06
feat: add periodic per-provider live model-list refresh

### LINKED ISSUE #2964: Model Catalog refresh
### Prerequisites

### Problem to solve

I use a local ollama Model provider and I there add and remove models quite frequently and if I use a direct connection to the ollama instance the model list is updated promptly. However, for clients connected via bifrost to the same ollama provider the model list is not updated when I add new models and I need to "resave the ollama provider" for the model list to update. 

I am a newbie to bifrost so it might be something that I have configured wrongly or is there a way to make the model list to update or disable the model cache for specific providers. 

### Proposed solution

A button to disable the model cache or set it to a configurable value per provider. 

### Alternatives considered

The only solutions I am evaluating are : 
1. Do not use bifrost for the ollama inference backends 
2. Implemet some type of cron job that "forces a refresh" of the ollama provider. 

### Area(s)

UI (React)

### Additional context

We use openwe-ui for frontend and a mix of cloaud and local model providers (ollama and vllm)

Shaik-Sirajuddin:
## Summary

Adds an optional `list_models_refresh_interval_sec` on providers (HTTP API, UI Performance tab, `config.json` schema) that enables periodic background refresh of a provider's live list-models cache, independent of the existing reactive refresh (on boot / provider add-update / key add-update).

Closes #2964

## Algorithm

- Dedicated 10s ticker (separate from the 5-min pricing/MCP ticker) fetches any provider past its `interval` since `lastFetchedAt`, flagging it `fetching` to prevent double-picks.
- All reactive refresh paths funnel through the same in-flight-guarded function as the ticker, preventing duplicate concurrent fetches.
- Disabled keys are skipped; a keyless provider with only a disabled key is skipped too, not fetched under the keyless sentinel.
- A 30s minimum interval is enforced everywhere (API, UI, `config.json`).

## Type of change

- [x] Feature

## Breaking changes

- [ ] Yes
- [x] No (opt-in field)

## Checklist

- [x] Added tests


akshaydeo:
@Shaik-Sirajuddin this misses a few aspects from structural point of view

we have started official work here https://github.com/maximhq/bifrost/pull/5561 and will be merged. Please look at this to construct a few things for future PRs 🙇

PR #1995, opened 2026-03-09
fix: map chat finish_reason to responses status and preserve terminal stream semantics

jerkeyray:
## Summary

Implements proper handling of incomplete responses in the OpenAI provider by mapping `length` finish reasons to incomplete status and preventing duplicate finish reason forwarding in streaming responses.

## Changes

- Added logic to map `length` finish reason to `incomplete` status with `max_output_tokens` reason in both streaming and non-streaming responses
- Introduced tracking to prevent forwarding finish reasons twice when they've already been sent in streaming chunks
- Added helper functions `responsesStatusFromChatFinishReason` and `responsesTerminalFromChatFinishReason` to centralize finish reason mapping logic
- Modified streaming response handling to emit `response.incomplete` events instead of `response.completed` when appropriate
- Added comprehensive test coverage for finish reason mapping in both response formats

## Type of change

- [ ] Feature
- [x] Bug fix
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [x] Core (Go)
- [ ] Transports (HTTP)
- [x] Providers/Integrations
- [ ] Plugins
- [ ] UI (Next.js)
- [ ] Docs

## How to test

Validate the finish reason mapping and incomplete response handling:

```sh
# Core/Transports
go version
go test ./...

# Test specific functionality
go test ./core/schemas -v -run TestToBifrostResponsesResponse_MapsLengthToIncomplete
go test ./core/schemas -v -run TestToBifrostResponsesStreamResponse_MapsLengthToIncompleteEvent
```

Test with OpenAI provider responses that hit token limits to verify `incomplete` status is properly set with `max_output_tokens` reason.

## Screenshots/Recordings

N/A

## Breaking changes

- [ ] Yes
- [x] No

## Related issues

Closes #1976
Closes #1821

## Security considerations

No security implications - this change only affects response status mapping and metadata handling.

## Checklist

- [x] I read `docs/contributing/README.md` and followed the guidelines
- [x] I added/updated tests where appropriate
- [ ] I updated documentation where needed
- [x] I verified builds succeed (Go and UI)
- [x] I verified the CI pipeline passes locally if applicable

PR #4964, opened 2026-07-06
fix: preserve streaming finish_reason in the accumulated response when forwarded on a content chunk

### LINKED ISSUE #4963: [Bug]: streaming finish_reason is dropped from the accumulated (logged) response when a provider sends it on a content chunk
### Prerequisites

### Description

When a streaming chat response is accumulated into a single response (the shape used for logging and post-hook plugins), the streaming accumulator reads `finish_reason` only from the chunk with the highest chunk index (`getLastChatChunkLocked`, called from `framework/streaming/chat.go`). That works for providers that send `finish_reason` on a dedicated terminal chunk with an empty delta, but not for providers that send it on the final content chunk.

The OpenAI-compatible streaming handler already supports the latter: when a chunk carries content and a `finish_reason` together, it forwards that chunk as-is and sets `forwardedTerminalFinishReason`, then appends a synthetic terminal chunk (`CreateBifrostChatCompletionChunkResponse`, chunk index + 1) whose `finish_reason` is deliberately nil to avoid emitting `finish_reason` twice to the client (the guard added in #1995). So the highest-index chunk carries a nil `finish_reason` while the real one sits on the content chunk one index lower.

The accumulator then records a nil `finish_reason`. Every sibling metadata field at the same site (`TokenUsage`, `Cost`, `CacheDebug`) is nil-guarded; only `finish_reason` is assigned unconditionally. The accumulated value feeds the logging plugin (`entry.StopReason`) and the Maxim plugin, so the stored/observed stop reason is null for these streams even though the client received the correct value in the stream. Non-streaming requests read `finish_reason` straight from the response and are unaffected, so this is a streaming-only observability gap.

### Steps to reproduce

Deterministic, no API key. Drive an OpenAI-compatible stream whose final content chunk carries `finish_reason` (a local httptest SSE server works), for example the events:

```
data: {"choices":[{"index":0,"delta":{"role":"assistant","content":"Hello"}}]}
data: {"choices":[{"index":0,"delta":{"content":" world"},"finish_reason":"stop"}]}
data: {"choices":[],"usage":{"prompt_tokens":10,"completion_tokens":2,"total_tokens":12}}
data: [DONE]
```

The handler emits chunk index 1 with `finish_reason="stop"` and a synthetic chunk index 2 with `finish_reason=null`. Feed that chunk sequence through the streaming accumulator and the accumulated `finish_reason` comes back `null`. With the logging plugin enabled, the log entry's `stop_reason` is empty for such a stream.

### Expected behavior

The accumulated (and therefore logged) response keeps the real `finish_reason` (`stop`, `length`, `tool_calls`, ...) whenever any chunk in the stream carried one, matching what the client received and what the non-streaming path records.

### Actual behavior

The accumulated `finish_reason` is null whenever the provider sends `finish_reason` on a content chunk, because the accumulator only looks at the highest-index chunk, which is the synthetic terminal chunk that intentionally has a nil `finish_reason`.

### Affected area(s)

Framework

### Version

Current dev (`7022fa17e`); the accumulator's finish_reason extraction predates the `forwardedTerminalFinishReason` handling added in #1995 and was never updated to match.

### Severity

fus3r:
## Summary

Fixes #4963. The streaming chat accumulator reconstructs a single response for logging and post-hook plugins, and reads `finish_reason` only from the highest-index chunk (`getLastChatChunkLocked`). For providers that send `finish_reason` on the final content chunk, the OpenAI-compatible handler forwards it on that chunk and appends a synthetic terminal chunk (chunk index + 1) whose `finish_reason` is nil, to avoid emitting `finish_reason` twice to the client (the `forwardedTerminalFinishReason` guard from #1995). The highest-index chunk therefore has a nil `finish_reason` while the real one sits one index lower, and the accumulator records null. Unlike the sibling `TokenUsage`, `Cost` and `CacheDebug` fields at the same site, `finish_reason` was assigned without a nil check. The accumulated value feeds the logging plugin (`entry.StopReason`) and Maxim, so streaming logs record an empty stop reason for these providers even though the client got the right value; non-streaming requests read it straight from the response and are unaffected.

## Changes

- Added `getChatFinishReasonLocked`, which returns the `finish_reason` of the highest-index chunk that actually carries one, mirroring the existing `getLastChatChunkLocked` helper.
- In `processAccumulatedChatStreamingChunks`, fall back to it only when the highest-index chunk has a nil `finish_reason`, so the common path (finish_reason on the terminal chunk) is unchanged and does no extra scan.
- Regression tests: one for `finish_reason` forwarded on a content chunk (fails on `dev`), one for the standard terminal-chunk shape (guards against over-reaching). The text-completion path already strips `finish_reason` from the forwarded chunk and defers it to the tail, so it is unaffected; the Responses accumulator never sets a per-chunk `finish_reason` from content, so it is out of scope.
- Changelog entry.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [ ] Core (Go)
- [ ] Transports (HTTP)
- [ ] Providers/Integrations
- [x] Plugins
- [ ] UI (React)
- [ ] Docs

The change is in the `framework/streaming` accumulator (there is no Framework box); it surfaces through the logging and Maxim plugins, which consume the accumulated response.

## How to test

```sh
cd framework
go test ./streaming/ -run 'TestChatStreamingFinishReason' -v
```

`TestChatStreamingFinishReasonForwardedOnContentChunk` fails on `dev` (the accumulated `finish_reason` is nil) and passes with this change; `TestChatStreamingFinishReasonOnTerminalChunk` passes before and after (it pins the standard shape). The full `./streaming/` suite is green, including under the race detector.

## Screenshots/Recordings

N/A

## Breaking changes

- [ ] Yes
- [x] No

Only the accumulated/logged response changes, and only in the case where it was previously losing the `finish_reason`.

## Related issues

Fixes #4963

## Security considerations

None.

## Checklist


R-droid101 (approved):
LGTM

akshaydeo (dismissed):

PR #4987, opened 2026-07-07
fix: add missing Gemini GenerationConfig/GoogleSearch/response fields from OpenAPI spec

### LINKED ISSUE #5166: [Bug]: Gemini generateContent request/response fields (translation_config, response_format, search_types, model_status) silently dropped
Bifrost's Gemini provider types (`core/providers/gemini`) don't model several fields present in Google's `generateContent` OpenAPI spec:

- `GenerationConfig.translation_config`
- `GenerationConfig.response_format`
- `GenerationConfig.enable_enhanced_civic_answers`
- `GoogleSearch.search_types`
- `GenerateContentResponse.model_status`

A client using the official `google-genai` SDK/CLI that sets any of the request fields above gets no error back — the fields simply aren't in Bifrost's Go structs, so they're dropped before the request reaches Gemini and the call silently falls back to default behavior (e.g. no translation, wrong search scope) instead of honoring what was sent.

`model_status` is likewise missing from Bifrost's response type, so any client code that inspects it (streaming or non-streaming) won't see it, even though Gemini returns it natively.

Shaik-Sirajuddin:
## Description

Closes #5166

Adds 4 real Gemini `generateContent` OpenAPI config/response fields that were missing from Bifrost's Gemini provider types, found via a schema-sync comparison against Google's published OpenAPI spec.

## Details

- Added `GenerationConfig.ResponseFormat`/`TranslationConfig`/`EnableEnhancedCivicAnswers`, `GoogleSearch.SearchTypes`, and `GenerateContentResponse.ModelStatus` (works for both streaming and non-streaming since both paths share the same struct).
- Wired `translation_config`/`enable_enhanced_civic_answers`/`response_format` as named `extra_params` keys on both chat completions and responses request paths, matching the existing `safety_settings`/`cached_content`/`labels` pattern.
- Scoped to `core/providers/gemini` only (Gemini-only wire types, Vertex has its own copy); fields with no OpenAI-canonical equivalent stay `extra_params`-only by design (see `gen/gemini-openapi-config-gaps.md`).

## Checklist

- [x] Tests added/updated and passing (`go test ./...`)
- [ ] Verified against a real endpoint (not just mocked/unit tests)
- [x] Docs updated (or N/A) — internal scope note in `gen/`, no user-facing docs impacted


Shaik-Sirajuddin:
## E2E verification (official Python OpenAI SDK, real Gemini key)

Ran before/after against two live Bifrost gateways (`dd5df2bb8` vs this branch), hitting `/v1/chat/completions` with `translation_config`, `enable_enhanced_civic_answers`, and Gemini's native `response_format` via `extra_body`.

All 4 cases returned `200` in both runs (Gemini is lenient about unknown fields), so response text alone wasn't a reliable signal. Decisive check was the actual outgoing `generationConfig` Bifrost builds:

**`response_format: {"text": {"mimeType": "text/plain"}}` sent as a real `response_format` param (as any OpenAI-SDK client would send it):**

```
before (dd5df2bb8, generic before this PR): GenerationConfig = {}
after  (a2b4d524e, this PR's original commit): GenerationConfig = {}   <- bug: dropped
after  (06fb0ad0f, latest): GenerationConfig = {"responseFormat":{"text":{"mimeType":"text/plain"}}}
```

Root cause: `/v1/chat/completions` treats `response_format` as a known top-level field (`chatParamsKnownFields` in `transports/bifrost-http/handlers/inference.go`), so it's parsed into the canonical `ChatParameters.ResponseFormat`, never into `ExtraParams`. The original extraction only read `ExtraParams`, so real clients silently lost this field — only reachable in unit tests that set `ExtraParams` directly. Fixed in `06fb0ad0f` by handling the Gemini-native shape (no `"type"` discriminator) inside `convertParamsToGenerationConfig`, without touching the existing OpenAI `json_schema`/`json_object` mapping (regression test added).

`translation_config` / `enable_enhanced_civic_answers` don't have this problem — neither key is in `chatParamsKnownFields`, so both land in `ExtraParams` correctly via the real HTTP path already.

PR #5011, opened 2026-07-07
fix: activate Anthropic fine-grained tool streaming on Bedrock Converse

### LINKED ISSUE #4965: [Bug]: Bedrock never activates fine-grained-tool-streaming — tool.eager_input_streaming is silently dropped
### Prerequisites

### Description

`ChatTool.EagerInputStreaming` (`core/schemas/chatcompletions.go:390`) is Bifrost's neutral-layer flag for Anthropic's `fine-grained-tool-streaming-2025-05-14` beta. The provider feature table (`core/providers/anthropic/types.go:199-216`) declares `EagerInputStreaming: true` for Bedrock, so the sanitizer (`anthropic/utils.go:396-397`) correctly preserves the flag instead of stripping it — but nothing on the Bedrock path ever reads it afterward.

On the direct Anthropic provider, the flag is consumed at `anthropic/utils.go:1101-1103`, which appends the `fine-grained-tool-streaming-2025-05-14` beta header when a tool has it set. Bedrock has no equivalent. `convertChatParameters` in `core/providers/bedrock/utils.go` builds tools into the typed Converse `toolConfig.tools[].toolSpec` shape, which has no eager-streaming slot, and never inspects `tool.EagerInputStreaming` at all. `grep -rn "EagerInputStreaming" core/providers/bedrock/` turns up only a test fixture (`bedrock_test.go:232`) asserting the feature-table entry — zero production code paths reference it.

This can't be worked around from the request side: Bedrock's Converse endpoint ignores the outer HTTP `anthropic-beta` header entirely (confirmed in-code at `bedrock/utils.go:685-687` — it's "consumed by Bedrock's edge and NOT forwarded to the underlying Claude model"); the only channel Bedrock honors is the body field `additionalModelRequestFields.anthropic_beta`, and the only code that currently writes to it is the server-tools tunnel (`utils.go:266-272`), gated on server-tool presence — not reachable for a plain custom tool with `eager_input_streaming: true`.

Net effect: setting `eager_input_streaming: true` on a custom tool is a silent no-op on Bedrock. No error, no fallback — the model just streams the full tool-call arguments as one blob instead of incremental `input_json_delta` chunks, adding multi-second latency to first-byte on large tool calls.

### Steps to reproduce

1. Send a streamed chat/messages request through Bifrost to a Bedrock-backed Claude model with a custom tool that sets `"eager_input_streaming": true`.
2. Capture the outbound Converse request body (or add a log at `bedrock/utils.go` before the request is sent).
3. Observe: no `additionalModelRequestFields` field is present, or if present (e.g. from server tools also in the request), it has no `anthropic_beta` entry containing `fine-grained-tool-streaming-2025-05-14`.
4. Observe the response: `input_json_delta` chunks for the custom tool call arrive as a single burst instead of incrementally, versus the same request against the direct Anthropic provider or a `/anthropic` messages request where the beta header path applies.

### Expected behavior

When any tool in the request has `eager_input_streaming: true` and the target provider supports the feature (Bedrock does, per the feature table), Bifrost should inject `fine-grained-tool-streaming-2025-05-14` into `additionalModelRequestFields.anthropic_beta` on the Converse request, and the model should stream `input_json_delta` increments for that tool.

### Actual behavior

The flag is silently ignored on Bedrock; tool-call arguments are never streamed incrementally regardless of the flag.

### Suggested fix

In `convertChatParameters` (`bedrock/utils.go`), add a sibling check to the existing server-tools tunnel block (~line 266-272): if any tool in `filteredTools` has `EagerInputStreaming != nil && *EagerInputStreaming`, ensure `AdditionalModelRequestFields` is initialized and call the existing `appendAnthropicBetaToFields(bedrockReq.AdditionalModelRequestFields, anthropic.AnthropicEagerInputStreamingBetaHeader)` helper — the same helper the server-tools path already uses, so it merges correctly instead of clobbering.

### Affected area(s)

Core (Go)

### Version

dev @ e0c541da8 (also present, same line numbers modulo drift, on transports/v1.6.2)

### Environment

```text
N/A — traced via source inspection, not a runtime-environment-specific bug.
```

### Relevant logs/output

```shell
N/A
```

### Regression?

_No response_

### Severity

fus3r:
## Summary

Fixes #4965. `eager_input_streaming` (Anthropic's `fine-grained-tool-streaming-2025-05-14` beta) is a silent no-op on Bedrock. `ChatTool.EagerInputStreaming` is preserved by the sanitizer (the feature table declares the beta supported for Bedrock), but nothing on the Converse path ever reads it: Converse can't express the beta in its typed `toolSpec` shape, and Bedrock's edge strips the outer `anthropic-beta` HTTP header. Only `additionalModelRequestFields.anthropic_beta` is honored, and the only code writing that field was the server-tools tunnel, which a plain custom tool never reaches.

The Responses path had the same gap: `ResponsesTool.EagerInputStreaming` survives provider filtering, but the Bedrock Responses converter never reads it. As a result, setting `eager_input_streaming: true` on a Bedrock-backed Claude model streams tool-call arguments as one blob instead of incremental `input_json_delta` chunks through both Chat Completions and Responses.

The direct Anthropic and Azure paths send the beta on the `anthropic-beta` header and Vertex embeds it in the request body via its own `anthropic_beta` injection. Bedrock Converse builds a typed request that carries the flag on neither channel, so it is silently dropped without this tunnel.

## Changes

- In `convertChatParameters` (`bedrock/utils.go`), mirror the server-tools tunnel: when a tool sets `EagerInputStreaming`, append `fine-grained-tool-streaming-2025-05-14` to `additionalModelRequestFields.anthropic_beta` through the shared `appendAnthropicBetaToFields` helper.
- In `ToBedrockResponsesRequest` (`bedrock/responses.go`), apply the same tunnel to the provider-filtered Responses tools after `ExtraParams` are processed, preserving caller-provided beta values.
- Gate both injections on the Anthropic model family so Nova, Llama, and other Converse models never receive an Anthropic-only field.
- Add Chat-path regression coverage for activation, flag-off/absent behavior, the Nova gate, and coexistence with a server-tool beta.
- Add equivalent Responses-path coverage, including preservation of an existing caller-provided beta.
- Add a changelog entry.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [x] Core (Go)
- [ ] Transports (HTTP)
- [x] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

```sh
cd core
go test ./providers/bedrock/ -run 'Test(BedrockConverseEagerInputStreaming|ToBedrockResponsesRequest_EagerInputStreaming)' -count=1
go test ./providers/bedrock/ -count=1
go test -race ./providers/bedrock/ -run 'Test(BedrockConverseEagerInputStreaming|ToBedrockResponsesRequest_EagerInputStreaming)' -count=1
```

All targeted tests, the full `./providers/bedrock/` package, and the targeted race-detector run pass locally.

## Screenshots/Recordings

N/A

## Breaking changes

- [ ] Yes
- [x] No

Only adds beta activation for a flag Bedrock previously ignored; requests without `eager_input_streaming` are unchanged.

## Related issues

Fixes #4965

## Security considerations

None.

## Checklist


Kunal-Shahi (approved):

public-aanp-tanium (approved):
LGTM, just needs a rebase/address conflicts.

jeremym-tanium (approved):

PR #5018, opened 2026-07-07
Filter out provider-level keys from selector in prompt manager

rlex:
## Summary

In current bifrost code, when VK is enforced, prompt manager allows you to select provider key - but request will fail, since prompt manager is basically a web client, asking you to pass VK header instead. It's a bit confusing for user.

## Changes

Tiny change, really.
UI now filters out provider keys during load - if VK is enforced, only VK keys are displayed.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [ ] Core (Go)
- [ ] Transports (HTTP)
- [ ] Providers/Integrations
- [ ] Plugins
- [x] UI (React)
- [ ] Docs

## How to test

Add one provider with key.
Add one VK key, attach provider key to it.

Without VK enabled, should show both VK and provider keys.
With VK enabled, should show only VK ones.

## Screenshots/Recordings

VK enabled:
Screenshot 2026-07-08 at 02 31 51
VK disabled:
Screenshot 2026-07-08 at 02 33 19

## Breaking changes

- [ ] Yes
- [x] No

## Related issues

## Security considerations

## Checklist

### DIFF
--- a/ui/components/prompts/fragments/settingsPanel.tsx
+++ b/ui/components/prompts/fragments/settingsPanel.tsx
@@ -8,12 +8,13 @@ import { Skeleton } from "@/components/ui/skeleton";
 import { getProviderLabel } from "@/lib/constants/logs";
 import { Input } from "@/components/ui/input";
 import { useGetVirtualKeysQuery } from "@/lib/store";
+import { useGetCoreConfigQuery } from "@/lib/store/apis/configApi";
 import { useGetAllKeysQuery, useGetProvidersQuery } from "@/lib/store/apis/providersApi";
 import { ModelProviderName } from "@/lib/types/config";
 import { ModelParams } from "@/lib/types/prompts";
 import { cn } from "@/lib/utils";
 import { PromptDeploymentsAccordionItem } from "@enterprise/components/prompt-deployments/promptDeploymentsAccordionItem";
-import { useCallback, useMemo, useState } from "react";
+import { useCallback, useEffect, useMemo, useState } from "react";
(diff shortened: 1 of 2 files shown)


KSHITIZ6341 (commented):
One enforced virtual-key state note.

ui/components/prompts/fragments/settingsPanel.tsx
```diff
@@ -200,13 +211,14 @@ export function SettingsPanel() {
 									/>
 								</div>
 
-								{(providerKeys.length > 0 || providerVirtualKeys.length > 0) && !!provider && (
+								{((!enforceVirtualKeys && providerKeys.length > 0) || providerVirtualKeys.length > 0) && !!provider && (
 									<ApiKeySelectorView
 										providerKeys={providerKeys}
 										virtualKeys={providerVirtualKeys}
 										value={apiKeyId}
```
KSHITIZ6341, line 218: Could we also handle the `__auto__` state when virtual keys are enforced? The selector now hides provider keys, but the context can still keep `apiKeyId === "__auto__"`. In that state `executePrompt` sends no `Authorization` header, so the request still fails in an enforced-VK environment even though the UI no longer shows provider keys. One option would be to require/select a valid `providerVirtualKey` when `enforceVirtualKeys` is true, or disable model/send until a VK is selected.

rlex: Oh, that's good suggestion. I'll try to tackle it a bit later.

akshaydeo (approved):

Issue #5019, opened 2026-07-08
Audio/video billing truncates fractional seconds, undercounting cost


Shaik-Sirajuddin:
### Description

Bifrost's cost-calculation pipeline treats audio/video duration usage as whole seconds (`int`) everywhere, then multiplies by the per-second rate. Any fractional part of a duration is truncated before the cost is computed, so requests are silently undercharged relative to the provider's actual bill.

This also caused a related crash: OpenAI's duration-based usage returns `seconds` as a float (e.g. `521.5`), but Bifrost's usage struct originally only accepted whole numbers there — a fractional value would fail to parse and abort the request (same class of bug as #5002).

### Why it matters

Provider billing granularity varies a lot, so a single hardcoded whole-second truncation doesn't reflect any of it accurately:

| Provider | Billing granularity |
|---|---|
| Deepgram | Fractional-second billing |
| AssemblyAI | Fractional-second billing |
| AWS (Transcribe) | Rounds to whole seconds (with a 15s floor) |
| Azure | Rounds to whole seconds |
| OpenAI | Token-quantized audio/video (fine-grained rounding, not continuous) |
| Gemini | Token-quantized audio/video |
| Sora / Veo (video) | Billed per exact second of output |
| ElevenLabs | Per character, not time |
| AWS Polly | Per character, not time |
| Azure TTS | Per character, not time |
| Anthropic | No audio/video billing |

This skews cost tracking away from the real invoice as usage scales.

### Suggested fix

Carry fractional seconds through the cost pipeline instead of truncating early, and where a provider is genuinely known to round for billing, model that explicitly per-provider rather than relying on an incidental `int` type.

Note this should **not** be implemented by adding another raw `float64` multiply into the cost path. #4755 tracks removing `float64` entirely from monetary values (cost/budget columns, the cost engine's return type, budget accumulation) due to float drift, and this fix touches that same multiplication step (`duration × per-second rate`). To avoid reintroducing the exact antipattern #4755 flags:

- Represent fractional duration at fixed precision (e.g. integer milliseconds, or whatever fixed-point/decimal type #4755 lands on) rather than as a `float64` seconds value.
- Feed that fixed-point duration into the cost multiplication so the result lands directly in the non-float representation #4755 is introducing, instead of round-tripping through `float64`.
- If #4755 hasn't landed yet, sequence this fix to build on top of it (or land it using the same decimal type) rather than shipping a float64-based fractional-seconds fix that #4755 would then have to re-touch.

Related: #4755

PR #5020, opened 2026-07-08
fix(transcription): support diarized_json segments, fix ElevenLabs speaker passthrough

### LINKED ISSUE #5002: [Bug]: OpenAI `gpt-4o-transcribe-diarize` transcription fails through Bifrost due to string segment IDs
### Prerequisites

## Description

   When sending an audio transcription request through Bifrost using OpenAI's `gpt-4o-transcribe-diarize` model with `response_format=diarized_json`, Bifrost returns HTTP 500.

   The same request works when sent directly to OpenAI.

   Bifrost appears to fail while unmarshalling the OpenAI provider response because OpenAI returns diarized segment IDs as strings, for example `"id":"seg_154"`, while Bifrost seems to expect an `int64`.

 The error points to this part of the OpenAI response:

 ```json
   {
     "end": 520.968,
     "id": "seg_154"
   }
 ```

 This suggests Bifrost has a strict transcription segment schema where id is typed as int64. For OpenAI diarized responses, id can be a string.

### Steps to reproduce

   ```bash
   curl --fail-with-body -X POST "http://<bifrost-host>:8080/v1/audio/transcriptions" \
     -H "Authorization: Bearer <redacted>" \
     -F "file=@./sample.mp3" \
     -F "model=openai/gpt-4o-transcribe-diarize" \
     -F "response_format=diarized_json" \
     -F "chunking_strategy=auto"
 ```

### Expected behavior

Bifrost should return the OpenAI transcription response instead of returning HTTP 500.

### Actual behavior

```json
   {
     "is_bifrost_error": true,
     "error": {
       "error": "Mismatch type int64 with value string \"at index 27473: mismatched type with value\\n\\n\\td\\\":520.968,\\\"id\\\":\\\"seg_154\\\"}],\\\"usa\\n\\t................^...............\\n\"",
       "message": "failed to unmarshal response from provider API"
     },
     "extra_fields": {
       "routing_info": {
         "provider": "openai",
         "model": "gpt-4o-transcribe-diarize",
         "key": "Test Key"
       },
       "provider": "openai",
       "original_model_requested": "gpt-4o-transcribe-diarize",
       "resolved_model_used": "gpt-4o-transcribe-diarize",
       "request_type": "transcription",
       "latency": 152788
     }
   }
 ```

### Affected area(s)

Core (Go)

### Version

1.6.3

### Environment

```text

```

### Relevant logs/output

```shell

```

### Regression?

_No response_

### Severity

Low (minor issue or cosmetic)

Shaik-Sirajuddin:
Closes #5002

## Summary

`response_format=diarized_json` (OpenAI's `gpt-4o-transcribe-diarize`) returns segments with a string `id` plus `speaker`/`type` fields, which crashed unmarshalling into `TranscriptionSegment`'s int `id`. Adds a distinct diarized segment type and fixes the decode path, plus a related gap in ElevenLabs' speaker passthrough.

## Changes

- Added `schemas.TranscriptionDiarizedSegment` (string id/speaker/type) and decode it separately from `TranscriptionSegment` in the OpenAI provider's transcription paths (Azure inherits the fix via the shared handler). `BifrostTranscriptionResponse` gets a matching `MarshalJSON`/`UnmarshalJSON` pair since both segment kinds share the `"segments"` key, including correct round-tripping through `framework/logstore`'s persist/reload cycle.
- ElevenLabs decoded `speaker_id` per word but never propagated it into the canonical `TranscriptionWord`; added a `Speaker` field and wired it through.

## Type of change

- [x] Bug fix

## Affected areas

- [x] Core (Go)
- [x] Transports (HTTP)
- [x] Providers/Integrations

## Breaking changes

- [x] No

## Related issues

Related but out of scope for this PR: #5019 (fractional-second billing truncation, a broader pre-existing pipeline behavior).

## Checklist


LauraGPT:
I reviewed this from the downstream OpenAI-compatible STT gateway angle. This looks good for clients and self-hosted transcription providers that need Bifrost to preserve provider-specific transcription shape rather than normalizing it too early.

The important compatibility pieces for that path are covered here:

- `diarized_json` segments with string IDs such as `seg_154` decode into a separate diarized segment type instead of crashing the canonical verbose segment parser.
- Empty diarized `segments: []` still round-trips as an explicit `segments` key, which matters for SDK/client validation.
- multipart transcription params now keep repeated `timestamp_granularities[]`, `temperature`, `chunking_strategy`, and provider-specific extras instead of dropping them before the provider adapter sees them.
- ElevenLabs `speaker_id` is carried through to the canonical word-level `speaker` field.

No FunASR/SenseVoice-side blocker from my pass. The only follow-up I would watch after merge is whether additional OpenAI-compatible transcription backends return diarized segment IDs as numbers or strings interchangeably; this PR already protects the currently failing string-ID case without regressing the existing verbose path.

akshaydeo (approved):

PR #4954, opened 2026-07-06
fix: marshal `tool_sync_interval` as duration string, support negative sync intervals, and fix `tool_execution_timeout` round-trip for MCP clients

Pratham-Mishra04:
## Summary

`tool_sync_interval` was being serialized as raw nanoseconds (Go's default `time.Duration` JSON encoding) while `tool_execution_timeout` already emitted a human-readable duration string. This mismatch caused `tool_sync_interval` to round-trip incorrectly between the API and the UI, and the storage layer was treating the interval as seconds instead of minutes. This PR aligns both fields to use Go duration strings on the wire, fixes the minutes/seconds unit confusion in storage and config application, and hardens the API handlers with overflow-safe validation.

## Changes

- Added `MarshalJSON` to `MCPConfig` so `tool_sync_interval` is emitted as a duration string (e.g. `"10m0s"`) instead of nanoseconds, matching the existing `MCPClientConfig.MarshalJSON` behavior.
- Extended `MCPClientConfig.MarshalJSON` to also emit `tool_sync_interval` as a duration string; negative values (sync disabled) are preserved correctly.
- Fixed `GetMCPConfig` in the RDB store to multiply `MCPToolSyncInterval` by `time.Minute` instead of `time.Second`, matching the stored unit (whole minutes).
- Fixed `updateConfig` in the HTTP config handler to apply the same `time.Minute` multiplier when reading `MCPToolSyncInterval` from the client config.
- Fixed `applyMCPGlobalSettingsToClientConfig` to mirror `ToolSyncInterval` in whole minutes rather than whole seconds, and updated the granularity warning accordingly.
- Removed the negative-interval rejection from `toolSyncIntervalDurationToStoredSeconds`; negative values are now valid and mean "sync disabled."
- Added `validateToolSyncIntervalMinutes` and `validateToolExecutionTimeoutSeconds` helpers in the MCP HTTP handler to guard against `int64` overflow when converting request integers to `time.Duration`.
- Applied the new validators in both `addMCPClient` and `updateMCPClient`, and ensured `ToolExecutionTimeout` is populated from the request in `addMCPClient` (it was previously missing).
- Propagated `TLSConfig` and `ToolExecutionTimeout` into the OAuth completion response structs where they were previously omitted.
- Simplified the disable-toggle call in `mcpClientsTable.tsx` to send only `{ disabled }`, avoiding the unsafe practice of echoing GET response values (duration strings) back as request units.
[...]

PR #5049, opened 2026-07-09
fix(mcp): stop enable/disable toggle from corrupting tool_sync_interval

### LINKED ISSUE #5026: [Bug]: Toggling an MCP client's enable/disable switch corrupts its tool_sync_interval (nanoseconds resent as minutes)
### Prerequisites

### Description

In the MCP Registry clients table, the enable/disable **Switch** resends the client config on toggle, and that payload includes `tool_sync_interval`. The value it sends is the raw value returned by `GET`, which the API serializes as a Go `time.Duration` in **nanoseconds**. The `PUT` handler, however, interprets `tool_sync_interval` as **minutes** and multiplies by `time.Minute`.

For any client that has a **non-zero per-client sync-interval override**, this overflows `int64` and persists a garbage negative interval, which can break the client's tool sync/connection. Clients using the global default (override `0`) are unaffected (`0 × minute = 0`). The corruption is silent — nothing in the UI indicates the interval was mangled by the toggle.

### Related

Related to maximhq/bifrost#4998 / maximhq/bifrost#4314 (the `/mcp` gateway tool-registry desync). Those issues recommend a disable→enable toggle as the recovery workaround — note that performing that toggle **via the dashboard switch** is what triggers this corruption, so the workaround is only safe via a raw `PUT {"disabled": ...}` that omits `tool_sync_interval`.

### Steps to reproduce

1. Create or edit an HTTP/SSE MCP client and, in the edit sheet, set a **Tool Sync Interval** override to e.g. `10` (minutes). Save.
2. In the MCP clients table, flip that client's enable/disable **Switch**.
3. Reopen the client (or `GET /api/mcp/clients`) and inspect `tool_sync_interval`.

### Expected behavior

`tool_sync_interval` is unchanged by the toggle (still 10 minutes / `600000000000` ns).

### Actual behavior

`tool_sync_interval` becomes a huge negative value, e.g. `-8044431881000000000`, leaving the client with an invalid sync interval.

### Affected area(s)

UI (React), Transports (HTTP)

### Version

dev @ c20f3836 (2026-07-08); also present on the v1.6.x release line.

### Environment

* Bifrost HTTP transport + UI (React), self-hosted; any OS/browser.
* Reproduced against current `dev`.

### Relevant logs/output

```
# Corruption math for a 10-minute override:
600000000000 ns  (GET value, unconverted)
  × time.Minute  -> -8044431881044754432 ns  (int64 overflow)
  / time.Second  -> -8044431881            (DB write, truncated)
  × time.Second  -> -8044431881000000000    (reload)
```

### Regression?

Latent unit mismatch between the ns-on-wire serialization and the minutes-on-write handler; the table toggle exposes it whenever a client has a non-default sync interval.

### Severity

---

## Root cause (source anchors, dev @ c20f3836)

* `ui/app/workspace/mcp-registry/views/mcpClientsTable.tsx:510` — the toggle resends `tool_sync_interval: c.config.tool_sync_interval ?? 0`, i.e. the raw `GET` value (nanoseconds), with no conversion.
* `core/schemas/mcp.go:427-438` — `MCPClientConfig.MarshalJSON` overrides **only** `tool_execution_timeout` (→ seconds string); `tool_sync_interval` (a `time.Duration`) falls through to default int64 **nanoseconds** marshaling. The frontend documents this: `mcpClientSheet.tsx:54` — "API sends tool_sync_interval as nanoseconds".
* `transports/bifrost-http/handlers/mcp.go:1098` — the `PUT` handler does `resolvedToolSyncInterval = time.Duration(*req.ToolSyncInterval) * time.Minute`, treating the field as minutes.

The edit **sheet** is correct and not the culprit: `toolSyncIntervalToMinutes` (`mcpClientSheet.tsx:55-61`) converts ns→minutes on load and submits minutes. Only the table Switch path skips the conversion.

## Proposed fix

Have the toggle send only the field it changes plus identity, e.g. `{ name, ..., disabled: !checked }`, and **not** resend `tool_sync_interval`. The `PUT` handler is effectively PATCH — omitted fields are preserved (`transports/bifrost-http/handlers/mcp.go:1044-1107`). (Optional secondary: fix the optimistic cache write in `ui/lib/store/apis/mcpApi.ts` which writes minutes into the cached raw config.) A deeper fix would (de)serialize `tool_sync_interval` in one stable unit the way `tool_execution_timeout` already does.

Shaik-Sirajuddin:
## Summary

The MCP client table's enable/disable Switch resent `tool_sync_interval` (nanoseconds, from `GET`) into `PUT`, which parses that field as minutes — overflowing `int64` into a garbage negative duration for clients with a non-default sync interval.

## Changes

- Toggle now sends only `{ disabled }`; the update handler already has PATCH semantics.
- Fixed the optimistic RTK Query cache write (minutes → nanoseconds) to match.
- Added `DurationFromUnits`, an overflow-guarded helper, and used it everywhere `tool_sync_interval`/`tool_execution_timeout` are derived from request or config data.
- Fixed two more real unit bugs found while auditing: global `tool_sync_interval` reload used seconds instead of minutes in two places (`config.go`, `lib/config.go`).
- Fixed a pointer-aliasing hazard in `config.go`'s `updateConfig` (a rejected request could leave live state partially mutated) and a validate-after-DB-write ordering issue in the client update handler.

## How to test

```sh
go test ./core/schemas/... ./transports/bifrost-http/handlers/... ./transports/bifrost-http/lib/...
cd ui && npm run typecheck
```

Added `tests/e2e/features/mcp-registry/mcp-registry-tool-sync-interval.spec.ts` (verified it reproduces the exact corrupted value from the bug report against pre-fix code, passes against the fix) and `core/schemas/duration_test.go` for the overflow-guard helper.

## Related issues

Closes #5026


Pratham-Mishra04:
Hey @Shaik-Sirajuddin this is already fixed in this pr - https://github.com/maximhq/bifrost/pull/4954

Shaik-Sirajuddin:
> Hey @Shaik-Sirajuddin this is already fixed in this pr - #4954

Missed to check that
Closing this as in wip here https://github.com/maximhq/bifrost/pull/4954

PR #5092, opened 2026-07-10
fix: support GA transcription-type sessions in POST /v1/realtime/client_secrets

Shaik-Sirajuddin:
## Summary

`POST /v1/realtime/client_secrets` mints both full realtime sessions and GA transcription-only sessions (`session.type: "transcription"`), but model extraction only checked `session.model` — the transcription schema has no such field, only `session.audio.input.transcription.model`. Every transcription-type request 400'd; confirmed live against real OpenAI.

## Changes

- `ExtractRealtimeClientSecretModel` falls back to the nested transcription model when `session.model` is absent.
- Added `schemas.IsGATranscriptionSessionBody`, a single classifier shared by extraction and both normalization layers (handler + provider), so they can't disagree on the same body. Explicit `session.type` wins when present; otherwise `session.model`/legacy root `model` beat the structural fallback, so a full session with transcription enabled as a sibling feature is never misclassified.
- Writes the model into the correct nested slot for transcription sessions instead of a bogus top-level `session.model`; the alias-resolution rewrite does the same.
- A first pass introduced its own classification bugs, caught by `codex --yolo` review and fixed in a follow-up commit.

## Type of change

- [x] Bug fix

## How to test

```sh
cd core && go test ./schemas/... ./providers/openai/... ./providers/azure/...
cd ../transports/bifrost-http && go test ./handlers/...
```

Also live-verified against real OpenAI's platform API using the `openai` Python SDK (and curl): minted a real GA transcription-session `client_secret`, and confirmed the regression scenario (full session + transcription sibling) keeps `type: "realtime"` and the correct model.

## Breaking changes

- [x] No

### DIFF
--- a/core/providers/openai/realtime.go
+++ b/core/providers/openai/realtime.go
@@ -320,14 +320,32 @@ func normalizeRealtimeClientSecretsRequest(
 	if marshalErr != nil {
 		return nil, "", newRealtimeClientSecretError(fasthttp.StatusInternalServerError, "server_error", "failed to encode normalized model", marshalErr)
 	}
-	session["model"] = modelJSON
-	StripNestedModelPrefixes(session)
-	if _, ok := session["type"]; !ok {
-		typeJSON, marshalErr := json.Marshal("realtime")
-		if marshalErr != nil {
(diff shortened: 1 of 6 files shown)


danpiths (approved):
looks good! thanks a lot for this PR :)

PR #5094, opened 2026-07-10
fix: round-trip anthropic redacted_thinking blocks on the responses surface

### LINKED ISSUE #5093: [Bug]: Streaming /v1/responses drops Anthropic redacted_thinking blocks, breaking multi-turn tool use with extended thinking
### Prerequisites

### Description

When Anthropic's safety systems flag part of Claude's reasoning, the API returns it as a `redacted_thinking` content block (an encrypted `data` payload) instead of a `thinking` block, and requires these blocks to be passed back unmodified in multi-turn tool use; their docs call out dropping them as the most common cause of the 400 error `` `thinking` or `redacted_thinking` blocks in the latest assistant message cannot be modified ``.

On the responses surface (`/v1/responses` with an Anthropic provider) the blocks are lost twice, in `core/providers/anthropic/responses.go`:

1. Streaming responses drop them entirely. The `content_block_start` handler in `ToBifrostResponsesStream` switches over the block type with cases for text, tool_use, MCP, compaction, and thinking, but has no case for `redacted_thinking`, so the block is silently discarded (`redacted_thinking` arrives complete in `content_block_start`; the SDKs model it as a start-event union member with no deltas). The non-streaming converter preserves the same block as a reasoning output item carrying `encrypted_content`.

2. Request conversion drops them again on replay. The reasoning items both converters produce for redacted thinking carry an empty, non-nil `summary` list next to `encrypted_content`, and `convertBifrostReasoningToAnthropicThinking` checks `Summary != nil` before falling back to `encrypted_content`. An empty list passes the nil check, the loop emits zero thinking blocks, and the encrypted payload is never converted back into a `redacted_thinking` block. So even the item the non-streaming path preserves is stripped from the outgoing request when the client echoes it back.

The result: a streaming client never receives the item (no `output_item.added` for it, only an orphan `output_item.done` carrying an empty assistant message shell, nothing in `response.completed.output`), the accumulated message that feeds logging, observability, and the semantic cache loses it too, and even a client that faithfully replays its history sends Anthropic an assistant message with the redacted block gone. When all of the turn's reasoning was redacted, the replayed message has no thinking block at all in a thinking-enabled tool-use conversation. The chat completions surface had the same class of bug and was fixed in #4943; the responses surface needs the equivalent treatment.

### Steps to reproduce

Deterministic, no API key: replay an Anthropic SSE sequence containing a `content_block_start` with a `redacted_thinking` block through `ToBifrostResponsesStream` and observe that no output item is produced for it, while the equivalent non-streaming message through `ToBifrostResponsesResponse` produces a reasoning item with `encrypted_content`. Then take that non-streaming reasoning item (empty `summary` list plus `encrypted_content`), put it in front of a `function_call` item in a `/v1/responses` input, convert with `ToAnthropicResponsesRequest`, and observe the outgoing Anthropic request contains the `tool_use` block but no `redacted_thinking` block.

Against the live API: send a streamed `/v1/responses` request to an Anthropic model with `reasoning` enabled and the documented test string that forces redaction (`ANTHROPIC_MAGIC_STRING_TRIGGER_REDACTED_THINKING_46C9A13E193C177646C7398A98432ECCCE4C1253D5E2D82641AC0E52CC2876CB`) as the input, and compare the stream (and the logged message) with the same request non-streamed.

### Expected behavior

The streamed response carries the redacted reasoning item like the non-streamed one does: an `output_item.added`/`output_item.done` pair for a reasoning item with `encrypted_content`, the item present in `response.completed.output` and in the accumulated/logged message. Replaying a reasoning item that carries `encrypted_content` (with or without an empty summary list) converts it back into the original `redacted_thinking` block, placed before the `tool_use` block as Anthropic requires.

### Actual behavior

The block contributes nothing to the converted stream: no `output_item.added`, nothing in `response.completed.output`, nothing in the accumulated/logged message. The output index was already allocated when the start event arrived, so the block's `content_block_stop` still goes through the generic done path and the stream emits an orphan `output_item.done` with no matching added, carrying an empty assistant message shell (`{"type": "message", "status": "completed", "role": "assistant", "content": []}`). Observed against the live API with the magic string: 9 done events for a single added on current dev. On the way back, replayed reasoning items with an empty summary list lose their `encrypted_content` and the outgoing request carries the `tool_use` block with no `redacted_thinking` before it. Multi-turn tool use with extended thinking breaks with an upstream 400 when redaction occurs and the model loses that reasoning context.

### Affected area(s)

Core (Go)

### Version

Current dev (reproduced at 3fbd9a4f7); the responses stream converter never handled `redacted_thinking`, and the request-side summary-list check predates it (the chat completions surface got the equivalent round-trip fix in #4943).

### Severity

fus3r:
## Summary

Fixes #5093. When Anthropic's safety systems flag part of Claude's reasoning, the API returns it as a `redacted_thinking` block carrying an encrypted `data` payload and requires the block to be replayed unmodified on later turns; their docs call out dropping these blocks as the most common cause of the 400 `` `thinking` or `redacted_thinking` blocks in the latest assistant message cannot be modified ``.

On the responses surface the blocks were lost twice. The inbound streaming converter dropped them: the `content_block_start` handler in `ToBifrostResponsesStream` has cases for text, tool_use, MCP, compaction, and thinking, but none for `redacted_thinking` (which arrives complete in the start event, with no deltas), so the block fell through to the default and was discarded. Streaming clients never received the item, and it was missing from `response.completed.output` and from the accumulated message that feeds logging, observability, and the semantic cache. The block's `content_block_stop` still went through the generic done path (the output index was allocated at the start event), so each dropped block also produced an orphan `output_item.done` with no matching added, carrying an empty assistant message shell. And the request converter dropped the payload again on replay: `convertBifrostReasoningToAnthropicThinking` checks `Summary != nil` before falling back to `encrypted_content`, while the reasoning items both response converters produce for redacted thinking carry an empty, non-nil summary list, so the encrypted branch was unreachable for exactly the items that need it and even the item the non-streaming path preserves was stripped from the outgoing request. The chat completions surface had the same class of bug, fixed in #4943; this is the responses equivalent.

## Changes

- Add a `redacted_thinking` case to the `content_block_start` handler that emits `output_item.added` with a reasoning item in the shape the non-streaming converter produces (`encrypted_content` carrying the payload), and stores the item so the block's `content_block_stop` emits the matching `output_item.done` and `response.completed` carries it in `output`, following the existing server-tool idiom in this handler.
- Fix the replay side: `convertBifrostReasoningToAnthropicThinking` now gates the summary loop on the list having entries instead of on nil, so reasoning items that carry an empty summary list next to a non-empty `encrypted_content` convert back into `redacted_thinking` blocks, placed before `tool_use` as Anthropic requires.
- Guard: blocks with no `data` contribute nothing at all, including their `content_block_stop` (no orphan done), mirroring the chat-surface guard from #4943.
- Regression tests, all six failing on `dev`: the full streamed lifecycle (added, done, and completed all carry the payload), the no-data guards including the stop, a mixed stream (redacted plus visible thinking plus text keep distinct output items and the payload does not bleed across), an egress round trip (the emitted item converts back to a `redacted_thinking` content block on the Anthropic-format egress), a full replay chain (stream ingress, a JSON round trip standing in for the client echo, then `ToAnthropicResponsesRequest` placing `redacted_thinking` before `tool_use`), and the same replay for the exact item shape the non-streaming converter produces.
- Changelog entry.

No framework change needed: the streaming accumulator already deep-copies `encrypted_content` on reasoning items, so the accumulated/logged message picks the block up once the converter emits it.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [x] Core (Go)
- [ ] Transports (HTTP)
- [x] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

```sh
cd core
go test ./providers/anthropic/ -run 'Redacted' -v
```

All six new tests fail on `dev` and pass with this change: the streamed lifecycle, mixed stream, and egress tests fail because the redacted block produces no output item at all, the two replay tests fail because the outgoing request carries the `tool_use` block with no `redacted_thinking` before it, and the no-data guard fails on the orphan `output_item.done` its stop used to synthesize. The full `./providers/anthropic/` suite is green, including under the race detector.

Also verified end to end against a real Anthropic backend (claude-sonnet-4-5, `reasoning` enabled, using the test string Anthropic documents to deterministically trigger redacted thinking, `ANTHROPIC_MAGIC_STRING_TRIGGER_REDACTED_THINKING_46C9A13E193C177646C7398A98432ECCCE4C1253D5E2D82641AC0E52CC2876CB`): on `dev` the streamed request delivers zero `encrypted_content` occurrences on the client SSE and zero in the logged accumulated message, while the identical non-streamed request keeps the redacted block; with this change the streamed run carries the redacted reasoning items on the wire (18 `encrypted_content` occurrences across the added, done, and completed events for 6 blocks) and the logged accumulated message keeps all 6, with non-streamed behavior unchanged.

## Screenshots/Recordings

N/A

## Breaking changes

- [ ] Yes
- [x] No

It only emits items that were previously dropped from the converted stream and replays encrypted payloads that were previously stripped from outgoing requests. Streams and requests without redacted thinking are unchanged; reasoning items with a populated summary list keep their existing thinking-block mapping.

## Related issues

Fixes #5093

## Security considerations

None. The payload is opaque to Bifrost and already flows through the non-streaming response path; this change gives the streaming and replay paths the same behavior.

## Checklist

### DIFF
--- a/core/changelog.md
+++ b/core/changelog.md
@@ -1,1 +1,2 @@
+- fix: round-trip Anthropic `redacted_thinking` blocks on the responses surface so multi-turn tool use with redacted reasoning can be replayed (closes #5093) [@fus3r](https://github.com/fus3r)
 - fix: reset `HasEmittedWebSearch` when recycling pooled Gemini responses stream state so grounded streaming requests keep emitting `web_search_call` items (closes #5113) [@fus3r](https://github.com/fus3r)
--- a/core/providers/anthropic/redactedthinkingresponses_test.go
+++ b/core/providers/anthropic/redactedthinkingresponses_test.go
@@ -193,6 +193,27 @@ func TestToBifrostResponsesStream_RedactedThinkingWithoutDataSkipped(t *testing.
 			if len(state.ContentIndexToBlockType) != 0 {
 				t.Errorf("expected block-type tracking to be cleaned up on stop, got %d entries", len(state.ContentIndexToBlockType))
 			}
+			if state.CurrentOutputIndex != 0 {
+				t.Errorf("expected no output index to be reserved for a data-less block, got CurrentOutputIndex=%d", state.CurrentOutputIndex)
+			}
+
+			// A following block must keep its position: the skipped block did not
+			// shift output indices.
+			textStart := &AnthropicStreamEvent{
+				Type:  AnthropicStreamEventTypeContentBlockStart,
+				Index: schemas.Ptr(1),
+				ContentBlock: &AnthropicContentBlock{
+					Type: AnthropicContentBlockTypeText,
+					Text: schemas.Ptr(""),
+				},
+			}
+			responses, err, _ = textStart.ToBifrostResponsesStream(context.Background(), 0, state)
+			if err != nil {
+				t.Fatalf("unexpected error on following text start: %v", err)
+			}
+			if len(responses) == 0 || responses[0].OutputIndex == nil || *responses[0].OutputIndex != 0 {
+				t.Errorf("expected the following text block to take output index 0, got %+v", responses)
+			}
 		})
 	}
 }
--- a/core/providers/anthropic/redactedthinkingresponses_test.go
+++ b/core/providers/anthropic/redactedthinkingresponses_test.go
@@ -1,0 +1,481 @@
+package anthropic
+
+import (
+	"context"
+	"encoding/json"
+	"testing"
+	"time"
+
+	"github.com/maximhq/bifrost/core/schemas"
+)
+
+// driveResponsesStream replays a sequence of Anthropic SSE events through
+// ToBifrostResponsesStream against a fresh stream state and returns every
+// Bifrost stream response emitted, in order.
+func driveResponsesStream(t *testing.T, events []*AnthropicStreamEvent) []*schemas.BifrostResponsesStreamResponse {
+	t.Helper()
+
+	state := AcquireAnthropicResponsesStreamState()
+	defer ReleaseAnthropicResponsesStreamState(state)
+
+	var out []*schemas.BifrostResponsesStreamResponse
+	seq := 0
+	for _, event := range events {
+		responses, err, _ := event.ToBifrostResponsesStream(context.Background(), seq, state)
+		if err != nil {
+			t.Fatalf("unexpected error converting %s: %v", event.Type, err)
+		}
+		out = append(out, responses...)
+		seq += len(responses)
+	}
+	return out
+}
+
+// redactedThinkingLifecycle returns the real Anthropic SSE shape of a response
+// whose only content block is a redacted_thinking block: the block arrives
+// complete in content_block_start (no deltas follow).
+func redactedThinkingLifecycle(data string) []*AnthropicStreamEvent {
+	stopReason := AnthropicStopReasonEndTurn
+	return []*AnthropicStreamEvent{
+		{
+			Type: AnthropicStreamEventTypeMessageStart,
+			Message: &AnthropicMessageResponse{
+				ID:    "msg_redacted_stream",
+				Model: "claude-sonnet-4-5-20250929",
+			},
+		},
+		{
+			Type:  AnthropicStreamEventTypeContentBlockStart,
+			Index: schemas.Ptr(0),
+			ContentBlock: &AnthropicContentBlock{
+				Type: AnthropicContentBlockTypeRedactedThinking,
+				Data: schemas.Ptr(data),
+			},
+		},
+		{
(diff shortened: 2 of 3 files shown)


R-droid101 (approved):
LGTM

Pratham-Mishra04:
@TejasGhatte can you review this

akshaydeo (dismissed):

akshaydeo (approved):

PR #5106, opened 2026-07-10
fix: never persist masked provider key previews

### LINKED ISSUE #4353: config.db corruption from masked-key preview in provider_configs JSON column (v1.5.10)
# Upstream Issue Draft — maximhq/bifrost
**Status**: DRAFT — operator review required before posting
**Parent NocoDB row**: 2297

---

## Title

config.db governance store enters "invalid character '*' looking for beginning of value" loop after provider PUT — recovers itself but creates `corrupt-*` artefacts and 2-minute outage

## Severity

**HIGH** — causes
- 500s on `/api/providers` GET (UI breaks)
- governance plugin re-init failures every ~2 minutes
- repeated `config.db.corrupt-*` snapshots (disk pressure: each ~26 MB)
- intermittent inability to authenticate VK-scoped requests during the window

Data is not permanently lost — Bifrost self-recovers by writing the in-memory snapshot back to a fresh `config.db` — but the recovery loop can pile up `corrupt-*` files until disk fills.

## Environment

- Bifrost: latest stable image as of 2026-06-11 (single-replica docker container)
- SQLite WAL mode, synchronous=FULL, wal_autocheckpoint=1000
- Storage: ext4 on QEMU virtual disk, no I/O errors in dmesg
- Container: RestartCount=0, OOMKilled=false, no memory cap
- Mount: `/srv/ashhl/bifrost/data` → `/app/data` (single bind, no overlay)

## Reproducer (observed, not isolated)

1. Bifrost running normally with 16 providers + 5 governance virtual keys (config.db ~26 MB).
2. Operator issues `PUT /api/providers/zai` to update provider config (~20:30 UTC).
3. PUT returns `500 {"error":{"message":"Failed to update provider: failed to update provider config in store: invalid character '*' looking for beginning of value; invalid character '*' looking for beginning of value; ..."}}`. The error repeats once per provider config row (~5 occurrences).
4. ~10 minutes later, Bifrost writes `config.db.corrupt-<unix-epoch>` + `config.db.pre-recover-<unix-epoch>` snapshots (binary identical to live db, integrity_check OK).
5. Governance plugin re-initialises and emits the same "invalid character '*'" payload one error per VK row (28 in our case = 5 VKs × 5 provider configs + 3 misc).
6. ~2 minutes of 500s on `/api/providers` and `/v1/embeddings` (governance auth fails).
7. Bifrost self-stabilises. Live `config.db` is clean afterwards (`sqlite3 -readonly ... 'PRAGMA integrity_check'` returns `ok`; no row has `value` starting with `*`).

## Root-cause hypothesis

The masked-key preview that GETs render (e.g. `/api/keys` returns `{"value": "sk-s************************EicA"}`) appears to be the same object passed through the governance-store reload path. When a config PUT lands and triggers a governance refresh, the in-memory representation briefly holds the masked string in the spot where the raw key belongs. Subsequent JSON-unmarshal of that struct fails because the leading `*` is not valid JSON whitespace/value.

The bug is **not** in SQLite — the on-disk row never actually receives the masked preview. It's in the **in-memory marshalling layer** that conflates "display" with "persist" representations of a key value.

Confirming evidence:
- `config_keys.value` column on disk: 17 rows, all clean (`sk-...`, `env.*`, hex hashes).
- `governance_virtual_keys.value`: 5 rows, all `sk-bf-...` clean.
- All recent backups (W71, W95-byteplus, W95-perplexity, W100): clean.
- Byte-offset grep of `logs.db` finds masked-key strings only inside upstream-OpenAI error payloads (`{"error":{"message":"Incorrect API key provided: sk-59f54****a997"}}`) which is provider-side masking and unrelated.
- `governance.db` (deprecated since 06-10 migration into config.db) is **0 bytes** on disk but appears in container `ls /app/data/`. It may still be touched on startup paths.

## Related upstream issues

- [#1336](https://github.com/maximhq/bifrost/issues/1336) — configstore migration  
- [#3510](https://github.com/maximhq/bifrost/issues/3510) — VK 401 errors  

Both bear the fingerprint of the same in-memory render/persist conflation.

## Workaround applied locally

- No DB writes were made (read-only investigation under credential moratorium).
- Staged VK quarantine recipe at `/opt/ashhl/staged/bifrost-vk-quarantine-20260611.md` — DELETE+RECREATE per VK with the ORIGINAL env value from master.env (no new credential needed). Will not execute unless the storm recurs.
- Optional preventive watcher (hourly SELECT-only on `config.db`) staged in the same doc.

## Asks of upstream

1. **Audit GET render paths**: ensure `key.MaskedValue()` (or equivalent) is never written back to disk or shared with the persistence layer.
2. **Add a server-side guard**: reject any PUT/POST whose `value` field matches `^\*+[A-Za-z0-9]{4,16}$` with HTTP 400 ("masked preview cannot be persisted").
3. **Stop writing 26 MB `corrupt-*` snapshots when the db is structurally OK**: integrity_check passes on all artefacts we captured, so the recovery heuristic is over-eager.
4. **Decide whether `governance.db` (0-byte) should be deleted on startup** post-migration, or removed from the container image entirely.

## Reproducer artefact bundle (available on request)

- Live db: `/srv/ashhl/bifrost/data/config.db` (sha256, schema dump, row counts)
- Corrupt snapshots: `config.db.corrupt-20260610-230851`, `config.db.corrupt-1781145015`, `config.db.pre-recover-1781145057`
- Container logs (last 24h, filtered to "invalid character" lines): 22 events
- This audit document: `/opt/ashhl/AUDIT/bifrost-upstream-issue-draft-20260611.md`

eyeveil:
## Summary

Prevents provider-key updates from persisting a redacted preview as the real credential. A stale or differently rendered mask could previously bypass the exact-match check, land in config storage, and trigger repeated JSON parse failures and governance reload outages.

## Changes

- Preserve the stored credential for every recognized redacted, non-secret update value, without requiring byte equality with the current server-rendered mask.
- Continue accepting genuine plaintext rotations and env/vault secret references.
- Add regression coverage for current masks, stale/different masks, plaintext rotations, env references, and empty-value validation behavior.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [ ] Core (Go)
- [x] Transports (HTTP)
- [ ] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

```sh
cd transports
go test ./bifrost-http/handlers -count=1
```

Expected: the handler package passes, including `TestMergeUpdatedKey_Value`.

## Screenshots/Recordings

Not applicable; no UI changes.

## Breaking changes

- [ ] Yes
- [x] No

## Related issues

Fixes #4353

## Security considerations

This closes a credential write-path integrity issue: display-only masked values can no longer replace stored provider credentials. Secret references and genuine plaintext rotations remain supported.

## Checklist

### DIFF
--- a/transports/bifrost-http/handlers/provider_keys.go
+++ b/transports/bifrost-http/handlers/provider_keys.go
@@ -504,7 +504,10 @@ func getKeyIDFromCtx(ctx *fasthttp.RequestCtx) (string, error) {
 	return decoded, nil
 }
 
-// validateProviderKeyURL checks that Ollama/SGL keys have a server URL configured.
+// validateProviderKeyURL checks that provider keys carry the nested fields
+// config.schema.json marks as required, so a create or merge can never persist
+// a key missing them (a masked update against a stored key lacking the section
+// would otherwise only surface later as a downstream 500).
 func validateProviderKeyURL(provider schemas.ModelProvider, key schemas.Key) error {
 	switch provider {
 	case schemas.Ollama:
@@ -514,6 +517,25 @@ func validateProviderKeyURL(provider schemas.ModelProvider, key schemas.Key) err
 	case schemas.SGL:
 		if key.SGLKeyConfig == nil || !key.SGLKeyConfig.URL.IsSet() {
 			return fmt.Errorf("sgl_key_config.url is required for SGL keys")
+		}
+	case schemas.Azure:
+		if key.AzureKeyConfig == nil || !key.AzureKeyConfig.Endpoint.IsSet() {
+			return fmt.Errorf("azure_key_config.endpoint is required for Azure keys")
+		}
+	case schemas.Bedrock:
+		if key.BedrockKeyConfig == nil || key.BedrockKeyConfig.Region == nil || !key.BedrockKeyConfig.Region.IsSet() {
+			return fmt.Errorf("bedrock_key_config.region is required for Bedrock keys")
+		}
+	case schemas.BedrockMantle:
+		if key.BedrockMantleKeyConfig == nil || key.BedrockMantleKeyConfig.Region == nil || !key.BedrockMantleKeyConfig.Region.IsSet() {
+			return fmt.Errorf("bedrock_mantle_key_config.region is required for Bedrock Mantle keys")
+		}
+	case schemas.VLLM:
+		if key.VLLMKeyConfig == nil || !key.VLLMKeyConfig.URL.IsSet() {
+			return fmt.Errorf("vllm_key_config.url is required for VLLM keys")
+		}
+		if key.VLLMKeyConfig.ModelName == "" {
+			return fmt.Errorf("vllm_key_config.model_name is required for VLLM keys")
 		}
 	}
 	return nil
--- a/transports/bifrost-http/handlers/provider_keys.go
+++ b/transports/bifrost-http/handlers/provider_keys.go
@@ -338,149 +338,139 @@ func (h *ProviderHandler) mergeUpdatedKey(oldRawKey, oldRedactedKey, updateKey s
 	}
 
 	if updateKey.AzureKeyConfig != nil && oldRedactedKey.AzureKeyConfig != nil && oldRawKey.AzureKeyConfig != nil {
-		if updateKey.AzureKeyConfig.Endpoint.IsRedacted() &&
-			updateKey.AzureKeyConfig.Endpoint.Equals(&oldRedactedKey.AzureKeyConfig.Endpoint) {
+		if updateKey.AzureKeyConfig.Endpoint.IsRedacted() && !updateKey.AzureKeyConfig.Endpoint.IsFromSecret() {
 			mergedKey.AzureKeyConfig.Endpoint = oldRawKey.AzureKeyConfig.Endpoint
 		}
 		if updateKey.AzureKeyConfig.ClientID != nil &&
 			oldRedactedKey.AzureKeyConfig.ClientID != nil &&
 			oldRawKey.AzureKeyConfig != nil &&
 			updateKey.AzureKeyConfig.ClientID.IsRedacted() &&
-			updateKey.AzureKeyConfig.ClientID.Equals(oldRedactedKey.AzureKeyConfig.ClientID) {
+			!updateKey.AzureKeyConfig.ClientID.IsFromSecret() {
 			mergedKey.AzureKeyConfig.ClientID = oldRawKey.AzureKeyConfig.ClientID
 		}
 		if updateKey.AzureKeyConfig.ClientSecret != nil &&
 			oldRedactedKey.AzureKeyConfig.ClientSecret != nil &&
 			oldRawKey.AzureKeyConfig != nil &&
 			updateKey.AzureKeyConfig.ClientSecret.IsRedacted() &&
-			updateKey.AzureKeyConfig.ClientSecret.Equals(oldRedactedKey.AzureKeyConfig.ClientSecret) {
+			!updateKey.AzureKeyConfig.ClientSecret.IsFromSecret() {
 			mergedKey.AzureKeyConfig.ClientSecret = oldRawKey.AzureKeyConfig.ClientSecret
 		}
 		if updateKey.AzureKeyConfig.TenantID != nil &&
 			oldRedactedKey.AzureKeyConfig.TenantID != nil &&
 			oldRawKey.AzureKeyConfig != nil &&
 			updateKey.AzureKeyConfig.TenantID.IsRedacted() &&
-			updateKey.AzureKeyConfig.TenantID.Equals(oldRedactedKey.AzureKeyConfig.TenantID) {
+			!updateKey.AzureKeyConfig.TenantID.IsFromSecret() {
 			mergedKey.AzureKeyConfig.TenantID = oldRawKey.AzureKeyConfig.TenantID
 		}
 	}
 
 	if updateKey.VertexKeyConfig != nil && oldRedactedKey.VertexKeyConfig != nil && oldRawKey.VertexKeyConfig != nil {
-		if updateKey.VertexKeyConfig.ProjectID.IsRedacted() &&
-			updateKey.VertexKeyConfig.ProjectID.Equals(&oldRedactedKey.VertexKeyConfig.ProjectID) {
+		if updateKey.VertexKeyConfig.ProjectID.IsRedacted() && !updateKey.VertexKeyConfig.ProjectID.IsFromSecret() {
 			mergedKey.VertexKeyConfig.ProjectID = oldRawKey.VertexKeyConfig.ProjectID
 		}
-		if updateKey.VertexKeyConfig.ProjectNumber.IsRedacted() &&
-			updateKey.VertexKeyConfig.ProjectNumber.Equals(&oldRedactedKey.VertexKeyConfig.ProjectNumber) {
+		if updateKey.VertexKeyConfig.ProjectNumber.IsRedacted() && !updateKey.VertexKeyConfig.ProjectNumber.IsFromSecret() {
 			mergedKey.VertexKeyConfig.ProjectNumber = oldRawKey.VertexKeyConfig.ProjectNumber
 		}
-		if updateKey.VertexKeyConfig.Region.IsRedacted() &&
-			updateKey.VertexKeyConfig.Region.Equals(&oldRedactedKey.VertexKeyConfig.Region) {
+		if updateKey.VertexKeyConfig.Region.IsRedacted() && !updateKey.VertexKeyConfig.Region.IsFromSecret() {
 			mergedKey.VertexKeyConfig.Region = oldRawKey.VertexKeyConfig.Region
 		}
-		if updateKey.VertexKeyConfig.AuthCredentials.IsRedacted() &&
-			updateKey.VertexKeyConfig.AuthCredentials.Equals(&oldRedactedKey.VertexKeyConfig.AuthCredentials) {
+		if updateKey.VertexKeyConfig.AuthCredentials.IsRedacted() && !updateKey.VertexKeyConfig.AuthCredentials.IsFromSecret() {
 			mergedKey.VertexKeyConfig.AuthCredentials = oldRawKey.VertexKeyConfig.AuthCredentials
 		}
 	}
 
 	if updateKey.BedrockKeyConfig != nil && oldRedactedKey.BedrockKeyConfig != nil && oldRawKey.BedrockKeyConfig != nil {
-		if updateKey.BedrockKeyConfig.AccessKey.IsRedacted() &&
-			updateKey.BedrockKeyConfig.AccessKey.Equals(&oldRedactedKey.BedrockKeyConfig.AccessKey) {
+		if updateKey.BedrockKeyConfig.AccessKey.IsRedacted() && !updateKey.BedrockKeyConfig.AccessKey.IsFromSecret() {
 			mergedKey.BedrockKeyConfig.AccessKey = oldRawKey.BedrockKeyConfig.AccessKey
 		}
-		if updateKey.BedrockKeyConfig.SecretKey.IsRedacted() &&
-			updateKey.BedrockKeyConfig.SecretKey.Equals(&oldRedactedKey.BedrockKeyConfig.SecretKey) {
+		if updateKey.BedrockKeyConfig.SecretKey.IsRedacted() && !updateKey.BedrockKeyConfig.SecretKey.IsFromSecret() {
 			mergedKey.BedrockKeyConfig.SecretKey = oldRawKey.BedrockKeyConfig.SecretKey
 		}
 		if updateKey.BedrockKeyConfig.SessionToken != nil &&
 			oldRedactedKey.BedrockKeyConfig.SessionToken != nil &&
 			oldRawKey.BedrockKeyConfig != nil &&
 			updateKey.BedrockKeyConfig.SessionToken.IsRedacted() &&
-			updateKey.BedrockKeyConfig.SessionToken.Equals(oldRedactedKey.BedrockKeyConfig.SessionToken) {
+			!updateKey.BedrockKeyConfig.SessionToken.IsFromSecret() {
 			mergedKey.BedrockKeyConfig.SessionToken = oldRawKey.BedrockKeyConfig.SessionToken
 		}
 		if updateKey.BedrockKeyConfig.Region != nil &&
 			oldRedactedKey.BedrockKeyConfig.Region != nil &&
 			oldRawKey.BedrockKeyConfig != nil &&
 			updateKey.BedrockKeyConfig.Region.IsRedacted() &&
-			updateKey.BedrockKeyConfig.Region.Equals(oldRedactedKey.BedrockKeyConfig.Region) {
+			!updateKey.BedrockKeyConfig.Region.IsFromSecret() {
 			mergedKey.BedrockKeyConfig.Region = oldRawKey.BedrockKeyConfig.Region
 		}
 		if updateKey.BedrockKeyConfig.ARN != nil &&
 			oldRedactedKey.BedrockKeyConfig.ARN != nil &&
 			oldRawKey.BedrockKeyConfig != nil &&
 			updateKey.BedrockKeyConfig.ARN.IsRedacted() &&
-			updateKey.BedrockKeyConfig.ARN.Equals(oldRedactedKey.BedrockKeyConfig.ARN) {
+			!updateKey.BedrockKeyConfig.ARN.IsFromSecret() {
 			mergedKey.BedrockKeyConfig.ARN = oldRawKey.BedrockKeyConfig.ARN
 		}
 		if updateKey.BedrockKeyConfig.RoleARN != nil &&
 			oldRedactedKey.BedrockKeyConfig.RoleARN != nil &&
 			oldRawKey.BedrockKeyConfig != nil &&
 			updateKey.BedrockKeyConfig.RoleARN.IsRedacted() &&
-			updateKey.BedrockKeyConfig.RoleARN.Equals(oldRedactedKey.BedrockKeyConfig.RoleARN) {
+			!updateKey.BedrockKeyConfig.RoleARN.IsFromSecret() {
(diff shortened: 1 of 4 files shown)


eyeveil:
Audited both findings against the current `SecretVar` semantics and provider key update flow.

The sibling-field finding was valid. I pushed `b1e0df315`, which applies the same redacted, non-secret preservation rule to Azure, Vertex, Bedrock, Bedrock Mantle, VLLM, Ollama, and SGL key config fields. I also added regression coverage for stale masks across those config families and verified that nested env references still apply normally.

On the plaintext-mask finding, I am keeping the fail-closed behavior. Once an incoming plaintext value matches the public redaction grammar, the payload carries no provenance that can distinguish a real credential from a stale UI preview. Accepting that value would reopen the write-path corruption this PR is intended to close. Genuine plaintext values outside the redaction grammar, plus env and vault references, continue to rotate normally.

Validation: `go test ./bifrost-http/handlers -count=1` passes.

transports/bifrost-http/handlers/provider_keys.go
```diff
@@ -325,154 +325,152 @@ func (h *ProviderHandler) deleteProviderKey(ctx *fasthttp.RequestCtx) {
...
+	// current redaction is deliberately NOT required: a stale or differently-masked
+	// preview (e.g. a different asterisk count, or a render from another replica)
+	// must be rejected too, otherwise the mask lands in the store and later breaks
+	// JSON re-parsing on reload ("invalid character '*'"). Genuine env/vault refs
+	// report IsRedacted() but are intentional, so IsFromSecret() lets them through.
+	if updateKey.Value.IsRedacted() && !updateKey.Value.IsFromSecret() {
 		mergedKey.Value = oldRawKey.Value
 	}
 
 	if updateKey.AzureKeyConfig != nil && oldRedactedKey.AzureKeyConfig != nil && oldRawKey.AzureKeyConfig != nil {
```
akshaydeo: **Finding 1 (High, blocking): masked values still persist when the old key lacks the section or field.**

Every nested block is gated on `oldRedactedKey.X != nil && oldRawKey.X != nil` (here, and at lines 367, 382), and optional pointer fields additionally require the old field to be non-nil (e.g. Bedrock SessionToken at 389-394). Since `mergedKey := updateKey`, when a gate fails the incoming masked value is persisted verbatim: `updateProviderKey` validates only Value-emptiness afterwards and `UpdateProviderKey` writes the key as-is.

Concrete trigger: `PUT .../keys/{id}` with `"bedrock_key_config": {"session_token": "ABCD************************WXYZ", ...}` while the stored key has no SessionToken (stale UI render from before the token was cleared, or a payload copied from another key). The mask is stored as the literal session token, reproducing the #4353 corruption class this PR closes for the other paths.

Suggested shape:
```go
// after the per-field preserve guards, for each nested masked non-secret
// value with no stored counterpart, fail loudly instead of persisting:
if updateKey.BedrockKeyConfig != nil && updateKey.BedrockKeyConfig.SessionToken != nil &&
    updateKey.BedrockKeyConfig.SessionToken.IsRedacted() && !updateKey.BedrockKeyConfig.SessionToken.IsFromSecret() &&
    (oldRawKey.BedrockKeyConfig == nil || oldRawKey.BedrockKeyConfig.SessionToken == nil) {
    return ..., errMaskedPreviewWithoutStoredValue // -> 400
}
```
(or centralize this in the shared helper suggested in the other comment, applied unconditionally to the update payload).

transports/bifrost-http/handlers/provider_keys.go
```diff
@@ -325,154 +325,152 @@ func (h *ProviderHandler) deleteProviderKey(ctx *fasthttp.RequestCtx) {
...
-	if updateKey.Value.IsRedacted() && updateKey.Value.Equals(&oldRedactedKey.Value) {
+	// Never persist a masked preview as the real key value. A redacted, non-secret
+	// incoming value can only be the GET-render placeholder echoed back by a client,
+	// so keep the stored credential instead of writing the mask. Matching the exact
+	// current redaction is deliberately NOT required: a stale or differently-masked
+	// preview (e.g. a different asterisk count, or a render from another replica)
+	// must be rejected too, otherwise the mask lands in the store and later breaks
+	// JSON re-parsing on reload ("invalid character '*'"). Genuine env/vault refs
+	// report IsRedacted() but are intentional, so IsFromSecret() lets them through.
+	if updateKey.Value.IsRedacted() && !updateKey.Value.IsFromSecret() {
```
akshaydeo: **Finding 5 (Medium): extract a shared predicate instead of 28 inline copies.**

`core/schemas/secretvar.go:455` already exports `ShouldPreserveStored()` with exactly this intent, used by `handlers/config.go`, `handlers/plugins.go`, and `lib/config.go`. It differs from this inline predicate only on empty values (it preserves; your empty-value subtest requires pass-through so validation fires), so reusing it verbatim would be wrong here - but that is an argument for a named sibling, not for 28 hand-inlined copies:

```go
// core/schemas/secretvar.go
// IsMaskedPlaceholder reports whether the value is a client-side redaction
// placeholder (not a secret ref) that must never overwrite a stored credential.
func (e *SecretVar) IsMaskedPlaceholder() bool {
    return e != nil && e.IsRedacted() && !e.IsFromSecret()
}
```

One tested helper also covers the ~15 individual guards the new test does not exercise, and gives the follow-up fixes in `providers.go:486-502` and `mcp.go:1086,1652` (which still carry the old `Equals`-gated bug) the same single source of truth. When the redaction shapes change next (e.g. a new sentinel), only one place needs updating.

Also a tradeoff worth a log line: any redacted-shaped intentional input (literal `[REDACTED]`, all-asterisk string) is now silently dropped with a 200; a warning log when the guard fires would make that observable.

transports/bifrost-http/handlers/provider_keys.go
```diff
@@ -325,154 +325,152 @@ func (h *ProviderHandler) deleteProviderKey(ctx *fasthttp.RequestCtx) {
 func (h *ProviderHandler) mergeUpdatedKey(oldRawKey, oldRedactedKey, updateKey schemas.Key) schemas.Key {
```
akshaydeo: **Finding 6 (Low): `oldRedactedKey` is now vestigial.**

After removing the `Equals` comparisons, all 22 remaining uses of `oldRedactedKey` in this function are nil checks that mirror `oldRawKey`'s structure 1:1 (`ProviderConfig.Redacted()` builds each sub-config and pointer field only when the raw one is non-nil). The parameter, the `GetProviderKeyRedacted` fetch in `updateProviderKey`, and the 10 tautological inner `oldRawKey.XKeyConfig != nil` re-checks inside already-guarded blocks can all be dropped. Beyond dead weight, checking `oldRedactedKey.X != nil` while reading `oldRawKey.X` couples correctness to the redaction staying structurally 1:1 with the raw key forever.

transports/bifrost-http/handlers/provider_keys_test.go
```diff
@@ -0,0 +1,203 @@
...
+			t.Fatalf("expected env ref applied, got ref=%q fromEnv=%v", merged.Value.GetRawRef(), merged.Value.IsFromEnv())
+		}
+		if merged.Value.GetValue() == rawValue {
+			t.Fatalf("stored raw value leaked into an env-ref update")
+		}
+	})
+
+	t.Run("empty value is not treated as redacted", func(t *testing.T) {
+		// Empty non-secret values must stay empty so the downstream
+		// "must not be empty" validation still fires — the merge must not
```
akshaydeo: **Finding 8 (Nit):** this comment contains a U+2014 em dash ("validation still fires — the merge"); house style is " - ", a colon, or a sentence break.

```go
// "must not be empty" validation still fires - the merge must not
```

akshaydeo (changes requested):
## Review

**Overview.** This PR replaces the byte-exact `Equals(currentRedaction)` guard in `mergeUpdatedKey` with `IsRedacted() && !IsFromSecret()` across all 27 credential fields, so any recognized mask preserves the stored credential instead of being persisted, while env/vault refs and plaintext rotations still apply. The direction is right and the common echo case is well tested, but the fix does not fully enforce its own stated invariant, and an almost-identical shared helper already exists in schemas.

### Findings

| 1 | High | `provider_keys.go:340` | Masked values still persist when the stored key lacks the config section (gates at 340, 367, 382) or the optional field is nil (e.g. Bedrock SessionToken, 389-394). `mergedKey := updateKey` shares pointers, the guard is skipped, and nothing downstream checks `IsRedacted` before persistence. Contradicts the new comment that stale previews "must be rejected too"; #4353 corruption remains reachable. | CONFIRMED |
| 2 | High | `providers.go:486` | Identical un-fixed bug one file over: NetworkConfig.CACertPEM and ProxyConfig URL/Username/Password/CACertPEM (486-502) still use `IsRedacted() && Equals(...)`; a stale or `<redacted>`-literal echo is persisted as the real proxy password / CA cert. | CONFIRMED |
| 3 | High | `mcp.go:1652` | Same un-fixed pattern in `mergeMCPHeaders` (1652) and TLSConfig.CACertPEM (1086): a stale masked Authorization header gets stored and then sent verbatim as the bearer token. | CONFIRMED |
| 4 | Medium | `provider_keys.go:101` | Create path (POST) has no mask guard: only non-emptiness is validated, so a GET-rendered mask POSTed as a new key (clone-by-copy workflow) is stored verbatim. Pre-existing, unchanged by this PR. | CONFIRMED |
| 5 | Medium | `provider_keys.go:336` | The predicate is hand-inlined 28 times, while `core/schemas/secretvar.go:455` already exports `ShouldPreserveStored()` (used by config.go, plugins.go, lib/config.go). It differs only on empty values, so extract a shared sibling helper (e.g. `IsMaskedPlaceholder()`) and use it everywhere, including the fixes for findings 2-4. | CONFIRMED |
| 6 | Low | `provider_keys.go:325` | `oldRedactedKey` is now vestigial: all 22 remaining uses are nil checks that mirror `oldRawKey`'s structure 1:1 (`Redacted()` preserves nil-ness). The parameter, its store fetch, and 10 tautological inner nil re-checks can be dropped. | CONFIRMED |
| 7 | Low | `provider_keys.go:336` | A redacted-shaped intentional input (literal `[REDACTED]`, all-asterisk string) is silently swallowed with HTTP 200 and the old credential stays active, with no signal to the caller. Accepted design tradeoff of the fix (the old path persisted the mask, which was worse); consider a warning log. | CONFIRMED (accepted tradeoff) |
| 8 | Nit | `provider_keys_test.go:94` | New comment contains a U+2014 em dash; house style uses " - ", a colon, or a sentence break instead. | CONFIRMED |

### Merge recommendation

**Not yet - request changes.** Finding 1 is a confirmed correctness gap inside the very function the PR rewrote: the "never persist a masked preview" invariant it documents and tests is bypassed whenever the stored key lacks the section or optional field, which is exactly the stale-echo scenario the PR exists to close. Fixing it pairs naturally with finding 5.

### Followups

1. **In this PR (blocking)**: enforce the invariant on nil-gated paths (`provider_keys.go:340-491`). When the incoming value is a masked non-secret and there is no stored raw counterpart, reject the update with 400 (or clear the field) instead of persisting the mask.
2. **In this PR**: extract the predicate as a `SecretVar` method next to `ShouldPreserveStored()` (`core/schemas/secretvar.go:455`), use it at all 28 sites, and test the helper once (this also covers the untested vault-ref branch of `IsFromSecret`).
3. **In this PR**: drop the vestigial `oldRedactedKey` parameter, its `GetProviderKeyRedacted` fetch in `updateProviderKey` (~line 201), and the tautological nil re-checks.
4. **In this PR**: replace the U+2014 in `provider_keys_test.go:94` with " - ".
5. **Follow-up PR**: apply the same helper to `providers.go:486-502` (proxy/CA cert merge) and `mcp.go:1086,1652` (MCP TLS/headers merge); both carry the identical live bug today.
6. **Follow-up PR**: reject `key.Value.IsRedacted()` plaintext in `createProviderKey` (`provider_keys.go:101`) so the invariant holds on create, and optionally add a warning log when an update drops a masked preview (finding 7).

### Refuted / cleared candidates

- Incomplete guard conversion: all 27 `Equals` guards were converted on this branch.
- Env/vault echo persisting a mask: `SecretVar.UnmarshalJSON` re-resolves refs at parse time, so a masked `Val` never survives for secret refs.
- Legitimate values accidentally matching `IsRedacted()`: regions, URLs, and ARNs cannot hit the asterisk patterns; only the literal `<redacted>`/`[REDACTED]` sentinels are human-reachable (finding 7).
- Test API mismatches: none; all SecretVar methods exist with the used signatures, and the empty-value and mismatched-mask fixtures behave as the test assumes.
- Pointer aliasing in the test (merge mutates `update`'s nested structs in place): real but currently harmless; noted in case future subtests re-assert on `update` after merging.

transports/bifrost-http/handlers/provider_keys.go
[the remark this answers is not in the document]
eyeveil: Keeping the fail-closed behavior here. The payload has no provenance that can distinguish a real redacted-shaped credential from a stale preview. Accepting it would reopen the corruption path. The update now fails with 400 when no stored counterpart exists, so this can no longer silently persist or clear a credential.

transports/bifrost-http/handlers/provider_keys.go
eyeveil: Fixed in `eebf54f87`. Masked placeholders now require a non-empty stored counterpart. Missing config sections, missing optional fields, and empty stored values return HTTP 400 before persistence. Regression tests cover all three cases.

transports/bifrost-http/handlers/provider_keys.go
eyeveil: Fixed in `eebf54f87`. Added `SecretVar.IsMaskedPlaceholder()` as the shared predicate and covered nil, empty, plaintext, mask, sentinel, env, and vault cases. The provider-key merge now uses it through one preservation helper.

transports/bifrost-http/handlers/provider_keys.go
eyeveil: Fixed in `eebf54f87`. Removed the update-path `GetProviderKeyRedacted` fetch, dropped the `oldRedactedKey` parameter, and removed the structural coupling to redacted config nil checks.

transports/bifrost-http/handlers/provider_keys_test.go
eyeveil: Fixed in `eebf54f87`. Replaced the character and ran formatting.

transports/bifrost-http/handlers/provider_keys.go
```diff
@@ -320,160 +314,143 @@ func (h *ProviderHandler) deleteProviderKey(ctx *fasthttp.RequestCtx) {
...
+		}
+	}
+
+	if mergedKey.VLLMKeyConfig != nil {
+		var stored *schemas.SecretVar
+		if oldRawKey.VLLMKeyConfig != nil {
+			stored = &oldRawKey.VLLMKeyConfig.URL
+		}
+		if err := preserve(&mergedKey.VLLMKeyConfig.URL, stored, "vllm_key_config.url"); err != nil {
+			return schemas.Key{}, err
```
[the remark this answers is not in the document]
eyeveil: Added — `validateProviderKeyURL` now enforces every schema-required nested field before persistence (Azure `endpoint`, Bedrock/Bedrock Mantle `region`, VLLM `url` + `model_name`), covering both create and the post-merge update path; table-driven tests for all cases. Note: the branch is based on `3fbd9a4f7` because current dev HEAD (`dbe8ab9d9`) doesn't compile (`core/schemas/responses.go:1132` references removed field `rawToolSearch`).

transports/bifrost-http/handlers/provider_keys.go
```diff
@@ -532,6 +518,25 @@ func validateProviderKeyURL(provider schemas.ModelProvider, key schemas.Key) err
...
+		if key.AzureKeyConfig == nil || !key.AzureKeyConfig.Endpoint.IsSet() {
+			return fmt.Errorf("azure_key_config.endpoint is required for Azure keys")
+		}
+	case schemas.Bedrock:
+		if key.BedrockKeyConfig == nil || key.BedrockKeyConfig.Region == nil || !key.BedrockKeyConfig.Region.IsSet() {
+			return fmt.Errorf("bedrock_key_config.region is required for Bedrock keys")
+		}
+	case schemas.BedrockMantle:
+		if key.BedrockMantleKeyConfig == nil || key.BedrockMantleKeyConfig.Region == nil || !key.BedrockMantleKeyConfig.Region.IsSet() {
+			return fmt.Errorf("bedrock_mantle_key_config.region is required for Bedrock Mantle keys")
```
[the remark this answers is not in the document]
eyeveil: Fixed — both call sites (create at 106, post-merge update at 228) now pass the resolved `baseProvider`, matching how the empty-value and alias checks already resolve it. Added `TestCreateProviderKey_CustomBedrockRequiresRegion`: a custom provider with `base_provider_type: bedrock` posting a key without a region now 400s at the handler.

transports/bifrost-http/handlers/provider_keys.go
```diff
@@ -320,160 +314,143 @@ func (h *ProviderHandler) deleteProviderKey(ctx *fasthttp.RequestCtx) {
...
-			mergedKey.VLLMKeyConfig.URL = oldRawKey.VLLMKeyConfig.URL
+	preserve := func(incoming, stored *schemas.SecretVar, field string) error {
+		if !incoming.IsMaskedPlaceholder() {
+			return nil
+		}
+		if stored == nil || !stored.IsSet() {
+			return fmt.Errorf("masked preview cannot be used for %s without a stored value", field)
+		}
+		*incoming = *stored
+		return nil
```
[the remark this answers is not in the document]
eyeveil: This is the accepted tradeoff from the maintainer review (finding 7 there): mask-shaped plaintext (`[REDACTED]`, all-asterisk, prefix+24-asterisk+suffix) is reserved as the display placeholder, because persisting an echoed mask corrupts the config store — the exact #4353 bug. Real credentials can't collide with those shapes (regions/URLs/ARNs/api keys never match), and rotating to an intentionally mask-shaped test value isn't a supported flow. Please remember this so it doesn't re-flag on every push.

BearTS (approved):
LGTM

akshaydeo (approved):

PR #5107, opened 2026-07-10
fix: repair bare wildcard model-list rows

### LINKED ISSUE #4318: [Bug]: governance_virtual_key_provider_configs.allowed_models persisted as bare "*" string instead of JSON array, blocks all subsequent PUT /api/providers updates with "invalid character '*' in literal"
### Affected area(s)

- Framework
- Transports (HTTP)

### Version

v1.5.10

### Environment

- OS: Linux x86_64 (Ubuntu 22.04 LTS)
- Deployment: docker compose, single node
- Storage: SQLite (`config.db`), WAL mode, default GORM driver
- Bifrost components: core/transports/ui/framework all v1.5.10
- Go: 1.22.x (vendor in image)
- No replicas, no Postgres config_store

### Description

After creating five virtual-key provider-config rows for a single provider (`zai`) via the admin UI and a follow-up `POST /api/virtual-keys` call, the rows in `governance_virtual_key_provider_configs` store `allowed_models` as the literal one-character string `*` (no JSON quotes, no array brackets) instead of the documented `["*"]` form.

Once any such row exists, every subsequent administrative PUT against the same provider (or the VK key plumbing under it) fails at request decode time with:

```
invalid character '*' in literal true (expecting 'r')
```

The error is surfaced on responses to both:

- `PUT /api/providers/{name}` (changing provider top-level config)
- `PUT /api/providers/{name}/keys/{id}` (rotating or editing a key under that provider)

This means a single corrupted row poisons the provider's entire admin surface; the admin REST API can no longer fix what its own write path produced.

### Steps to reproduce

1. Run Bifrost v1.5.10 with SQLite config store (default docker-compose, no Postgres).
2. Create a custom OpenAI-compatible provider (e.g. `zai`) with one key via the admin UI.
3. Create a virtual key (POST `/api/virtual-keys`) with a single `provider_configs` entry:
   ```json
   {
     "provider": "zai",
     "weight": 1.0,
     "allowed_models": ["*"],
     "key_ids": ["*"]
   }
   ```
4. Inspect the row:
   ```bash
   sqlite3 /path/to/config.db \
     "SELECT id, provider, allowed_models FROM governance_virtual_key_provider_configs WHERE provider='zai';"
   ```
   Observed: `allowed_models` column contains `*` (one byte), NOT `["*"]` (five bytes).
5. Attempt any subsequent admin PUT against the provider, e.g.:
   ```bash
   curl -X PUT http://127.0.0.1:11010/api/providers/zai \
     -H 'content-type: application/json' \
     -d '{"network_config":{"default_request_timeout_in_seconds":120}}'
   ```
   Response: HTTP 400 with body containing `invalid character '*' in literal true (expecting 'r')`.

### Expected behavior

`allowed_models` is always persisted as a JSON-encoded array (`["*"]` for unrestricted, `["model-a","model-b"]` for restricted, `[]` for deny-all per `docs/changelogs/ent-v1.4.0-prerelease1.mdx` semantics). The GORM tag `gorm:"type:text;serializer:json"` on `framework/configstore/tables/virtualkey.go:TableVirtualKeyProviderConfig.AllowedModels` should guarantee this round-trip.

### Actual behavior

At least one write path stores `allowed_models` as the bare character `*`. When the row is later loaded for an admin-API request body, the JSON deserializer hits the bare `*` and returns `invalid character '*' in literal true`, which propagates as the HTTP 400 above.

### Workaround

Direct SQLite repair (single transaction is safe — column is plain TEXT with no FK):

```sql
UPDATE governance_virtual_key_provider_configs
SET    allowed_models = '["*"]'
WHERE  allowed_models = '*';
```

After this, all admin PUTs against the affected provider succeed and the routing logic still treats the VK as unrestricted (matches the documented `IsUnrestricted()` check in `core/schemas/account.go`).

### Suggested fix

Three layers to investigate, in order of likely impact:

1. **Storage write path** — confirm `gorm:"type:text;serializer:json"` is actually being honored for `TableVirtualKeyProviderConfig.AllowedModels`. The bug pattern (raw Go `string` written instead of JSON-encoded `[]string`) suggests a code path that bypasses the GORM serializer — possibly a hand-rolled `db.Exec("UPDATE ... SET allowed_models=?", "*")` or a `BeforeSave` hook that mutates the field to a raw string when it sees the unrestricted case.
2. **Validation hook** — `BeforeSave` on the row (already present per the grep showing `// BeforeSave validates WhiteList and BlackList fields`) should reject any non-JSON-array form before it reaches the DB. Currently it appears to permit (or produce) the bare string.
3. **DB-level guard** — add a SQLite `CHECK (json_valid(allowed_models))` constraint on the column so the next regression is caught at the storage boundary rather than the deserializer boundary.

### Related

- Closed issue [#2717](https://github.com/maximhq/bifrost/issues/2717) — symptomatically related (also about `allowed_models` wildcard semantics on v1.4.x); cited as similar-but-distinct because the v1.4.x report was about `[]` being treated as "deny all" through `config.json`, whereas this report is about a v1.5.10 OSS SQLite write producing a bare `*` value via the admin API.
- Merged PR [#4221](https://github.com/maximhq/bifrost/pull/4221) — `publishRoutingAllowlist` extraction; touches the same `allowed_models` path on the publish side.
- Sister issue (config_keys.value serialization) — will be filed separately and cross-linked.

### Logs/output

```
{"level":"error","ts":"2026-06-11T...","caller":"handlers/providers.go:...",
 "msg":"failed to decode provider update","err":"invalid character '*' in literal true (expecting 'r')"}
```

### Severity

High — corrupts admin surface and is not self-recoverable without DB access.

### Regression?

Cannot confirm — first instance observed at v1.5.10.

eyeveil:
## Summary

Repairs legacy virtual-key provider-config rows that contain a bare `*` in JSON-serialized model-list columns. Those rows cannot be deserialized by GORM and can block the provider administration surface until the database is repaired manually.

## Changes

- Add an idempotent data-repair migration converting bare `*` values to canonical `["*"]` JSON in `allowed_models` and `blacklisted_models`.
- Leave valid JSON and all other values untouched.
- Add regression coverage for the current serializer round trip and for loading both repaired legacy corruption variants.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [ ] Core (Go)
- [ ] Transports (HTTP)
- [ ] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [ ] Docs

Framework/configstore is affected.

## How to test

```sh
cd framework
go test ./configstore -count=1
```

Expected: the configstore package passes, including the wildcard round-trip and repair migration tests.

## Screenshots/Recordings

Not applicable; no UI changes.

## Breaking changes

- [ ] Yes
- [x] No

## Related issues

Fixes #4318

## Security considerations

No credential material is read or changed. The migration only canonicalizes two malformed model-list values so affected configuration rows can be loaded again.

## Checklist

### DIFF
--- a/framework/configstore/migrations.go
+++ b/framework/configstore/migrations.go
@@ -441,6 +441,7 @@ var configstoreMigrationSteps = []migrationStep{
 	{IDs: []string{"add_sidekiq_kind_status_created_index"}, run: migrationAddSidekiqKindStatusCreatedIndex},
 	{IDs: []string{"add_fast_mode_cache_pricing_columns"}, run: migrationAddFastModeCachePricingColumns},
 	{IDs: []string{"add_inference_geo_multiplier_column"}, run: migrationAddInferenceGeoMultiplierColumn},
+	{IDs: []string{"repair_bare_wildcard_allowed_models"}, run: migrationRepairBareWildcardAllowedModels},
 }
 
 // quoteSQLiteIdentifier quotes a SQLite identifier, escaping any double quotes.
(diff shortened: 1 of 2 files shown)


akshaydeo:
@eyeveil we never allowed blank * in that field - how did you land up in this scenario ?

akshaydeo (approved):

PR #5151, opened 2026-07-13
[fix]: OpenAI Responses - preserve JSON schema key order in structured outputs

### LINKED ISSUE #5148: [Bug]: Responses API alphabetizes structured-output JSON schema keys, breaking OpenAI's documented key-ordering contract
### Prerequisites

### Description

Bifrost does not forward OpenAI `/v1/responses` structured-output schemas transparently: every nested object inside `text.format.schema` (`properties`, `$defs`, `items`, `anyOf` members) reaches the provider with **alphabetized keys**.

Two mechanisms in core combine to cause this:

1. `ResponsesTextConfigFormatJSONSchema` stores its schema-bearing fields as `map[string]any` (`Schema`, `Properties`, `Defs`, `Definitions`, `Items`, `AnyOf`/`OneOf`/`AllOf`), so the client's key order is destroyed at decode time (`core/schemas/responses.go`).
2. The outbound request is serialized with `MarshalSorted` (`sonic.Config{SortMapKeys: true}`), which emits plain-map values alphabetized (`core/providers/utils/utils.go`).

For most JSON APIs, key order is semantically irrelevant, and sorting for deterministic serialization (the prompt-caching rationale in the `MarshalSorted` comment) would be harmless. For OpenAI Structured Outputs it is not harmless — OpenAI documents key order as observable output behavior ([Structured Outputs — Key ordering](https://developers.openai.com/api/docs/guides/structured-outputs#key-ordering)):

> When using Structured Outputs, outputs will be produced in the same order as the ordering of keys in the schema.

Structured Outputs use constrained decoding, so schema key order changes which token sequences are valid at each position. Alphabetizing a union member such as `TextPart {"type": ..., "text": ...}` into `{"text": ..., "type": ...}` moves the discriminator behind the payload field. When the model opens a union object with the natural `{"type":` token, the reordered branch is grammatically eliminated and generation is locked into a sibling branch — producing schema-valid but semantically wrong output (wrong `anyOf` branch, fabricated content, runaway repetition until `max_output_tokens`, or completed responses missing required properties).

Since `anyOf` is the only union mechanism OpenAI strict mode supports, schema key order is the *only* branch-steering signal available to the provider (OpenAI), and Bifrost currently destroys it.

Notably, the chat path already handles this correctly: `ToolFunctionParameters` uses `OrderedMap` for its schema fields. The Responses `text.format.schema` path was left behind.

### Steps to reproduce

The schema below is a MWE for a structured output of the form `list[TextPart | WebCitation]` (i.e. a list, where the items are a discriminated union). In the request, the properties are in declaration order, each union member starting with its `type` discriminator. Run it against the official OpenAI API and the model does exactly what the prompt says; run the identical request through a Bifrost deployment (OpenAI provider configured) and the model emits fabricated citations instead. In our tests: 5/5 correct via OpenAI directly, 0/5 correct via Bifrost.

```bash
#!/usr/bin/env bash
# Direct (OPENAI_BASE_URL=https://api.openai.com/v1):  {"parts":[{"type":"text","text":"hello world"}]}
# Via Bifrost (OPENAI_BASE_URL=<bifrost>/v1):          {"parts":[{"type":"cite:web","url":"https://example.com"}]}

curl -s "${OPENAI_BASE_URL:-https://api.openai.com/v1}/responses" \
  -H "Authorization: Bearer $OPENAI_API_KEY" \
  -d @- <<'JSON'
{
  "model": "gpt-5.2",
  "input": "Respond with a single text part whose content is \"hello world\". Do not emit any citation parts.",
  "reasoning": {"effort": "low"},
  "max_output_tokens": 512,
  "text": {"format": {"type": "json_schema", "name": "reply", "strict": true, "schema": {
    "$defs": {
      "TextPart": {
        "additionalProperties": false,
        "properties": {
          "type": {"const": "text", "title": "Type", "type": "string"},
          "text": {"title": "Text", "type": "string"}
        },
        "required": ["type", "text"],
        "title": "TextPart",
        "type": "object"
      },
      "WebCitation": {
        "additionalProperties": false,
        "properties": {
          "type": {"const": "cite:web", "title": "Type", "type": "string"},
          "url": {"title": "Url", "type": "string"}
        },
        "required": ["type", "url"],
        "title": "WebCitation",
        "type": "object"
      }
    },
    "additionalProperties": false,
    "properties": {
      "parts": {
        "items": {"anyOf": [{"$ref": "#/$defs/TextPart"}, {"$ref": "#/$defs/WebCitation"}]},
        "title": "Parts",
        "type": "array"
      }
    },
    "required": ["parts"],
    "title": "Reply",
    "type": "object"
  }}}
}
JSON
```

To make the mutation explicit, here is the schema as the client sends it to Bifrost, and what Bifrost actually forwards to OpenAI (captured from the outbound request serializer on `main`).

Schema sent to Bifrost (pydantic declaration order — `TextPart` starts with its `type` discriminator):

```json
{
  "$defs": {
    "TextPart": {
      "additionalProperties": false,
      "properties": {
        "type": {"const": "text", "title": "Type", "type": "string"},
        "text": {"title": "Text", "type": "string"}
      },
      "required": ["type", "text"],
      "title": "TextPart",
      "type": "object"
    },
    "WebCitation": {
      "additionalProperties": false,
      "properties": {
        "type": {"const": "cite:web", "title": "Type", "type": "string"},
        "url": {"title": "Url", "type": "string"}
      },
      "required": ["type", "url"],
      "title": "WebCitation",
      "type": "object"
    }
  },
  "additionalProperties": false,
  "properties": {
    "parts": {
      "items": {"anyOf": [{"$ref": "#/$defs/TextPart"}, {"$ref": "#/$defs/WebCitation"}]},
      "title": "Parts",
      "type": "array"
    }
  },
  "required": ["parts"],
  "title": "Reply",
  "type": "object"
}
```

Schema Bifrost forwards to OpenAI — note `TextPart.properties` is now `{"text", "type"}` while `WebCitation.properties` remains `{"type", "url"}` (already alphabetical), so only the citation branch can still begin with `"type"`:

```json
{
  "additionalProperties": false,
  "properties": {
    "parts": {
      "items": {"anyOf": [{"$ref": "#/$defs/TextPart"}, {"$ref": "#/$defs/WebCitation"}]},
      "title": "Parts",
      "type": "array"
    }
  },
  "required": ["parts"],
  "type": "object",
  "$defs": {
    "TextPart": {
      "additionalProperties": false,
      "properties": {
        "text": {"title": "Text", "type": "string"},
        "type": {"const": "text", "title": "Type", "type": "string"}
      },
      "required": ["type", "text"],
      "title": "TextPart",
      "type": "object"
    },
    "WebCitation": {
      "additionalProperties": false,
      "properties": {
        "type": {"const": "cite:web", "title": "Type", "type": "string"},
        "url": {"title": "Url", "type": "string"}
      },
      "required": ["type", "url"],
      "title": "WebCitation",
      "type": "object"
    }
  },
  "title": "Reply"
}
```

### Expected behavior

the client's `text.format.schema` reaches the provider with its key order intact.

### Actual behavior

every nested schema object is alphabetized, changing what the model generates.

### Affected area(s)

Core (Go)

### Version

main (`59096c8e8`); also reproduced on v1.5.x

### Environment

```text
- OS: macOS 15 (arm64) and Linux (arm64 container)
- Go: 1.26.x
- Components: core, transports (bifrost-http), OpenAI provider
```

### Relevant logs/output

```shell
# direct OpenAI, 5/5:
{"parts":[{"type":"text","text":"hello world"}]}

# via Bifrost, 0/5 — fabricated citations despite the prompt forbidding them,
# sometimes repeating until max_output_tokens ("status": "incomplete"):
{"parts":[{"type":"cite:web","url":"https://example.com"},{"type":"cite:web","url":"https://example.org"}]}
```

### Regression?

Not a recent regression as far as we can tell; the Responses schema fields have used plain maps since their introduction.

### Severity

georg-wolflein:
## Summary

OpenAI Structured Outputs are order-sensitive by documented contract ([Structured Outputs — Key ordering](https://developers.openai.com/api/docs/guides/structured-outputs#key-ordering)):

> When using Structured Outputs, outputs will be produced in the same order as the ordering of keys in the schema.

Bifrost currently alphabetizes every nested object inside `text.format.schema` on the Responses path: the schema-bearing fields of `ResponsesTextConfigFormatJSONSchema` are typed `map[string]any`, so the client's key order is destroyed at decode time, and the sorted provider marshal (`MarshalSorted`, `SortMapKeys: true`) emits the nested objects alphabetized.

Structured Outputs use constrained decoding, so schema key order changes which token sequences are valid at each position. For discriminated `anyOf` unions this flips branch selection: alphabetizing `TextPart {"type", "text"}` into `{"text", "type"}` means that branch can no longer *begin* with `"type"`, so the natural `{"type":` opener locks generation into a sibling branch — producing schema-valid but semantically wrong output (fabricated content, runaway repetition until `max_output_tokens`). Since `anyOf` is the only union mechanism OpenAI strict mode supports (`oneOf` is rejected with `invalid_json_schema`), key order is the only branch-steering signal clients have. See #5148 for the full analysis.

This PR migrates the schema-bearing fields to `OrderedMap`, exactly mirroring `ToolFunctionParameters` (which chat and Responses function tools already use). `OrderedMap` round-trips JSON in insertion order and its `MarshalJSON` bypasses `SortMapKeys`, so deterministic serialization of the surrounding request (prompt-cache stability) is unchanged while the schema stays byte-faithful.

### Reproduction

The schema is a `parts: list[TextPart | WebCitation]` union in pydantic declaration order, each member starting with its `type` discriminator; the prompt explicitly asks for a single text part (from #5148):

```bash
#!/usr/bin/env bash
# OPENAI_BASE_URL=https://api.openai.com/v1  -> direct OpenAI
# OPENAI_BASE_URL=<bifrost>/v1               -> via Bifrost

curl -s "${OPENAI_BASE_URL:-https://api.openai.com/v1}/responses" \
  -H "Authorization: Bearer $OPENAI_API_KEY" \
  -d @- <<'JSON'
{
  "model": "gpt-5.2",
  "input": "Respond with a single text part whose content is \"hello world\". Do not emit any citation parts.",
  "reasoning": {"effort": "low"},
  "max_output_tokens": 512,
  "text": {"format": {"type": "json_schema", "name": "reply", "strict": true, "schema": {
    "$defs": {
      "TextPart": {
        "additionalProperties": false,
        "properties": {
          "type": {"const": "text", "title": "Type", "type": "string"},
          "text": {"title": "Text", "type": "string"}
        },
        "required": ["type", "text"],
        "title": "TextPart",
        "type": "object"
      },
      "WebCitation": {
        "additionalProperties": false,
        "properties": {
          "type": {"const": "cite:web", "title": "Type", "type": "string"},
          "url": {"title": "Url", "type": "string"}
        },
        "required": ["type", "url"],
        "title": "WebCitation",
        "type": "object"
      }
    },
    "additionalProperties": false,
    "properties": {
      "parts": {
        "items": {"anyOf": [{"$ref": "#/$defs/TextPart"}, {"$ref": "#/$defs/WebCitation"}]},
        "title": "Parts",
        "type": "array"
      }
    },
    "required": ["parts"],
    "title": "Reply",
    "type": "object"
  }}}
}
JSON
```

Results (5 runs each):

| Route | Result |
|---|---|
| OpenAI directly | 5/5 correct: `{"parts":[{"type":"text","text":"hello world"}]}` |
| Bifrost **before** this PR | 0/5 — fabricated citations instead, e.g. `{"parts":[{"type":"cite:web","url":"https://example.com"}]}`, sometimes repeating until `max_output_tokens` (`"status": "incomplete"`) |
| Bifrost **after** this PR | 5/5 correct: `{"parts":[{"type":"text","text":"hello world"}]}` |

## Changes

- `core/schemas/responses.go`: `ResponsesTextConfigFormatJSONSchema` schema-bearing fields (`Schema`, `Properties`, `$defs`, `definitions`, `items`, `anyOf`/`oneOf`/`allOf`) migrated from `map[string]any` to `OrderedMap`, mirroring `ToolFunctionParameters`; `JSONSchemaFromMap`/`ToMap` adapted via the existing `SafeExtractOrderedMap` idiom.
- `core/providers/gemini/`: schema normalization (`normalizeSchemaForGemini`) is now OrderedMap-aware and order-preserving — `propertyOrdering` now reflects the client's true declaration order (previously it captured randomized Go map iteration order); `extractSchemaMapFromResponseFormat` no longer drops OrderedMap schemas on the Responses→Chat fallback path.
- `core/providers/anthropic/`: schema extraction adapted via `SafeExtractOrderedMap`; the Responses→Chat fallback path (`convertChatResponseFormatToTool`, GA `output_config`) now handles OrderedMap schemas with order-preserving normalization instead of silently dropping them or skipping nested normalization.
- `core/providers/cohere/`: `response_format` schema extraction handles OrderedMap schemas.
- `plugins/compat/requestcopy.go`: request clone deep-copies `OrderedMap` schema fields.
- Regression tests:
  - `TestPayloadOrdering_ResponsesTextFormatJSONSchema` (golden): a decoded Responses request must re-serialize with all nested schema objects in the client's original, non-alphabetical key order, plus a 100-iteration determinism check. Fails on current `dev` without this change.
  - `TestConvertChatResponseFormatToTool_OrderedMapSchema`: the Responses→Chat fallback must preserve property order and still normalize nested multi-type unions to `anyOf`.
- Changelog entries in `core/changelog.md` and `plugins/compat/changelog.md`.

Design note: `OrderedMap` was chosen over raw JSON passthrough because it keeps the schema introspectable for cross-provider conversion (gemini/anthropic/cohere) and reuses the established idiom from `ToolFunctionParameters`. Deterministic serialization for prompt caching is preserved — `OrderedMap.MarshalJSON` output is deterministic for identical client input, and plain-map paths still sort.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [x] Core (Go)
- [ ] Transports (HTTP)
- [x] Providers/Integrations
- [x] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

Run the reproduction script above against a Bifrost build of this branch with an OpenAI provider configured — the model returns the requested text part instead of fabricated citations (5/5 in our runs), matching direct-to-OpenAI behavior.

## Screenshots/Recordings

N/A (no UI changes).

## Breaking changes

- [ ] Yes
- [x] No

Wire format is unchanged (except that schema key order is now preserved, which is the fix). For Go library consumers constructing `ResponsesTextConfigFormatJSONSchema` directly, the field types change from `map[string]any` to `OrderedMap` — the same source-visible migration `ToolFunctionParameters` already went through.

## Related issues

Closes #5148

## Security considerations

None — no changes to auth, secrets, or request routing; the request body is forwarded more faithfully than before.

## Checklist


akshaydeo (commented):
Went through this in depth - traced every migrated field and helper across core, the providers, and the compat plugin, built all modules at the PR head, and ran the provider unit tests (the two openai failures reproduce identically on dev, unrelated to this change).

Overall a solid fix for a real bug, and the migration is thorough: the compat plugin covers all eight migrated fields, marshaling stays on MarshalSorted everywhere, and the golden ordering test is a nice touch.

One actual behavior change introduced here that needs a decision: `Schema` going from `*any` to `*OrderedMap` means a non-object value at the nested `text.format.schema.schema` position (the wrapped `{name, strict, schema: true}` shape) used to decode and get forwarded on dev, and now fails the whole request with `orderedmap: expected '{'`. The top-level `text.format.schema: true` case already failed before this PR, so this only bites the wrapped shape with a boolean/scalar schema. I'm fine accepting the narrowing, but then it should be called out in the breaking changes section.

Small nits, ideally in this PR (inline comments on each):
- `buildJSONSchemaFromMap` in gemini/utils.go still only accepts anyOf/oneOf/allOf as `[]interface{}` - its twin `schemas.JSONSchemaFromMap` got a `case []OrderedMap` in this PR, this one should match.
- the `*OrderedMap` arm in `extractSchemaMapFromResponseFormat` should go through `SafeExtractOrderedMap` (or a nil guard) so a typed-nil pointer can't pass the caller's `!= nil` check and emit `"responseJsonSchema": null`.
- leftover comment stubs `// Recursively normalize oneOf` / `// Recursively normalize allOf` with empty bodies in anthropic/utils.go after the loop consolidation.

Pre-existing gaps this PR makes more visible (not caused here, fine as follow-ups):
- gemini `reconstructSchemaFromJSONSchema`'s spread-fields branch duplicates `ToMap()` but emits only 8 of its ~24 fields - items, enum, anyOf/oneOf/allOf, $ref, definitions, format, pattern, min/max get silently dropped from schemas sent to Gemini. Replacing it with `ToMap()` plus the two gemini extras (title from Name, propertyOrdering) would fix that, with its own tests since it changes behavior.
- anthropic's `convertResponsesTextConfigToAnthropicOutputFormat` rebuilds the schema from only 6 fields and never reads `JSONSchema.Schema`, while the reverse converter (updated here) extracts everything - the round trip is asymmetric.
- compat's `cloneResponsesTextConfig` still shares the `AdditionalProperties` pointer between clone and original (latent - nothing mutates it in place today).
- the OrderedMap-aware normalize walker is now ~90 lines duplicated between anthropic and gemini (identical except the leaf call), and compat's `cloneOrderedMap` re-implements the unexported `schemas.deepCopyOrderedMap` - worth extracting/exporting once so the next provider doesn't grow a third copy.

Things I chased and cleared, for the record: the `[]interface{}` assertions in `convertMapToToolFunctionParameters` look scary next to `ToMap()` now emitting typed slices, but that output never reaches the map-based converter (Vertex/BedrockMantle Responses requests go through the typed `convertJSONSchemaToToolFunctionParameters`, which handles `[]OrderedMap` and `[]string` correctly, and JSON-decoded chat traffic keeps `[]interface{}` shapes - the residual exposure is identical on dev). Likewise `normalizeSchemaTypes` skipping nested OrderedMaps is unreachable: OrderedMap schemas only exist on the egress direction and never carry uppercase Gemini types.

Happy to merge once the nits are addressed.

core/schemas/responses.go
```diff
@@ -517,30 +517,34 @@ type ResponsesTextConfigFormat struct {
...
 // ResponsesTextConfigFormatJSONSchema represents a JSON schema specification
 // It supports JSON Schema fields used by various providers for structured outputs.
+// Schema-bearing fields use OrderedMap (mirroring ToolFunctionParameters) because
+// structured-output generation is sensitive to JSON schema property order: providers
+// like OpenAI follow the literal key order of the schema, so decoding into plain Go
+// maps (and re-marshaling them sorted) degrades output quality.
 type ResponsesTextConfigFormatJSONSchema struct {
 	Name                 *string                     `json:"name,omitempty"`
-	Schema               *any                        `json:"schema,omitempty"`
+	Schema               *OrderedMap                 `json:"schema,omitempty"`
```
akshaydeo, line 526: This narrows the wire contract: a non-object value at the nested `text.format.schema.schema` position (wrapped `{name, strict, schema: true}` shape) decoded and was forwarded on dev, and now fails the whole request decode with `orderedmap: expected '{'`. The top-level `"schema": true` case already failed before, so the blast radius is just the wrapped shape with a boolean/scalar schema - fine to accept, but then let's note it under breaking changes.

core/providers/gemini/utils.go
```diff
@@ -2362,10 +2363,10 @@ func buildJSONSchemaFromMap(schemaMap map[string]interface{}) *schemas.Responses
 
 	// Extract anyOf
 	if anyOf, ok := normalizedSchemaMap["anyOf"].([]interface{}); ok {
-		anyOfMaps := make([]map[string]any, 0, len(anyOf))
+		anyOfMaps := make([]schemas.OrderedMap, 0, len(anyOf))
```
akshaydeo, line 2366: This extraction still only accepts `anyOf`/`oneOf`/`allOf` as `[]interface{}`, while the twin `schemas.JSONSchemaFromMap` got a `case []OrderedMap:` in this PR for exactly this shape. Nothing routes typed slices here today, but the asymmetry will bite when something does:

```go
switch anyOf := normalizedSchemaMap["anyOf"].(type) {
case []interface{}:
    // existing per-item SafeExtractOrderedMap loop
case []schemas.OrderedMap:
    jsonSchema.AnyOf = anyOf
}
```

core/providers/gemini/utils.go
```diff
@@ -2708,13 +2775,12 @@ func extractSchemaMapFromResponseFormat(responseFormat *interface{}) map[string]
 		return nil
 	}
 
-	schemaMap, ok := schemaObj.(map[string]interface{})
-	if !ok {
-		return nil
+	switch schemaObj.(type) {
+	case map[string]interface{}, *schemas.OrderedMap, schemas.OrderedMap:
```
akshaydeo, line 2779: A typed-nil `*schemas.OrderedMap` matches this arm, and `normalizeOrderedSchemaForGemini` returns a nil pointer wrapped in a non-nil interface - which passes the caller's `schemaMap != nil` check at line 1255 and would serialize as `"responseJsonSchema": null`. Routing this through `schemas.SafeExtractOrderedMap` (which rejects typed-nil) closes it:

```go
if om, ok := schemas.SafeExtractOrderedMap(schemaObj); ok {
    return normalizeSchemaValueForGemini(om)
}
if m, ok := schemaObj.(map[string]interface{}); ok {
    return normalizeSchemaValueForGemini(m)
}
return nil
```

core/providers/anthropic/utils.go
```diff
@@ -2738,90 +2740,162 @@ func normalizeSchemaForAnthropic(schema map[string]interface{}) map[string]inter
...
+	// be []interface{} (JSON-decoded) or []schemas.OrderedMap (typed struct fields).
+	for _, key := range []string{"anyOf", "oneOf", "allOf"} {
+		switch schema[key].(type) {
+		case []interface{}, []schemas.OrderedMap:
+			normalized[key] = normalizeSchemaValueForAnthropic(schema[key])
 		}
-		normalized["anyOf"] = newAnyOf
 	}
 
 	// Recursively normalize oneOf
```
akshaydeo, line 2782: Leftover stubs from the loop consolidation - `// Recursively normalize oneOf` and `// Recursively normalize allOf` now sit over empty bodies and read like the handling was deleted. Safe to drop both.

core/providers/gemini/responses.go
```diff
@@ -2786,55 +2786,50 @@ func convertTextConfigToGenerationConfig(textConfig *schemas.ResponsesTextConfig
...
+		// If Schema field is set, use it directly. Normalize via the
+		// OrderedMap-aware path so the client's key order survives end-to-end.
+		return normalizeSchemaValueForGemini(jsonSchema.Schema)
+	}
 
-		if jsonSchema.Defs != nil {
-			schema["$defs"] = *jsonSchema.Defs
-		}
+	// New format: Schema is spread across individual fields
+	schema := make(map[string]interface{})
```
akshaydeo, line 2796: Follow-up rather than this PR: this spread branch duplicates `ResponsesTextConfigFormatJSONSchema.ToMap()` but emits only 8 of its ~24 fields - `items`, `enum`, `anyOf`/`oneOf`/`allOf`, `$ref`, `definitions`, `format`, `pattern`, and the min/max constraints are silently dropped from schemas sent to Gemini (pre-existing, this PR just had to touch both spreaders). Calling `ToMap()` and layering the two gemini extras (`title` from `Name`, `propertyOrdering`) on top would close the gap, though it changes behavior and deserves its own tests.

georg-wolflein:
Closing: this change landed in dev via #5408 (with follow-up test fixes in #5411) and was released in transports/v1.6.5, which resolves #5148. The content of #5408 is substantially this PR's implementation (e.g. the payload-ordering test changes are identical), with the review nits from this thread fixed on top.

PR #5197, opened 2026-07-14
fix(gemini): harden processGeminiStreamChunk for non-object stream chunks

### LINKED ISSUE #5188: [Bug]: Gemini streaming intermittently fails with "failed to parse Gemini stream response: Mismatch type  ... with value string" when a chunk is not a JSON object
### Prerequisites

### Description

`processGeminiStreamChunk` in `core/providers/gemini/gemini.go` unconditionally decodes every SSE `data:` payload into a `GenerateContentResponse` struct via `sonic.Unmarshal`.
If the payload is not a JSON object, for example a bare JSON string like `"message"`, or a JSON array `[{...}]` sonic returns an error and the whole stream fails.

```
Failed to process chunk: failed to parse Gemini stream response: Mismatch type struct { CreateTime **time.Time "json:\"createTime,omitempty\""; **gemini.Alias } with value string "at index 0: mismatched type with value ... \"error\" ..."
```

I am not sure in what case can gemini return streamed error messages. I can see this bug report in google python sdk https://github.com/googleapis/python-genai/issues/1162

### Steps to reproduce

Minimal Go reproduction (I have copied the exact `GenerateContentResponse` type, its `UnmarshalJSON`, and `processGeminiStreamChunk`):

  ```go
  package main

  import (
        "bytes"
        "fmt"
        "reflect"
        "time"

        "github.com/bytedance/sonic"
  )

  type Candidate struct{}
  type GenerateContentResponsePromptFeedback struct{}
  type GenerateContentResponseUsageMetadata struct{}

  type GenerateContentResponse struct {
        Candidates     []*Candidate                           `json:"candidates,omitempty"`
        CreateTime     time.Time                              `json:"createTime,omitempty"`
        ModelVersion   string                                 `json:"modelVersion,omitempty"`
        PromptFeedback *GenerateContentResponsePromptFeedback `json:"promptFeedback,omitempty"`
        ResponseID     string                                 `json:"responseId,omitempty"`
        UsageMetadata  *GenerateContentResponseUsageMetadata  `json:"usageMetadata,omitempty"`
  }

  func (g *GenerateContentResponse) UnmarshalJSON(data []byte) error {
        type Alias GenerateContentResponse
        aux := &struct {
                CreateTime *time.Time `json:"createTime,omitempty"`
                *Alias
        }{Alias: (*Alias)(g)}
        if err := sonic.Unmarshal(data, &aux); err != nil {
                return err
        }
        if !reflect.ValueOf(aux.CreateTime).IsZero() {
                g.CreateTime = time.Time(*aux.CreateTime)
        }
        return nil
  }

  func processGeminiStreamChunk(jsonData []byte) (*GenerateContentResponse, error) {
        if bytes.Contains(jsonData, []byte(`"error"`)) {
                var errorCheck map[string]interface{}
                if err := sonic.Unmarshal(jsonData, &errorCheck); err == nil {
                        if errValue, hasError := errorCheck["error"]; hasError {
                                return nil, fmt.Errorf("gemini api error: %v", errValue)
                        }
                }
        }
        var geminiResponse GenerateContentResponse
        if err := sonic.Unmarshal(jsonData, &geminiResponse); err != nil {
                return nil, fmt.Errorf("failed to parse Gemini stream response: %v", err)
        }
        return &geminiResponse, nil
  }

  func main() {
        // 1. Healthy object — parses fine.
        ok := []byte(`{"candidates":[],"createTime":"2026-02-11T14:30:56Z","modelVersion":"gemini-2.0-flash"}`)
        if _, err := processGeminiStreamChunk(ok); err != nil {
                fmt.Println("Case1 err:", err)
        } else {
                fmt.Println("Case1 ok")
        }

        // 2. Array-framed success chunk.
        arrayChunk := []byte(`[{"candidates":[],"createTime":"2026-02-11T14:30:56Z"}]`)
        if _, err := processGeminiStreamChunk(arrayChunk); err != nil {
                fmt.Println("Case2 err:", err)
        }

        // 3. Array-wrapped error — the "error" fast-path misses it.
        errorArrayChunk := []byte(`[{"error":{"code":400,"message":"bad thing"}}]`)
        if _, err := processGeminiStreamChunk(errorArrayChunk); err != nil {
                fmt.Println("Case3 err:", err)
        }

        // 4. Exact production repro: a bare JSON string fragment from a split error body.
        stringChunk := []byte(`"message"`)
        if _, err := processGeminiStreamChunk(stringChunk); err != nil {
                fmt.Println("Case4 err:", err)
        }
  }
```

  Output:
  Case1 ok
  Case2 err: failed to parse Gemini stream response: Mismatch type struct { CreateTime *time.Time "json:\"createTime,omitempty\""; *main.Alias } with value array ...
  Case3 err: failed to parse Gemini stream response: Mismatch type struct { ... } with value array ...
  Case4 err: failed to parse Gemini stream response: Mismatch type struct { ... } with value  "at index 1: mismatched type with value\n\n\t\"message\"\n\t.^.......\n"

  Case 4 matches the production error signature.

### Expected behavior

Bifrost should not fail the entire stream when a single chunk is malformed or non-object-shaped. Specifically it should: (a) skip or buffer chunks whose top-level JSON is not {...}, (b)    handle array-framed responses by iterating each element, and (c) surface real Gemini API errors (including array-wrapped ones) instead of a generic struct-mismatch error.

### Actual behavior

 processGeminiStreamChunk returns failed to parse Gemini stream response: Mismatch type struct { CreateTime *time.Time ...; *gemini.Alias } with value string "message" (or "with value   array" for array-framed payloads) and the stream is aborted. The raw offending chunk is not logged, making the problem hard to diagnose in production.

### Affected area(s)

Core (Go)

### Version

core/v1.5.10, also present in main

### Environment

```text
- OS: Linux (production), macOS 14.5 (repro)
- Go: 1.23
- Bifrost components: core v1.5.10
```

### Relevant logs/output

```shell

```

### Regression?

_No response_

### Severity

axelray-dev:
## Summary

Hardens Gemini stream chunk parsing so non-object SSE payloads no longer surface as struct mismatch failures.

Fixes maximhq/bifrost#5188

## Changes

* Trim empty stream payloads and return a clear soft parse error.
* Decode top-level JSON arrays by taking the first object element.
* Surface array-wrapped Gemini API errors as `gemini api error: ...`.
* Soft-fail bare JSON strings, numbers, booleans, and null with an unexpected-type message.
* Keep the existing object decode and object-error path for healthy chunks.
* Add table-driven unit tests for the helper.

## Verification

* `go test ./core/providers/gemini/ -count=1 -run ProcessGeminiStreamChunk`

## Backwards compatibility

* Healthy object chunks are unchanged.
* Call sites that already warn and continue on non-`gemini api error` parse failures keep that behavior for soft failures.


yuvarajl:
@akshaydeo it would be great if it could land on next release ? It would be great as some failures we are seeing in the systems

axelray-dev:
Addressed the array-handling review feedback.

The array branch now processes every object element, preserves all response content in the stream response, and propagates an API error from any later element instead of returning after the first object.

Added regression coverage for:
- multiple object elements preserving both content parts and the final finish reason
- a later Gemini API error being returned to the stream

Validation on the updated branch:
- `go test ./providers/gemini -count=1` -- passed
- `git diff --check` -- passed

Please re-review the thread.

axelray-dev:
Updated the branch with the current `dev` branch and added the requested regression assertion that the first merged Gemini content part remains `"first"`.

Validation on the updated branch:
- `go test -p 1 ./providers/gemini -count=1` -- passed
- `git diff --check` -- passed

Please re-review the updated branch.

axelray-dev:
Verified the current head already contains the requested first-part assertion and the array merge/error propagation coverage. VPS validation: go test -p 1 ./providers/gemini -count=1 passed. No additional code change was needed.

axelray-dev:
I rechecked the current head against the stream-array review. processGeminiStreamChunk iterates every object in an array, merges all candidate parts, and propagates a later error instead of returning after the first element. The regression test now also asserts that the first irst part is preserved, followed by second. On the VPS, the full TestProcessGeminiStreamChunk group and go vet ./providers/gemini pass. No additional code change was needed.

PR #5230, opened 2026-07-15
fix(openrouter): merge embeddings catalog into ListModels discovery

### LINKED ISSUE #5224: [Bug]: Forbidden error ("model_blocked") when using OpenRouter embedding model via virtual key
### Prerequisites

### Description

I'm using a virtual key that is authorized to use all models from OpenRouter, but I receive a model_blocked error when requesting an embedding from an OpenRouter model.

### Steps to reproduce

1. Configure a virtual key with access to all OpenRouter models.
2. Send a request to the /v1/embeddings endpoint with the following payload:
```{
  "model": "openrouter/intfloat/multilingual-e5-large",
  "input": ["sample text 1"]
}```

### Expected behavior

The request should succeed, since the virtual key is authorized for all OpenRouter models.

### Actual behavior

The request is rejected with a model_blocked error, despite the virtual key having full access to OpenRouter models.

### Affected area(s)

Core (Go)

### Version

v1.6.3

### Environment

```text
Running in Kubernetes 1.35 with Graviton 3 processor
```

### Relevant logs/output

```shell

```

### Regression?

_No response_

### Severity

Shaik-Sirajuddin:
## Summary

OpenRouter publishes embedding models via a separate `/v1/embeddings/models` endpoint rather than including them in `/v1/models`. Bifrost's model catalog only ever fetched the latter, so a virtual key with a wildcard allow-list (`["*"]`) for OpenRouter still got `model_blocked` on legitimate embedding requests (e.g. `openrouter/intfloat/multilingual-e5-large`) — the model simply never appeared in the catalog governance checks against. Closes #5224.

## Changes

- Fetch and merge OpenRouter's `/v1/embeddings/models` catalog into the existing `/v1/models` discovery in `listModelsByKey`, reusing the same filter/alias/backfill pipeline (`core/providers/openrouter/openrouter.go`).
- Embeddings fetch is best-effort: a failure there never blocks chat-model discovery.
- Always fetches the fixed `/v1/embeddings/models` path — a `BifrostContextKeyURLPath` override meant for the primary model-list request is not replayed against it.

## Type of change

- [x] Bug fix

## How to test

```sh
go test ./core/providers/openrouter/...
```

Manually verified end-to-end against a live Bifrost instance + real OpenRouter key: reproduced #5224 (`model_blocked` on `openrouter/intfloat/multilingual-e5-large` with a wildcard-allowed VK), confirmed it now succeeds, confirmed a genuinely nonexistent model is still correctly blocked, and confirmed normal chat completions are unaffected.

## Breaking changes

- [x] No

## Related issues

Closes #5224


Shaik-Sirajuddin:
Verified against a live Bifrost instance + real OpenRouter API key, rebuilding both the pre-fix (`upstream/dev`) and post-fix binaries and re-running the exact repro from #5224 against each.

**Before fix** — `POST /v1/embeddings` with `openrouter/intfloat/multilingual-e5-large` on a VK with `allowed_models: ["*"]`:
```json
{
  "type": "model_blocked",
  "status_code": 403,
  "error": {
    "message": "Model 'intfloat/multilingual-e5-large' is not allowed for this virtual key"
  }
}
```

**After fix** — same request, same VK:
```json
{
  "data": [{ "index": 0, "object": "embedding", "embedding": [0.0166..., "...", 1024 dims] }],
  "model": "intfloat/multilingual-e5-large",
  "usage": { "prompt_tokens": 5, "total_tokens": 5, "cost": { "total_cost": 5e-08 } }
}
```

Model Catalog UI after the fix, showing the embedding model discovered and actively serving traffic under OpenRouter:

Model Catalog showing intfloat/multilingual-e5-large under OpenRouter

PR #5317, opened 2026-07-17
feat: support GET /v1/models/{model_id} retrieval with optional custom-field preservation

### LINKED ISSUE #5261: [Feature]: Support /v1/models/{model_id} retrieval
### Prerequisites

### Problem to solve

Like [LiteLLM](https://github.com/BerriAI/litellm/pull/13268), it would be good that bifrost support this feature with all custom metadata fields included. I have some openai api compatible models which also have non-standard metadata fields. I would like to serve them through bifrost.

### Proposed solution

The  /v1/models/{model_id} should return model related fields without truncating any non-standard field 

### Alternatives considered

_No response_

### Area(s)

Core (Go)

### Additional context

_No response_

Shaik-Sirajuddin:
## Summary

`/v1/models/{model_id}` retrieval didn't exist, and models list/retrieve silently dropped any non-standard field an upstream OpenAI-compatible backend (e.g. self-hosted vLLM) returned. Closes #5261.

## Changes

- New `GET /v1/models/{model_id}` on the native surface and the `/openai`, `/cursor` OpenAI-compatible surfaces — reuses the existing list-models call + filters for the match, no new provider-interface method.
- New per-provider `include_custom_model_fields` flag (default off, mirrors `send_back_raw_response`) — preserves non-standard upstream fields on retrieval instead of dropping them; list endpoint output is unchanged either way.
- Extra fields kept as `json.RawMessage` (not `interface{}`) so large integers round-trip without float64 precision loss.
- Wired through all ~20 providers sharing the OpenAI list-models path, configstore + DB migration, and the UI (new toggle in the provider Debugging tab).

## Type of change

- [x] Feature

## Affected areas

- [x] Core (Go)
- [x] Transports (HTTP)
- [x] Providers/Integrations
- [x] UI (React)

## How to test

```sh
go test ./core/... ./framework/configstore/... ./transports/bifrost-http/...
```

Test report and raw logs: https://github.com/maximhq/bifrost/pull/5317#issuecomment-4999286218

## Breaking changes

- [x] No


Shaik-Sirajuddin:
## Test report

| Suite | Result |
|---|---|
| `core/providers/openai` | 1 pre-existing failure (unrelated, confirmed via `git stash` on `dev`) |
| `core/providers/{ollama,groq,vllm,sgl,bedrock,bedrockmantle,xai,cerebras,nebius,parasail,perplexity,sarvam,opencode,deepseek,fireworks}` | pass |
| `core` (incl. VK-filter/nil-ctx regression tests below) | pass |
| `core/schemas` | pass |
| `framework/configstore` (incl. new migration) | pass |
| `transports/bifrost-http/handlers` (incl. new E2E tests) | pass |
| `transports/bifrost-http/integrations` | pass |

Also verified against real software: a local Ollama instance, the real OpenAI API (byte-identical retrieve response with the flag on vs. off), and a mock vLLM-style upstream with raw JSON diffing for exact field/precision preservation.

### Before/after — reviewer-flagged fix (VK provider allow-list bypass + nil-context panic)

Ran the two new regression tests against the pre-fix code, then against the fix, same command:

**Before:**
```
=== RUN   TestRetrieveModelRequest_RespectsVirtualKeyProviderFilter
    bifrost_test.go:3060: expected 404, got status 0x285db598458 (message: invalid x-api-key)
--- FAIL: TestRetrieveModelRequest_RespectsVirtualKeyProviderFilter (5.15s)
=== RUN   TestRetrieveModelRequest_NilContextDoesNotPanic
    bifrost_test.go:3083: RetrieveModelRequest panicked with nil context: runtime error: invalid memory address or nil pointer dereference
--- FAIL: TestRetrieveModelRequest_NilContextDoesNotPanic (0.00s)
FAIL
```
(the "invalid x-api-key" error confirms it actually attempted the call to the disallowed provider instead of blocking it)

**After:**
```
=== RUN   TestRetrieveModelRequest_RespectsVirtualKeyProviderFilter
--- PASS: TestRetrieveModelRequest_RespectsVirtualKeyProviderFilter (0.00s)
=== RUN   TestRetrieveModelRequest_NilContextDoesNotPanic
--- PASS: TestRetrieveModelRequest_NilContextDoesNotPanic (5.02s)
PASS
```


### Before/after — follow-up: same bug at the HTTP-handler level

Greptile's re-review after the first fix (core-level only) found the HTTP handler still didn't
resolve the VK's allowed-provider set when a provider was explicitly named. Same before/after:

**Before:**
```
retrieve_model_e2e_test.go:225: expected 404 (provider outside VK allow-list), got 200: {"id":"groq/llama3-groq","owned_by":"groq"}
--- FAIL: TestRetrieveModel_E2E_VirtualKeyProviderRestrictionBlocksExplicitProvider
```
(disallowed provider's model returned in full - genuine leak, not just a filtering gap)

**After:**
```
--- PASS: TestRetrieveModel_E2E_VirtualKeyProviderRestrictionBlocksExplicitProvider (0.01s)
```

PR #5318, opened 2026-07-17
fix: make Gemini isImageEditRequest order-insensitive across parts

### LINKED ISSUE #5314: [Bug]: Gemini native generateContent — image-to-image silently dropped when a text part precedes the image (image-edit detection only inspects Parts[0])
## Description

On the native Gemini ingress (`/genai/v1beta/models/{model}:generateContent`), Bifrost decides between **image generation** (text→image) and **image edit** (image→image) by inspecting **only the first part** of the first content.

If the caller places the text prompt before the inline image — a perfectly valid ordering that the native Google API accepts — the request is misclassified as image **generation**, and **all `inlineData` image parts are silently discarded**. The model then produces a fresh text-to-image result that is completely unrelated to the input image.

This breaks every image-to-image / reference-image flow whenever the client emits parts in `[text, image]` order. Text-only generation is unaffected (nothing to drop), which makes it look like *"image-to-image is broken but text-to-image works"*.

> Related to but distinct from #2283 (which covered `size` / `imageConfig` mapping on the OpenAI-style `/v1/images/*` endpoints and `imageConfig` passthrough). This report is specifically about **edit-vs-generation classification dropping the input image on the native `generateContent` endpoint**, which is still present on latest `dev`.

## Environment

- Reproduced against a self-hosted Bifrost gateway, model `gemini-3.1-flash-image`.
- Root cause confirmed present on `dev` @ `c0909f9` (2026-07-15).

## Reproduction

Use a distinctive base image, and a **neutral** prompt that does **not** describe the image — so that if the image is dropped, a text-only fallback cannot accidentally "guess" the input and mask the bug. Only the **part order** differs between the two requests.

Generate a 512×512 input (left half magenta `#FF00FF`, right half cyan `#00FFFF`):

```python
import zlib, struct, base64
W = H = 512
def chunk(t, d): return struct.pack(">I", len(d)) + t + d + struct.pack(">I", zlib.crc32(t + d) & 0xffffffff)
raw = bytearray()
for y in range(H):
    raw.append(0)
    for x in range(W):
        raw += bytes((255, 0, 255)) if x < W // 2 else bytes((0, 255, 255))
png = b"\x89PNG\r\n\x1a\n" + chunk(b"IHDR", struct.pack(">IIBBBBB", W, H, 8, 2, 0, 0, 0)) + \
      chunk(b"IDAT", zlib.compress(bytes(raw), 9)) + chunk(b"IEND", b"")
open("input.png", "wb").write(png)
open("input.b64", "w").write(base64.b64encode(png).decode())
```

**Case A — text first (typical SDK ordering) → input image dropped:**

```bash
B64=$(cat input.b64)
curl -sS -X POST "https://<bifrost>/genai/v1beta/models/gemini-3.1-flash-image:generateContent" \
  -H "x-goog-api-key: <vk>" -H "Content-Type: application/json" \
  -d "{\"contents\":[{\"role\":\"user\",\"parts\":[
        {\"text\":\"Add one small yellow circle at the exact center. Do not change anything else in the image.\"},
        {\"inlineData\":{\"mimeType\":\"image/png\",\"data\":\"$B64\"}}
      ]}],
      \"generationConfig\":{\"responseModalities\":[\"IMAGE\"]}}"
```

**Result:** an unrelated scene; the magenta/cyan input is nowhere in the output. Bifrost routed this to `ToBifrostImageGenerationRequest`, which extracts only the text prompt.

**Case B — image first → works:**

Identical request, but `parts` = `[ {inlineData…}, {text…} ]`.

**Result:** the exact magenta/cyan image is preserved, with a yellow circle added. Bifrost routed this to `ToBifrostImageEditRequest`, which forwards the image.

(Confirmed from gateway logs: Case A = image generation, Case B = image edit.)

## Observed vs Expected

| Part order | Observed | Expected |
|---|---|---|
| `[text, image]` | image **generation** — `inlineData` dropped → output unrelated to input | image **edit** — input image used |
| `[image, text]` | image **edit** — input image used | image **edit** — input image used |

The native Google API is order-insensitive for this; Bifrost should be too.

## Root cause

`transports/bifrost-http/integrations/genai.go` — `isImageEditRequest()` inspects only `Parts[0]`:

```go
if len(req.Contents) > 0 && len(req.Contents[0].Parts) > 0 &&
   req.Contents[0].Parts[0].InlineData != nil &&                 // ← only the FIRST part
   strings.Contains(req.Contents[0].Parts[0].InlineData.MIMEType, "image") {
    for _, modality := range req.GenerationConfig.ResponseModalities {
        if modality == gemini.ModalityImage {
            return true
        }
    }
}
```

When this returns `false`, the request falls through to `isImageGenerationRequest()` → `ToBifrostImageGenerationRequest()` in `core/providers/gemini/images.go`, whose "standard Gemini contents" branch extracts **only** the text prompt and never reads `part.InlineData` — so every inline image is dropped:

```go
// ToBifrostImageGenerationRequest — contents branch
for _, content := range request.Contents {
    for _, part := range content.Parts {
        if part != nil && part.Text != "" {
            bifrostReq.Input.Prompt = part.Text   // only text; InlineData never read
            break
        }
    }
    ...
}
```

The correct path, `ToBifrostImageEditRequest()`, already collects `InlineData` from **all** parts regardless of position — it is simply never reached when the text part happens to come first.

## Suggested fix

Make `isImageEditRequest()` scan **all** parts (across all contents) for an image `InlineData`, not just `Parts[0]`:

```go
func isImageEditRequest(req *gemini.GeminiGenerationRequest) bool {
    if schemas.IsImagenModel(req.Model) && len(req.Instances) > 0 && req.Instances[0].ReferenceImages != nil {
        return true
    }
    hasImage := false
    for _, content := range req.Contents {
        for _, part := range content.Parts {
            if part.InlineData != nil && strings.Contains(part.InlineData.MIMEType, "image") {
                hasImage = true
                break
            }
        }
        if hasImage {
            break
        }
    }
    if hasImage {
        for _, modality := range req.GenerationConfig.ResponseModalities {
            if modality == gemini.ModalityImage {
                return true
            }
        }
    }
    return false
}
```

This makes classification order-insensitive and matches the native Google API's behavior.

As a client-side mitigation we now emit image parts before the text part, which restores correct behavior — but the gateway ideally shouldn't depend on part order for a request that is valid either way.

Shaik-Sirajuddin:
## Description

Fix Gemini `generateContent` image-edit detection to be part-order-insensitive within a single content, instead of only checking `Contents[0].Parts[0]`.

## Details

- `isImageEditRequest()` now scans all parts of `Contents[0]` for inline image data, not just `Parts[0]`.
- Scope stays limited to `Contents[0]` (matches original behavior) — multi-turn/cross-content handling is unchanged and out of scope.
- MIME type match is now case-insensitive.
- Added unit + end-to-end tests covering the regression and edge cases.

Fixes #5314

## Verification

Reproduced live against the real Gemini API, via `curl` and the official `google-genai` SDK, by inspecting the outgoing request Bifrost sends to Google: before the fix the image was dropped, after the fix it's preserved byte-identical.

## Checklist

- [x] Tests added/updated and passing
- [x] Verified against a real endpoint
- [x] Docs updated (or N/A) — N/A


Shaik-Sirajuddin:
## Test report

**Before (unfixed):**
```
=== RUN   TestIsImageEditRequest_TextFirst
--- FAIL: TestIsImageEditRequest_TextFirst (0.00s)
=== RUN   TestIsImageEditRequest_ImageAcrossMultipleContents
--- FAIL: TestIsImageEditRequest_ImageAcrossMultipleContents (0.00s)
FAIL
```

**After (fixed):**
```
--- PASS: TestIsImageEditRequest_ImageFirst (0.00s)
--- PASS: TestIsImageEditRequest_TextFirst (0.00s)
--- PASS: TestIsImageEditRequest_TextOnly_NotEdit (0.00s)
--- PASS: TestIsImageEditRequest_ImageOnlyInLaterContent_NotEdit (0.00s)
--- PASS: TestIsImageEditRequest_HistoricalImageInEarlierContent_NotEdit (0.00s)
--- PASS: TestIsImageEditRequest_CaseInsensitiveImageMimeType (0.00s)
--- PASS: TestIsImageEditRequest_NilPartInList (0.00s)
--- PASS: TestIsImageEditRequest_NonImageInlineData_NotEdit (0.00s)
--- PASS: TestIsImageEditRequest_ImageWithoutImageModality_NotEdit (0.00s)
--- PASS: TestIsImageEditRequest_TextFirst_EndToEndPreservesImage (0.00s)
PASS
```

| Check | Before | After |
|---|---|---|
| `TestIsImageEditRequest*` | 2 FAIL / 6 | 10 PASS / 10 |
| `integrations` package | — | PASS |
| `core/providers/gemini` package | — | PASS |
| Live (`curl`), text-first, outgoing `inlineData` | absent | present, byte-identical |
| Live (`google-genai` SDK), text-first, outgoing `inlineData` | absent | present, byte-identical |

PR #5322, opened 2026-07-17
fix: mint fresh Responses ID on chat bridge, preserve repeated transcription form fields

### LINKED ISSUE #5319: Chat-only providers bridged into /v1/responses return chatcmpl-* IDs instead of resp_*
## Description

`ToBifrostResponsesResponse` (`core/schemas/mux.go`) reuses the source chat completion's `chatcmpl-*` ID verbatim instead of minting a fresh `resp_*` ID when bridging a chat-only provider's response into the `/v1/responses` shape.

Affects any chat-only provider bridged into the Responses API (Ollama, Groq, DeepSeek, Cerebras, Nebius, sgl, vllm, Mistral, Parasail, Sarvam, Opencode). Gemini and Anthropic already mint fresh `resp_*` IDs on their own bridge paths, so this is an inconsistency between providers, not a universal gap.

## Steps to reproduce

```
POST /v1/responses
{"model":"ollama/qwen2.5:0.5b","input":"hi"}
```

Response `id` comes back as `chatcmpl-<n>` instead of `resp_<random>`.

## Expected behavior

`id` should always be `resp_*` for any `/v1/responses` response, regardless of which provider served the underlying request.

## Affected area(s)

Core (Go)

### LINKED ISSUE #5320: Repeated transcription multipart form fields collapse to a single value
## Description

`prepareTranscriptionRequest` (`transports/bifrost-http/handlers/inference.go`) reads unknown multipart form fields into `ExtraParams` as `value[0]`, always taking only the first value of a repeated field. Array-style transcription params (e.g. `known_speaker_names[]`, `known_speaker_references[]`, as used by the OpenAI diarization transcription params) silently lose every value past the first.

The outbound OpenAI-compatible multipart writer (`core/providers/openai/transcription.go`, `ParseTranscriptionFormDataBodyFromRequest`) has the same one-value assumption on the way back out: it only special-cases `string` in `ExtraParams`, so even an in-process `[]string` value would get JSON-encoded into a single field instead of re-emitted as repeated multipart fields.

## Steps to reproduce

```
POST /v1/audio/transcriptions
Content-Type: multipart/form-data

model=openai/whisper-1
known_speaker_names=alice
known_speaker_names=bob
known_speaker_references=refA
known_speaker_references=refB
```

Only `alice`/`refA` reach the upstream provider; `bob`/`refB` are silently dropped.

## Expected behavior

All repeated values for a given field should be preserved end-to-end and forwarded to the provider as repeated multipart fields under the same key.

## Affected area(s)

Transports (HTTP), Core (Go)

Shaik-Sirajuddin:
## Summary

Two Responses/transcription pipeline bugs: chat-only providers bridged into `/v1/responses` leaked a `chatcmpl-*` ID instead of `resp_*` (non-streaming and streaming), and repeated multipart transcription form fields (e.g. speaker-diarization arrays) collapsed to a single value both on the way in and on the way back out to an OpenAI-compatible upstream.

## Changes

- `ToBifrostResponsesResponse` and `ToBifrostResponsesStreamResponse` mint a fresh `resp_*` ID instead of reusing the source chat completion's ID — matches the existing convention already used by the Gemini/Anthropic bridge paths.
- `prepareTranscriptionRequest` preserves repeated multipart form fields as `[]string` in `ExtraParams` instead of collapsing to the first value, dropping blank entries; single-valued fields keep the existing scalar behavior.
- `ParseTranscriptionFormDataBodyFromRequest` re-emits `[]string` extra params as repeated multipart fields under the same key, instead of JSON-encoding them into one field — otherwise the parsing fix wouldn't actually reach the upstream provider correctly.
- Added a regression test asserting bridged responses get a `resp_`-prefixed id distinct from the source chat completion id.

## Type of change

- [x] Bug fix

## Breaking changes

- [x] No

## How to test

```sh
go build ./...
go test ./core/schemas/... ./core/providers/openai/... ./transports/bifrost-http/handlers/...
```

## Test report

| Case | Before | After |
|---|---|---|
| `/v1/responses` via a chat-only provider | `id` = `chatcmpl-*` | `id` = `resp_*` |
| Repeated speaker-diarization fields via `/v1/audio/transcriptions` | only 1st value reaches upstream | all values reach upstream |

Full raw before/after logs (official SDK, live temp instance) in a PR comment below.

Fixes #5319
Fixes #5320


Shaik-Sirajuddin:
### Raw test logs (official `openai` Python SDK, live temp instance)

**`/v1/responses` id — before/after**
```
before: resp.id = 'chatcmpl-425'
after:  resp.id = 'resp_582097718f2bc82d283420cb34f728cfe98553c5ea39c3b702'
```

**Speaker-diarization fields captured at upstream — before/after**
```
before: known_speaker_names[]=['alice']       known_speaker_references[]=['refA']
after:  known_speaker_names[]=['alice','bob'] known_speaker_references[]=['refA','refB']
```

PR #5323, opened 2026-07-17
fix: strip temperature for OpenAI/Azure reasoning models on Chat Completions and Responses paths

### LINKED ISSUE #5321: OpenAI/Azure reasoning models reject non-default temperature, Bifrost forwards it unstripped
## Description

`ToOpenAIChatRequest` and `ToOpenAIResponsesRequest` (`core/providers/openai/chat.go`, `core/providers/openai/responses.go`) strip `top_p` for reasoning models (o1/o3/o4, gpt-oss, GPT-5.x) on OpenAI/Azure, but never strip `temperature`. These models reject any non-default `temperature` on OpenAI/Azure's actual endpoints with a 400 error.

Affects both `/v1/chat/completions` and `/v1/responses` for the `openai`/`azure` providers.

## Steps to reproduce

```
POST /v1/chat/completions
{"model":"azure/gpt-5.5","messages":[{"role":"user","content":"hi"}],"temperature":0.3}
```

`temperature` is forwarded verbatim to Azure, which rejects it.

## Expected behavior

`temperature` should be stripped for reasoning models on OpenAI/Azure, consistent with the existing `top_p` handling, without affecting other providers that host reasoning-named models (e.g. Fireworks/OpenRouter `gpt-oss`) and don't share this restriction.

## Affected area(s)

Core (Go)

Shaik-Sirajuddin:
## Summary

OpenAI/Azure reasoning models (o1/o3/o4, gpt-oss, GPT-5.x) reject any non-default `temperature`, but Bifrost forwarded it unstripped on both the Chat Completions and Responses request paths.

## Changes

- `ToOpenAIChatRequest`: strip `temperature` for reasoning models, scoped to the `schemas.OpenAI`/`schemas.Azure` case only, mirroring the existing `reasoning_effort` normalization in the same switch branch.
- `ToOpenAIResponsesRequest`: strip `temperature` alongside the existing `top_p`-stripping block for reasoning models — scoped specifically to `schemas.OpenAI`/`schemas.Azure` (unlike the pre-existing `top_p` strip, which applies to any provider sharing this request builder) since other providers hosting reasoning-named models (e.g. Fireworks/OpenRouter `gpt-oss`) don't share this restriction.

## Type of change

- [x] Bug fix

## Breaking changes

- [x] No

## How to test

```sh
go build ./...
go test ./core/providers/openai/...
```

## Test report

| Path | Before | After |
|---|---|---|
| `/v1/chat/completions` (`model=openai/gpt-5.5`, `temperature=0.3`) | `temperature` present in outbound body | `temperature` absent |
| `/v1/responses` (`model=openai/gpt-5.5`, `temperature=0.3`) | `temperature` present in outbound body | `temperature` absent |

Full raw before/after logs (official SDK, live temp instance) in a PR comment below.

Fixes #5321

### DIFF
--- a/core/providers/openai/chat.go
+++ b/core/providers/openai/chat.go
@@ -71,6 +71,7 @@ func ToOpenAIChatRequest(ctx *schemas.BifrostContext, bifrostReq *schemas.Bifros
 	switch bifrostReq.Provider {
 	case schemas.OpenAI, schemas.Azure:
 		openaiReq.normalizeReasoningEffort(caps)
+		openaiReq.stripTemperatureForReasoningModels(capModel)
 		// URL-sourced documents are NOT inlined here. Chat Completions rejects file_url, so they
 		// still have to be resolved before the request goes out - but that is a network fetch that
 		// can fail, and this function has no way to report a failure. Callers invoke
@@ -191,6 +192,17 @@ func (req *OpenAIChatRequest) normalizeReasoningEffort(caps schemas.ModelCaps) {
 			// Clear max_tokens since OpenAI doesn't use it
 			req.ChatParameters.Reasoning.MaxTokens = nil
 		}
+	}
+}
+
+// stripTemperatureForReasoningModels drops the temperature parameter for OpenAI/Azure
+// reasoning models (o1/o3/o4 series, gpt-oss, GPT-5.x), which reject any non-default
+// temperature value with a 400 error. Unlike top_p, GPT-5.x rejects temperature even
+// at the default "none" reasoning effort, so this strips unconditionally rather than
+// only for always-reasoning (-pro/-codex) variants.
+func (req *OpenAIChatRequest) stripTemperatureForReasoningModels(capModel string) {
+	if req.ChatParameters.Temperature != nil && IsOpenAIReasoningModel(capModel) {
+		req.ChatParameters.Temperature = nil
 	}
 }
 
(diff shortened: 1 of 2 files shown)


Shaik-Sirajuddin:
### Raw test logs (official `openai` Python SDK, live temp instance)

**outbound `/v1/chat/completions` body — before/after**
```
before: {"model":"gpt-5.5","temperature":0.3,"messages":[{"role":"user","content":"hi"}]}
after:  {"model":"gpt-5.5","messages":[{"role":"user","content":"hi"}]}
```

**outbound `/v1/responses` body — before/after**
```
before: {"model":"gpt-5.5","temperature":0.3,"input":[{"role":"user","content":"hi"}]}
after:  {"model":"gpt-5.5","input":[{"role":"user","content":"hi"}]}
```

Shaik-Sirajuddin:
Acknowledged, but this is a structural limitation shared by the neighboring `reasoning_effort` normalization and `top_p`-stripping logic in the same functions — both already rely on `isOpenAIReasoningModel(capModel)` with `capModel := schemas.ResolveCanonicalModel(ctx, bifrostReq.Model)`, and neither can see through an opaque Azure deployment name like `production-reasoning` either.

`ResolveCanonicalModel` already exists specifically to solve this class of problem (its doc comment calls out the identical "Claude-on-Azure" case: an opaque deployment ID with an admin-configured `ModelName` alias carrying the real model string). Without that alias configured, there's no way for Bifrost to know what model backs a deployment short of calling Azure's management API at request time to discover it — a separate, much larger feature than this fix.

Not expanding scope here since it's pre-existing and not introduced by this change; the mitigation is configuring a `ModelName` alias for the deployment, same as any other Azure model-family detection in this codebase.

PR #5338, opened 2026-07-17
fix: round-trip MCPToolManagerConfig.tool_execution_timeout via MarshalJSON

### LINKED ISSUE #5337: [Bug]: MCPToolManagerConfig.tool_execution_timeout corrupts to a negative timeout on a JSON marshal/unmarshal round trip
### Prerequisites

- [x] I have tested against the current `dev` branch

### Description

`schemas.MCPToolManagerConfig` defines a custom `UnmarshalJSON` that reads a bare-integer `tool_execution_timeout` as **seconds** (matching `config.schema.json`, which documents the integer form as "an integer number of seconds"). However, the type has **no** `MarshalJSON`, so it inherits `schemas.Duration`'s default encoding, which emits the value as **nanoseconds**.

The two directions therefore disagree, and the type does not survive a JSON marshal → unmarshal round trip: a value marshaled as nanoseconds is read back as seconds, which multiplies it by 1e9 and overflows `int64`, producing a **negative** `time.Duration`. A negative timeout yields an already-expired context deadline, so every MCP tool execution that uses it fails immediately.

The sibling type `MCPClientConfig` was deliberately given a `MarshalJSON` "so it round-trips correctly" (it emits a duration string); `MCPToolManagerConfig` has the same seconds-based `UnmarshalJSON` but was missed.

### Steps to reproduce

Minimal, self-contained reproduction:

```go
package main

import (
	"fmt"
	"time"

	"github.com/bytedance/sonic"
)

func main() {
	cfg := schemas.MCPToolManagerConfig{
		ToolExecutionTimeout: schemas.Duration(30 * time.Second),
		MaxAgentDepth:        10,
	}

	data, _ := sonic.Marshal(cfg)
	fmt.Printf("marshaled:       %s\n", data)

	var back schemas.MCPToolManagerConfig
	_ = sonic.Unmarshal(data, &back)
	fmt.Printf("after round trip: %v\n", back.ToolExecutionTimeout.D())
}
```

Output:

```
marshaled:       {"tool_execution_timeout":30000000000,"max_agent_depth":10}
after round trip: -1914857h49m7.419103232s
```

Any code path that marshals the struct to JSON and later unmarshals it hits this. Note the generic `DeepCopy[T]` helper in `transports/bifrost-http/lib/config.go` is exactly a `sonic.Marshal` → `sonic.Unmarshal` round trip, so copying any value that embeds `MCPToolManagerConfig` through it would corrupt the timeout. The DB persistence path stores the value in an integer-seconds column and is unaffected.

### Expected behavior

`tool_execution_timeout` survives a marshal/unmarshal round trip unchanged (e.g. `30s` stays `30s`), like the sibling `MCPClientConfig`.

### Actual behavior

The value is reinterpreted from nanoseconds to seconds on the second unmarshal, overflows `int64`, and becomes a large negative duration (`-1914857h49m7.419103232s`), i.e. an already-expired timeout.

### Affected area(s)

Core (Go) — `core/schemas/mcp.go`

### Version

`core@v1.7.1` and current `dev` (the `MarshalJSON`/`UnmarshalJSON` asymmetry is present in both).

### Environment

```
- Go: 1.26.x
- Component: core/schemas (MCPToolManagerConfig)
```

### Severity

Medium (some functionality impaired) — MCP tool execution breaks for any config that round-trips the tool-manager timeout through JSON.

### Suggested fix

Add a `MarshalJSON` to `MCPToolManagerConfig` that emits `tool_execution_timeout` as a duration string, mirroring `MCPClientConfig.MarshalJSON`. Happy to open a PR (with round-trip regression tests).

glitch-ux:
## Summary

`MCPToolManagerConfig.tool_execution_timeout` does not survive a JSON marshal/unmarshal round trip: it is read back as a large negative (already-expired) `time.Duration`, which fails every MCP tool execution that relies on it.

`MCPToolManagerConfig.UnmarshalJSON` reads a bare-integer `tool_execution_timeout` as **seconds** (matching `config.schema.json`, which documents the integer form as seconds), but the type has **no** `MarshalJSON`, so it inherits `schemas.Duration`'s **nanosecond** encoding. Marshaling `30s` therefore emits `30000000000`, which the next unmarshal reinterprets as `30000000000` *seconds*, overflowing `int64`:

```
marshaled:       {"tool_execution_timeout":30000000000,"max_agent_depth":10}
after round trip: -1914857h49m7.419103232s
```

The sibling `MCPClientConfig` was already given a `MarshalJSON` "so it round-trips correctly"; `MCPToolManagerConfig` has the same seconds-based `UnmarshalJSON` but was missed.

## Changes

- Add `MCPToolManagerConfig.MarshalJSON` that emits `tool_execution_timeout` as a duration string, mirroring the existing `MCPClientConfig.MarshalJSON`. The value now round-trips, and the serialized form (a duration string) is consistent with the sibling type and permitted by `config.schema.json` (`"type": ["integer", "string"]`).
- Add regression tests: a marshal→unmarshal round-trip test, plus tests asserting the existing integer-seconds and duration-string unmarshal behavior is preserved.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [x] Core (Go)
- [ ] Transports (HTTP)
- [ ] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

```sh
go version   # go1.26.x
cd core
go vet ./schemas/
go test ./schemas/ -run MCPToolManagerConfig -v
```

Expected: all three `TestMCPToolManagerConfig*` tests pass. The round-trip test fails on `dev` (marshals `30000000000`, unmarshals to a negative duration) and passes with this change (marshals `"30s"`, round-trips to `30s`). The full `core/schemas` package suite also passes.

## Breaking changes

- [ ] Yes
- [x] No

`tool_execution_timeout` now serializes as a duration string (e.g. `"30s"`) instead of an integer nanosecond count. Reads already accept both forms, the schema permits both, and this matches how `MCPClientConfig` already serializes the same field.

## Related issues

Closes #5337

## Security considerations

None (auth/secrets/PII unaffected). The fix prevents a corrupted (negative) timeout that would otherwise cause MCP tool executions to fail with an already-expired deadline.

## Checklist

- [x] I read the contributing guidelines and followed them
- [ ] I updated documentation where needed (not needed — behavior now matches the documented schema)
- [x] I verified builds succeed (Go; `go vet` + `core/schemas` tests)
- [x] I verified the affected package tests pass locally


akshaydeo:
Hi @glitch-ux — thanks for the contribution! This PR is currently blocked because our CLA bot shows the Contributor License Agreement as not yet signed. Could you sign it here so we can move this forward: https://cla-assistant.io/maximhq/bifrost?pullRequest=5338

PR #5339, opened 2026-07-17
fix: apply routing rules to WebSocket Responses requests

etnperlong:
## Summary

WebSocket Responses requests (`GET /v1/responses` upgrade + `response.create` events) bypassed governance routing rules and load balancing. The handler parsed the provider directly from the model string (defaulting to OpenAI) and dialed the native WS upstream without ever running `PreRequestHook`, where routing rules are evaluated. Clients like Codex sending `gpt-*` models therefore always hit OpenAI regardless of configured routing rules.

The realtime WS handler already solved this (`wsrealtime.go` `handleUpgrade` runs `RunPreRequestHooks` explicitly); this PR brings the Responses WS handler in line with it.

## Changes

- capture upgrade-request query params in `authHeaders` (the fasthttp ctx is recycled after the handler returns, so they must be snapshotted at upgrade time)
- in `handleResponseCreate`, surface upgrade-request headers/query on the context so CEL routing rules (`headers[...]` / `params[...]`) see the same shape as normal HTTP requests
- run `RunPreRequestHooks` before upstream selection so governance routing rules / load balancing can rewrite provider/model (matched rules also commit their key pin via `BifrostContextKeyRoutingPinnedAPIKeyID`)
- read back the routed provider/model into the request; fail with 400 when no provider resolves (mirrors the empty-provider check in core `handleRequest`)
- recompute the per-provider `DisableStore` store override when routing changes the provider

Routing to a provider without native WS support falls back to the HTTP bridge as before, which handles format translation via the standard inference pipeline. The second `RunPreRequestHooks` inside the HTTP bridge path is idempotent: rules are evaluated against the already-rewritten model.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [ ] Core (Go)
- [x] Transports (HTTP)
- [ ] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

```sh
go build ./transports/bifrost-http/handlers/
go vet ./transports/bifrost-http/handlers/
go test ./transports/bifrost-http/handlers/
```

Manual verification: configure a routing rule targeting `gpt-*` models to a non-OpenAI provider, connect a WS Responses client (e.g. Codex), and send a `response.create` event with a `gpt-*` model. Before this fix the request always went to OpenAI via the native WS upstream; after the fix it follows the routing rule (falling back to the HTTP bridge when the routed provider has no native WS support).

## Screenshots/Recordings

N/A

## Breaking changes

- [ ] Yes
- [x] No

## Related issues

WebSocket Responses requests ignore configured routing rules and always route to OpenAI.

## Security considerations

Headers surfaced on the routing context are the same set already captured for auth handling; forwarded-header allowlist/denylist filtering (`isSecurityDeniedExtraHeader`, header matcher) is unchanged.

## Checklist

### DIFF
--- a/transports/bifrost-http/handlers/wsresponses.go
+++ b/transports/bifrost-http/handlers/wsresponses.go
@@ -113,6 +113,7 @@ type authHeaders struct {
 	googAPIKey    string
 	baggage       string
 	headers       map[string][]string
+	queryParams   map[string]string
 }
 
 // captureAuthHeaders captures the auth headers from the request.
@@ -129,6 +130,12 @@ func captureAuthHeaders(ctx *fasthttp.RequestCtx) *authHeaders {
 	for key, value := range ctx.Request.Header.All() {
 		k := strings.ToLower(string(key))
 		ah.headers[k] = append(ah.headers[k], string(value))
+	}
+	if queryArgs := ctx.URI().QueryArgs(); queryArgs.Len() > 0 {
+		ah.queryParams = make(map[string]string, queryArgs.Len())
+		for key, value := range queryArgs.All() {
+			ah.queryParams[strings.ToLower(string(key))] = string(value)
+		}
 	}
 	return ah
 }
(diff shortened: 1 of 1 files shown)


akshaydeo:
Hi @etnperlong — thanks for the contribution! This PR is currently blocked because our CLA bot shows the Contributor License Agreement as not yet signed. Could you sign it here so we can move this forward: https://cla-assistant.io/maximhq/bifrost?pullRequest=5339

PR #5355, opened 2026-07-18
[fix]: Anthropic provider - stop double-counting cache-read tokens in streaming usage

### LINKED ISSUE #5354: [Bug]: Anthropic streaming usage double-counts cache-read tokens (~2× prompt_tokens inflation)
### Prerequisites
- [x] I am using the latest version (tested against `dev` @ `a0f0afb84` / `core/v1.7.2`)

### Description

Bifrost over-counts prompt tokens by ~2× on any Anthropic-protocol stream where the provider emits `cache_read_input_tokens` only in the final `message_delta` event (not at `message_start`). This is a protocol-level bug — any provider using Anthropic streaming semantics with this `message_start` / `message_delta` usage split is affected. Reproduced with the `kimi-code` provider (`k3` model via an Anthropic-compatible endpoint), but the flaw is in Bifrost's Anthropic streaming accumulator, not the provider.

Per the Anthropic Messages API streaming contract, a response emits two usage snapshots:
1. `message_start` — `message.usage`: full prompt as `input_tokens`, with `cache_creation_input_tokens` and `cache_read_input_tokens` often still `0`.
2. `message_delta` (final event before `message_stop`) — `usage`: the **authoritative** post-cache breakdown: `input_tokens` = uncached-only tail, plus real `cache_creation_input_tokens` / `cache_read_input_tokens`.

The authoritative total prompt = `message_delta.input_tokens + cache_creation_input_tokens + cache_read_input_tokens`. The `message_delta` usage **overwrites**, it does not merge-with, the `message_start` snapshot.

### Steps to reproduce

Use any Anthropic-protocol streaming provider that emits cache stats only at `message_delta` (kimi-code/k3 exhibits this). Send 3 consecutive turns in the same session so prompt caching kicks in from turn 2 onward. Inspect the recorded `token_usage` in the exported request logs.

Stream events observed (verbatim from logs):

| Turn | `message_start` (input / cache_read) | `message_delta` (input / cache_read) | Correct total prompt | Bifrost recorded |
|---|---|---|---|---|
| 1 | 20458 / 0 | 20458 / 0 | 20458 | **20458** ✓ |
| 2 | 20532 / 0 | 308 / 20224 | 20532 | **40756** ✗ (= 20532 + 20224) |
| 3 | 20621 / 0 | 141 / 20480 | 20621 | **41101** ✗ (= 20621 + 20480) |

- Turn 1 is correct only because no cache read occurred.
- From turn 2 on, Bifrost records `message_start.input_tokens` (the full prompt, cache fields still 0) **plus** `message_delta.cache_read_input_tokens` — double-counting the cached portion.

Reproduction signature: first request accurate, every subsequent request ~2×, growing by ~2× the real per-turn conversation growth.

### Expected behavior

`prompt_tokens = message_delta.input_tokens + message_delta.cache_creation_input_tokens + message_delta.cache_read_input_tokens`

(i.e., `message_delta.usage` is authoritative for all input-side fields; the `message_start` snapshot is overwritten, not merged-with.)

### Actual behavior

`prompt_tokens = message_start.input_tokens + message_delta.cache_read_input_tokens`

### Root cause (one line)

The streaming usage accumulator in `HandleAnthropicChatCompletionStreaming` (`core/providers/anthropic/anthropic.go`) uses a `max-keep` guard on `input_tokens` that does NOT branch on event type — it treats `message_start.input_tokens` (full prompt) and `message_delta.input_tokens` (uncached tail, smaller) identically, so the delta's authoritative smaller value is rejected and the stale start value is retained; then `normalizeCachedUsage` folds `cache_read_input_tokens` on top at stream end → double-count.

Before/after formula:
- **BEFORE (buggy):** `prompt_tokens = message_start.input_tokens + message_delta.cache_read_input_tokens`
- **AFTER (correct):** `prompt_tokens = message_delta.input_tokens + message_delta.cache_creation_input_tokens + message_delta.cache_read_input_tokens`

### Affected area(s)
- Core (Go) — `core/providers/anthropic/anthropic.go`

### Version
- `core/v1.7.2`; reproduced against `dev` @ `a0f0afb84`

### Environment
- OS: macOS (reproduced); Go 1.26.x
- Provider: kimi-code (k3) via Anthropic-compatible endpoint
- Stream: `stream:true` with Anthropic Messages API event semantics (`message_start` → content deltas → `message_delta` → `message_stop`)
- Bifrost components: core

### Relevant logs/output

Three consecutive exported request logs from the same session (turn 1 correct, turns 2 & 3 double-counted) — see the evidence table above. Each log's top-level `token_usage` field holds the recorded `prompt_tokens` and `prompt_tokens_details.cached_read_tokens`; `raw_response` (JSON-line-delimited stream events) holds the `message_start` and `message_delta` snapshots. Happy to share the raw logs if helpful.

### Regression?
Unknown — the `max-keep` + `normalizeCachedUsage` pattern appears long-standing in the Anthropic provider path.

### Severity
**High** — silent ~2× inflation of prompt-token accounting for any Anthropic-protocol stream with prompt caching. Directly impacts cost calculation, governance budget tracking, telemetry, and any downstream consumer of `token_usage`. The error is systematic (every cached request from turn 2 onward) and invisible to the caller (no error surfaced).

### Proposed fix (prepared, happy to open a PR)

Make the `message_delta` usage authoritative for all input-side fields via event-level overwrite. Concretely:
- Branch on event type in the accumulator.
- On `message_delta`: overwrite `InputTokens`, `CacheCreationInputTokens`, `CacheReadInputTokens` from delta and reconstruct `PromptTokens = delta.input + delta.cache_creation + delta.cache_read` (event-level unconditional overwrite — not per-field `if != 0` checks, which would conflate absent-with-zero for non-pointer Go `int` fields).
- On `message_start`: preserve existing `max-keep` as the fallback for early-terminated streams.
- Add an impossible-zero guard (`PromptTokens == 0 && startSnapshot.InputTokens > 0` → restore start-side) for non-conformant Anthropic-compatible providers that emit a partial-usage `message_delta`. Dead code for spec-conformant providers.
- Skip the end-of-stream `normalizeCachedUsage` fold when `message_delta` was processed (the cache breakdown is already folded into `PromptTokens` by the overwrite).

I have a minimal, tested fix ready in a fork (2 files: `anthropic.go` + a new `streamcacheusage_test.go` with 3 fixtures, ~348 insertions / 43 deletions). New fixture `TestAnthropicStreamCacheReadNotDoubleCounted` uses the turn-2 numbers above and fails pre-fix (records 40756), passes post-fix (records 20532). Existing `core/providers/anthropic/...` tests remain green. Happy to open the PR against `dev` on request.

### Open questions for maintainers
- Should the impossible-zero guard emit a structured-log line when it fires (non-conformance detection)? The accumulator is currently logger-free; adding a logger dependency is a trade-off.
- The Anthropic Responses API path (`HandleAnthropicResponsesStream` + `accumulateAnthropicResponsesUsage`) uses a separate accumulator — out of scope for this report, but may warrant a parallel audit.

is911:
## Summary

Fixes a ~2× prompt-token inflation in Anthropic-protocol streaming usage accounting. The streaming accumulator in `HandleAnthropicChatCompletionStreaming` used a `max-keep` guard on `input_tokens` that did not branch on event type, so the authoritative `message_delta.input_tokens` (the uncached tail, smaller) was rejected in favor of the stale `message_start.input_tokens` (the full prompt). Then `normalizeCachedUsage` folded `cache_read_input_tokens` on top at stream end → `message_start.input_tokens + message_delta.cache_read_input_tokens` → double-count on every cached request from turn 2 onward.

Closes #5354.

**One-line root cause:** the accumulator merged the two Anthropic usage snapshots via `max-keep` instead of treating `message_delta.usage` as authoritative, then folded `cache_read` on top of an already-inclusive prompt total.

**Before / after formula:**
- BEFORE (buggy):  `prompt_tokens = message_start.input_tokens + message_delta.cache_read_input_tokens`
- AFTER (fixed):   `prompt_tokens = message_delta.input_tokens + message_delta.cache_creation_input_tokens + message_delta.cache_read_input_tokens`

### Evidence (3 consecutive turns, same session, provider: bifrost-kimi-anthropic / kimi-code/k3)

| Turn | `message_start` (input/cache_read) | `message_delta` (input/cache_read) | Correct total | Bifrost recorded (pre-fix) | Post-fix |
|------|-----------------------------------|-----------------------------------|---------------|----------------------------|----------|
| 1    | 20458 / 0                         | 20458 / 0                         | 20458         | 20458 ✓                    | 20458 ✓  |
| 2    | 20532 / 0                         | 308 / 20224                       | 20532         | 40756 ✗ (=20532+20224)     | 20532 ✓  |
| 3    | 20621 / 0                         | 141 / 20480                       | 20621         | 41101 ✗ (=20621+20480)     | 20621 ✓  |

Turn 1 was correct only because no cache read occurred (delta.cache_read == 0, so the buggy sum reduced to start.input). From turn 2 on, the buggy accumulator double-counted the cached portion. Reproduction signature: first request accurate, every subsequent request ~2×, growing by ~2× the real per-turn conversation growth.

## Changes

- **`core/providers/anthropic/anthropic.go`** — Extracted the event-loop usage-accumulation block into a new helper `applyStreamUsageEvent(usage, eventType, usageToProcess, startSnapshot)` that branches on the Anthropic stream event type:
  - On `message_delta`: event-level unconditional overwrite of `InputTokens`, `CacheCreationInputTokens`, `CacheReadInputTokens` from delta (NOT per-field `if != 0` checks — those would conflate absent-with-zero for non-pointer Go `int` fields); reconstruct `PromptTokens = delta.InputTokens + delta.CacheCreationInputTokens + delta.CacheReadInputTokens`.
  - On `message_start` (or any non-delta event carrying usage): preserves existing `max-keep` as the fallback for early-terminated streams.
  - Impossible-zero guard (`PromptTokens == 0 && startSnapshot.InputTokens > 0` → restore start-side values) for non-conformant Anthropic-compatible providers that emit a partial-usage `message_delta` (input=0 with cache fields omitted, deserializing to all-zeros). Dead code for spec-conformant providers — a real Anthropic prompt always has `PromptTokens >= 1`.
  - End-of-stream `normalizeCachedUsage` fold is skipped when `message_delta` was processed (`deltaProcessed` flag) — the cache breakdown is already folded into `PromptTokens` by the overwrite.
- **`core/providers/anthropic/streamcacheusage_test.go`** (new) — Three fixtures (see How to test).
- **`core/changelog.md`** — Entry at the top per contributing guide.

### Design decision / trade-off
The fix uses an **event-level overwrite** rather than per-field `if != 0` merge. Per-field checks would conflate absent-with-zero for non-pointer Go `int` fields (JSON deserialization of an omitted field produces 0), so a delta that legitimately omits `cache_read` would be indistinguishable from a delta that reports `cache_read=0`. The event-level overwrite (with the impossible-zero guard as a safety net) is the only spec-correct option for the non-pointer field shape.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [x] Core (Go)
- [ ] Transports (HTTP)
- [x] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

```sh
go version    # 1.22+ (tested on 1.26)
go test ./core/providers/anthropic/... -run TestAnthropicStream -v
go test ./core/providers/anthropic/...
go vet ./...
go build ./...
```

**New fixtures:**

1. `TestAnthropicStreamCacheReadNotDoubleCounted` — the primary regression guard. Turn-2 stream: `message_start` input=20532/cache_read=0 → `message_delta` input=308/cache_read=20224/cache_creation=0. Asserts `prompt_tokens == 20532` (i.e., 308 + 0 + 20224). Pre-fix this test FAILS recording 40756 (proves the test bites); post-fix PASSES recording 20532.
2. `TestAnthropicStreamNoCacheReadUnchanged` — turn-1 non-regression (no cache). `message_start` input=20458 → `message_delta` input=20458/cache_read=0. Asserts `prompt_tokens == 20458`. Passes both pre- and post-fix.
3. `TestAnthropicStreamUsage_NonConformantDeltaOmitsCacheFields` — proves the impossible-zero guard is load-bearing. `message_start` input=20532/cache_read=20224 → `message_delta` input=0 with cache fields omitted (→ deserialize to 0). Asserts `prompt_tokens == 20532` (start-side restored by the guard), NOT 0.

**Existing tests:** `go test ./core/providers/anthropic/...` shows zero pass/fail diff vs `dev` baseline.

## Screenshots/Recordings

N/A (no UI changes).

## Breaking changes

- [ ] Yes
- [x] No

Behavior change: `prompt_tokens` for Anthropic-protocol streams with prompt caching will be ~50% lower than before (correct instead of ~2× inflated). Any downstream consumer that calibrated against the inflated values (cost baselines, governance budgets, telemetry dashboards) will see a one-time step-change to the correct values. The corrected values match the Anthropic Messages API spec.

## Related issues

Closes #5354

## Security considerations

None. This is a usage-accounting fix; no auth, secrets, PII, or sandboxing surfaces are touched. The corrected `prompt_tokens` flows into existing cost/governance/telemetry paths unchanged.

## Checklist

- [x] I read `docs/contributing/README.md` and followed the guidelines (commit format `[fix]:`, affected packages listed, `core/changelog.md` updated)
- [x] I added/updated tests where appropriate (3 new fixtures, including a turn-2 regression guard that fails pre-fix)
- [x] I updated documentation where needed (changelog entry)
- [x] I verified builds succeed (Go) — `go build ./...`, `go vet ./...` green
- [ ] I verified the CI pipeline passes locally if applicable — CI runs on push; local `go test ./core/providers/anthropic/...` green

## Follow-up flags (out of scope for this PR)

1. **Anthropic Responses API path** (`HandleAnthropicResponsesStream` + `accumulateAnthropicResponsesUsage`) uses a separate accumulator that was NOT audited for the same bug class. Worth a parallel audit if the Responses API exhibits similar double-counting.
2. **OpenAI streaming path** — independently audited and confirmed **structurally immune** (single-snapshot emission + no cache-fold step; `normalizeCachedUsage` has 0 matches in `core/providers/openai/`). No fix needed.
3. **Bedrock provider** (`core/providers/bedrock/bedrock.go`) has its own `normalizeCachedUsage` — not audited; recommend a separate diagnostic if Bedrock routes Anthropic-protocol streams.
4. `cache_creation_input_tokens` is consistently 0 from the kimi-code provider — upstream provider quirk, not a Bifrost bug. Not fixed.

## Known limitations (provider conformance assumption)

The fix assumes Anthropic-compatible providers either (a) repeat cache fields at `message_delta` per the Anthropic Messages API spec, or (b) omit the entire `message_delta.usage` (delta-absent fallback). Non-conformant providers that emit a PARTIAL-usage `message_delta` (input_tokens=0 with cache fields omitted — deserializing to all-zeros for non-pointer Go `int`) are handled by the impossible-zero guard. The guard is dead code for spec-conformant providers (zero behavior change). The guard does not log when it fires (the accumulator is currently logger-free; a structured-log line for non-conformance detection is a possible future enhancement).

### DIFF
--- a/core/changelog.md
+++ b/core/changelog.md
@@ -1,3 +1,4 @@
+- fix: stop double-counting cache-read tokens in Anthropic streaming usage accumulator (thanks [@is911](https://github.com/is911)!)
 - feat: add Sarvam AI provider with chat, text-to-speech, and speech-to-text support (thanks [@Purvi09](https://github.com/Purvi09)!)
 - feat: add ElevenLabs sound effects (text-to-sound) support (thanks [@SecretSun](https://github.com/SecretSun)!)
 - feat: add `ProjectID` to Bedrock and Bedrock Mantle key configs with per-alias overrides for Bedrock, Bedrock Mantle, and Vertex
--- a/core/providers/anthropic/anthropic.go
+++ b/core/providers/anthropic/anthropic.go
@@ -608,6 +608,13 @@ func (provider *AnthropicProvider) ChatCompletionStream(ctx *schemas.BifrostCont
 // totals only after the full stream, so the streaming accumulator must apply the
 // same fold before billing - including on a mid-stream cancel/timeout. The += is
 // not idempotent; callers guard with a flag to apply it exactly once.
+//
+// NOTE: this fold is ONLY correct when usage.PromptTokens still holds the
+// PRELIMINARY message_start.input_tokens (the full prompt before the cache
+// breakdown is known). When applyStreamUsageEvent has already processed a
+// message_delta event, the cache breakdown is already folded into PromptTokens
+// and the caller MUST skip this fold (else it double-counts). The streaming
+// chat handler tracks this via the deltaProcessed flag.
 func normalizeCachedUsage(usage *schemas.BifrostLLMUsage) {
 	if usage == nil || usage.PromptTokensDetails == nil {
 		return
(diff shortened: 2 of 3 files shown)


akshaydeo:
@TejasGhatte have a look once

TejasGhatte:
Hey @is911 according to anthropic api input tokens field is cache exclusive. Also message delta usage counts are cumulative.
It seems like kimi endpoint is sending cache inclusive input tokens which breaks the anthropic spec.

wangkanai:
Independent confirmation from a production deployment, plus data that I think bears directly on the open question in this thread.

> @TejasGhatte: according to anthropic api input tokens field is cache exclusive. Also message delta usage counts are cumulative. It seems like kimi endpoint is sending cache inclusive input tokens which breaks the anthropic spec.

Both halves of that are right — and the conclusion I'd draw is that Bifrost should be robust to it anyway, because **this is not Kimi-specific**. I hit the identical defect on **QwenCloud**, an unrelated vendor, and the two providers deviate in exactly the same way.

Captured verbatim from `https://token-plan.<region>.maas.aliyuncs.com/apps/anthropic/v1/messages`, with Bifrost out of the path entirely:

```jsonc
// message_start — prompt-scale input_tokens, NO cache counters at all
{"type":"message_start","message":{"id":"msg_ba0b5ef1","model":"qwen3.7-plus","content":[],
 "usage":{"input_tokens":8834,"output_tokens":0}}}

// message_delta — the authoritative, cache-aware split (NOT an increment)
{"type":"message_delta","delta":{"stop_reason":"end_turn"},
 "usage":{"input_tokens":18,"output_tokens":426,"cache_creation_input_tokens":0,
          "cache_read_input_tokens":8831,"cache_creation":{"ephemeral_5m_input_tokens":0}}}
```

The non-streaming response for the same request agrees with `message_delta`: `18 + 0 + 8831 = 8849`.

For contrast, the compliant shape — MiniMax on the same code path, whose `message_start` reports `input_tokens: 0` and which therefore never triggered the bug:

```jsonc
{"type":"message_start","message":{"usage":{"input_tokens":0,"output_tokens":0,"service_tier":"standard"}}}
{"type":"message_delta","usage":{"cache_read_input_tokens":8990,"input_tokens":1,"output_tokens":2}}
```

So the trigger condition is precise and provider-shaped: **`message_start` carries a prompt-scale `input_tokens` with no cache counters, and the real split arrives only in `message_delta`.** At least two major Anthropic-dialect providers do this today. A gateway that silently doubles `prompt_tokens` against them is expensive: over one week of agent traffic, **42% of every logged prompt token was invented** — which corrupts cost attribution, budgets, rate limiting and dashboards, and inverts the apparent cache-hit rate (a real ~99.7% displayed as ~52%, sending you off hunting a provider caching problem that does not exist).

## On the undercount concern raised by the bot review

> Positive partial deltas can undercount prompt usage.

That is a real risk with "message_delta overwrites unconditionally". The variant I deployed avoids it by staying **monotone — it never lowers a value** — while still preferring the cache-aware number. An event is authoritative for `input_tokens` only if it actually carries cache counters; otherwise it is a fallback used until an authoritative event turns up. `authSeen` is derived from the accumulator itself, so the merge also stays order-independent and needs no new state:

```go
// core/providers/anthropic/anthropic.go — accumulateAnthropicResponsesUsage
authoritative := usageToProcess.CacheReadInputTokens > 0 || usageToProcess.CacheCreationInputTokens > 0
authSeen := usage.InputTokensDetails != nil &&
	(usage.InputTokensDetails.CachedReadTokens > 0 || usage.InputTokensDetails.CachedWriteTokens > 0)
if authoritative || (!authSeen && usageToProcess.InputTokens > usage.InputTokens) {
	usage.InputTokens = usageToProcess.InputTokens
	if billedUsage != nil {
		billedUsage.PromptTokens = usageToProcess.InputTokens
	}
}
```

Compliant upstreams are provably unaffected: real Anthropic puts the cache counters on `message_start`, so that frame is authoritative and the later `message_delta` (output only, no cache keys) cannot clobber it.

## Production verification

Deployed to a live v1.6.11 gateway. Identical bytes, `stream` the only variable — streaming now agrees with non-streaming exactly on every provider tested:

| provider | `prompt_tokens` before | after | non-streaming truth |
|---|---:|---:|---:|
| QwenCloud `qwen3.7-plus` | 17,669 | **8,849** | 8,849 |
| Moonshot `kimi-k3` | 17,862 | **8,932** | 8,932 |
| MiniMax `M3` (control) | 8,992 | **8,992** | 8,992 |

Organic traffic at ~300K-token prompts went from `prompt_tokens ≈ 2 × cached_read` to a residual of 411–1,369 tokens, matching the providers that were always correct. Full `providers/anthropic`, `providers/azure`, `providers/utils` and `schemas` suites pass.

## One thing this PR does not cover

There is a **second** accumulator with the same flaw: `AnthropicPassthroughStreamUsage.ObserveEvent` in `core/providers/anthropic/passthrough_usage.go`, tracked separately as #5510. It max-merges `input_tokens` across events in the same way. I verified they are independent — deploying a fix to only the passthrough one changed nothing for my traffic (17,667 → 17,669), which is how I found that the native accumulator here was the one in play. Whichever approach lands, both sites need it.

Happy to open a PR for the passthrough half, or to rebase this one — it has been merge-conflicted since 08-19.

PR #5356, opened 2026-07-18
[fix]: tracing - concurrent requests sharing a W3C traceID no longer lose log rows

### LINKED ISSUE #5256: [Bug]: Concurrent HTTP requests sharing a W3C trace ID lose LLM log rows
### Prerequisites

- [x] Reproduced on latest official Docker image: `maximhq/bifrost:latest`, runtime version `v1.6.4`, digest `sha256:5f1fed63b5c2c78c51f25d80d5d2d683d79f74587ea0906e8c03fbd256ab52e0`.

## Description

When multiple concurrent HTTP requests carry the same valid W3C trace ID, Bifrost handles every request successfully but **persists only some LLM log rows**.

Each request has:

- the same 32-hex-character trace ID;
- a unique parent span ID in `traceparent`;
- a unique `x-request-id`;
- a unique metadata request index.

With 18 requests and concurrency 10, all 18 return HTTP 200. On latest Bifrost v1.6.4, only 10 of 18 appear in `/api/logs` and the LLM Logs UI. Repeated tests on v1.5.7 persisted 9 or 10 of 18.

This causes request counts, token usage, and cost totals to be underreported.

No Dash0 collector or OpenTelemetry SDK is involved in this reproducer. It sends raw `traceparent` headers directly to Bifrost and verifies Bifrost's own Logs API.

This looks related to, but distinct from:

- [#3003](https://github.com/maximhq/bifrost/issues/3003), which covers orphaned pending logs after client disconnects;
- [#5095](https://github.com/maximhq/bifrost/issues/5095), which covers logs missing after the pending-entry TTL expires;
- [#2999](https://github.com/maximhq/bifrost/issues/2999), which covers WebSocket trace completion.

## Suspected cause

Current `main` still keys active traces only by the inherited trace ID:

- [`TraceStore.CreateTrace` stores by `trace.TraceID`](https://github.com/maximhq/bifrost/blob/c0909f9752156121c6c775694df6e656a6ad3860/framework/tracing/store.go#L75-L112).
- [`TraceStore.CompleteTrace` uses `LoadAndDelete(traceID)`](https://github.com/maximhq/bifrost/blob/c0909f9752156121c6c775694df6e656a6ad3860/framework/tracing/store.go#L151-L162).
- [The logging plugin also groups and drains pending entries by trace ID](https://github.com/maximhq/bifrost/blob/c0909f9752156121c6c775694df6e656a6ad3860/plugins/logging/main.go#L1328-L1367).

Concurrent sibling requests therefore overwrite one active trace. The first completion can remove the shared trace and drain only the log entries accumulated at that moment. Later completions cannot inject remaining entries. This is a hypothesis based on the observed behavior and current source, not a confirmed root cause.

An internal per-request key, such as trace ID plus request ID or root span ID, would preserve distributed-trace correlation without treating the trace ID as a unique request key.

### Steps to reproduce

This reproducer uses short, successful, non-streaming HTTP requests. It does not rely on a disconnect, WebSocket, or TTL expiry.

Run Bifrost with logging enabled and at least one configured chat-completion model.

```python
import asyncio
from collections import Counter
import json
import os
import secrets

import httpx


BASE_URL = os.getenv("BIFROST_BASE_URL", "http://localhost:8080").rstrip("/")
MODEL = os.environ["BIFROST_MODEL"]
VIRTUAL_KEY = os.environ["BIFROST_VK"]
REQUESTS = 18
CONCURRENCY = 10


async def send(
    client: httpx.AsyncClient,
    semaphore: asyncio.Semaphore,
    trace_id: str,
    index: int,
) -> httpx.Response:
    headers = {
        "Authorization": f"Bearer {VIRTUAL_KEY}",
        "x-bf-vk": VIRTUAL_KEY,
        "traceparent": f"00-{trace_id}-{secrets.token_hex(8)}-01",
        "x-request-id": f"shared-trace-repro-{index:02d}",
    }
    body = {
        "model": MODEL,
        "messages": [{"role": "user", "content": f"Reply only with {index}."}],
        "max_tokens": 1,
        "temperature": 0,
        "metadata": {"repro_id": trace_id, "request_index": str(index)},
    }
    async with semaphore:
        return await client.post(
            f"{BASE_URL}/v1/chat/completions",
            headers=headers,
            json=body,
        )


async def main() -> None:
    trace_id = secrets.token_hex(16)
    semaphore = asyncio.Semaphore(CONCURRENCY)

    async with httpx.AsyncClient(timeout=90) as client:
        responses = await asyncio.gather(
            *(send(client, semaphore, trace_id, index) for index in range(REQUESTS))
        )
        statuses = Counter(response.status_code for response in responses)

        # Default log writer batch interval is 5 seconds.
        await asyncio.sleep(7)
        log_response = await client.get(
            f"{BASE_URL}/api/logs",
            params={
                "limit": 100,
                "offset": 0,
                "sort_by": "timestamp",
                "order": "desc",
                "period": "1h",
            },
        )
        log_response.raise_for_status()

    logs = log_response.json()["logs"]
    persisted = sum(trace_id in json.dumps(log) for log in logs)
    print(f"shared trace ID: {trace_id}")
    print(f"HTTP statuses: {dict(statuses)}")
    print(f"persisted logs: {persisted}/{REQUESTS}")


asyncio.run(main())
```

### Expected behavior

All 18 successful requests persist as 18 independent LLM logs, even though they are sibling requests in one distributed trace.

### Actual behavior

Latest-version test:

```text
Bifrost runtime version: v1.6.4
Docker image: maximhq/bifrost:latest
Image digest: sha256:5f1fed63b5c2c78c51f25d80d5d2d683d79f74587ea0906e8c03fbd256ab52e0
shared trace ID: f875c9095ab398fe46556d10939099ae
requests: 18, concurrency: 10, HTTP statuses: {200: 18}
persisted logs: 10/18
```

Earlier v1.5.7 tests produced 10/18 and 9/18 persisted logs. The varying count suggests a concurrency race.

### Affected area(s)

Framework, Transports (HTTP), Plugins

### Version

latest: v1.6.4 (`maximhq/bifrost:latest` at test time) - Also reproduced: v1.5.7

### Environment

```text
- Docker Desktop on macOS
- OpenAI-compatible `POST /v1/chat/completions`
- Non-streaming requests
- AWS Bedrock-backed Claude Haiku model alias
- 18 requests, concurrency 10, `max_tokens: 1`
- Logging enabled with default batch interval
```

### Severity

High (major functionality broken). Requests complete successfully, but observability and billing/cost reporting silently lose data.

AdityaPainuli:
## Summary

Concurrent HTTP requests carrying the same W3C `traceparent` trace ID (normal for sibling calls of one distributed trace) lose LLM log rows. All requests return 200, but only some appear in `/api/logs`, so request counts, token usage, and cost totals silently underreport.

Root cause: `TraceStore` keyed active traces by the trace ID alone. `CreateTrace` did `traces.Store(traceID, ...)`, so each concurrent sibling overwrote the previous entry, and `CompleteTrace`'s `LoadAndDelete(traceID)` handed the entry to whichever request completed first. Every other sibling got nil back and flushed nothing. `deferredSpans` and the stream accumulator were keyed the same way, so streaming requests hit the same collision.

Reproduced at store level with the issue's shape (18 concurrent create/complete cycles, one shared trace ID, unique request IDs): 4/18 flushed, 14 lost. Control with unique trace IDs: 18/18. After the fix: 18/18.

## Changes

* `TraceStore.CreateTrace` now stores each trace under a unique per-request key (`Trace.InternalID`, new field) and returns that key. Every consumer already treats the returned value as an opaque handle carried through context, so no call sites change.
* `trace.TraceID` keeps the inherited W3C value, span export and distributed-trace linking are untouched (OTEL converter reads `trace.TraceID`, verified `span.TraceID` is never read outside tests).
* Logging plugin `Inject` joins pending entries by `InternalID` (this was the second half of the row loss), with a `TraceID` fallback for traces created by paths that predate the field.
* `x-bifrost-trace-id` response header still exposes the W3C trace ID (that is what Tempo/Grafana index), not the internal key.
* Deferred spans and stream accumulators ride the new key automatically, fixing the streaming variant.
* New regression test: 18 concurrent requests sharing one inherited trace ID, asserts every `CompleteTrace` returns the correct request's trace with the shared W3C ID preserved. Passes with `-race`.
* Updated three existing tests that asserted the old "returned value == W3C trace ID" contract, with comments documenting the new one.

## Type of change

- [X] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [X] Core (Go)
- [X] Transports (HTTP)
- [ ] Providers/Integrations
- [X] Plugins
- [ ] UI (React)
- [ ] Docs

(core/schemas, framework/tracing, plugins/logging, HTTP middleware)

## How to test

```sh
cd framework
go test ./tracing/ -count=1
go test ./tracing/ -count=1 -race -run TestCreateTrace_ConcurrentSharedInheritedTraceID

cd ../plugins/logging && go test ./...
cd ../otel && go test ./...
```

End-to-end: send N concurrent requests with the same traceparent trace ID and unique parent span IDs, then check /api/logs, all N rows should appear.

## Breaking changes

- [X] No

The value flowing through BifrostContextKeyTraceID is now an internal handle rather than the W3C trace ID. Everything in this repo treats it as opaque and passes it back into tracer methods, verified by sweeping all consumers. One thing reviewers should check: if the enterprise overlay has a connector that joins its own pending state on trace.TraceID the way the logging plugin did, it should move to InternalID too (Inject receives it on the trace snapshot).

## Related issues

Closes maximhq/bifrost#5256

## Security considerations

None. Trace IDs were never used for auth; the internal key is random and revealed only via context within the process.

## Checklist

- [X] I read `docs/contributing/README.md` and followed the guidelines
- [X] I added/updated tests where appropriate
- [X] I updated documentation where needed
- [X] I verified builds succeed (Go and UI)
- [X] I verified the CI pipeline passes locally if applicable

### DIFF
--- a/core/schemas/trace.go
+++ b/core/schemas/trace.go
@@ -11,7 +11,8 @@ import (
 // Trace represents a distributed trace that captures the full lifecycle of a request
 type Trace struct {
 	RequestID             string            // Request ID for the trace
-	TraceID               string            // Unique identifier for this trace
+	TraceID               string            // Exported trace identifier; inherited from the W3C traceparent header when present, so concurrent requests of one distributed trace may share it
+	InternalID            string            // Unique per-request handle the trace is stored under; unlike TraceID it is never shared across requests
 	ParentID              string            // Parent trace ID from incoming W3C traceparent header
 	RootSpan              *Span             // The root span of this trace
 	Spans                 []*Span           // All spans in this trace
@@ -171,6 +172,7 @@ func (t *Trace) SnapshotForExport() *Trace {
 	clone := &Trace{
 		RequestID:      t.RequestID,
 		TraceID:        t.TraceID,
+		InternalID:     t.InternalID,
 		ParentID:       t.ParentID,
 		StartTime:      t.StartTime,
 		EndTime:        t.EndTime,
@@ -254,6 +256,7 @@ func (t *Trace) Reset() {
 	defer t.mu.Unlock()
 	t.RequestID = ""
 	t.TraceID = ""
+	t.InternalID = ""
 	t.ParentID = ""
 	t.RootSpan = nil
 	for i := range t.Spans {
--- a/framework/tracing/store.go
+++ b/framework/tracing/store.go
@@ -67,19 +67,27 @@ func NewTraceStore(ttl time.Duration, logger schemas.Logger) *TraceStore {
 	return store
 }
 
-// CreateTrace creates a new trace and stores it, returns trace ID only.
+// CreateTrace creates a new trace and stores it, returns the trace's store key.
 // The inheritedTraceID parameter is the trace ID from an incoming W3C traceparent header.
-// If provided, this trace will use that ID to continue the distributed trace.
-// If empty, a new trace ID will be generated.
+// If provided, the trace's exported TraceID uses that value to continue the
+// distributed trace. If empty, a new trace ID is generated.
+//
+// The returned key is unique per call, NOT the (possibly shared) W3C trace ID:
+// concurrent sibling requests of one distributed trace legitimately carry the
+// same inherited trace ID, and keying the store by it made siblings overwrite
+// each other — the first CompleteTrace stole the entry and the rest flushed
+// nothing (lost log rows, issue #5256). Callers treat the returned value as an
+// opaque handle; the W3C ID lives on trace.TraceID for span export and linking.
 // Note: The parent span ID (for linking to upstream spans) is handled separately
 // via context in StartSpan, not stored on the trace itself.
 func (s *TraceStore) CreateTrace(inheritedTraceID string, requestID ...string) string {
 	trace := s.tracePool.Get().(*schemas.Trace)
 	// Reset and initialize the trace
+	trace.InternalID = generateTraceID()
 	if inheritedTraceID != "" {
 		trace.TraceID = inheritedTraceID
 	} else {
-		trace.TraceID = generateTraceID()
+		trace.TraceID = trace.InternalID
 	}
 	// Note: trace.ParentID is intentionally not set here.
 	// Parent-child relationships are between spans, not traces.
@@ -109,8 +117,8 @@ func (s *TraceStore) CreateTrace(inheritedTraceID string, requestID ...string) s
 	// Reset request headers
 	trace.RequestHeaders = nil
 
-	s.traces.Store(trace.TraceID, trace)
-	return trace.TraceID
+	s.traces.Store(trace.InternalID, trace)
+	return trace.InternalID
 }
 
 // GetTrace retrieves a trace by ID
--- a/framework/tracing/store.go
+++ b/framework/tracing/store.go
@@ -286,9 +286,11 @@ func (s *TraceStore) StartSpan(traceID, name string, kind schemas.SpanKind) *sch
 
 	span := s.spanPool.Get().(*schemas.Span)
 
-	// Reset and initialize the span
+	// Reset and initialize the span. Span.TraceID carries the exported W3C
+	// trace identity (shared across sibling requests of a distributed trace);
+	// store lookups use the unique key passed as traceID, never this field.
 	span.SpanID = generateSpanID()
-	span.TraceID = traceID
+	span.TraceID = trace.TraceID
 	span.Name = name
 	span.Kind = kind
 	span.StartTime = time.Now()
@@ -333,10 +335,11 @@ func (s *TraceStore) StartChildSpan(traceID, parentSpanID, name string, kind sch
 
 	span := s.spanPool.Get().(*schemas.Span)
 
-	// Reset and initialize the span
+	// Reset and initialize the span. As in StartSpan, Span.TraceID carries the
+	// exported W3C trace identity, not the store key.
 	span.SpanID = generateSpanID()
 	span.ParentID = parentSpanID
-	span.TraceID = traceID
+	span.TraceID = trace.TraceID
 	span.Name = name
 	span.Kind = kind
 	span.StartTime = time.Now()
--- a/framework/tracing/store_test.go
+++ b/framework/tracing/store_test.go
@@ -169,10 +169,10 @@ func TestStartChildSpan_WithExternalParentSpanID(t *testing.T) {
 		t.Errorf("root span.ParentID = %q, want external parent %q", rootSpan.ParentID, externalParentSpanID)
 	}
 
-	// span.TraceID is internal linkage to the store entry (the per-request key),
-	// not the W3C ID — exporters read the W3C ID from trace.TraceID.
-	if rootSpan.TraceID != traceID {
-		t.Errorf("root span.TraceID = %q, want store key %q", rootSpan.TraceID, traceID)
+	// Span.TraceID carries the exported W3C identity even though the store
+	// key returned by CreateTrace is a distinct per-request handle.
+	if rootSpan.TraceID != inheritedTraceID {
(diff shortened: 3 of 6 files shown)


akshaydeo:
@AdityaPainuli in which case trace id would be same? Ideally that setup seems wrong in the first place and we are fixing the issue that ideally should not be there in the first place

AdityaPainuli:
@akshaydeo This is standard OTEL/W3C behavior, not a broken setup. Trace ID = whole transaction, span ID = one call within it. 

So when an instrumented app (agent doing parallel tool calls, RAG doing embeddings + chat, etc.) fans out into multiple LLM calls for one user action, every request hits Bifrost with the same trace ID and a different span ID. That's exactly what makes them show up as one trace in Tempo/Datadog.

The issue's repro (18 requests, same trace ID, unique span IDs) is just simulating that fan-out. Today each sibling overwrites the previous one in the store, hence the missing log rows. The fix keeps the shared ID for export, only the internal store key becomes per-request.

akshaydeo:
Sir I know whats the difference between a span id and a trace id. What im saying is - if its w3c path - and parent spanid in the header is not present we generate a span id [https://github.com/maximhq/bifrost/blob/d60f2969451a77de1d05ae9db243c5643f12e656/framework/tracing/store.go#L282.](<https://github.com/maximhq/bifrost/blob/d60f2969451a77de1d05ae9db243c5643f12e656/framework/tracing/store.go#L282.>) Did you read up the code and were able to reproduce what you are fixing?<br><br>can you share a quick video of the same so I understand whats the gap here?

AdityaPainuli:
Sure, I'll put together concrete evidence ~~first thing tomorrow~~. Reproduction of the issue against the current dev and a video comparing it to my branch.

AdityaPainuli:
The case: any OTEL-instrumented app making parallel calls inside one traced operation. The SDK propagates the same trace ID on all of them and mints a unique span ID per call. 

On reproduction: reproduced it against current dev  (d60f296). 18 concurrent requests with that traffic shape, 17 lost their trace (CompleteTrace returns nil, so nothing reaches Inject, no log row), and the one that flushed contained another request's span. The span IDs are unique as mentioned, the collision is the traces map itself: [L112](https://github.com/maximhq/bifrost/blob/d60f2969451a77de1d05ae9db243c5643f12e656/framework/tracing/store.go#L112.) by trace.TraceID so siblings evict each other, then [L151](https://github.com/maximhq/bifrost/blob/d60f2969451a77de1d05ae9db243c5643f12e656/framework/tracing/store.go#L151.) LoadAndDelete gives the entry to the first completer. 

Here's the recording I can commit the demo test case I also prepared lmk if required.

https://github.com/user-attachments/assets/bc3f8839-3dfb-4989-b0ac-d791da275431

akanz1:
Thanks, @AdityaPainuli! any updates here and when we can expect this fix to be released?

AdityaPainuli:
All changes related to fix already done. Once the review is done by @akshaydeo I think we can merge this.

roroghost17:
Hey @AdityaPainuli, I went through the PR and it looks good. I will just be adding a small update on top of it that is required and will merge it together. Can you please just rebase this once and ping me. Thanks

AdityaPainuli:
@roroghost17 rebased to current dev; the branch is now linear (the two fixed commits on top of dev tip). All yours, thanks for picking it up, and lmk if anything else is needed from my end.

akshaydeo (approved):

PR #5377, opened 2026-07-20
fix: exclude unresolved VKs from scope ID set and normalize empty routing rule scope to `global`

Pratham-Mishra04:
## Summary

Fixes two related bugs in governance config file merging: (1) brand-new virtual keys whose secret-backed value fails to resolve were being counted as valid `scope_id` targets even though they never received a DB row, allowing routing rules to pass sanitization while pointing at a virtual key that doesn't exist; and (2) routing rules with an empty `scope` field were being persisted verbatim with `scope=""` instead of being normalized to `"global"`, causing them to silently match zero requests at runtime.

## Changes

- Introduces `skippedNewVirtualKeyIDs` to track brand-new virtual keys that were skipped due to an unresolvable secret reference. These IDs are now excluded from the `vkScopeIDs` set used to validate routing rule `scope_id` targets, so a routing rule cannot pass sanitization by referencing a virtual key that was never written to the DB.
- Adds a normalization pass over routing rules before sanitization: any rule with `scope=""` is coerced to `scope="global"`, and any rule with `scope="global"` has its `scope_id` cleared to prevent dangling references.
- Introduces `routingScopeResolves` as a named helper to centralize the check for whether a rule's scope target will still exist after the current sync. This is now used both to reject an incoming file rule and to validate any fallback-to-persisted candidate — a persisted rule whose own `scope_id` is about to be pruned by this same sync is removed rather than restored, preventing a dangling-rule bug one level down.
- Moves the `mergeGovernanceConfig` comment to sit directly above the function it describes (it had drifted above `entityIDSet`).

## Type of change

- [x] Bug fix

## Affected areas

- [x] Transports (HTTP)

## How to test

```sh
go test ./...
```

- Configure a virtual key in `config.json` whose `Value` references an env/vault secret that cannot be resolved, then define a routing rule targeting that virtual key's ID. Verify the routing rule is rejected and does not appear in the persisted config.
- Configure a routing rule in `config.json` with `scope` omitted or set to `""`. Verify it is persisted with `scope="global"` and no `scope_id`.
- Configure a routing rule targeting a team or virtual key that is being removed in the same sync. Verify the rule is removed rather than restored from the prior persisted version.

## Breaking changes

- [ ] Yes
- [x] No

## Security considerations

Prevents routing rules from referencing virtual keys that were never persisted due to unresolvable secret references, which could otherwise cause unexpected routing behavior or expose unintended fallback paths.

## Checklist

### DIFF
--- a/transports/bifrost-http/lib/config.go
+++ b/transports/bifrost-http/lib/config.go
@@ -2335,7 +2335,6 @@ func resolveGovernanceKeyReferences(ctx context.Context, config *Config, governa
 	})
 }
 
-// mergeGovernanceConfig merges governance config from file with store
 // entityIDSet builds a lookup set of IDs from an already-synced slice plus a
 // slice of entries still pending insertion, so newly-added-in-this-file entities
 // are valid scope_id targets even before they're persisted.
@@ -2350,6 +2349,7 @@ func entityIDSet[T any](existing []T, id func(T) string, toAdd []T) map[string]b
 	return set
 }
 
+// mergeGovernanceConfig merges governance config from file with store
 func mergeGovernanceConfig(ctx context.Context, config *Config, configData *ConfigData, governanceConfig *configstore.GovernanceConfig) {
 	logger.Debug("merging governance config from config file with store")
 	// When config.json is the source of truth, file-present entities must be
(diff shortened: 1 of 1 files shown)


Pratham-Mishra04:
* **#5377** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/5377?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/maximhq/bifrost/5377?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `dev`


<h2></h2>

Pratham-Mishra04:
### Merge activity

* **Jul 20, 9:27 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/5377).
* **Jul 20, 9:28 AM UTC**: @Pratham-Mishra04 merged this pull request with [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/5377).

PR #5384, opened 2026-07-20
fix: gate matview read path on shape check to prevent "column does not exist" during rolling deploys

impoiler:
## Summary

During a rolling deploy, a replica that loses the `ensureMatViews` advisory lock race could enable the matview read path while the lock-holding replica was still mid-repair. This caused queries like `getModelHistogramFromMatView` to hit columns (`cancelled_count`, `throughput_*`) that the old-schema view did not yet have, producing "column does not exist" errors.

## Changes

- `ensureMatViews` now returns `(bool, error)` — the bool indicates whether this process actually performed the create/repair work. A `false` return with a `nil` error means another replica held the lock; callers must not treat this as "views are usable."
- Added `matViewShapesReady`, a read-only `pg_catalog` check that verifies every managed matview exists and carries the full column set this build requires. It is safe and cheap to call from replicas that skipped the lock, and is used as the gate before enabling the matview read path.
- Added `pqTextArray`, a minimal helper that renders a `[]string` as a Postgres `text[]` literal for use as a query parameter, avoiding a driver-specific array type dependency. Backslashes and double-quotes are escaped correctly.
- In `newPostgresLogStore`, when `ensureMatViews` returns `false` (lock held elsewhere), the startup path now calls `matViewShapesReady` before enabling the read path. If the shapes are not current yet, it starts the refresher's recovery path instead of flipping `matViewsReady` immediately.
- In `startMatViewRefresher`, a successful `refreshMatViews` tick no longer unconditionally flips `matViewsReady`. It now calls `matViewShapesReady` first, since `REFRESH` succeeds on an old-shape view and `refreshMatViews` also returns `nil` when it skipped due to the lock or activity gate.
- Added `pg_catalog.pg_table_is_visible` filters to the existence and column-listing queries in `matViewNeedsRebuild` to avoid matching same-named views in other schemas.
- Updated all call sites and tests to handle the new `(bool, error)` signature.

## Type of change

- [x] Bug fix

## Affected areas

- [x] Core (Go)

## How to test

```sh
go test ./framework/logstore/...
```

The existing `TestEnsureMatViewsSharesRefreshAdvisoryLock` test now asserts the `bool` return value in both the lock-held and lock-free cases. The scale migration test and perf tests have been updated to match the new signature.

To validate the rolling-deploy scenario specifically: run two instances sharing the same Postgres database against a schema that is missing the new matview columns, confirm that the replica that loses the lock race does not enable the matview read path until `matViewShapesReady` returns true.

## Breaking changes

- [x] No

Internal function signature change only (`ensureMatViews` return type). No exported API is affected.

## Security considerations

`pqTextArray` escapes backslashes and double-quotes. Column names passed to it come from package-level constants, not user input, but the escaping is correct regardless.

## Checklist

### DIFF
--- a/framework/logstore/matviews.go
+++ b/framework/logstore/matviews.go
@@ -78,7 +78,12 @@ ON mv_logs_hourly (hour, provider, model, status, object_type, selected_key_id, 
 
 // mvLogsHourlyRequiredColumns is the canonical column set used by
 // repairMatViewShapes to detect old-shape mv_logs_hourly views from prior
-// schema versions and drop them so they get rebuilt on startup.
+// schema versions and drop them so they get rebuilt on startup, and by
+// matViewShapesReady to gate the matview read path.
+//
+// Must mirror every output column of mvLogsHourlyDDL. A partial list would let
+// a view missing an unlisted column pass both checks, and readers selecting it
+// would fail with "column does not exist" — see TestMvLogsHourlyRequiredColumnsMatchDDL.
 var mvLogsHourlyRequiredColumns = []string{
 	"hour",
 	"provider",
@@ -94,10 +99,22 @@ var mvLogsHourlyRequiredColumns = []string{
 	"business_unit_id",
 	"alias",
 	"canonical_model_name",
+	"count",
+	"success_count",
+	"error_count",
 	"cancelled_count",
+	"avg_latency",
+	"p90_latency",
+	"p95_latency",
+	"p99_latency",
+	"total_prompt_tokens",
+	"total_completion_tokens",
 	"throughput_completion_tokens",
 	"throughput_latency_ms",
 	"throughput_request_count",
+	"total_tokens",
+	"total_cached_read_tokens",
+	"total_cost",
 }
 
 // legacyMatViewNames are matviews from previous schema versions that no longer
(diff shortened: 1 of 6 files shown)


impoiler:
* **#5397** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/5397?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5384** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/5384?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/maximhq/bifrost/5384?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `dev`


<h2></h2>

Pratham-Mishra04 (approved):

Pratham-Mishra04:
### Merge activity

* **Jul 21, 7:38 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/5384).
* **Jul 21, 7:39 AM UTC**: [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/5384) rebased this pull request as part of a merge.
* **Jul 21, 7:39 AM UTC**: @Pratham-Mishra04 merged this pull request with [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/5384).

PR #5457, opened 2026-07-22
fix: resolve model aliases in Azure passthrough requests

hugochinchilla:
## Problem

Given an Azure key config with aliases:

```yaml
providers:
  azure:
    keys:
      - name: foundry-swedencentral-1
        value: env.AZURE_KEY
        models:
          - mistral-document-ai-2512-1-swedencentral-14
          - mistral-ocr-3-0
        aliases:
          mistral-ocr-3-0: mistral-document-ai-2512-1-swedencentral-14
```

calling `/azure_passthrough` with the alias name fails:

```
POST /azure_passthrough/providers/mistral/azure/ocr?api-version=2025-03-01-preview
→ 404 DeploymentNotFound: "mistral-ocr-3-0 does not exist"
```

`AzureProvider.Passthrough`/`PassthroughStream` forward `req.Path` and `req.Body` verbatim, so the user-facing alias name (in the `/openai/deployments/{name}` path segment or the top-level `"model"` body field) reaches Azure unresolved. Core already resolves each attempt's key alias into the request context (`ResolvedAlias`) and rotates the key pool for passthrough requests — the provider just never applied it to the wire request.

Same problem is described in #5458 

## Fix

New `resolvePassthroughAlias(ctx, path, body)` in `core/providers/azure/utils.go`, called from both `Passthrough` and `PassthroughStream`:

- rewrites the `/deployments/{alias}` path segment to the alias `model_id` (case-insensitive, matching `KeyAliases.ResolveConfig` semantics)
- rewrites the top-level `"model"` JSON body field when it equals the alias

It follows the same ctx-based pattern as the existing `resolveAzureEndpoint` / `resolveAPIVersion` per-alias overrides, which the passthrough path already honors. No-op when no alias matched — clients sending real deployment names get untouched requests. Non-JSON/multipart bodies pass through unchanged (those routes carry the deployment in the path). Usage extraction and logging still see the user-facing model name.

## Load balancing

This also makes load balancing work for aliased models through passthrough: the key pool already filters by the requested name against each key's `models` list and rotates with weights. Since the `ResolvedAlias` in ctx is refreshed per attempt for the selected key, multiple keys sharing an alias (e.g. `mistral-ocr-3-0` on Sweden Central and East US resources) each rewrite to their own regional deployment — including on rate-limit retries.

## Tests

`TestResolvePassthroughAlias` (10 subtests): path rewrite mid-path / end-of-path / case-insensitive, body rewrite preserving sibling fields, no-op guards (no alias, different model, no model field, non-JSON body), and per-key resolution to distinct regional deployments.

Full `core/providers/azure` suite passes; `go build ./...` and `go vet` clean.

Note: the OpenAI, Anthropic, and Gemini passthrough endpoints share this bug class (model in body for OpenAI/Anthropic, in path for Gemini). Kept out of scope here since each needs its own verification (multipart model field for OpenAI audio, `:action` suffix for Gemini) — happy to follow up in a separate PR.

### DIFF
--- a/core/providers/azure/utils.go
+++ b/core/providers/azure/utils.go
@@ -1,11 +1,12 @@
 package azure
 
 import (
-	"encoding/json"
+	"bytes"
 	"strings"
 
-	"github.com/bytedance/sonic"
 	schemas "github.com/maximhq/bifrost/core/schemas"
+	"github.com/tidwall/gjson"
+	"github.com/tidwall/sjson"
 )
 
 // getAzureScopes returns the configured scopes or the default scope if none are valid.
@@ -106,24 +107,16 @@ func rewriteDeploymentSegment(path, alias, modelID string) string {
 // untouched. ponytail: top-level JSON only — multipart passthrough routes
 // carry the deployment in the path, not the body.
 func rewriteBodyModel(body []byte, alias, modelID string) []byte {
-	if len(body) == 0 {
+	// JSON objects only: gjson/sjson do not validate, and a multipart body
+	// carrying a "model" form field would otherwise get corrupted.
+	if trimmed := bytes.TrimLeft(body, " \t\r\n"); len(trimmed) == 0 || trimmed[0] != '{' {
 		return body
 	}
-	var m map[string]json.RawMessage
-	if err := sonic.Unmarshal(body, &m); err != nil {
+	current := gjson.GetBytes(body, "model")
+	if current.Type != gjson.String || !strings.EqualFold(current.Str, alias) {
 		return body
 	}
-	var current string
-	raw, ok := m["model"]
-	if !ok || sonic.Unmarshal(raw, &current) != nil || !strings.EqualFold(current, alias) {
-		return body
-	}
-	quoted, err := sonic.Marshal(modelID)
-	if err != nil {
-		return body
-	}
-	m["model"] = quoted
-	out, err := sonic.Marshal(m)
+	out, err := sjson.SetBytes(body, "model", modelID)
 	if err != nil {
 		return body
 	}
--- a/core/providers/azure/utils.go
+++ b/core/providers/azure/utils.go
@@ -1,8 +1,10 @@
 package azure
 
 import (
+	"encoding/json"
 	"strings"
 
+	"github.com/bytedance/sonic"
 	schemas "github.com/maximhq/bifrost/core/schemas"
 )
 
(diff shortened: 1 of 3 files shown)


hugochinchilla:
This PR closes #5458

core/providers/azure/utils.go
```diff
@@ -65,3 +67,65 @@ func resolveAzureEndpoint(ctx *schemas.BifrostContext, key schemas.Key) string {
...
+	if len(body) == 0 {
+		return body
+	}
+	var m map[string]json.RawMessage
+	if err := sonic.Unmarshal(body, &m); err != nil {
+		return body
+	}
+	var current string
+	raw, ok := m["model"]
+	if !ok || sonic.Unmarshal(raw, &current) != nil || !strings.EqualFold(current, alias) {
```
akshaydeo: please use gjson and sjson here to avoid parsing entire body

hugochinchilla: Done — switched to `gjson`/`sjson` in 9f7dc1f, no full-body unmarshal now. Added a leading-`{` guard since they do not validate JSON, so multipart bodies stay untouched.

core/providers/azure/utils.go
```diff
@@ -65,3 +68,57 @@ func resolveAzureEndpoint(ctx *schemas.BifrostContext, key schemas.Key) string {
...
+	// JSON objects only: gjson/sjson do not validate, and a multipart body
+	// carrying a "model" form field would otherwise get corrupted.
+	if trimmed := bytes.TrimLeft(body, " \t\r\n"); len(trimmed) == 0 || trimmed[0] != '{' {
+		return body
+	}
+	current := gjson.GetBytes(body, "model")
+	if current.Type != gjson.String || !strings.EqualFold(current.Str, alias) {
+		return body
+	}
+	out, err := sjson.SetBytes(body, "model", modelID)
```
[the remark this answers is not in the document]
hugochinchilla: I checked this against the actual library behaviour rather than assuming, and I'm going to skip `ValidBytes` — but the comment did surface something worth pinning, so thanks.

What `gjson` v1.18.0 / `sjson` v1.2.5 actually do:

| input | `gjson.GetBytes(_, "model")` | `sjson.SetBytes(_, "model", "DEP")` |
|---|---|---|
| `{"model":"a"} trailing` | `"a"` | `{"model":"DEP"} trailing` |
| `{"model":"a", "x":` | `"a"` | `{"model":"DEP", "x":` |
| multipart w/ `name="model"` | Null | **`{"model":"DEP"}`** |

On the valid-prefix-plus-trailing-data case the rewrite is faithful: it swaps the alias and leaves the malformed remainder malformed. Azure returns the same 400 parse error before and after, so `ValidBytes` buys no change in outcome — only a full O(n) scan of every passthrough body. That matters here specifically: the route this PR exists for is `/providers/mistral/azure/ocr`, whose bodies are base64-encoded PDFs, and the review ask on this line was to stop scanning the whole body.

The genuinely destructive row is the third one — on a non-JSON input `sjson.SetBytes` discards the body and returns a bare `{"model":"DEP"}`, which on `/audio/transcriptions` would throw away the uploaded file. That is already blocked twice over (leading-`{` guard, and `gjson` returns Null anyway), but it had no test. Added in 95dddec: the non-JSON subtest now uses a realistic multipart body with a `name="model"` form field carrying the alias, so the guard fails loudly if anyone relaxes it.

The `json.Unmarshal(gotBody, &...)` validation in `TestResolvePassthroughAlias` is untouched.

akshaydeo:
Hi @hugochinchilla — thanks for the contribution! This PR is currently blocked because our CLA bot shows the Contributor License Agreement as not yet signed. Could you sign it here so we can move this forward: https://cla-assistant.io/maximhq/bifrost?pullRequest=5457

hugochinchilla:
Hi @akshaydeo I already signed the CLA, the bot messaged me about this blocked status but because it does not detect you and @Pratham-Mishra04 as signers as you figure in the PR as contributors for some reason. 

imagen

I guess I should rebase the branch on dev.

PR #5462, opened 2026-07-22
fix(vertex): support API-key/context-header auth in cached content methods

TransactCharlie:
## Summary

The Vertex cached content methods (create / list / retrieve / update / delete) all authenticate through the shared `vertexAuthHeaders` helper, which unconditionally fetched an OAuth token from the key credentials and set it on the `Authorization` header. Because `SetExtraHeaders` runs first and `vertexAuthHeaders` runs after, any `Authorization` header a caller injected via context extra headers was overwritten.

This is the same limitation that affected `Embedding` before #4200. It prevents a caller that holds short-lived credentials out of band (e.g. a proxy/gateway) from supplying its own bearer token — the exact pattern the Gemini generation endpoints already support via the API-key query-parameter escape hatch.

## Changes

`vertexAuthHeaders` now takes the API-key query-parameter path when the key carries a value (`?key=<value>`), mirroring the Gemini generation endpoints, and leaves an existing `Authorization` header (set from context extra headers) intact. When the key has no value it falls back to the previous OAuth-token behaviour.

Because every cached content method funnels through this one helper, the single change covers all five operations.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [ ] Core (Go)
- [ ] Transports (HTTP)
- [x] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [ ] Docs

## How to Test

Added `TestVertexAuthHeaders_APIKeyPreservesInjectedAuthHeader` (internal `package vertex`) asserting that, with an API-key-valued key, the injected `Authorization` header is preserved and the `key` query parameter is set.

## Breaking changes

- [ ] Yes
- [x] No

## Related issues

Follow-up to #4200, which applied the same escape hatch to the `Embedding` method.

## Security considerations

No new auth mechanisms introduced. The change reuses the existing `authQuery` pattern already present in `ChatCompletion()`, `Responses()`, and `ResponsesStream()`. The API key is URL-encoded via `url.QueryEscape` consistent with all other call sites.

## Checklist


TransactCharlie:
I updated this branch to resolve conflicts. Please review and merge if you are happy with this change.

TransactCharlie:
Heya -- Gentle prod to see if this could be merged if it's acceptable please.

TransactCharlie:
Hi again -- we'd *really* appreciate this getting merged if you think it is acceptable. We are considering having to fork and we would really rather avoid that. Appreciate there are hundreds of open PR's but this one I think is pretty tiny with a scoped blast radius.

🙏 

image

akshaydeo (approved):

PR #5505, opened 2026-07-23
fix: forward vLLM responses-stream chunks and completion events

### LINKED ISSUE #5504: [Bug]: vLLM streaming Responses API hangs forever - chunks are silently discarded and the request never completes
### Prerequisites

### Description

`POST /v1/responses` with `stream: true` against the `vllm` provider hangs forever - no chunks are ever delivered to the client, and the request never times out or errors. This is a complete regression of VLLM's streaming Responses API, introduced by `c6777d43b` ("fix: moves vllm to native responses api", #5306), merged 2026-07-18.

**Root cause, traced and confirmed via an instrumented debug rebuild (not inferred):**

That PR switched `VLLMProvider.Responses`/`ResponsesStream` from converting requests to Chat Completions (the previous, working approach) to calling vLLM's native `/v1/responses` endpoint directly, via the shared `core/providers/openai/openai.go`'s `HandleOpenAIResponsesStreaming`, passing `HandleVLLMResponse` as a custom per-chunk response handler. VLLM is now the *only* provider that passes a non-nil `customResponseHandler` into this shared function - every other OpenAI-compatible provider (Azure, Bedrock Mantle, Fireworks, OpenRouter, Perplexity, xAI) passes `nil`.

The per-chunk SSE loop in `HandleOpenAIResponsesStreaming` has two branches:

https://github.com/maximhq/bifrost/blob/main/core/providers/openai/openai.go#L1877-L1979 (line numbers approximate to `main` @ 907ae1fcf)

```go
if customResponseHandler != nil {
    rawRequest, rawResponse, bifrostErr := customResponseHandler([]byte(jsonData), &response, nil, false, false)
    if bifrostErr != nil {
        ...
        return
    }
    // <-- nothing else happens here. Falls straight through to the next loop iteration.
} else {
    // ... sets response.ExtraFields.ChunkIndex
    // ... detects response.Type == Error / Failed and converts to BifrostError + return
    // ... detects response.Type == Completed / Incomplete and calls ProcessAndSendResponse + return
    // ... otherwise calls ProcessAndSendResponse for the delta chunk
}
```

For VLLM (`customResponseHandler != nil`), a successfully-parsed chunk (`bifrostErr == nil`) is **never sent anywhere** — `ProcessAndSendResponse` is only called inside the `else` branch. This branch also never inspects `response.Type`, so it never recognizes `response.completed`/`failed`/`error` and never `return`s early like every other provider does. The loop just keeps reading SSE lines until the connection's genuine EOF.

That's where it gets worse. Once EOF is reached and the loop breaks, deferred cleanup runs in this order: `stopCancellation` → `stopIdleTimeout` → `releaseGzip` → `ReleaseStreamingResponse`. The last one contains:

```go
// core/providers/utils/utils.go:2904
// Drain any remaining data from the body stream before releasing.
// This prevents "whitespace in header" errors when the connection is reused
// (see: https://github.com/valyala/fasthttp/issues/1743).
if _, err := io.Copy(io.Discard, bodyStream); err != nil { ... }
```

This drain exists specifically for callers that `return` *before* fully consuming the body (which is what every other provider's code path does - they return the moment they see `response.completed`). Because VLLM's loop instead reads all the way to genuine EOF itself, this second `io.Copy` read against an already-fully-drained fasthttp streaming body reader **blocks forever instead of returning immediately**. Confirmed via debug logging (`logger.Warn` inserted before/after each cleanup step, rebuilt locally with `make setup-workspace`) - execution reaches `"calling ReleaseStreamingResponse"` and never logs `"ReleaseStreamingResponse returned"`.

Because `ReleaseStreamingResponse` never returns, the next deferred call — `CloseStream(ctx, responseChan)`, which closes the channel the client-facing handler is waiting on - never runs either. A goroutine dump (`SIGQUIT`) on a hung process confirms the HTTP-handling goroutine is parked in `core/bifrost.go:6122`, inside `executeRequestWithRetries`, blocked on `providers/utils/stream.go:33`'s `CheckFirstStreamChunkForError`'s `<-stream` — waiting on a channel that is both empty (no chunk ever sent) and never closed (cleanup never completes).

**Net effect:** every streaming Responses API request to VLLM hangs indefinitely - no data, no error, no timeout - and leaks a goroutine plus an unreleased connection per occurrence, since nothing in the call chain ever unblocks.

---

### Steps to reproduce

vLLM's own server isn't required to reproduce this - the bug is entirely in Bifrost's SSE-consumption code and triggers on any spec-valid OpenAI-compatible Responses API stream. 

Disclosing plainly: the repro below uses a small local mock server emitting valid Responses API SSE events (not a live GPU-backed vLLM instance), which keeps the repro to a few seconds and needs no GPU/vLLM install. Validated the mock's event format is spec-correct by running the identical stream through the real `openai` provider as a control - it parsed and delivered every chunk correctly, so the difference is provably in VLLM's code path, not the test data.

```bash
git clone https://github.com/maximhq/bifrost.git && cd bifrost
make setup-workspace   # or just `make build` if not doing local-module debugging
make build

# minimal local mock: valid OpenAI-compatible Responses API SSE stream
mkdir -p /tmp/fakevllm && cat > /tmp/fakevllm/main.go <<'EOF'
package main
import ("fmt"; "log"; "net/http")
func flush(w http.ResponseWriter, s string) {
    fmt.Fprintf(w, "data: %s\n\n", s)
    w.(http.Flusher).Flush()
}
func main() {
    http.HandleFunc("/v1/responses", func(w http.ResponseWriter, r *http.Request) {
        w.WriteHeader(200)
        flush(w, `{"type":"response.created","sequence_number":0,"response":{"id":"r1","object":"response","created_at":1,"model":"m","output":[],"status":"in_progress"}}`)
        flush(w, `{"type":"response.output_text.delta","sequence_number":1,"delta":"hello"}`)
        flush(w, `{"type":"response.completed","sequence_number":2,"response":{"id":"r1","object":"response","created_at":1,"model":"m","status":"completed","output":[{"type":"message","role":"assistant","status":"completed","content":[{"type":"output_text","text":"hello"}]}]}}`)
    })
    log.Fatal(http.ListenAndServe("127.0.0.1:8092", nil))
}
EOF
cd /tmp/fakevllm && go mod init fakevllm && go build -o fakevllm . && ./fakevllm &
cd -

mkdir -p /tmp/bf-repro && cat > /tmp/bf-repro/config.json <<'EOF'
{
  "providers": {
    "vllm": {
      "keys": [{
        "name": "vllm-key", "value": "x", "weight": 1.0, "models": ["*"],
        "vllm_key_config": { "url": "http://127.0.0.1:8092", "model_name": "fake-model" }
      }]
    }
  }
}
EOF
./tmp/bifrost-http -port 8100 -app-dir /tmp/bf-repro &

curl -v -m 15 -X POST http://127.0.0.1:8100/v1/responses \
  -d '{"model":"vllm/fake-model","stream":true,"input":[{"role":"user","content":"hi"}]}'
```

Result: `curl: (28) Operation timed out ... 0 bytes received`. Reproduces every time.

**Control test** (proves it's VLLM's code path, not the mock server): repeat the exact same request but with an `openai` provider pointed at the same mock server (`network_config.base_url` override) - it streams back correctly, with every chunk delivered and `chunk_index` populated sequentially (0-2).

**Confirmed reproducible against the compiled binary two additional ways**, both with identical results: (1) VLLM as the only configured provider (rules out any multi-provider interaction), (2) with an instrumented debug rebuild showing the exact call sequence up to the hang point (goroutine dump + `logger.Warn` trace available on request).

---

### Expected behavior

A streaming `/v1/responses` request to the `vllm` provider should stream response chunks to the client and complete normally, the same as it does for every other configured provider (openai, azure, fireworks, openrouter, perplexity, xai, etc.)

### Actual behavior

The request hangs forever. Zero chunks are delivered, no error is returned, and there is no timeout - confirmed hanging past 25+ seconds in testing, with nothing to suggest it would ever resolve on its own. Each hung request leaks a goroutine and an unreleased connection.

### Affected area(s)

Core (Go)

### Version

Introduced in commit `c6777d43b` (PR #5306). Confirmed absent in `core/v1.7.2`, present starting `core/v1.7.3` (the current latest tagged core release) and on `main` HEAD.

### Environment

```text
- OS: macOS 15 (Darwin 25.5.0), arm64
- Go: 1.26.5
- Reproduced via: compiled bifrost-http binary (make build), curl, against a local mock SSE server
```

### Relevant logs/output

```shell
DEBUG-RCA responses-stream: about to call activeClient.Do url=http://127.0.0.1:8092/v1/responses
DEBUG-RCA responses-stream: activeClient.Do returned err=<nil> status=200
DEBUG-RCA responses-stream: read SSE line: {"type":"response.created",...}
DEBUG-RCA responses-stream: customResponseHandler succeeded, bifrostErr=nil -- falling through WITHOUT sending chunk
DEBUG-RCA responses-stream: read SSE line: {"type":"response.output_text.delta",...}
DEBUG-RCA responses-stream: customResponseHandler succeeded, bifrostErr=nil -- falling through WITHOUT sending chunk
DEBUG-RCA responses-stream: read SSE line: {"type":"response.completed",...}
DEBUG-RCA responses-stream: customResponseHandler succeeded, bifrostErr=nil -- falling through WITHOUT sending chunk
DEBUG-RCA responses-stream: ReadDataLine returned readErr=EOF
DEBUG-RCA responses-stream: breaking out of SSE loop, readErr=EOF
DEBUG-RCA responses-stream: calling stopCancellation
DEBUG-RCA responses-stream: stopCancellation returned
DEBUG-RCA responses-stream: calling stopIdleTimeout
DEBUG-RCA responses-stream: stopIdleTimeout returned
DEBUG-RCA responses-stream: calling releaseGzip
DEBUG-RCA responses-stream: releaseGzip returned
DEBUG-RCA responses-stream: calling ReleaseStreamingResponse
[... never returns ...]


Corroborating goroutine dump (`SIGQUIT`) frame:

goroutine ... [chan receive]:
runtime.chanrecv(...)
github.com/maximhq/bifrost/core/providers/utils.CheckFirstStreamChunkForError(...)
	.../core/providers/utils/stream.go:33
github.com/maximhq/bifrost/core.executeRequestWithRetries[...](...)
	.../core/bifrost.go:6122
```

### Regression?

Yes. Worked correctly (via the chat-completion fallback path) through `core/v1.7.2`. Broke starting `core/v1.7.3`, introduced by `c6777d43b` / PR #5306.

### Severity

sammaji:
## Summary

Fixes the `customResponseHandler` path in `HandleOpenAIResponsesStreaming` where `sendBackRawRequest` and `sendBackRawResponse` were hardcoded to `false`, and where subsequent error/completion handling was incorrectly scoped inside the `else` block, causing it to be skipped entirely when a custom response handler was provided. Fixes #5504

## Changes

- Replaced hardcoded `false, false` arguments in the `customResponseHandler` call with the actual `sendBackRawRequest` and `sendBackRawResponse` values.
- Added `RawResponse` population for the custom handler path, mirroring the existing behavior in the non-custom handler path.
- Moved error type handling (`ResponsesStreamResponseTypeError`, `ResponsesStreamResponseTypeFailed`), completion handling (`ResponsesStreamResponseTypeCompleted`, `ResponsesStreamResponseTypeIncomplete`), and chunk dispatch outside of the `else` block so they execute regardless of whether a custom response handler is used.
- Removed the `// TODO fix this` comment that tracked this known issue.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [x] Core (Go)
- [ ] Transports (HTTP)
- [x] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

```sh
go test ./...
```

## Breaking changes

- [ ] Yes
- [x] No

## Security considerations

None.

## Checklist

### DIFF
--- a/core/providers/openai/openai.go
+++ b/core/providers/openai/openai.go
@@ -1871,9 +1871,8 @@ func HandleOpenAIResponsesStreaming(
 
 			// Parse into bifrost response
 			var response schemas.BifrostResponsesStreamResponse
-			// TODO fix this
 			if customResponseHandler != nil {
-				rawRequest, rawResponse, bifrostErr := customResponseHandler([]byte(jsonData), &response, nil, false, false)
+				rawRequest, rawResponse, bifrostErr := customResponseHandler([]byte(jsonData), &response, nil, sendBackRawRequest, sendBackRawResponse)
 				if bifrostErr != nil {
 					if sendBackRawRequest {
 						bifrostErr.ExtraFields.RawRequest = rawRequest
@@ -1884,6 +1883,9 @@ func HandleOpenAIResponsesStreaming(
 					ctx.SetValue(schemas.BifrostContextKeyStreamEndIndicator, true)
 					providerUtils.ProcessAndSendBifrostError(ctx, postHookRunner, providerUtils.EnrichError(ctx, bifrostErr, jsonBody, nil, sendBackRawRequest, sendBackRawResponse, latency), responseChan, logger, postHookSpanFinalizer)
 					return
+				}
+				if sendBackRawResponse {
+					response.ExtraFields.RawResponse = jsonData
 				}
 			} else {
 				if err := sonic.UnmarshalString(jsonData, &response); err != nil {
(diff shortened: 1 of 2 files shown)


sammaji:
* **#5505** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/5505?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/maximhq/bifrost/5505?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `dev`


<h2></h2>

akshaydeo:
### Merge activity

* **Jul 23, 8:42 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/5505).
* **Jul 23, 8:42 PM UTC**: @akshaydeo merged this pull request with [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/5505).

PR #5563, opened 2026-07-26
adds count token api support for bedorck mantle

akshaydeo:
## Summary

Implements `CountTokens` for the Bedrock Mantle provider using the native-Anthropic `/anthropic/v1/messages/count_tokens` endpoint. This is the only available token-counting path for Claude models that are cross-Region-inference-only on `bedrock-runtime` and therefore have no regional `bedrock-runtime` endpoint for the Bedrock `CountTokens` API.

## Changes

- Removed the unsupported-operation stub for `CountTokens` from `bedrockmantle.go` and replaced it with a full implementation in a new `counttokens.go` file.
- The implementation routes requests to `https://bedrock-mantle.<region>.api.aws/anthropic/v1/messages/count_tokens`, honouring the same region-resolution precedence (model prefix → alias → key config → default) as other mantle surfaces.
- Non-Anthropic models (e.g. `gpt-*`, Gemma) are rejected locally with an unsupported-operation error rather than sending a request that AWS would reject upstream, since the `/anthropic/v1/messages` surface is Claude-specific.
- Auth follows the existing mantle pattern: API key via `Authorization` header or SigV4 signing when no API key is configured. The project header is included because the `bedrock-mantle:CountTokens` IAM action is scoped to a Bedrock Project resource.
- Fields rejected by the count_tokens schema (`max_tokens`, `temperature`, `fallback_credit_token`) are stripped from the request body before sending.
- Region prefixes in the model ID are split off so only the bare model ID appears in the request body.
- Error responses are parsed using the Anthropic error envelope (`{"type":"error","error":{...}}`) rather than the AWS JSON error shape.
- Unit tests cover URL construction, region-resolution precedence, non-Anthropic model rejection, and request body field stripping/retention.
- Integration test harness updated to enable `CountTokens: true` for the Claude scenario.
- Provider capability table and Bedrock Mantle documentation updated to reflect the new supported operation.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [x] Core (Go)
- [ ] Transports (HTTP)
- [x] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [x] Docs

## How to test

```sh
# Unit tests (no AWS credentials required)
go test ./core/providers/bedrockmantle/...

# Integration test (requires AWS credentials and a Bedrock Mantle-enabled account)
go test ./core/providers/bedrockmantle/ -run TestBedrockMantle -v
```

The integration test exercises `CountTokens` against a Claude model ID via the native-Anthropic mantle surface. Pointing the scenario at a `gpt-oss` or Gemma model ID should return an unsupported-operation error, confirming the local guard works correctly.

## Breaking changes

- [ ] Yes
- [x] No

## Security considerations

SigV4 signing is applied after the request body is fully set, ensuring the signature covers exactly the bytes sent on the wire. API keys flow through the `Authorization` header only and are never written into the request body. No PII or secrets are logged; error bodies are gated behind the existing `sendBackRawRequest`/`sendBackRawResponse` flags.

## Checklist


akshaydeo:
* **#5563** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/5563?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/maximhq/bifrost/5563?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `dev`


<h2></h2>

akshaydeo:
@coderabbitai full review

akshaydeo:
@coderabbitai full review

akshaydeo:
@coderabbitai full review

akshaydeo:
@coderabbitai full review

akshaydeo:
@coderabbitai full review

core/providers/bedrockmantle/counttokens.go
```diff
@@ -0,0 +1,180 @@
...
+// only), which is why it is not redundant with the bedrock provider's Converse count-tokens.
+func mantleAnthropicCountTokensURL(region string) string {
+	return fmt.Sprintf("https://bedrock-mantle.%s.api.aws/anthropic/v1/messages/count_tokens", region)
+}
+
+// parseMantleAnthropicError parses an error response from the Bedrock Mantle native-Anthropic
+// surface. Errors there follow the Anthropic envelope
+// ({"type":"error","error":{"type":...,"message":...}}) rather than the AWS JSON error shape
+// that bedrock-runtime returns, so the Anthropic error type is what must be unmarshalled.
+func parseMantleAnthropicError(resp *fasthttp.Response) *schemas.BifrostError {
```
Pratham-Mishra04: this we can reuse from anthropic module

core/providers/bedrockmantle/counttokens.go
```diff
@@ -0,0 +1,180 @@
...
+	// Sign after the body is set so the SigV4 signature covers exactly what is sent. The signer
+	// is nil when a Bedrock Mantle API key is configured (auth is the Authorization header then).
+	if signer := provider.mantleSigner(ctx, key, url, "application/json", region); signer != nil {
+		sigHeaders, signErr := signer(jsonBody)
+		if signErr != nil {
+			return nil, signErr
+		}
+		for headerKey, headerValue := range sigHeaders {
+			req.Header.Set(headerKey, headerValue)
+		}
```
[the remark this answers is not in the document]
akshaydeo: Fixed, though not with the suggested patch - unconditionally sending `jsonBody` would have been incorrect here.

In large payload mode the router skips body parsing entirely (`integrations/router.go:709-712`) and only `hydrateAnthropicRequestFromLargePayloadMetadata` runs, which sets `Model` and `Stream` and nothing else. So the converted `jsonBody` carries no messages in that mode - sending it would trade a 403 for a 400 (or a meaningless count).

The actual defect was narrower: only the signature was wrong. `req.Body()` drains a fasthttp body stream into the request buffer (`fasthttp/http.go:449-466`), so materializing it before signing makes the signed bytes and the sent bytes identical:

```go
signedBody := jsonBody
if usedLargePayloadBody {
    signedBody = req.Body()
}
sigHeaders, signErr := signer(signedBody)
```

The API-key path keeps streaming - there is no signature to invalidate, and the raw body is the only place the real messages live. A `max_tokens` in a body the client posted to `/v1/messages/count_tokens` earns a 400 that is correctly attributable to the caller; silently rewriting a passthrough body would be the surprising behaviour.

Added `TestCountTokensSignsTheBodyItSends` to pin the fasthttp draining behaviour the fix depends on.

Note for follow-up: `anthropic.completeRequest` (`core/providers/anthropic/anthropic.go:241-253`) has the identical pattern and is what the Bedrock Mantle *chat* surface uses via `HandleAnthropicChatCompletionRequest` + `mantleSigner`, so the same 403 is latent there. Not touched in this PR.

core/providers/bedrockmantle/counttokens.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+
+	// Count-tokens responses are always tiny, so the large-response streaming client is skipped
+	// and the response is buffered normally.
+	latency, bifrostErr, wait := providerUtils.MakeRequestWithContext(ctx, provider.mantleClient, req, resp)
```
Pratham-Mishra04: I think we can use HandleAnthropicCountTokensRequest() from anthropic

core/providers/bedrockmantle/counttokens.go
akshaydeo: fixed

core/providers/bedrockmantle/counttokens.go
akshaydeo: fixed

core/providers/bedrockmantle/counttokens.go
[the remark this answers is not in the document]
akshaydeo: Follow-up: this fix has moved. Per the review thread below, `bedrockmantle.CountTokens` now routes through the shared `anthropic.HandleAnthropicCountTokensRequest`, so the signed-body fix lives in `anthropic.completeRequest` rather than in this file - the code you commented on no longer exists here.

Same behaviour, one level up:

```go
signedBody := jsonBody
if usedLargePayloadBody {
    signedBody = req.Body()
}
sigHeaders, bErr := signer(signedBody)
```

Relocating it also fixed the identical latent issue on the Bedrock Mantle chat surface, which goes through the same `completeRequest` with a signer.

Pratham-Mishra04 (approved):

akshaydeo:
### Merge activity

* **Jul 27, 7:13 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/5563).
* **Jul 27, 7:14 AM UTC**: @akshaydeo merged this pull request with [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/5563).

PR #5599, opened 2026-07-27
adds CalculateCost and GetModelConfig for ctx in plugins

akshaydeo:
## Summary

Exposes the model catalog to plugins via `ctx.GetModelInfo` and `ctx.CalculateCost` inside any hook — pre-request, post-LLM, streaming, realtime, MCP tool execution, and HTTP transport pre-hooks. Previously, plugins had no first-class way to look up per-token pricing or model capability metadata from within a hook without importing framework packages directly.

## Changes

- **`core/schemas/modelcatalog.go`** — Introduces `ModelInfoProvider`, a narrow interface (`GetModelInfo`, `CalculateRequestCost`) that core exposes without importing the framework. Mirrors the same pattern used for `schemas.Tracer`.
- **`core/schemas/context.go`** — Adds `BifrostContext.GetModelInfo` and `BifrostContext.CalculateCost` as first-class accessors. Both delegate to the catalog stamped on the context under `BifrostContextKeyModelCatalog`, returning zero values when no catalog is wired so plugins are safe in standalone SDK usage.
- **`core/bifrost.go`** — Adds `SetModelCatalog` / `getModelCatalog` / `setModelCatalogOnContext` on the `Bifrost` struct. The catalog is stored in an `atomic.Pointer` via a `modelCatalogHolder` box to avoid the typed-nil-in-interface trap. `setModelCatalogOnContext` is called at the start of every request path (`handleRequest`, `handleStreamRequest`, `RunPreRequestHooks`, `RunStreamPreHooks`, `RunRealtimeTurnPreHooks`, `ExecuteChatMCPTool`, `ExecuteResponsesMCPTool`).
- **`framework/modelcatalog/modelinfo.go`** — Implements `ModelInfoProvider` on `*ModelCatalog`. `GetModelInfo` resolves through pricing rows then capability entries. `CalculateRequestCost` wraps the existing `CalculateCost` path so tiered pricing, batch/flex rates, cache costs, and governance overrides all apply. `ApplyModelInfo` is extracted as an exported function so the list-models enrichment path and `GetModelInfo` share identical mapping logic.
- **`transports/bifrost-http/handlers/inference.go`** — `enrichListModelsResponse` now calls `modelcatalog.ApplyModelInfo` instead of duplicating the field-mapping inline, eliminating ~50 lines and ensuring the two paths never drift.
- **`transports/bifrost-http/handlers/middlewares.go`** — Stamps the catalog onto the context in `TransportInterceptorMiddleware` so `HTTPTransportPreHook` sees it before the inference path runs.
- **`transports/bifrost-http/server/server.go`** — Calls `s.Client.SetModelCatalog` during `Bootstrap` after the catalog is confirmed non-nil.

**Design decisions:**
- The `modelCatalogHolder` box is intentional: storing a `schemas.ModelInfoProvider` interface directly in `atomic.Pointer` would require `*ModelInfoProvider`, which is error-prone. The holder also makes a nil pointer unambiguously mean "not configured."
- The catalog is stamped on the root context once per request; plugin-scoped contexts inherit it through `WithPluginScope`'s `valueDelegate` chain without any extra wiring.
- `CalculateCost` must be called synchronously in a hook — the governance scope lookup reads request identity off the live context.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [x] Core (Go)
- [x] Transports (HTTP)
- [ ] Providers/Integrations
- [x] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

```sh
go test ./core/schemas/... ./framework/modelcatalog/... ./transports/bifrost-http/...
```

New tests cover:
- `TestModelInfoAccessorsNoCatalogWired` — no panic when no catalog is configured
- `TestModelInfoAccessorsDelegate` — correct delegation to the provider
- `TestModelInfoAccessorsSkipEmptyArgs` — empty model / nil response never reach the catalog
- `TestModelInfoVisibleFromPluginScope` — scoped and nested plugin contexts inherit the handle
- `TestModelInfoVisibleFromDerivedContext` — derived contexts (fallbacks, fan-out) inherit through the parent chain
- `TestGetModelInfoPopulatesPricingAndLimits` — pricing fields render in fixed-decimal notation
- `TestGetModelInfoReportsDeprecation` — deprecated flag propagates
- `TestGetModelInfoUnknownModel` — unknown and empty model IDs return nil
- `TestApplyModelInfoDoesNotOverwriteProviderValues` — catalog never overwrites provider-reported fields
- `TestApplyModelInfoNilSafe` — nil receiver and nil entry are safe

## Breaking changes

- [ ] Yes
- [x] No

## Security considerations

The catalog handle is read-only from a plugin's perspective (`GetModelInfo` returns a freshly allocated struct; `CalculateCost` reads but does not mutate). No credentials or PII flow through these accessors.

## Checklist

### DIFF
--- a/core/bifrost.go
+++ b/core/bifrost.go
@@ -88,6 +88,7 @@ type Bifrost struct {
 	bifrostRequestPool  sync.Pool                           // Pool for BifrostRequest objects
 	logger              schemas.Logger                      // logger instance, default logger is used if not provided
 	tracer              atomic.Value                        // tracer for distributed tracing (stores schemas.Tracer, NoOpTracer if not configured)
+	modelCatalog        atomic.Pointer[modelCatalogHolder]  // model pricing/capability catalog exposed to plugins via ctx.GetModelInfo (nil if not configured)
 	MCPManager          mcp.MCPManagerInterface             // MCP integration manager (nil if MCP not configured)
 	mcpCredStore        schemas.MCPCredentialStore          // Per-call credential resolver for MCP tool execution (wraps oauth2Provider for OAuth-flavored auth types)
 	mcpInitOnce         sync.Once                           // Ensures MCP manager is initialized only once
@@ -220,6 +221,14 @@ type pluginTimingAccumulator struct {
 // of different concrete types, even if they implement the same interface.
 type tracerWrapper struct {
 	tracer schemas.Tracer
+}
+
+// modelCatalogHolder boxes the ModelInfoProvider interface so it can live in an
+// atomic.Pointer. Storing the interface directly would need a *ModelInfoProvider,
+// which is easy to get wrong; the holder also lets a nil pointer mean
+// "no catalog configured" without the typed-nil-in-interface trap.
+type modelCatalogHolder struct {
+	catalog schemas.ModelInfoProvider
 }
 
 // INITIALIZATION
@@ -402,6 +411,51 @@ func (bifrost *Bifrost) SetTracer(tracer schemas.Tracer) {
 // getTracer returns the tracer from atomic storage with type assertion.
 func (bifrost *Bifrost) getTracer() schemas.Tracer {
 	return bifrost.tracer.Load().(*tracerWrapper).tracer
+}
+
+// SetModelCatalog sets the model pricing/capability catalog for the Bifrost
+// instance. Once set, plugins can reach it from any hook via ctx.GetModelInfo
+// and ctx.CalculateCost. Safe to call at any time; nil clears it.
+//
+// Callers passing a concrete pointer type must nil-check first: a nil
+// *modelcatalog.ModelCatalog boxed into the interface is not == nil.
+func (bifrost *Bifrost) SetModelCatalog(catalog schemas.ModelInfoProvider) {
+	if catalog == nil {
+		bifrost.modelCatalog.Store(nil)
+		return
+	}
+	bifrost.modelCatalog.Store(&modelCatalogHolder{catalog: catalog})
+}
+
+// getModelCatalog returns the configured catalog, or nil if none was set.
+func (bifrost *Bifrost) getModelCatalog() schemas.ModelInfoProvider {
+	holder := bifrost.modelCatalog.Load()
+	if holder == nil {
+		return nil
+	}
+	return holder.catalog
+}
+
+// setModelCatalogOnContext stamps the catalog handle onto a request context so
+// plugin hooks can reach it. Scoped plugin contexts delegate Value lookups to
+// the root (see BifrostContext.WithPluginScope), so stamping the root once is
+// enough for every plugin in the pipeline.
+//
+// The no-catalog case clears rather than skips: reused long-lived contexts
+// (realtime WS connections re-enter this per message) would otherwise keep
+// serving a catalog that a later SetModelCatalog(nil) already retired.
+// ClearValue is itself a no-op on a context with no values map, so SDK users
+// who never wire a catalog still pay nothing.
+func (bifrost *Bifrost) setModelCatalogOnContext(ctx *schemas.BifrostContext) {
+	if ctx == nil {
+		return
+	}
+	catalog := bifrost.getModelCatalog()
+	if catalog == nil {
+		ctx.ClearValue(schemas.BifrostContextKeyModelCatalog)
+		return
+	}
+	ctx.SetValue(schemas.BifrostContextKeyModelCatalog, catalog)
 }
 
 // ReloadConfig reloads the config from DB
@@ -3128,6 +3182,7 @@ func (bifrost *Bifrost) ExecuteChatMCPTool(ctx *schemas.BifrostContext, toolCall
 	} else {
 		ensureMCPRawStorageContext(ctx)
 		ensureMCPTracerContext(ctx, bifrost.getTracer())
+		bifrost.setModelCatalogOnContext(ctx)
 	}
 	if bifrost.MCPManager == nil {
 		return nil, &schemas.BifrostError{
@@ -3147,6 +3202,7 @@ func (bifrost *Bifrost) ExecuteResponsesMCPTool(ctx *schemas.BifrostContext, too
 	} else {
 		ensureMCPRawStorageContext(ctx)
 		ensureMCPTracerContext(ctx, bifrost.getTracer())
+		bifrost.setModelCatalogOnContext(ctx)
 	}
 	if bifrost.MCPManager == nil {
 		return nil, &schemas.BifrostError{
@@ -4802,6 +4858,7 @@ func (bifrost *Bifrost) RunPreRequestHooks(ctx *schemas.BifrostContext, req *sch
 	if _, ok := ctx.Value(schemas.BifrostContextKeyRequestID).(string); !ok {
 		ctx.SetValue(schemas.BifrostContextKeyRequestID, uuid.New().String())
 	}
+	bifrost.setModelCatalogOnContext(ctx)
 
 	pipeline := bifrost.getPluginPipeline()
 	defer bifrost.releasePluginPipeline(pipeline)
@@ -4826,6 +4883,7 @@ func (bifrost *Bifrost) RunStreamPreHooks(ctx *schemas.BifrostContext, req *sche
 
 	tracer := bifrost.getTracer()
 	ctx.SetValue(schemas.BifrostContextKeyTracer, tracer)
+	bifrost.setModelCatalogOnContext(ctx)
 
 	// Create a trace so the logging plugin can accumulate streaming chunks.
 	// The traceID is used as the accumulator key in ProcessStreamingChunk.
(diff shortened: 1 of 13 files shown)


akshaydeo:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/maximhq/bifrost/5599?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#5599** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/5599?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/maximhq/bifrost/5599?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#5565** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/5565?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#5640](https://github.com/maximhq/bifrost/pull/5640) <a href="https://app.graphite.com/github/pr/maximhq/bifrost/5640?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#5562** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/5562?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5561** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/5561?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `dev`


<h2></h2>

core/bifrost.go
```diff
@@ -390,6 +399,43 @@ func (bifrost *Bifrost) getTracer() schemas.Tracer {
...
+// the root (see BifrostContext.WithPluginScope), so stamping the root once is
+// enough for every plugin in the pipeline. No-op when no catalog is configured,
+// which keeps the context map clean for SDK users who never wire one.
+func (bifrost *Bifrost) setModelCatalogOnContext(ctx *schemas.BifrostContext) {
+	if ctx == nil {
+		return
+	}
+	if catalog := bifrost.getModelCatalog(); catalog != nil {
+		ctx.SetValue(schemas.BifrostContextKeyModelCatalog, catalog)
+	}
```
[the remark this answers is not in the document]
akshaydeo: Fixed - `setModelCatalogOnContext` now clears the key instead of skipping when no catalog is configured:

```go
catalog := bifrost.getModelCatalog()
if catalog == nil {
    ctx.ClearValue(schemas.BifrostContextKeyModelCatalog)
    return
}
ctx.SetValue(schemas.BifrostContextKeyModelCatalog, catalog)
```

The reachable path is the reused-context one: `RunPreHooks` (core/bifrost.go) re-enters this per WS message on a long-lived context, so a context stamped before a `SetModelCatalog(nil)` would have kept serving the retired catalog to every later hook. That also contradicted `SetModelCatalog`'s own doc ("Safe to call at any time; nil clears it").

Two clarifications on the finding: `BifrostContextKeyModelCatalog` is not in `reservedKeys` (core/schemas/context.go:14), so there is no restricted-writes gate on either call. And `ClearValue` is a no-op when the context has no values map, so SDK users who never wire a catalog still pay nothing - the doc comment was updated to say so rather than claiming the whole function no-ops.

Read-back stays safe: `GetModelInfo`/`CalculateCost` use the comma-ok assertion, so the cleared nil yields `catalog == nil` and returns nil/0.

framework/modelcatalog/modelinfo.go
```diff
@@ -0,0 +1,144 @@
...
+	if entry == nil {
+		return nil
+	}
+
+	info := &schemas.Model{ID: model}
+	ApplyModelInfo(info, entry)
+
+	if params := mc.datasheet.GetSupportedParameters(model); len(params) > 0 {
+		info.SupportedParameters = params
+	}
```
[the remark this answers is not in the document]
akshaydeo: Fixed - the clones now happen in `ApplyModelInfo` rather than in `GetModelInfo`, because that is where the aliasing originates and it has a second caller (`transports/bifrost-http/handlers/inference.go`, the list-models path) that escapes the same pointers into an HTTP response.

Root cause confirmed: `datasheet/store.go` does `pricing, ok := s.pricingData[key]` then `convertTablePricingToEntry(&pricing)`. Reading a struct out of a map copies it, but a Go struct copy is shallow - the map header and pointer fields still point at the row living under `s.mu`. `convertTablePricingToEntry` (datasheet/types.go) then assigns `ContextLength`, `MaxInputTokens`, `MaxOutputTokens`, `Architecture` and `AdditionalAttributes` straight through, so the chain handed callers a live handle into catalog state.

Clone depth is per-type and deliberate:
- `AdditionalAttributes` - `maps.Clone`. Shallow is correct here: the values are strings, which are immutable, so there is nothing further down to alias.
- `Architecture` - struct copy **plus** `slices.Clone` on `InputModalities` and `OutputModalities`. A shallow struct copy alone would have left both slices aliased.
- The three `*int` limits - fresh allocations via `new(*entry.X)`.

Worth noting the severity is understated at Minor. The lock in `store.go` is released before `GetModelInfo` returns, so a plugin writing to the escaped attributes map while the 24h pricing sync reads it is an unsynchronized concurrent map access - that is a `fatal error: concurrent map writes` that takes the process down, not a recoverable panic. It also contradicted two ownership promises already in this PR (`modelinfo.go` "freshly allocated and owned by the caller" and `context.go` "Mutating it does not affect the catalog or any other caller").

Tests added as requested:
- `TestGetModelInfoReturnsCallerOwnedCopy` - mutates every returned field, asserts the next call returns clean values and that the pointers differ between calls.
- `TestApplyModelInfoClonesMutableEntryFields` - covers `Architecture` and the attributes map directly against a shared entry, since the seeded `GetModelInfo` rows do not carry those.

`make test-framework` reports 384 tests / 0 failures for the modelcatalog package.

framework/modelcatalog/modelinfo.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+// formatCost renders a per-unit rate the way the models API reports it. Fixed
+// 10-decimal notation rather than %g so sub-cent token rates never surface in
+// scientific notation, which clients parsing these as decimals choke on.
+func formatCost(v float64) string {
+	return fmt.Sprintf("%.10f", v)
+}
+
+func ptr[T any](v T) *T {
```
Pratham-Mishra04, line 158: new()

core/bifrost.go
```diff
@@ -390,6 +399,51 @@ func (bifrost *Bifrost) getTracer() schemas.Tracer {
 	return bifrost.tracer.Load().(*tracerWrapper).tracer
 }
 
+// SetModelCatalog sets the model pricing/capability catalog for the Bifrost
+// instance. Once set, plugins can reach it from any hook via ctx.GetModelInfo
+// and ctx.CalculateCost. Safe to call at any time; nil clears it.
+//
+// Callers passing a concrete pointer type must nil-check first: a nil
+// *modelcatalog.ModelCatalog boxed into the interface is not == nil.
+func (bifrost *Bifrost) SetModelCatalog(catalog schemas.ModelInfoProvider) {
```
Pratham-Mishra04, line 422: we already have catalog before initialising core client so we can directly take it in .Init() instead of a setter

framework/modelcatalog/modelinfo.go
akshaydeo: aa fuck - ye claude instructions follow hi nahi karata :cry: fixing

framework/modelcatalog/modelinfo.go
```diff
@@ -0,0 +1,160 @@
...
+
+	providerUtils "github.com/maximhq/bifrost/core/providers/utils"
+	"github.com/maximhq/bifrost/core/schemas"
+)
+
+// ModelCatalog implements schemas.ModelInfoProvider, which is what backs
+// ctx.GetModelInfo / ctx.CalculateCost inside plugin hooks. Core cannot import
+// this package (framework depends on core, not the reverse), so the wiring is
+// an interface assertion here plus a Bifrost.SetModelCatalog call at boot.
+var _ schemas.ModelInfoProvider = (*ModelCatalog)(nil)
```
Pratham-Mishra04, line 16: this placeholder is not required now right?

Pratham-Mishra04 (changes requested):
added comments

core/schemas/modelcatalog.go
```diff
@@ -0,0 +1,25 @@
...
+// to plugins. The catalog itself lives in framework/modelcatalog, which core
+// cannot import (framework depends on core, not the other way round), so the
+// concrete instance arrives as an interface value stamped onto the request
+// context under BifrostContextKeyModelCatalog.
+//
+// This mirrors how Tracer is wired: interface declared here, implemented in
+// framework, reached from BifrostContext via a lookup-and-delegate accessor.
+//
+// Implemented by *modelcatalog.ModelCatalog.
+type ModelInfoProvider interface {
```
Pratham-Mishra04, line 13: lets call it modelCatalog only then to make it more extensible for datasheet changes?

Issue #5630, opened 2026-07-28
[Docs]: logs_store.retention_days has no effect on postgres, but is documented next to the Postgres example


jitokim:
### Prerequisites

- [x] I have searched existing issues and docs to avoid duplicates

### Affected page(s)

- `docs/deployment-guides/config-json/storage.mdx` — "Log Retention"
- `docs/enterprise/log-exports.mdx` — "Configuration"
- `transports/config.schema.json` — `logs_store.retention_days`

### What's wrong or missing?

Two docs pages give opposite instructions, and the one shown next to the PostgreSQL example is the wrong one.

`storage.mdx` documents `retention_days` as a general purge control, inside the `logs_store` section that directly follows the PostgreSQL config example:

> Set `retention_days` to automatically purge old log entries. `0` disables retention-based cleanup.

`log-exports.mdx` says the opposite:

> Retention is configured on the **client config**, not on `logs_store`. The active key is `log_retention_days`.

`log-exports.mdx` is the correct one.

#### Reproduction

`maximhq/bifrost:latest` + `postgres:16-alpine`.

**config.json — only `logs_store.retention_days` set:**

```json
{
  "logs_store": {
    "enabled": true,
    "type": "postgres",
    "retention_days": 1,
    "config": {
      "host": "...", "port": "5432", "user": "...",
      "password": "...", "db_name": "bifrost", "ssl_mode": "disable"
    }
  }
}
```

**Startup log:**

```
{"level":"info","message":"log retention days: 365"}
{"level":"info","message":"log retention cleaner initialized with 365 days retention"}
{"level":"info","message":"starting log cleanup: deleting logs older than 2025-07-28T17:36:01Z (retention: 365 days)"}
```

The configured `1` is ignored and the cleaner runs at 365. Rows seeded at 30 days old survived the pass.

**Then adding `client.log_retention_days: 1`** (leaving `logs_store.retention_days: 1` in place):

```
{"level":"info","message":"log retention days: 1"}
{"level":"info","message":"starting log cleanup: deleting logs older than 2026-07-27T17:36:32Z (retention: 1 days)"}
{"level":"info","message":"log cleanup completed: deleted 10 logs in 1 batches"}
```

#### Why

[...]

Issue #1665, opened 2026-02-16
PostgreSQL TOAST bloat causes unbounded PVC growth despite log retention cleanup


Edward-Upton:
## Problem

When using the PostgreSQL log store with log retention enabled, the PVC storage grows continuously and never shrinks, even though the number of live rows in the `logs` table remains consistent.

In our production deployment (3 replicas, 14-day log retention), the `logs` table is **123 GB** on disk, but the actual live data is a fraction of that:

| Component | Size |
|---|---|
| Main table (live rows) | 1.3 GB |
| Indexes | 523 MB |
| **TOAST table** | **120 GB** |

The `logs` table has ~20 large `text` columns (`input_history`, `output_message`, `raw_request`, `raw_response`, `tools`, `tool_calls`, etc.) that store full LLM request/response payloads. PostgreSQL stores these oversized values in a TOAST table. When log retention deletes old rows, the TOAST chunks become dead tuples — but they're never reclaimed.

### Root Cause

**1. No VACUUM after bulk deletions**

The `LogsCleaner` in `framework/logstore/cleaner.go` deletes rows in batches of 100 via `DeleteLogsBatch`, but never runs `VACUUM` or `VACUUM ANALYZE` afterward. Dead tuples accumulate in both the main table and its TOAST table.

**2. Default PostgreSQL autovacuum thresholds are too high for this table**

Autovacuum triggers when: `dead_tuples > threshold + (scale_factor × live_tuples)`

With defaults (`threshold=50`, `scale_factor=0.2`) and ~51M live TOAST tuples:

```
50 + (0.2 × 51,000,000) = 10,200,050 dead tuples needed to trigger
```

In our deployment, the TOAST table had **7.6M dead tuples** but autovacuum hadn't run in 3 days because it hadn't hit the 10.2M threshold yet.

**3. Regular VACUUM doesn't return space to the OS**

Even when autovacuum does run, it only marks dead tuples as reusable for future inserts. It does not shrink the on-disk files. Only `VACUUM FULL` (which takes an exclusive table lock) actually returns space to the filesystem. This means the PVC will only ever grow.

### 1. Run `VACUUM ANALYZE` after each cleanup cycle

After the batch deletion loop in `cleanupOldLogs()` completes, execute a `VACUUM ANALYZE logs` to immediately reclaim dead tuples and update query planner statistics. This could be done via a raw SQL exec on the GORM `*sql.DB`:

```go
// After cleanup loop finishes
[...]

PR #5634, opened 2026-07-28
fix: prune mcp_tool_logs on the retention schedule

### LINKED ISSUE #5631: [Bug]: mcp_tool_logs is never purged by retention cleanup on postgres/sqlite
### Prerequisites

### Description

`docs/enterprise/log-exports.mdx` states, under "What gets deleted":

> It deletes from the main `Log` table as well as the MCP tool logs table.

On postgres the retention cleaner only deletes from `logs`. `mcp_tool_logs` is never purged by age, so on any deployment that logs MCP tool calls it grows without bound regardless of the retention setting.

### Steps to reproduce

Reproduced on `maximhq/bifrost:latest` + `postgres:16-alpine`.

1. Run with `logs_store.type: postgres` and `client.log_retention_days: 1`.
2. Seed rows aged 30 days into both tables:

```sql
INSERT INTO logs (id, timestamp, object_type, provider, model, status, created_at)
SELECT 'seed-log-'||g, now()-interval '30 days', 'chat.completion', 'openai', 'gpt-4o', 'success', now()-interval '30 days'
FROM generate_series(1,10) g;

INSERT INTO mcp_tool_logs (id, timestamp, tool_name, status, created_at)
SELECT 'seed-mcp-'||g, now()-interval '30 days', 'test_tool', 'success', now()-interval '30 days'
FROM generate_series(1,10) g;
```

3. Restart Bifrost — a cleanup pass runs immediately on boot.
4. Count the seeded rows in both tables.

### Expected behavior

Rows older than the retention window are deleted from both tables, as documented.

### Actual behavior

Cleanup log:

```
{"level":"info","message":"log retention days: 1"}
{"level":"info","message":"starting log cleanup: deleting logs older than 2026-07-27T17:36:32Z (retention: 1 days)"}
{"level":"info","message":"log cleanup completed: deleted 10 logs in 1 batches"}
```

Row counts afterwards:

```
logs           remaining seed rows = 0
mcp_tool_logs  remaining seed rows = 10
mcp_tool_logs  oldest row age      = 1 mon
```

### Notes from reading the code

`DeleteLogsBatch` — the function the cleaner calls — selects and deletes from `Log` only (`framework/logstore/rdb.go`). The only deletes touching `MCPToolLog` on the RDB path are `DeleteMCPToolLogs` (explicit ID list) and a stale sweep scoped to `status = 'processing'`. Neither is retention-based.

ClickHouse does get an `mcp_tool_logs` TTL (`clickhousemigrate.go`), but only via `logs_store.retention_days`, which has no effect on postgres/sqlite — see #5630.

### Affected area(s)

Framework, Docs

### Additional context

A further signal that the omission is unintentional: `chLogsTTL` in `framework/logstore/clickhousemigrate.go` builds the TTL clause for **both** tables and documents itself as

> derives the logs/mcp_tool_logs TTL clause from the configured retention. Values < 1 leave TTL unset (the LogsCleaner still prunes via DeleteLogsBatch).

i.e. the cleaner was expected to cover `mcp_tool_logs` on backends that have no table TTL. Combined with ClickHouse already applying the TTL to `mcp_tool_logs`, this reads as a backend parity gap rather than a deliberate exclusion, so the fix belongs in the cleaner rather than in the docs.

Found while reading into #1665.

jitokim:
## Summary

The retention cleaner only pruned the `logs` table. `mcp_tool_logs` was never deleted by age on the RDB path, so it grew without bound on postgres and sqlite regardless of the configured retention.

Closes #5631.

## Why this is a parity gap, not a deliberate exclusion

- ClickHouse already applies the same TTL to both tables (`migrationClickHouseMCPToolLogsTable`).
- `chLogsTTL`, which builds that TTL clause for **both** `logs` and `mcp_tool_logs`, documents itself as:
  > derives the logs/mcp_tool_logs TTL clause from the configured retention. Values < 1 leave TTL unset (the LogsCleaner still prunes via DeleteLogsBatch).

  i.e. the cleaner was expected to cover `mcp_tool_logs` on backends without a table TTL.
- `docs/enterprise/log-exports.mdx` already documents the cleaner as deleting "the main `Log` table as well as the MCP tool logs table".

So the documented and ClickHouse behaviour was already the intended one; only the RDB path was missing it.

## Changes

- `framework/logstore/cleaner.go` — new `MCPToolLogRetentionManager` optional interface, plus `cleanupOldMCPToolLogs` which runs the same batched loop against the same cutoff and reports its counts separately.
- `framework/logstore/rdb.go`, `clickhousestore.go`, `hybrid.go` — `DeleteMCPToolLogsBatch` on each store.
- `framework/logstore/clickhouse.go`, `clickhousemigrate.go` — doc comments updated to name both methods (1 line each).
- `docs/enterprise/log-exports.mdx` — bucket lifecycle rules must cover `<prefix>/mcp-logs/`, which only becomes reachable now that MCP rows are pruned (`MCPToolObjectKey` writes there).

### Two design decisions worth reviewing

**1. A separate optional interface rather than widening `LogRetentionManager`.**

`transports/bifrost-http/server/server.go` gates the entire cleaner on `s.Config.LogsStore.(logstore.LogRetentionManager)` and silently skips starting it when the assertion fails. Adding a method to that interface would therefore silently disable *all* retention for any store that does not implement it. With a separate interface, such a store skips MCP pruning only.

Trade-off, documented on the interface: a downstream type embedding `logstore.LogStore` satisfies `LogRetentionManager` for free but not `MCPToolLogRetentionManager`, so it keeps pruning `logs` while skipping `mcp_tool_logs`. Partial skip seemed clearly better than silently disabling retention entirely, but happy to change this if you prefer the interface widened.

**2. The MCP sweep is `defer`red inside `cleanupOldLogs`.**

The logs loop returns early on its first delete error. Running the MCP sweep after it sequentially would mean any persistent condition on `logs` (lock contention, statement timeout, restricted grants) aborts every daily run before MCP is reached — reintroducing the same unbounded growth through another path, on exactly the deployments most likely to hit it. The `defer` sits immediately after the cutoff is computed so no existing line moved. A cancelled context still short-circuits inside `cleanupOldMCPToolLogs`.

### Reproduction: same test file, same database, only the branch differs

A temporary in-package test (not included in this PR) seeded the rows from #5631 into real Postgres from `framework/docker-compose.yml`, ran one `cleanupOldLogs` pass with `RetentionDays: 1`, then counted both tables.

**On `dev`:**

```
SEEDED       logs=10  mcp_tool_logs=10
AFTER SWEEP  logs=0   mcp_tool_logs=10
--- FAIL: TestZZReproMCPToolLogsRetention
    Should be zero, but was 10
    Messages: mcp_tool_logs should be pruned on the same schedule
```

**On this branch:**

```
SEEDED       logs=10  mcp_tool_logs=10
{"level":"info","message":"starting log cleanup: deleting logs older than 2026-07-27T18:37:36Z (retention: 1 days)"}
{"level":"info","message":"log cleanup completed: deleted 10 logs in 1 batches"}
{"level":"info","message":"mcp tool log cleanup completed: deleted 10 mcp tool logs in 1 batches"}
AFTER SWEEP  logs=0   mcp_tool_logs=0
--- PASS: TestZZReproMCPToolLogsRetention
```

### Suite

Run against live Postgres + ClickHouse (`framework/docker-compose.yml`), since the parity suite skips silently otherwise:

```sh
cd framework
gofmt -l logstore/     # no output
go vet ./logstore/...  # clean
go test ./logstore/... -count=1
# ok  github.com/maximhq/bifrost/framework/logstore  103.686s
```

The new parity case is confirmed not skipped:

```
--- PASS: TestLogStoreParity/Deletes/delete_logs_batch (0.04s)
--- PASS: TestLogStoreParity/Deletes/delete_mcp_logs_batch (0.03s)
```

### Tests added — 21, of which 16 are failure/edge cases

`cleaner_test.go` (new): store without MCP support still prunes logs; MCP batch error stops only the MCP loop; logs batch error still prunes MCP; cancelled sweep skips MCP deletes; zero rows issues a single call; fewer-than-batchSize stops after one pass; more-than-batchSize iterates until drained; already-cancelled context issues no calls; cancellation mid-loop returns promptly; MCP sweep uses the same cutoff as the logs sweep; default retention when unset; both tables pruned on a real SQLite store.

`rdb.go` (real SQLite): keeps rows newer than cutoff; no matching rows returns zero; empty table returns zero; respects batchSize; cancelled context errors; leaves the logs table alone.

`hybrid.go`: delegates to inner; unsupported inner returns zero; propagates inner error.

`logstoreparity_test.go`: `delete_mcp_logs_batch` across backends, plus compile-time `var _ MCPToolLogRetentionManager` assertions for all three stores.

Two mutation checks confirm the tests bite rather than merely pass:
- removing the ClickHouse override → `FAIL … delete_mcp_logs_batch: postgres vs clickhouse mismatch`
- flipping `created_at <` to `<=` in `rdb.go` → `FAIL … KeepsRowsNewerThanCutoff`

## Breaking changes

- [x] No

`LogRetentionManager` is unchanged, so existing implementations keep working. Stores that do not implement the new optional interface simply skip MCP pruning, which is the current behaviour.

Behaviour change worth calling out: on an existing deployment with a backlog of `mcp_tool_logs`, the first cleanup pass after upgrading will delete every row older than the retention window, in batches of 100.

## Known limitations (deliberate, flagged for follow-up)

1. **Shared 30-minute budget.** Both sweeps share one `context.WithTimeout(30*time.Minute)` and `logs` always spends it first, so a very large first-run backlog can leave nothing for MCP. It self-corrects on subsequent runs. Splitting the budget felt out of scope here.
2. **`DeleteMCPToolLogsBatch` duplicates `DeleteLogsBatch`** in each store. Deliberate — structural de-duplication (generics or a shared helper) would mix a refactor into a behavioural fix. Happy to follow up separately if you want it.

## Type of change

- [x] Bug fix

## Affected areas

- [x] Core (Go) — `framework/logstore`
- [x] Docs

## Related issues

- Closes #5631
- Context: #1665, #5630

## Security considerations

None. No new data is exposed; a table that was previously retained indefinitely is now subject to the retention already configured for logs.

## Checklist

- [x] I verified builds succeed (Go) — `go vet` and the full `framework/logstore` suite; the UI was untouched

### DIFF
--- a/docs/enterprise/log-exports.mdx
+++ b/docs/enterprise/log-exports.mdx
@@ -227,7 +227,7 @@ If you have object storage enabled, configure a lifecycle / object-lifecycle-man
     {
       "ID": "bifrost-logs-expire-30d",
       "Status": "Enabled",
-      "Filter": { "Prefix": "bifrost/logs/" },
+      "Filter": { "Prefix": "bifrost/" },
       "Expiration": { "Days": 30 }
     }
   ]
@@ -246,7 +246,7 @@ Apply with `aws s3api put-bucket-lifecycle-configuration --bucket <bucket> --lif
         "action": { "type": "Delete" },
         "condition": {
           "age": 30,
-          "matchesPrefix": ["bifrost/logs/"]
+          "matchesPrefix": ["bifrost/"]
         }
       }
     ]
--- a/docs/enterprise/log-exports.mdx
+++ b/docs/enterprise/log-exports.mdx
@@ -217,7 +217,7 @@ The cleaner deletes **database rows** older than `now - retention_days` (UTC). I
 
 > **Important**: the retention cleaner does **not** delete the corresponding payloads in S3 or GCS. The hybrid log store explicitly delegates object cleanup to the bucket's own lifecycle policy.
 
-If you have object storage enabled, configure a lifecycle / object-lifecycle-management rule on the bucket to expire objects under your `prefix/`. Pick a duration that matches (or exceeds) `log_retention_days`.
+If you have object storage enabled, configure a lifecycle / object-lifecycle-management rule on the bucket to expire objects under your `prefix/`. Pick a duration that matches (or exceeds) `log_retention_days`. Scope the rule to the whole `prefix/`, not just `prefix/logs/` — MCP tool call payloads live under `prefix/mcp-logs/` and would otherwise be orphaned once their rows are pruned.
 
 **Amazon S3 lifecycle rule (example)**
 
--- a/framework/logstore/cleaner.go
+++ b/framework/logstore/cleaner.go
@@ -22,12 +22,27 @@ type LogRetentionManager interface {
 	DeleteLogsBatch(ctx context.Context, cutoff time.Time, batchSize int) (deletedCount int64, err error)
 }
 
+// MCPToolLogRetentionManager is implemented by log stores that can prune MCP
+// tool logs by age. It is intentionally separate from LogRetentionManager so
+// that a store which does not implement it keeps working (it simply skips MCP
+// pruning) instead of failing the LogRetentionManager type assertion that
+// gates the cleaner being started at all.
+//
+// Note for decorators: a type that embeds LogStore satisfies
+// LogRetentionManager for free (DeleteLogsBatch is part of LogStore) but does
+// NOT satisfy this interface, so it keeps pruning logs while silently skipping
+// mcp_tool_logs. Forward DeleteMCPToolLogsBatch explicitly, as HybridLogStore
+// does.
+type MCPToolLogRetentionManager interface {
+	DeleteMCPToolLogsBatch(ctx context.Context, cutoff time.Time, batchSize int) (deletedCount int64, err error)
+}
+
 // CleanerConfig holds configuration for the log cleaner
 type CleanerConfig struct {
 	RetentionDays int
 }
 
-// LogsCleaner manages the cleanup of old logs
+// LogsCleaner manages the cleanup of old logs and MCP tool logs
 type LogsCleaner struct {
 	manager     LogRetentionManager
 	config      CleanerConfig
@@ -46,6 +61,7 @@ func NewLogsCleaner(manager LogRetentionManager, config CleanerConfig, logger sc
 }
 
 // StartCleanupRoutine starts a goroutine that periodically cleans up old logs
+// and MCP tool logs
 func (c *LogsCleaner) StartCleanupRoutine() {
 	c.mu.Lock()
 	defer c.mu.Unlock()
@@ -102,7 +118,8 @@ func (c *LogsCleaner) StopCleanupRoutine() {
 	c.stopCleanup = nil
 }
 
-// cleanupOldLogs deletes logs older than the retention period in batches
+// cleanupOldLogs deletes logs older than the retention period in batches, and
+// prunes MCP tool logs against the same cutoff when the store supports it
 func (c *LogsCleaner) cleanupOldLogs(ctx context.Context) {
 	retentionDays := c.config.RetentionDays
 	if retentionDays < 1 {
@@ -112,6 +129,13 @@ func (c *LogsCleaner) cleanupOldLogs(ctx context.Context) {
 	// Calculate cutoff time
 	cutoff := time.Now().UTC().AddDate(0, 0, -retentionDays)
 	c.logger.Info("starting log cleanup: deleting logs older than %s (retention: %d days)", cutoff.Format(time.RFC3339), retentionDays)
+
+	// Deferred so the MCP sweep runs on every exit path below. The two tables
+	// are independent, and the logs loop bails out on the first delete error:
+	// letting that starve MCP pruning would leave mcp_tool_logs growing without
+	// bound, which is the bug this sweep exists to fix. A cancelled ctx still
+	// short-circuits inside cleanupOldMCPToolLogs.
+	defer c.cleanupOldMCPToolLogs(ctx, cutoff)
 
 	totalDeleted := int64(0)
 	batchCount := 0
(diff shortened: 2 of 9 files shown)


jitokim:
Thanks — all three addressed in 3686e32.

1. **Lifecycle examples (major)** — correct, and the inconsistency was mine: I added the paragraph about covering the whole prefix but left both examples scoped to `bifrost/logs/`. Both the S3 `Filter.Prefix` and the GCS `matchesPrefix` now use `bifrost/`.

2. **Sentinel assertion** — agreed, cancelling the context proved only that *an* error surfaced, not that the inner store's error survived the hop. The test now wraps an MCP-capable inner store that returns a sentinel and asserts with `require.ErrorIs`.

3. **Unclosed hybrid store** — fixed; it now closes on exit like the two tests next to it.

Re-verified against live Postgres + ClickHouse from `framework/docker-compose.yml`:

```
gofmt -l logstore/     # no output
go vet ./logstore/...  # clean
go test ./logstore/... -count=1
# ok  github.com/maximhq/bifrost/framework/logstore  102.200s
```

jitokim:
Re-verified against current `dev` (231 commits on since this branch point). The gap this PR closes is still open, and nothing here has been superseded.

- `framework/logstore/cleaner.go` is byte-identical to the merge base — `LogRetentionManager` still exposes only `DeleteLogsBatch`.
- Both implementations still scope that to the `logs` table alone: `RDBLogStore.DeleteLogsBatch` selects `Model(&Log{})` (`rdb.go:4469`), and `ClickHouseLogStore.DeleteLogsBatch` (`clickhousestore.go:492`) is the same shape.
- `DeleteMCPToolLogsBatch` and `MCPToolLogRetentionManager` do not exist anywhere on `dev`, so there is no naming collision.

Worth narrowing the scope claim, because it is not symmetric across backends. On ClickHouse, `mcp_tool_logs` is created with `ttl: chLogsTTL(retentionDays)` (`clickhousemigrate.go:271`), so it is already pruned by the table's own TTL. The unbounded growth is specifically the **RDB (Postgres/SQLite)** path, which has no TTL and is not reached by `DeleteLogsBatch`. The comment on `chLogsTTL` itself points at this split — it notes that below the TTL threshold "the LogsCleaner still prunes via DeleteLogsBatch", which is exactly the path that does not cover this table.

On the migrations that landed since — `logs_add_upstream_and_overhead_latency_columns`, `logs_add_batch_debug_column`, `logs_add_cost_breakdown_columns`, `logs_recreate_matviews_with_cost_breakdown`, `logs_add_overhead_breakdown_column` — all five add columns to `logs` rather than new tables, so they are already covered by the existing row-level delete and do not widen the coverage gap. This PR adds no migration of its own, so there is no version collision with them either.

One more thing worth surfacing: `docs/enterprise/log-exports.mdx:214` already says *"It deletes from the main `Log` table as well as the MCP tool logs table"*, and this PR does not touch that line — it predates the branch. So on `dev` today the documented behaviour and the code disagree, and this PR is what makes the existing sentence true rather than adding a new promise.

Still `MERGEABLE`, CodeRabbit approved. @akshaydeo @Pratham-Mishra04 — happy to adjust anything here if the shape is not what you want.

PR #5665, opened 2026-07-29
changelog updates

### LINKED ISSUE #4215: [Bug]: HuggingFace models show provider ID twice in /v1/models, which breaks requests

### LINKED ISSUE #4851: v1.6.2 and v1.6.3 governance rate-limit reset causes high CPU in BumpRateLimitUsage/updateRateLimitReferences
## Summary

After upgrading Bifrost from `v1.6.0` to `v1.6.2`, a production Kubernetes deployment started burning sustained CPU in the governance rate-limit reset path, despite very low request volume.

Image

The CPU profile points at request-time rate-limit reset side effects:

```text
github.com/maximhq/bifrost/plugins/governance.(*LocalGovernanceStore).BumpRateLimitUsage
  -> ResetExpiredRateLimitsInMemory
  -> resetExpiredRateLimitFromSnapshot
  -> updateRateLimitReferences
  -> sync.Map/HashTrieMap Range + high allocation/GC
```

This looks correlated with the `v1.6.2` source change that makes `BumpRateLimitUsage` call `ResetExpiredRateLimitsInMemory(ctx, rateLimitID)` when a reset target is due. In `v1.6.0`, `BumpRateLimitUsage` reset the counters inside the local CAS clone and did not call `updateRateLimitReferences` from the request hot path.

## Environment

Redacted production environment:

- Image with issue: `docker.io/maximhq/bifrost:v1.6.2`
- Previous image: `docker.io/maximhq/bifrost:v1.6.0`
- Helm chart: `bifrost` chart `2.1.25`
- Deployment: Kubernetes, single Bifrost replica
- Storage/config store: PostgreSQL
- Governance plugin: enabled
- Semantic cache/vector store: disabled
- pprof: captured from loopback-only authenticated pprof endpoint

Sensitive values have been redacted/omitted: hostnames, IPs, pod names, DB connection details, credentials, virtual key IDs/values, customer data, and request payloads.

## Observed behavior

CPU before the `v1.6.2` rollout was low and stable:

```text
2026-07-01T17:00:00Z 0.044 cores
2026-07-01T18:00:00Z 0.045 cores
2026-07-01T18:30:00Z 0.040 cores
```

The `v1.6.2` ReplicaSet was created around `2026-07-01T18:32:14Z`. CPU jumped immediately after:

```text
2026-07-01T18:40:00Z 1.988 cores
2026-07-01T18:50:00Z 1.988 cores
2026-07-01T19:00:00Z 1.991 cores
2026-07-01T19:30:00Z 1.991 cores
2026-07-01T20:00:00Z 1.992 cores
```

Traffic did not increase during this window:

```text
bifrost_upstream_requests_total rate:
2026-07-01T18:30:00Z 0.119 req/s
2026-07-01T18:40:00Z 0.005 req/s
2026-07-01T19:00:00Z 0.015 req/s
2026-07-01T19:30:00Z 0.026 req/s

http_requests_total rate:
2026-07-01T18:30:00Z 0.511 req/s
2026-07-01T18:40:00Z 0.364 req/s
2026-07-01T19:00:00Z 0.285 req/s
2026-07-01T19:30:00Z 0.281 req/s
```

We also observed occasional DB deadlocks during expired rate-limit reset persistence:

```text
failed to reset expired rate limits: failed to persist rate limit resets to database: failed to reset rate limit <redacted-rate-limit-id>: ERROR: deadlock detected (SQLSTATE 40P01)
```

## CPU profile

15-second Go CPU profile from `v1.6.2`:

```text
Duration: 15.01s, Total samples = 54.01s (359.91%)
Showing nodes accounting for 45.41s, 84.08% of 54.01s total

flat     flat%   cum      cum%    function
7.46s    13.81%  34.09s   63.12%  github.com/maximhq/bifrost/plugins/governance.(*LocalGovernanceStore).updateRateLimitReferences.func1
1.71s     3.17%  36.07s   66.78%  internal/sync.(*HashTrieMap[...]).iter
0.10s     0.19%  38.53s   71.34%  github.com/maximhq/bifrost/plugins/governance.(*LocalGovernanceStore).resetExpiredRateLimitFromSnapshot
0.02s     0.04%  38.66s   71.58%  github.com/maximhq/bifrost/plugins/governance.(*LocalGovernanceStore).ResetExpiredRateLimitsInMemory.func1
0.01s     0.02%  40.03s   74.12%  github.com/maximhq/bifrost/plugins/governance.(*LocalGovernanceStore).BumpRateLimitUsage
0.01s     0.02%  38.93s   72.08%  github.com/maximhq/bifrost/plugins/governance.(*LocalGovernanceStore).ResetExpiredRateLimitsInMemory
0.01s     0.02%  36.12s   66.88%  github.com/maximhq/bifrost/plugins/governance.(*LocalGovernanceStore).updateRateLimitReferences

Allocation/GC-heavy runtime cost was also visible:
20.33s cum runtime.newobject
20.23s cum runtime.mallocgc
19.36s cum runtime.mallocgcSmallScanNoHeader
7.63s  cum runtime.gcDrain
```

The raw pprof file is 57 KB. I attempted to attach it as a redacted Gist, but GitHub rejected Gist creation with `HTTP 403: Rate Limit Exceeded`. I can provide the raw profile privately, or attach it manually through the GitHub web UI if needed. The profile summary above contains the relevant stack and timings.

## Source-level suspicion

`plugins/governance/v1.6.2` changes `BumpRateLimitUsage` to call the full reset path from request handling:

```go
if tokenNewLastReset, requestNewLastReset := gs.rateLimitResetTargets(old, time.Now()); tokenNewLastReset != nil || requestNewLastReset != nil {
    gs.ResetExpiredRateLimitsInMemory(ctx, rateLimitID)
    continue
}
```

That full reset path then reaches:

```go
resetRateLimit, ok := gs.ResetRateLimitAt(ctx, rateLimit.ID, tokenNewLastReset, requestNewLastReset)
...
gs.updateRateLimitReferences(ctx, resetRateLimit)
```

In the profile, that makes `updateRateLimitReferences` and `sync.Map.Range` dominate CPU during normal requests.

By contrast, `plugins/governance/v1.6.0` reset token/request counters inside `BumpRateLimitUsage`'s local clone before incrementing, without calling `ResetExpiredRateLimitsInMemory` or `updateRateLimitReferences` from the request hot path.

Relevant refs:

- `plugins/governance/v1.6.2`: `BumpRateLimitUsage` around `plugins/governance/store.go:438-450`
- `plugins/governance/v1.6.2`: `resetExpiredRateLimitFromSnapshot` around `plugins/governance/store.go:1843-1864`
- `plugins/governance/v1.6.0`: `BumpRateLimitUsage` around `plugins/governance/store.go:418-455`

## Impact

With less than 1 HTTP req/s and often less than 0.1 upstream req/s, the process consumed about 2 CPU cores continuously, and during one profile about 3.6 CPU cores. This is disproportionate to traffic and appears to be internal governance/rate-limit work.

## Expected behavior

Request-time rate-limit usage updates should not trigger a global/reference-refresh-heavy path. If a reset is needed during `BumpRateLimitUsage`, it should be local/constant-time, or the expensive reference refresh should be deferred/coalesced outside the request hot path.

## Mitigation being considered

Rollback to `v1.6.0`, because the deployment config and chart stayed the same and only the image tag changed from `v1.6.0` to `v1.6.2`.

### LINKED ISSUE #5329: [Bug]: `/api/logs` returns an incorrect `total_count` for time ranges of 24 hours or longer
### Prerequisites

### Description

When PostgreSQL LogStore is used, GET `/api/logs` may return a `pagination.total_count` that does not match the logs accessible through pagination.

Log details are always queried from the raw logs table using the exact requested time range:
```
timestamp >= start_time
AND timestamp <= end_time
```
For materialized-view-eligible ranges of at least 24 hours, however, `total_count` switches to the hourly `mv_logs_hourly` materialized view:
```
hour >= date_trunc('hour', start_time)
AND hour <= end_time
```
Because each hour row represents a complete hourly bucket, this expands both boundaries:

- `start_time` is rounded down to the beginning of its hour, including logs before the requested start.
- The bucket containing `end_time` is included in full, including logs after the requested end.
- If `end_time` is exactly on an hour boundary, the complete hour beginning at `end_time` is still included because the predicate uses <=.

Consequently, `total_count` may include logs outside the requested interval even though those logs cannot appear in any page of the detail response.

### Steps to reproduce

1. Use PostgreSQL LogStore with `mv_logs_hourly` ready.
2. Create logs inside and outside the boundary portions of a 24-hour interval.
3. Call `/api/logs` with a range of at least 24 hours and non-hour-aligned boundaries.
4. Compare `pagination.total_count` with the number of rows accessible through pagination.

### Expected behavior

`total_count` matches the number of logs within the requested time range.

### Actual behavior

`total_count` includes logs outside the requested time range because both boundaries are expanded to hourly buckets.

### Affected area(s)

Framework

### Version

latest

### Environment

```text

```

### Relevant logs/output

```shell

```

### Regression?

_No response_

### Severity

### LINKED ISSUE #5433: /genai endpoint rejects valid minLength/maxLength in tool schemas (400 INVALID_ARGUMENT)
## Discord Thread Context
**Thread:** /genai endpoint rejects valid minLength/maxLength in tool schemas (400 INVALID_ARGUMENT)
**Link:** https://discord.com/channels/1371790330269007932/1529187196920139857
**Created by:** akshay_deo (361806850561146880)

## Recent Thread Messages
**fredzolio:** The `/genai/v1beta/models/{model}:generateContent` endpoint (Gemini-native format) rejects any request whose `functionDeclarations[].parameters` schema has `minLength`/`maxLength` on a string property...
**fredzolio:** **Steps to Reproduce**

```bash
curl -X POST "https://<bifrost-host>/genai/v1beta/models/gemini-3.5-flash:generateContent" \
  -H "x-goog-api-key: <key>" \
  -d...
**fredzolio:** Expected: 200 OK, minLength honored or safely ignored.
**fredzolio:** Actual:
```
{"error":{"code":400,"message":"value at properties.textSelection.minLength must be a number","status":"INVALID_ARGUMENT","details":null}}
```
minLength is a number (tested as both JSON nu...
**fredzolio:** Comparison — same schema on /openai/v1/chat/completions (works, 200 OK)
```bash
curl -X POST "https://<bifrost-host>/openai/v1/chat/completions" \
  -H "Authorization: Bearer <key>" \
  -H "Content-Ty...
**fredzolio:** Additional Notes
- Removing minLength/maxLength makes the /genai request succeed immediately.
- Affects any MCP/agent client whose tool schemas use string length constraints (e.g. opencode's own built...
**fredzolio:** Environment
- Endpoint: /genai/v1beta/models/{model}:generateContent
- Models tested: gemini-3.5-flash, gemini-3.1-pro-preview (both reproduce)
- Auth: x-goog-api-key header (also reproduced with Auth...


**Discord Thread:** https://discord.com/channels/1371790330269007932/1529187196920139857

### LINKED ISSUE #5504: [Bug]: vLLM streaming Responses API hangs forever - chunks are silently discarded and the request never completes

### LINKED ISSUE #5546: [Bug]: Upstream SSE stream death swallowed into a clean [DONE] — dead streams appear successful (v1.5.16 through v1.6.5)
### Prerequisites

### Description

When an upstream OpenAI-compatible provider's connection dies while
streaming a chat completion — either before sending any SSE data event, or
partway through — Bifrost does not propagate that as a transport error or
close the client connection. Instead it emits one synthetic, content-free
chunk (empty `id`/`created`/`model` if the death is pre-first-byte, or the
real `id`/`created`/`model` already established by earlier real chunks if
the death is mid-stream; either way, one choice with an empty `delta`, no
`finish_reason`, and `usage.total_tokens: 0`) followed by a normal `data:
[DONE]`. From the client's perspective this is indistinguishable from a
provider that generated zero tokens and stopped cleanly — there is no
signal anywhere in the response that the upstream connection actually
failed.

This defeats any client-side retry/failover logic keyed on stream health:
a caller that trusts the clean `[DONE]` has no way to know the completion
is truncated/empty due to a real upstream failure, and any request-level
logging built on "did the stream end with `[DONE]`" will record the
request as a normal success.

### Steps to reproduce

1. A minimal mock OpenAI-compatible provider that dies according to which
   fault path is requested:

   ```python
   # mock_provider.py — run with: uvicorn mock_provider:app --port 8000
   import time, json
   from fastapi import FastAPI
   from fastapi.responses import StreamingResponse

   app = FastAPI()

   def chunk(content=None, finish=None):
       delta = {"content": content} if content else {}
       return "data: " + json.dumps({
           "id": "chatcmpl-repro", "object": "chat.completion.chunk",
           "created": int(time.time()), "model": "repro-model",
           "choices": [{"index": 0, "delta": delta, "finish_reason": finish}],
       }) + "\n\n"

   async def die_before_first_byte():
       raise RuntimeError("simulated upstream death, pre-first-byte")
       yield  # pragma: no cover — makes this a generator

   async def die_mid_stream():
       yield chunk("partial answer")
       raise RuntimeError("simulated upstream death, mid-stream")

   @app.post("/v1/chat/completions")
   async def chat_completions(body: dict):
       model = body["model"]
       gen = die_before_first_byte() if "before" in model else die_mid_stream()
       return StreamingResponse(gen, media_type="text/event-stream")
   ```

2. Point Bifrost at it as a custom OpenAI-compatible provider (`config.json`):

   ```json
   {
     "providers": {
       "repro-provider": {
         "keys": [{"name": "k", "value": "env.REPRO_KEY",
                    "models": ["repro-die-before-first-byte", "repro-die-mid-stream"],
                    "weight": 1.0}],
         "network_config": {"base_url": "http://localhost:8000",
                              "allow_private_network": true},
         "custom_provider_config": {"base_provider_type": "openai"}
       }
     }
   }
   ```

3. Drive both fault paths through Bifrost:

   ```bash
   curl -sN http://localhost:8080/v1/chat/completions -H 'Content-Type: application/json' \
     -d '{"model":"repro-provider/repro-die-before-first-byte","messages":[{"role":"user","content":"hi"}],"stream":true,"stream_options":{"include_usage":true}}'

   curl -sN http://localhost:8080/v1/chat/completions -H 'Content-Type: application/json' \
     -d '{"model":"repro-provider/repro-die-mid-stream","messages":[{"role":"user","content":"hi"}],"stream":true,"stream_options":{"include_usage":true}}'

### Expected behavior

the client either sees the connection close/reset, or receives a documented error frame it can distinguish from a real completion (e.g. an SSE `error` event, or a non-2xx status if nothing has been sent yet).

### Actual behavior

(both cases, verbatim from a real run against `v1.5.16` and `v1.6.3`):  Pre-first-byte: ``` data: {"id":"","choices":[{"index":0,"delta":{}}],"created":0,"model":"","object":"chat.completion.chunk","system_fingerprint":"","usage":{"total_tokens":0},...}  data: [DONE] ```  Mid-stream (after one real content chunk, `"partial answer"`): ``` data: {"id":"chatcmpl-<real>","choices":[{"index":0,"delta":{"content":" streamed"}}],"created":<real>,"model":"<real>",...}  data: {"id":"chatcmpl-<real>","choices":[{"index":0,"delta":{}}],"created":<real>,"model":"<real>","usage":{"total_tokens":0},...}  data: [DONE] ```  Both a 200 status and a clean `[DONE]` — no error, no dropped connection, no distinguishing marker.

### Affected area(s)

Transports (HTTP)

### Version

v1.5.16, v1.6.3, v1.6.5

### Environment

```text

```

### Relevant logs/output

```shell

```

### Regression?

_No response_

### Severity

### LINKED ISSUE #5551: [Bug]: transports/bifrost-http/lib test package does not compile on dev (MockConfigStore missing UpdateBudgetOverride)
### Prerequisites

### Description

The `transports/bifrost-http/lib` test package does not compile on `dev`.

`UpdateBudgetOverride` is declared on the `ConfigStore` interface but `MockConfigStore` in the test file does not implement it, so every test in the package fails to build:

- `framework/configstore/store.go:356-357` — interface method:
  ```go
  // UpdateBudgetOverride updates only the override state and returns the refreshed budget.
  UpdateBudgetOverride(ctx context.Context, id string, amount float64, mode tables.BudgetOverrideMode, cyclesRemaining int, tx ...*gorm.DB) (*tables.TableBudget, error)
  ```
- `transports/bifrost-http/lib/config_test.go` — zero implementations of `func (m *MockConfigStore) UpdateBudgetOverride`, while `MockConfigStore` is assigned to a `configstore.ConfigStore` field at ~20 call sites (1665, 1713, 1768, 1806, 1849, 1896, 1935, 2852, 2883, 2899, ...).

This looks like the interface gained a method without the mock being updated. It likely went unnoticed because `.github/workflows/pr-tests.yml` is `workflow_dispatch` only ("PR Tests (Requires Approval)"), so Go tests do not run automatically on PRs.

Note that `go.work` is gitignored, so a contributor has to create one to build against the in-repo modules at all. Without it, the same package fails differently because the published `core` lags the in-repo test code (`undefined: schemas.AddUpstreamLatency`, `undefined: schemas.BifrostContextKeyRoutingInfo`). Either way the package cannot be built from a fresh clone.

### Steps to reproduce

```bash
git clone --depth 1 --branch dev https://github.com/maximhq/bifrost.git
cd bifrost
go work init ./core ./framework ./transports $(for d in plugins/*/; do [ -f "$d/go.mod" ] && echo ./$d; done)
go test ./transports/bifrost-http/lib/ -run XXX
```

### Expected behavior

The package compiles and its tests run.

### Actual behavior

Build fails: `*MockConfigStore does not implement configstore.ConfigStore (missing method UpdateBudgetOverride)`.

### Affected area(s)

- [x] Transports (HTTP)
- [x] Framework

### Version

`dev @ 2952aae628f93b82aea5c9a14e2abe844ece2c0e` (2026-07-24)

### Environment

Linux arm64, Go 1.26.5 (toolchain pulled by the repo's `go 1.26.5` directive), fresh shallow clone of `dev`.

### Relevant logs/output

```shell
$ go test ./transports/bifrost-http/lib/ -run XXX
# github.com/maximhq/bifrost/transports/bifrost-http/lib [github.com/maximhq/bifrost/transports/bifrost-http/lib.test]
transports/bifrost-http/lib/config_test.go:1665:21: cannot use store (variable of type *MockConfigStore) as configstore.ConfigStore value in struct literal: *MockConfigStore does not implement configstore.ConfigStore (missing method UpdateBudgetOverride)
transports/bifrost-http/lib/config_test.go:1713:21: cannot use store (variable of type *MockConfigStore) as configstore.ConfigStore value in struct literal: *MockConfigStore does not implement configstore.ConfigStore (missing method UpdateBudgetOverride)
...
transports/bifrost-http/lib/config_test.go:2899:30: too many errors
FAIL	github.com/maximhq/bifrost/transports/bifrost-http/lib [build failed]
```

### Regression?

Yes — the mock compiled before `UpdateBudgetOverride` was added to the interface.

### Severity

### Additional context

Separate observation from the same package, only visible once it builds: on a fork of `dev` where the mock is fixed, `go test ./transports/bifrost-http/lib/` takes **~536s**, because `LoadConfig` reaches `modelcatalog.Init` (`transports/bifrost-http/lib/config.go:4470`) which does a blocking datasheet fetch per call — `LoadFromURLIntoMemory` when there is no config store (`framework/modelcatalog/main.go:240`), or the startup URL sync when the fresh temp DB has no pricing rows (`main.go:~141`). Every `LoadConfig` in `config_test.go` therefore performs real network I/O against `getbifrost.ai`, putting the package at ~89% of the default 10m `go test` timeout and making it fail on a slow network. Injecting a `file:` pricing URL (already supported by `loadPricingFromURL`) in the test fixtures would remove the dependency. Happy to split this into its own issue if you'd prefer.

### LINKED ISSUE #5552: [Feature]: refresh the live model catalog in the background (models added after boot stay invisible until restart)
### Prerequisites

### Problem to solve

The live model catalog is only ever populated at bootstrap and on key edits, so a model an upstream provider starts serving after the gateway booted stays invisible until the gateway is restarted.

`framework/modelcatalog/live/store.go` says so itself:

```go
// The store is passive — it never calls the network. Callers (the HTTP server
// after key add/update, or a future background refresher) decide when to
// fetch and push results in via Upsert.
```

That "future background refresher" does not exist on `dev`. `RefreshLiveModelsForProvider` has exactly two call sites in `transports/bifrost-http/server/server.go`:

- `:678` — after a key is added/updated/deleted
- `:1869` — the bootstrap seed loop

So the only ways to pick up a newly served model are to restart the gateway or to touch a key to force a refresh.

This is most visible with OpenAI-compatible custom providers that front a fast-moving upstream. In our deployment three such providers refresh their own catalogs every 10 minutes, but the gateway kept serving the boot-time snapshot for as long as it stayed up, and `/v1/models` consumers downstream (an agent CLI reading the gateway catalog) inherited the stale list. It also affects first-party providers whenever a vendor ships a model mid-uptime.

### Proposed solution

Run the existing per-provider refresh on a ticker.

The bootstrap block at `server.go:1856-1873` already fans out across providers; extracting it into a reusable method and calling that from a ticker keeps a single fan-out path:

```go
func (s *BifrostHTTPServer) RefreshAllLiveModels(ctx context.Context)   // reused by the bootstrap seed
func (s *BifrostHTTPServer) startLiveModelRefresher(ctx context.Context) // ticker, exits on ctx.Done()
```

Notes from implementing this on a fork of `dev`:

- Read the provider set through `GetAllProviders` / `GetProviderConfigRaw` rather than ranging `s.Config.Providers` directly, so the background pass is safe while providers are being edited. (The bootstrap call site can range the map because it runs before serving starts.)
- The cost is one boot-equivalent fan-out per interval. On a 16-provider / 30-key deployment that is 60 list-models calls per pass (filtered + unfiltered per key), so the interval wants to be configurable and the feature opt-out-able for anyone whose providers meter list-models.
- A default in the 15m range lines up with how quickly aggregators tend to move without being noisy.

Happy to open a PR with this if the approach looks right.

### Alternatives considered

- **Invalidate on a `/v1/models` cache miss.** Does not help: the miss is for a model the caller does not know exists yet, and the common consumer lists the whole catalog rather than probing.
- **Ask operators to restart or re-save a key.** This is the current workaround, and it is not discoverable — nothing surfaces that the catalog is stale.
- **Shorten the pricing datasheet sync interval.** Unrelated: that path refreshes pricing metadata, not the per-provider live model list.

### Area(s)

- [x] Transports (HTTP)
- [x] Framework

### Additional context

Verified against `dev @ 2952aae628f93b82aea5c9a14e2abe844ece2c0e` (2026-07-24).

### LINKED ISSUE #5554: [Bug]: provider reload wipes the live model catalog before refetching, so a transient list-models failure empties it
## Prerequisites

## Description

`ReloadProvider` clears the provider's live model catalog **before** re-fetching it, so a transient list-models failure during a provider update leaves the provider with **no live models at all** until some later refresh succeeds.

`transports/bifrost-http/server/server.go:671-679` (dev @ `2952aae`):

```go
// Refresh keyconfig from the current key list, then drop any stale live
// entries (for keys removed in this update) before refetching per-key.
s.Config.ModelCatalog.SetKeyConfigForProvider(provider, inMemoryKeys)
s.Config.ModelCatalog.InvalidateLiveProvider(provider)   // ← wipes ALL keys' entries
if hasNoKeys {
    logger.Warn("model discovery skipped for provider %s: no keys configured", provider)
} else {
    s.RefreshLiveModelsForProvider(ctx, provider, inMemoryKeys)   // best-effort
}
```

`RefreshLiveModelsForProvider` is best-effort by contract — `FetchAndStoreLiveForKey` logs errors and "they do not abort the other call" (`server.go:1049-1052`), and it writes nothing when the upstream call fails. So the sequence "invalidate everything, then try to refill" turns any upstream hiccup into an empty catalog, even though a perfectly good one was in memory a millisecond earlier.

The stated intent of the invalidate is narrow and is quoted in the comment itself: *"drop any stale live entries (for keys removed in this update)"*. Wiping every key's entry is much broader than that intent, and the fan-out that is supposed to restore them can fail for reasons that have nothing to do with the edit (upstream 5xx, rate limit, timeout, a gateway restarting behind the provider's base URL).

Consequences for an operator: editing an unrelated provider field (a timeout, a description) while the upstream is briefly unhealthy empties `GET /v1/models` for that provider, and routing falls back to the static datasheet. There is no background refresher to heal it (see #5552), so it persists until the next successful key/provider edit or a restart.

## Steps to reproduce

1. Configure a custom OpenAI-compatible provider whose upstream `/v1/models` works; confirm `GET /v1/models?provider=<name>` lists its models.
2. Make that upstream fail or hang (stop it, or point `base_url` at a port that refuses).
3. `PUT /api/providers/<name>` with the provider's current body (any no-op edit).
4. Bring the upstream back up and query `GET /v1/models?provider=<name>`.

The provider's live entries are gone: step 3's invalidate removed them and the refetch in the same call could not repopulate them.

## Expected behavior

A failed re-discovery leaves the previous live catalog in place; only entries for keys that were removed or disabled are pruned.

## Actual behavior

The whole provider's live catalog is wiped before the refetch, so a failed refetch empties it.

## Affected area(s)

Transports (HTTP)

## Version

dev @ `2952aae628f93b82aea5c9a14e2abe844ece2c0e` (2026-07-25)

## Relevant logs/output

```shell
# from FetchAndStoreLiveForKey on the failing refetch — note the catalog is
# already empty at this point, so the "fallback" is the only thing left:
falling back onto the static datasheet for provider <name>
```

## Regression?

No — the ordering has been there since live per-key catalogs were introduced.

## Severity

## Additional context

Minimal fix: retain the entries for keys that are still configured **and enabled**, and let the per-key upsert overwrite them only on success (it already writes nothing on failure):

```go
s.Config.ModelCatalog.SetKeyConfigForProvider(provider, inMemoryKeys)
if hasNoKeys {
    s.Config.ModelCatalog.InvalidateLiveProvider(provider)
    logger.Warn("model discovery skipped for provider %s: no keys configured", provider)
} else {
    keep := make(map[string]struct{}, len(inMemoryKeys)+1)
    if isKeylessProvider { keep[""] = struct{}{} }
    for _, key := range inMemoryKeys {
        if keyEnabled(key) { keep[key.ID] = struct{}{} }
    }
    s.Config.ModelCatalog.RetainLiveKeys(provider, keep)   // new: prune only gone/disabled keys
    s.RefreshLiveModelsForProvider(ctx, provider, inMemoryKeys)
}
```

with `RetainKeys` on `framework/modelcatalog/live.Store` (drop entries for this provider whose `KeyID` is not in `keep`) and a thin `RetainLiveKeys` wrapper on `ModelCatalog`. `OnKeyUpdated` has the same shape for the enabled-key branch.

Implemented and tested locally (live-store unit test plus handler/server coverage); happy to open a PR.

### LINKED ISSUE #5555: [Bug]: *StreamRequest returns (nil, nil) for empty streams — consumers hang forever on a nil-channel receive
### Description

`ChatCompletionStreamRequest` / `ResponsesStreamRequest` (and the other `*StreamRequest` methods on `core`) can return **`(nil, nil)`** — a nil chunk channel together with a nil `*schemas.BifrostError` — whenever the provider's stream closes without emitting a single chunk. An integrator doing the natural thing:

```go
ch, bErr := client.ResponsesStreamRequest(ctx, req)
if bErr != nil { /* handle */ }
for chunk := range ch { /* ... */ } // ch == nil → blocks forever
```

hangs forever: a receive from a nil channel never returns. There is no error to act on and nothing in the API contract suggests a nil-check on the channel is required when the error is nil.

**Root cause chain:**

1. `providerUtils.CheckFirstStreamChunkForError` (`core/providers/utils/stream.go`) blocks reading the first chunk; when the source channel closes with zero chunks it returns `nil, done, nil` ("return nil so callers can distinguish this from a live stream channel").
2. Its only caller, `executeRequestWithRetries` (`core/bifrost.go`), never inspects that nil — it assigns the result straight into the public return value: `result = any(checkedStream).(T)`.
3. The public API surface therefore yields `(nil, nil)`.

A secondary effect: because the empty-stream detection blocks on the first chunk, the call also doesn't return until the underlying stream fully closes.

### Reproduction

Minimal pure-core probe (any provider whose stream can close with zero chunks; easiest to hit on `core` v1.7.3 with the vLLM provider, where the Responses-API streaming path emitted no chunks at all — that chunk-forwarding problem is already fixed on current `main`, but the `(nil, nil)` path remains reachable for any genuinely empty stream):

```go
client, _ := bifrost.Init(context.Background(), schemas.BifrostConfig{Account: acct{}})
ch, bErr := client.ResponsesStreamRequest(ctx, req)
fmt.Printf("ch=%v err=%v\n", ch == nil, bErr) // ch=true err=<nil>
for range ch {} // hangs forever
```

Observed output with a live vLLM 0.21 backend on v1.7.3:

```
ResponsesStreamRequest returned after 8.5s: ch=nil err=<nil>
```

(In our gateway, which embeds `bifrost/core`, this hung every Anthropic-dialect streaming request and leaked one goroutine per request.)

### Expected behavior

A nil `*schemas.BifrostError` implies a usable (non-nil) stream channel. An empty stream should surface as a clean zero-chunk stream: `range` exits immediately, a receive yields `(nil, false)`.

### Actual behavior

`(nil, nil)` is returned; consumers block forever on the nil channel.

### Area

Core (Go)

### Version / environment

- `github.com/maximhq/bifrost/core` v1.7.3; `(nil, nil)` path also present on `main` @ 2952aae
- Go 1.26.4 (darwin/arm64 and linux/amd64)

### Proposed fix

In `executeRequestWithRetries`, when `CheckFirstStreamChunkForError` reports an empty stream, substitute a closed, non-nil channel for the public result and complete the request span synchronously (the provider goroutine has already exited, so nothing remains to complete a deferred span — this also matters for the large-payload passthrough placeholder, which is a closed channel by design). PR incoming.

akshaydeo:
## Summary

This PR cuts the `core v1.7.5`, `framework v1.5.5`, `transports v1.6.7`, and associated plugin releases. It delivers a broad set of provider correctness fixes, new routing and governance capabilities, streaming reliability improvements, and database connection controls.

## Changes

- **Virtual Key Budget Overrides**: Temporary budget overrides for virtual keys are now supported across the DB schema, governance store, admin APIs, and UI. Overrides add `override_amount` on top of `max_limit` and run for a finite number of reset cycles or until removed, controlled via `override_mode`, `override_cycles_total`, and `override_anchor_reset`.
- **Stream Truncation Detection**: Added `SSETruncation` interface and EOF handler support across all providers so upstream stream death surfaces as an error instead of a clean `[DONE]`. Fixes silent discard of dead streams.
- **Empty Stream Nil Channel**: `*StreamRequest` now returns a closed non-nil channel for empty streams instead of `(nil, nil)`, preventing consumers from hanging on a nil-channel receive (thanks [@kharkevich](https://github.com/kharkevich)!).
- **User Scope for Routing and Pricing**: Routing rules and pricing overrides can now be scoped to individual users via a `user_id` CEL variable, an enterprise user picker in the pricing overrides UI, and streaming request-type options in the routing rules UI.
- **Background Model Catalog Refresh**: Provider model lists are now re-fetched in the background on a configurable `live_models_sync_interval` (default 1 hour). Provider reload no longer wipes the catalog before refetching, so a transient failure cannot empty it.
- **Responses Retrieve Stream**: Added a retrieve-stream method for the Responses API across core, framework, governance, and logging plugins.
- **Routing Info Headers**: Routing info headers are now emitted for streaming responses, inference and integration APIs, and error/passthrough paths, with corrected behavior on streaming fallbacks.
- **Bedrock Mantle**: Added count-tokens API support, merged additional tools into the tools list, and set structured outputs to false for Claude models.
- **Connector Latency Data**: Bifrost latency and overhead duration are now exported to connectors (OTel and others).
- **Database Connection Controls**: Added `conn_max_idle_time` (default 5m) to config and logs stores, `cache_ttl` (default 60s) for password-command credential resolution, and `matview_refresh_timeout` to bound a single materialized view refresh pass.
- **Matview Improvements**: Added cached tokens to the materialized view, enabled on-the-fly refresh, corrected hybrid matview count to prevent boundary bucket over-counting, added `customer_id`/`business_unit_id` to `scopeProjection`, and reverted the read-path shape check that gated reads during rolling deploys.
- **OTel Export Timeout**: Added an `export_timeout` setting (default 5s) to bound slow or unreachable collector exports, which previously had no timeout on gRPC exports.
- **Multinode Override Counts**: Corrected override counts for multinode setups, resolving high CPU in governance rate-limit reset.
- **vLLM Responses Streaming**: vLLM responses-stream chunks and completion events are now forwarded instead of silently discarded, and stream truncation is handled correctly.
- **Gemini Tool Schema Constraints**: Valid integer constraints in tool schemas are no longer rejected with a 400 INVALID_ARGUMENT.
- **HuggingFace Model IDs**: Backfilled model IDs no longer duplicate the inference-provider segment.
- **Azure Responses API Version**: Removed the default preview api-version for GA Azure Responses endpoints.
- **GenAI API**: Fixed image search for the GenAI search tool, honored `IncludeServerSideToolInvocations`, and followed redirects for downloads in GenAI passthrough.
- **Tool Search Wire Shape**: Preserved the `tool_search` Responses API wire shape (thanks [@devonpmack](https://github.com/devonpmack)!).
- **Replicate Image Generation**: Input images are now handled in image generation requests.
- **SDK Compatibility**: `strict: null` is now converted to `false`.
- **New Guardrails**: Added Lakera and Repello Argus as guardrail integrations with configuration docs and UI branding.
- **Access Profile Config Schema**: `config.schema.json` now accepts `blacklisted_models`, `weight`, and `model_budgets` on access profile provider configs.
- **Async Entity Selectors**: Teams, customers, and virtual keys are loaded through async selector components; the customer list returns a server-computed `virtual_key_count` instead of a full preload.
- **Go 1.26.5**: Upgraded the toolchain and all builder images.

## Type of change

- [x] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Documentation
- [x] Chore/CI

## Affected areas

- [x] Core (Go)
- [x] Transports (HTTP)
- [x] Providers/Integrations
- [x] Plugins
- [x] UI (React)
- [ ] Docs

## How to test

```sh
# Core/Transports
go version  # should report go1.26.5
go test ./...

# UI
cd ui
pnpm i
pnpm test
pnpm build
```

**New config knobs to validate:**
- `live_models_sync_interval` — background provider model-list refresh interval (default `1h`, `0` disables).
- `conn_max_idle_time` — idle physical connection lifetime for config and logs stores (default `5m`).
- `cache_ttl` — TTL for password-command credential resolution (default `60s`).
- `matview_refresh_timeout` — maximum duration for a single materialized view refresh pass.
- `export_timeout` (OTel plugin) — maximum duration for a single OTel export call (default `5s`).
- `override_mode`, `override_cycles_total`, `override_anchor_reset` — virtual key budget override fields.

## Breaking changes

- [ ] Yes
- [x] No

## Related issues

- Closes [#4215](https://github.com/maximhq/bifrost/issues/4215): HuggingFace models show provider ID twice in `/v1/models`
- Closes [#4851](https://github.com/maximhq/bifrost/issues/4851): Governance rate-limit reset causes high CPU in `BumpRateLimitUsage`/`updateRateLimitReferences`
- Closes [#5329](https://github.com/maximhq/bifrost/issues/5329): `/api/logs` returns incorrect `total_count` for time ranges ≥ 24 hours
- Closes [#5433](https://github.com/maximhq/bifrost/issues/5433): `/genai` endpoint rejects valid `minLength`/`maxLength` in tool schemas
- Closes [#5504](https://github.com/maximhq/bifrost/issues/5504): vLLM streaming Responses API hangs forever, chunks silently discarded
- Closes [#5546](https://github.com/maximhq/bifrost/issues/5546): Upstream SSE stream death swallowed into a clean `[DONE]`
- Closes [#5551](https://github.com/maximhq/bifrost/issues/5551): `transports/bifrost-http/lib` test package does not compile (`MockConfigStore` missing `UpdateBudgetOverride`)
- Closes [#5552](https://github.com/maximhq/bifrost/issues/5552): Models added after boot stay invisible until restart
- Closes [#5554](https://github.com/maximhq/bifrost/issues/5554): Provider reload wipes the live model catalog before refetching
- Closes [#5555](https://github.com/maximhq/bifrost/issues/5555): `*StreamRequest` returns `(nil, nil)` for empty streams, consumers hang forever

## Security considerations

- `cache_ttl` for password-command credential resolution reduces how frequently credentials are re-fetched from external commands; operators should set an appropriate TTL for their secret rotation cadence.
- The `user_id` virtual keys filter is enterprise-only and fails closed on OSS builds.
- Budget override fields are gated behind the governance store and admin APIs; no public-facing exposure without authentication.

## Checklist


akshaydeo:
* **#5665** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/5665?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/maximhq/bifrost/5665?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `dev`


<h2></h2>

akshaydeo:
### Merge activity

* **Jul 30, 9:45 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/5665).
* **Jul 30, 9:46 AM UTC**: @akshaydeo merged this pull request with [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/5665).

PR #5701, opened 2026-07-30
[fix]: builtin disable reverts on restart, custom plugin config updates fail

### LINKED ISSUE #5699: [Bug]: disabling a builtin reverts on restart, and custom plugin config updates fail
### Prerequisites

### Description

Two plugin bugs, found on 1.6.3 and still reproducing on `dev`.

**Disabling a builtin does not survive a restart.** Turn off `logging`, `governance`, `compat` or `prompts` in the UI. It works, then they are back after a restart. `GET /api/plugins` still says `enabled: false, status: disabled` while `GET /api/plugins/loaded` lists them as running.

Those four load based on `ClientConfig` and never look at the stored plugin row. `telemetry`, `otel`, `semantic_cache` and `maxim` do check it and disable fine.

**Editing a custom plugin config 500s and changes nothing.** The new config lands in the DB, but the running instance keeps using the old one, so from the UI it just looks like config changes do nothing. The row also ends up with `path=NULL` and `isCustom=false`, so the plugin fails to load on the next boot too.

Cause is `path` being a `*string`: omitted and explicit null look identical, so the handler clears it, and the reload then goes down the builtin branch. The main plugins view sends `path` and works; the Update Plugin sheet does not and breaks.

### Steps to reproduce

Builtin disable:

1. `PUT /api/plugins/logging {"enabled":false,"config":{}}`
2. `GET /api/plugins/loaded` - `logging` is gone, as expected
3. Restart the gateway
4. `GET /api/plugins/loaded` - `logging` is back, while `GET /api/plugins` still reports it disabled

Custom plugin config:

1. Build a `.so` plugin exporting `GetName`, `Cleanup` and `PreLLMHook`
2. `POST /api/plugins {"name":"probe","path":"/app/probe.so","enabled":true,"config":{"tag":"one"}}`
3. `PUT /api/plugins/probe {"enabled":true,"config":{"tag":"two"}}` (no `path`, which is what the Update Plugin sheet sends)
4. Observe the 500, and that the plugin keeps running with `tag=one`

### Expected behavior

A disabled plugin stays disabled across restarts, and a config update applies to the running plugin.

### Actual behavior

Builtins reload on restart while the API reports them disabled; custom plugin config updates return 500 and leave the old config live.

### Affected area(s)

Transports (HTTP), Plugins, UI (React)

### Version

v1.6.3, also reproduced on `dev` at 02ceef30d

### Environment

```text
- OS: Linux x86_64
- Go: 1.26
- Deployment: docker
- Bifrost: transports v1.6.3 / core v1.6.3 / framework v1.4.3, and current dev
```

### Relevant logs/output

```shell
PUT /api/plugins/probe {"enabled":true,"config":{"tag":"two"}}

{"is_bifrost_error":false,"status_code":500,"error":{"message":"Plugin updated in database but failed to load: unknown built-in plugin: probe"}}
```

### Regression?

Not that I can tell, both look long standing.

### Severity

eyeveil:
Fixes #5699

## Summary

Two plugin bugs, both reproducible on `dev`.

**`enabled=false` ignored at startup.** `logging`, `governance`, `compat` and `prompts` decide whether to load from `ClientConfig` and never read the stored row, so disabling them lasts until the next restart. The API reports them disabled the whole time, since the response uses the row rather than what is loaded. Startup now checks the row too. Missing row still means enabled, so nobody loses `/metrics` or logging on upgrade.

**Partial updates wiped `path`.** `path` is a `*string`, so a field that was never sent looks the same as an explicit null and got cleared either way. On a custom plugin the reload then took the builtin branch and failed with `unknown built-in plugin`, returning 500 while the old instance kept running with the old config. The row was left with `path=NULL` and `IsCustom=false`, breaking the next boot as well. Presence is now read off the raw body, so omitted `path`/`placement`/`order` keep their stored values. Sending `path: ""` still detaches it.

The Update Plugin sheet also sends `path` back now. The main plugins view already did, which is why this only broke on one of the two screens.

## Type of change

- [x] Bug fix

## Affected areas

- [x] Transports (HTTP)
- [x] Plugins
- [x] UI (React)

## How to test

```sh
go test ./bifrost-http/handlers/... ./bifrost-http/server/...
```

`TestUpdatePlugin_PreservesPathWhenOmitted` fails if you revert the handler change.

Manually: install a `.so` plugin, then `PUT {"enabled":true,"config":{...}}` without `path`. Before, 500 and the old config stays live. After, 200 and `Init` runs again with the new config. For the builtin half, disable `telemetry` and `logging`, restart, check `/api/plugins/loaded`. `/metrics` goes 404 while telemetry is off, 200 after re-enabling.

## Breaking changes

- [x] No

A row saying `enabled=false` now actually takes effect at startup. If you disabled `logging` or `governance` and were counting on a restart bringing it back, it will not.

## Security considerations

`governance` carries auth enforcement, so disabling it now sticks across restarts. That was already the runtime behaviour, this just makes restart agree.

## Checklist

### DIFF
--- a/transports/bifrost-http/handlers/plugins.go
+++ b/transports/bifrost-http/handlers/plugins.go
@@ -1,6 +1,7 @@
 package handlers
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"errors"
@@ -430,7 +431,8 @@ func (h *PluginsHandler) updatePlugin(ctx *fasthttp.RequestCtx) {
 	// neither, and nulling `path` turns a custom plugin into a built-in lookup that
 	// then fails with "unknown built-in plugin" while the old instance keeps running.
 	// Presence is detected on the raw body because absent and explicit null both
-	// decode to a nil pointer.
+	// decode to a nil pointer. An explicit null is treated as "not provided" as well;
+	// an empty string is the documented way to detach a path.
 	var rawFields map[string]json.RawMessage
 	if err := json.Unmarshal(ctx.PostBody(), &rawFields); err != nil {
 		logger.Error("failed to unmarshal update plugin request fields: %v", err)
@@ -438,13 +440,13 @@ func (h *PluginsHandler) updatePlugin(ctx *fasthttp.RequestCtx) {
 		return
 	}
 	if existingPlugin != nil {
-		if _, sent := rawFields["path"]; !sent {
+		if !fieldProvided(rawFields, "path") {
 			request.Path = existingPlugin.Path
 		}
-		if _, sent := rawFields["placement"]; !sent {
+		if !fieldProvided(rawFields, "placement") {
 			request.Placement = existingPlugin.Placement
 		}
-		if _, sent := rawFields["order"]; !sent {
+		if !fieldProvided(rawFields, "order") {
 			request.Order = existingPlugin.Order
 		}
 	}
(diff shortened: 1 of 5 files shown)


eyeveil:
Good catch on the null path, that was a real hole. An explicit `"path": null` was counted as provided, so it still wrote a nil path and sent a custom plugin down the built-in branch, same 500 as the omitted case.

Fixed in b66fae8e9, though slightly differently than suggested: null is now treated as "not provided" and keeps the stored value, rather than returning a 400. An empty string is already the documented detach signal, and rejecting null outright would break callers who currently send it. Added `TestUpdatePlugin_PreservesPathWhenNull`, which fails against the old presence check.

@coderabbitai review

eyeveil:
recheck

PR #5702, opened 2026-07-30
[fix]: classify dynamic plugins by the hooks they actually export

### LINKED ISSUE #5700: [Bug]: dynamic .so plugins are classified as every plugin type
### Prerequisites

### Description

`DynamicPlugin` satisfies `LLMPlugin`, `MCPPlugin` and `HTTPTransportPlugin` structurally no matter what the `.so` actually exports, and `rebuildInterfaceCaches` / `InferPluginTypes` classify with plain type assertions. So every dynamic plugin is treated as all three.

A plugin exporting only `PreLLMHook` and `HTTPTransportPreHook` gets reported as `["llm","mcp","http"]`, runs no-op hooks on every request, and on removal has `Cleanup()` called twice, once per list. That last one is a double close if you release anything in `Cleanup`.

`framework/plugins` already has `AsLLMPlugin` / `AsMCPPlugin` / `AsHTTPTransportPlugin` that check the hook pointers. The two call sites just do not use them.

### Steps to reproduce

1. Build a `.so` plugin exporting `GetName`, `Cleanup`, `PreLLMHook` and `HTTPTransportPreHook`, but no MCP hooks. Have `Cleanup` log a line.
2. `POST /api/plugins {"name":"probe","path":"/app/probe.so","enabled":true,"config":{}}`
3. `GET /api/plugins/probe` and look at `status.types`
4. `PUT /api/plugins/probe {"enabled":false,"path":"/app/probe.so","config":{}}` and count the `Cleanup` lines

### Expected behavior

Types match the exported symbols, and `Cleanup()` is called once when the plugin is removed.

### Actual behavior

Types are reported as `["llm","mcp","http"]` regardless of exported symbols, and `Cleanup()` is called twice.

### Affected area(s)

Transports (HTTP), Plugins

### Version

v1.6.3, also reproduced on `dev` at 02ceef30d

### Environment

```text
- OS: Linux x86_64
- Go: 1.26
- Deployment: docker, custom .so plugin loading
- Bifrost: transports v1.6.3 / core v1.6.3 / framework v1.4.3, and current dev
```

### Relevant logs/output

```shell
# probe.so exports PreLLMHook and HTTPTransportPreHook, no MCP hooks
GET /api/plugins/probe -> "status":{"types":["llm","mcp","http"]}

# on disable, the plugin's Cleanup runs once per list it was added to
CLEANUP
CLEANUP
```

### Regression?

Not that I can tell.

### Severity

eyeveil:
Fixes #5700

## Summary

`DynamicPlugin` satisfies all three hook interfaces regardless of which symbols the `.so` exports, and both classification sites use plain type assertions. So every dynamic plugin lands in all three interface caches, runs no-op hooks per request, reports types it does not implement, and gets `Cleanup()` called once per list on removal, which is a double close for anything that frees resources there.

Switched both sites to the `As*` helpers in `framework/plugins`, which also check the hook pointers. That is what they are there for. Builtins are concrete types, so nothing changes for them.

## Type of change

- [x] Bug fix

## Affected areas

- [x] Transports (HTTP)
- [x] Plugins

## How to test

Install a `.so` exporting `GetName`, `Cleanup` and `PreLLMHook` but no MCP hooks, then disable it. Before: `Cleanup` runs twice and the API lists `mcp` in its types. After: once, and the types match the exported symbols.

## Breaking changes

- [x] No

## Checklist

- [ ] I added/updated tests where appropriate - `DynamicPlugin` hook fields are unexported, so a test would need to live in `framework/plugins` next to the `As*` helpers. Happy to add one there if you would like.

### DIFF
--- a/transports/bifrost-http/lib/config.go
+++ b/transports/bifrost-http/lib/config.go
@@ -5131,13 +5131,17 @@ func (c *Config) rebuildInterfaceCaches() {
 	var httpTransport []schemas.HTTPTransportPlugin
 
 	for _, p := range *basePlugins {
-		if llmPlugin, ok := p.(schemas.LLMPlugin); ok {
+		// Use the framework's As* helpers rather than bare type assertions: a .so plugin
+		// is wrapped in a DynamicPlugin that satisfies every hook interface even when the
+		// .so exports none of those symbols, so bare assertions put every dynamic plugin
+		// into all three caches and invoke no-op hooks on every request.
+		if llmPlugin := plugins.AsLLMPlugin(p); llmPlugin != nil {
 			llm = append(llm, llmPlugin)
 		}
-		if mcpPlugin, ok := p.(schemas.MCPPlugin); ok {
+		if mcpPlugin := plugins.AsMCPPlugin(p); mcpPlugin != nil {
 			mcp = append(mcp, mcpPlugin)
 		}
-		if httpPlugin, ok := p.(schemas.HTTPTransportPlugin); ok {
+		if httpPlugin := plugins.AsHTTPTransportPlugin(p); httpPlugin != nil {
 			httpTransport = append(httpTransport, httpPlugin)
 		}
 		if cm, ok := p.(schemas.ConfigMarshallerPlugin); ok {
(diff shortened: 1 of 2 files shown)


eyeveil:
recheck

PR #5769, opened 2026-08-02
scope semanticache poisoning scoped to cache key

akshaydeo:
## Summary

Closes a cross-tenant cache leak in the semantic cache plugin where two requests carrying different virtual keys but the same `cache_key`/`default_cache_key` could be served each other's cached responses. Additionally tightens the per-request threshold override so callers can only raise the similarity bar, never lower it below the operator-configured floor.

## Changes

- **Virtual-key cache partitioning**: `resolveCacheKey` now automatically scopes the resolved cache bucket to the authenticated virtual key ID (prefixed as `vk:<id>:<raw_key>`) when one is present on the request. This is a single choke point shared by both the read path (`PreLLMHook`, direct and semantic modes) and the write path (`PostLLMHook`), so the isolation is enforced everywhere without any per-call-site changes. Requests without a virtual key retain the existing un-scoped behavior.
- **Threshold floor enforcement**: Threshold resolution is extracted into `resolveCacheThreshold`. A per-request override below the plugin's configured threshold is floored at the configured value rather than accepted — callers can make matching stricter but cannot loosen the operator's similarity gate. The inline threshold logic in `search.go` is replaced with a call to this method.
- **Documentation**: Added a `<Note>` to the semantic caching docs explaining the automatic VK-scoped partitioning behavior and updated the `x-bf-cache-threshold` header description to reflect the floor semantics.
- **Tests**: `TestResolveCacheKey` covers the no-VK fallback, explicit header, empty VK, same-tenant hit, and the core two-VK collision prevention case. `TestResolveCacheThreshold` covers no override, defeat-by-zero, below-floor, above-floor, exact-match, and non-float override cases.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [ ] Core (Go)
- [ ] Transports (HTTP)
- [ ] Providers/Integrations
- [x] Plugins
- [ ] UI (React)
- [x] Docs

## How to test

```sh
go test ./plugins/semanticcache/... -run "TestResolveCacheKey|TestResolveCacheThreshold" -v
```

To validate end-to-end isolation, send two requests with different virtual keys but the same `x-bf-cache-key` header. The second request must not receive the first tenant's cached response — it should produce a fresh LLM call and populate its own scoped bucket.

## Breaking changes

- [ ] Yes
- [x] No

The VK-scoped key format (`vk:<id>:<raw_key>`) is a new internal bucket name. Existing deployments without virtual key enforcement are unaffected. Deployments using virtual keys will see cache misses on first deploy as existing un-scoped entries are not migrated — this is intentional and the safe behavior given the security motivation.

## Security considerations

The primary motivation is preventing cross-tenant cache poisoning: a shared `default_cache_key` (a common deployment pattern) previously allowed one virtual key's cached response — potentially containing PII or sensitive tool-call payloads — to be served to a different virtual key's request. The VK-scoped partitioning closes this at the single key-resolution choke point, covering both read and write paths simultaneously.

## Checklist


akshaydeo:
* **#5769** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/5769?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/maximhq/bifrost/5769?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `dev`


<h2></h2>

PR #5772, opened 2026-08-02
protect proxy config from ssrf

akshaydeo:
## Summary

Closes an SSRF vulnerability in the MCP HTTP/SSE client registration path. Previously, an unauthenticated caller (default-open posture, no admin password set) could register an HTTP or SSE MCP client pointing at loopback, private-network, link-local, or cloud metadata addresses (e.g. `169.254.169.254`). Additionally, even for authenticated callers, the MCP HTTP transport fell back to the mcp-go library's own default HTTP client when no TLS config was provided, which carried no dial guard whatsoever.

## Changes

- **Handler-layer gate (`rejectPrivateMCPTargetIfAuthBypassed`)**: Added a pre-registration check in `addMCPClient` that resolves the `connection_string` hostname and rejects the request with HTTP 403 if any resolved IP is non-public and the caller reached the endpoint with no credential check (`BifrostContextKeyAuthBypassed`). Authenticated callers retain the documented ability to point MCP clients at local or private servers.

- **New `BifrostContextKeyAuthBypassed` context key**: Added to distinguish requests that were let through with no credential check at all from genuinely authenticated sessions (both previously set `IsLocalAdminContextKey`). The auth middleware sets this key in the fail-open branch. This scoping ensures only the unauthenticated path is restricted.

- **`buildTLSHTTPClient` always returns a guarded client**: Removed the early `return nil, nil` when `tlsCfg == nil`. The function now always constructs a transport routed through `network.PrivateNetworkDialContext`, which blocks link-local and unspecified addresses at dial time regardless of TLS configuration. TLS customization is layered on top when provided.

- **`network.PrivateNetworkDialContext`**: Added a new dialer variant that, unlike `SSRFSafeDialContext`, permits loopback and private-network targets (the documented primary use case for MCP HTTP clients) while still blocking link-local addresses (including `169.254.169.254`) and unspecified addresses. Includes the same DNS-rebinding protection as `SSRFSafeDialContext`.

- **Integration test fix (`TestMain`)**: The MCP integration tests connect to loopback-bound `httptest.Server` instances. The new `PrivateNetworkDialContext` guard in `buildTLSHTTPClient` would have blocked these. `TestMain` now installs an unguarded dialer for the test package only via `SetDialContextForTests`, with a comment explaining the scope.

- **Documentation**: Added a `<Note>` to `docs/mcp/connecting-to-servers.mdx` explaining that unauthenticated callers cannot register HTTP/SSE clients pointing at private addresses, and directing users to either enable auth or define the client in `config.json`.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [x] Core (Go)
- [x] Transports (HTTP)
- [ ] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [x] Docs

## How to test

```sh
go test ./core/network/... ./core/mcp/... ./transports/bifrost-http/handlers/... ./core/internal/mcptests/...
```

**Unauthenticated loopback rejected:**
Start Bifrost with no admin password configured. Attempt to register an HTTP MCP client pointing at `http://127.0.0.1:3001/mcp` via `POST /api/mcp/client`. Expect HTTP 403.

**Authenticated loopback allowed:**
Start Bifrost with an admin password set and authenticate. Register the same client. Expect the registration to succeed.

**Link-local blocked at dial time:**
Attempt to register an HTTP MCP client pointing at `http://169.254.169.254/latest/meta-data/` as an authenticated caller. The handler-layer check passes (public-IP check does not apply to authenticated callers), but the dial-time guard in `PrivateNetworkDialContext` blocks the connection.

**Public targets unaffected:**
Register an SSE MCP client pointing at a public internet address as an unauthenticated caller. Expect the registration to succeed.

## Breaking changes

- [x] No

The only behavioral change for existing deployments is that MCP HTTP/SSE clients registered by unauthenticated callers pointing at private/loopback/link-local addresses are now refused. Authenticated callers and all STDIO clients are unaffected. Deployments with auth enabled see no change in behavior.

## Security considerations

This PR addresses an SSRF vulnerability. Without this fix, an unauthenticated caller on a Bifrost instance running with the default-open posture (no admin password) could cause the gateway to dial arbitrary internal addresses by registering a crafted MCP HTTP/SSE client. The fix applies defense in depth at two layers: the HTTP handler (pre-registration IP check for unauthenticated callers) and the MCP transport dial function (link-local/unspecified block for all callers, including authenticated ones).

## Checklist


akshaydeo:
* **#5772** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/5772?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/maximhq/bifrost/5772?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `dev`


<h2></h2>

PR #5778, opened 2026-08-02
unauthenticated proxy blocks

akshaydeo:
## Summary

Setting the outbound proxy URL is at least as sensitive as setting a provider's base URL: once enabled, the proxy replaces the HTTP client's Dial function entirely, routing every outbound provider request — including `Authorization` and `x-api-key` headers — through the attacker-chosen host. Without this fix, an unauthenticated caller on the network could set the proxy URL while dashboard auth is disabled or unconfigured, since the auth middleware's fail-open branch previously granted the same `IsLocalAdminContextKey` flag as a genuinely authenticated session.

## Changes

- Introduced `BifrostContextKeyAuthBypassed` context key, set exclusively in the auth middleware's fail-open branch (when `authConfig == nil` or auth is disabled), distinct from `IsLocalAdminContextKey` which is also true for real authenticated sessions.
- Added `isAuthBypassed` helper in the config handler to read this key.
- Gated the "enable proxy with URL" path in `updateProxyConfig` behind `isAuthBypassed`: returns `403 Forbidden` if the request was not genuinely authenticated, while the rest of the management API remains accessible under the default-open posture.
- The pre-existing empty-URL `400` validation intentionally runs before the auth gate so a malformed request does not leak whether dashboard auth is configured.
- Added OpenAPI documentation for the new `403` response on `PUT /proxy-config`, including an updated endpoint description explaining the restriction.
- Added `config_proxy_ssrf_test.go` with two tests: one proving an auth-bypassed caller cannot set the proxy URL, and one proving the empty-URL `400` still fires first for an invalid request.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [x] Core (Go)
- [x] Transports (HTTP)
- [ ] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [x] Docs

## How to test

```sh
go test ./transports/bifrost-http/handlers/... -run TestUpdateProxyConfig
```

**Expected outcomes:**
- `TestUpdateProxyConfig_UnauthenticatedWriteRejected` — `PUT /proxy-config` with `enabled: true` and a URL while `BifrostContextKeyAuthBypassed` is set returns `403`.
- `TestUpdateProxyConfig_EmptyURLCheckRunsBeforeAuthGate` — `PUT /proxy-config` with `enabled: true` and an empty URL while `BifrostContextKeyAuthBypassed` is set returns `400`, not `403`.

To verify end-to-end: disable dashboard authentication, then attempt to `PUT /proxy-config` with a valid proxy URL. The request should be rejected with `403`. Disabling the proxy (setting `enabled: false`) should remain unrestricted.

## Breaking changes

- [ ] Yes
- [x] No

## Security considerations

This closes an SSRF-adjacent privilege escalation: without dashboard authentication configured, any network-reachable caller could previously redirect all outbound provider traffic — including API keys — through an arbitrary host. The fix ensures that enabling the outbound proxy always requires a genuinely authenticated admin session, independent of the global fail-open posture applied to the rest of the management API.

## Checklist


akshaydeo:
* **#5778** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/5778?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/maximhq/bifrost/5778?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `dev`


<h2></h2>

PR #5811, opened 2026-08-03
retain tool cache_control for bedrock converse

### LINKED ISSUE #5629: [Bug]: Bedrock invoke routes drop tool `cache_control` when converting to Converse, disabling prompt caching (full prefix re-billed every turn)
## Prerequisites

## Description

On the Bedrock invoke routes (`POST /bedrock/model/{id}/invoke` and `.../invoke-with-response-stream`), a `cache_control` marker in an Anthropic-native request body is **silently discarded** while the request is converted into a Converse call. Bedrock therefore receives a `toolConfig` with no cache breakpoint, nothing is cached, and the entire prefix is billed as fresh input tokens on **every turn**.

To be precise about routing, since it matters for reading the table below: these ingress routes accept an Anthropic-native (InvokeModel-shaped) body, but the **egress call is Converse** — `getModelPathAndRegion(ctx, "converse", …)` in `Responses()`/`ChatCompletion()`, `"converse-stream"` for streams. A body with a `messages` array on `/invoke` is detected as a Responses request, converted by `ToBedrockConverseRequest()`, and dispatched accordingly. The bug is in that ingress conversion, not in Converse and not in Bedrock.

This is a cost/performance bug only — responses are correct, so it produces no signal other than the token accounting. It affects any client that sets a `cache_control` breakpoint on this route, including Claude Code and anything else built on `AnthropicBedrock`, which send a tool array with a breakpoint and re-send that same prefix on every turn of a session: each turn pays full input price for the whole tool block instead of the cache-read price.

The identical marker is honored on three other arms, which localises the defect to the invoke ingress conversion. Cache spellings differ by request dialect: an Anthropic-native body marks a breakpoint with a per-tool `cache_control` and reports `cache_creation_input_tokens`/`cache_read_input_tokens`; a Bedrock-native Converse body uses a positional `{"cachePoint":{"type":"default"}}` in `toolConfig.tools` and reports `cacheWriteInputTokens`/`cacheReadInputTokens`. All numbers are from the reproduction script below:

| Arm | Request dialect | Turn 1 | Turn 2 |
|---|---|---|---|
| Direct to Bedrock, no gateway (`InvokeModel`) | Anthropic / `cache_control` | `input_tokens=330`, `cache_creation=8336` | `input_tokens=330`, **`cache_read=8336`** |
| Bifrost `/anthropic/v1/messages` | Anthropic / `cache_control` | `input_tokens=330`, `cache_creation=8336` | `input_tokens=330`, **`cache_read=8336`** |
| Bifrost `/bedrock/.../converse` | Bedrock / `cachePoint` | `inputTokens=330`, `cacheRead=8336` | `inputTokens=330`, **`cacheRead=8336`** |
| **Bifrost `/bedrock/.../invoke`** | Anthropic / `cache_control` | **`input_tokens=8666`**, no cache fields | **`input_tokens=8666`**, no cache fields |

`8666 = 330 + 8336` — exactly the prefix that should have been cached, re-billed in full.

**Row 2 is the decisive control:** the same Anthropic-native dialect, carrying the same `cache_control`, reaching the same Converse egress through the same gateway — and it caches. So the marker is translatable on this path, the payload is valid, and neither Converse nor Bedrock is at fault; only the invoke ingress parser loses it. Row 1 confirms Bedrock honors the marker when it receives one; row 3 excludes the `/bedrock` surface itself.

**The marker is inert, not merely unreported.** `input_tokens` is byte-identical (8666) whether `cache_control` is present or absent, which excludes "caching worked upstream and the usage fields just weren't surfaced". A `cache_control` on a `system` block is dropped the same way.

### Root cause

`convertAnthropicTools` in `core/providers/bedrock/invoke.go` (L819–845) copies only three fields per tool:

```go
spec := &BedrockToolSpec{}
if name, ok := toolMap["name"].(string); ok {
    spec.Name = name
}
if desc, ok := toolMap["description"].(string); ok {
    spec.Description = &desc
}
if inputSchema, ok := toolMap["input_schema"]; ok {
    inputSchemaBytes, _ := providerUtils.MarshalSorted(inputSchema)
    spec.InputSchema = BedrockToolInputSchema{JSON: json.RawMessage(inputSchemaBytes)}
}
bedrockTools = append(bedrockTools, BedrockTool{ToolSpec: spec})
```

`cache_control` is never read, and no `{"cachePoint": {"type": "default"}}` element is appended to `toolConfig.tools` — even though `ToBedrockResponsesRequest` does exactly that for the same logical input arriving via the Anthropic surface (appending a `BedrockTool{CachePoint: …}` after the marked tool). `invoke.go` L282 routes **both** unary and streaming through this function, so both are affected.

This is confirmable without any network call. Unmarshalling an Anthropic-native invoke body that carries `cache_control` on a tool and on a `system` block, then calling `ToBedrockConverseRequest()`, yields a `toolConfig.tools` of length 2 for 2 input tools — a breakpoint would make 3 — with `CachePoint == nil` on every tool and on the system message. Carrying the same request one step further through `ToBifrostResponsesRequest()`, the intermediate `BifrostResponsesRequest` has `Tools[0].CacheControl == nil`, so the marker is gone before the shared egress builder runs. The same tool through `AnthropicMessageRequest.ToBifrostResponsesRequest()` retains `CacheControl != nil`. Two ingress parsers, one egress builder, and only one of them carries the marker across.

Note this is adjacent to **#5560** — same file, same ingress conversion, same class of "Anthropic-native input silently dropped" — but a distinct field and code path. PR #5596 (`BedrockContentBlock.UnmarshalJSON`) does not touch `convertAnthropicTools`, so it will not fix this.

## Steps to reproduce

Point Bifrost at a Bedrock deployment of a caching-capable Anthropic model (below uses `us.anthropic.claude-haiku-4-5-20251001-v1:0`), then send the **same** request twice. Turn 1 should create a cache entry, turn 2 should read it.

The tool array must exceed the model's minimum cacheable prefix (2048 tokens for Haiku), so generate enough verbose tools — this script emits ~8.3k tokens of tools with a breakpoint on the last one:

```bash
MODEL=us.anthropic.claude-haiku-4-5-20251001-v1:0

python3 - > /tmp/body.json <<'PY'
import json
blurb = ("A tool that performs a well-documented operation with many described "
         "parameters and considerable descriptive text. ") * 10
tools = [{
    "name": f"cachetool_{i}",
    "description": blurb,
    "input_schema": {
        "type": "object",
        "properties": {
            "alpha": {"type": "string", "description": "The alpha parameter, fully described."},
            "beta":  {"type": "integer", "description": "The beta parameter, fully described."},
        },
        "required": ["alpha"],
    },
} for i in range(30)]
tools[-1]["cache_control"] = {"type": "ephemeral"}   # <-- the breakpoint that gets dropped
print(json.dumps({
    "anthropic_version": "bedrock-2023-05-31",
    "max_tokens": 16,
    "tools": tools,
    "messages": [{"role": "user", "content": "Say OK."}],
}))
PY

for turn in 1 2; do
  echo "--- turn $turn ---"
  curl -s -X POST "http://localhost:8080/bedrock/model/$MODEL/invoke" \
    -H 'content-type: application/json' --data @/tmp/body.json \
    | python3 -c 'import json,sys; print(json.load(sys.stdin)["usage"])'
  sleep 2
done
```

Observed (both turns identical, no cache fields at all):

```
--- turn 1 ---
{'input_tokens': 8666, 'output_tokens': 5}
--- turn 2 ---
{'input_tokens': 8666, 'output_tokens': 5}
```

**Control A — the same body through `/anthropic/v1/messages`** (swap `"model"` in for `"anthropic_version"`) caches correctly: turn 1 `cache_creation_input_tokens=8336`, turn 2 `cache_read_input_tokens=8336`, `input_tokens=330`. **This is the control that matters most:** same gateway, same Anthropic-native dialect, same `cache_control`, and the same Converse egress — so the marker is translatable on this path and the request is valid. The invoke ingress route is the only difference.

**Control B — the same body sent directly to Bedrock, no gateway** (`boto3` `invoke_model`, same `modelId`, same JSON) also caches correctly, with identical numbers — confirming Bedrock honors the marker when it arrives.

**Control C — the marker is inert.** Delete the `cache_control` line and re-run the `/invoke` case: `input_tokens` is still exactly `8666`. Identical billing with and without the marker shows it never reached Bedrock.

**Control D — native `Converse` through the same gateway caches.** Sending the same 30 tools as `toolConfig.tools` with a trailing `{"cachePoint": {"type": "default"}}` to `/bedrock/model/$MODEL/converse` yields `cacheReadInputTokens=8336`. Only the invoke ingress route is affected.

## Expected behavior

`cache_control` on an Anthropic-native body sent to the invoke routes is honored — the conversion appends a `cachePoint` to `toolConfig.tools` so the prefix is cached, and turn 2 reports a cache read (matching `/anthropic` and native Converse).

## Actual behavior

`cache_control` is silently dropped; nothing is cached and the full prefix is re-billed as fresh input tokens on every request. No warning or error is emitted.

## Affected area(s)

- Core (Go)

## Version

Reproduced on `transports/v1.6.3`. `core/providers/bedrock/invoke.go` is unchanged through `transports/v1.6.6` (identical blob), and `convertAnthropicTools` still contains no reference to `cache_control` on `dev` at `core/v1.7.4` — so this is present on latest, not a pin-age artifact, and not a recent regression either. `core/providers/bedrock/cache_points_test.go` covers only the `stripCachePoints*` helpers and never exercises `convertAnthropicTools`, so no existing test guards this path.

## Environment

```
- Bifrost: transports v1.6.3 (invoke.go unchanged through v1.6.6; also verified on dev @ core v1.7.4)
- Provider: AWS Bedrock, us-west-2
- Model: us.anthropic.claude-haiku-4-5-20251001-v1:0
- Route: POST /bedrock/model/{id}/invoke and /invoke-with-response-stream
- Client: raw curl with an Anthropic-native body (also reproduces via AnthropicBedrock SDK / Claude Code)
```

## Regression?

Not a regression — `convertAnthropicTools` has never read `cache_control`.

## Suggested fix

In `convertAnthropicTools`, carry the marker across. Anthropic attaches `cache_control` to an individual tool, whereas Converse expects a positional `cachePoint` element in the `tools` list, so the natural translation is: when a source tool carries `cache_control`, append `{"cachePoint": {"type": "default", ...ttl}}` **after** that tool's `toolSpec`.

Two details worth deciding deliberately:

- The provider already has `newBedrockCachePoint(ttl)` and a `BedrockModelSupportsCachePoints(model)` capability check used elsewhere (`chat.go`, `responses.go`), plus a Nova-family exclusion. Reusing those keeps behavior consistent with the Converse path rather than introducing a second policy.
- `system` blocks on this route drop `cache_control` too (`parseSystemMessages` unmarshals into `BedrockSystemMessage`, which only has a `cachePoint` field), and the Converse `system` array takes `cachePoint` the same way — worth fixing together, since a Claude Code-shaped request marks both.

Happy to open a PR with a regression test asserting turn-2 cache-read parity between `/anthropic` and `/bedrock/.../invoke` for the same body.

## Additional observation (same route) — streaming emits no `input_tokens`

While measuring the above: `POST /bedrock/model/{id}/invoke-with-response-stream` emits **no input token count at all**. Decoding the event stream with botocore's `EventStreamBuffer`, the `message_start` frame carries no `usage` and only `message_delta` appears, with `output_tokens` alone — the raw response bytes contain no occurrence of `input_tokens` or `usage` beyond that. The non-streaming `invoke` route on the same payload reports `input_tokens` normally, and **direct Bedrock `invoke_model_with_response_stream` on the same body does return** `message_delta.usage` with `input_tokens`, `cache_creation_input_tokens`, `cache_read_input_tokens`, plus `amazon-bedrock-invocationMetrics.inputTokenCount`. So the data exists upstream and is dropped in translation.

Source: `toAnthropicInvokeStreamBytes` in `core/providers/bedrock/invoke.go` builds the `message_delta` usage map with a single key (L1407–1411):

```go
if resp.Response != nil && resp.Response.Usage != nil {
    messageDelta["usage"] = map[string]interface{}{
        "output_tokens": resp.Response.Usage.OutputTokens,
    }
}
```

`message_start` likewise never sets `usage`. Anthropic's streaming format carries input tokens on `message_start.message.usage` and cumulative usage on `message_delta.usage`, so both are available slots.

Consequences: clients cannot compute request cost from the stream at all, and the `cache_control` bug above therefore produces no observable signal on the streaming route (`AnthropicBedrock` always streams). This looks like the invoke-route analogue of #4746, which fixed the same class of omission for Converse streaming. Happy to split this into its own issue if preferred.

akshaydeo:
## Summary

Prompt caching markers (`cache_control`) on Anthropic-native invoke requests were silently dropped when routing through Bedrock's Converse API. This meant cache breakpoints set on system blocks and tools never reached the model, and cache token counts (`cache_creation_input_tokens`, `cache_read_input_tokens`) were never surfaced in either streaming or non-streaming `/invoke` responses.

Closes #5629

## Changes

- **System block cache_control translation**: `parseSystemMessages` now detects a per-block `cache_control` field on Anthropic-dialect system entries and appends a positional `cachePoint` entry immediately after the marked block, matching the behavior `convertSystemMessages` already applies on the Bifrost→Bedrock egress path.
- **Tool cache_control translation**: `convertAnthropicTools` now detects `cache_control` on each Anthropic-native tool map and appends a positional `cachePoint` entry after the marked tool. Nova-family exclusion is handled downstream by the shared `ToBifrostResponsesRequest` builder, so no additional gating is needed here.
- **Cache token fields in non-streaming responses**: `toBedrockInvokeAnthropicResponse` now routes usage through `buildBedrockTokenUsage` and populates `cache_creation_input_tokens` and `cache_read_input_tokens` on `BedrockInvokeMessagesUsage`, which gains those two `omitempty` fields.
- **Cache token fields in streaming responses**: `toAnthropicInvokeStreamBytes` now includes `input_tokens` and, when non-zero, `cache_read_input_tokens` and `cache_creation_input_tokens` in the `message_delta` usage object. Because Bedrock Converse only reports usage on the terminal stream event, all token counts are consolidated there rather than split across `message_start`/`message_delta` as native Anthropic does.
- **`cacheControlTTL`** **helper**: Extracts the optional `"ttl"` string from an Anthropic-dialect `cache_control` map so it can be forwarded to `newBedrockCachePoint`, which already enforces Bedrock's supported TTL allow-list.
- **New test file** **`invoke_test.go`**: Covers no-cache-control regression, cache_control on tools and system blocks, TTL forwarding and fallback, Nova-family exclusion, the full invoke→Converse→Bifrost pipeline, streaming usage fields, and non-streaming usage fields.

## Type of change

- [x] Bug fix

## Affected areas

- [x] Core (Go)
- [x] Providers/Integrations

## How to test

```sh
go test ./core/providers/bedrock/... -v -run "TestConvertAnthropicTools|TestParseSystemMessages|TestToBedrockConverseRequest_InvokeCache|TestToAnthropicInvokeStreamBytes|TestToBedrockInvokeAnthropicResponse"
```

Expected: all new tests pass. Existing Bedrock tests remain unaffected.

To validate end-to-end, send an Anthropic-native `/invoke` request with `cache_control` on a system block or tool to a Claude model via Bedrock and confirm:

- The model receives the cache breakpoint (observable via `cache_creation_input_tokens` > 0 on the first turn).
- Subsequent identical requests show `cache_read_input_tokens` > 0 and reduced `input_tokens`.
- Streaming responses include all three token fields in the `message_delta` event.

## Breaking changes

- [x] No

## Related issues

Closes #5629

## Security considerations

None. Changes are limited to token-count field mapping and cache marker translation within the Bedrock provider.

## Checklist

### DIFF
--- a/core/providers/bedrock/invoke.go
+++ b/core/providers/bedrock/invoke.go
@@ -801,6 +801,13 @@ func (r *BedrockInvokeRequest) parseSystemMessages() []BedrockSystemMessage {
 					continue
 				}
 				result = append(result, msg)
+
+				// Anthropic-dialect blocks mark a breakpoint with a per-block "cache_control"
+				// instead of Converse's positional cachePoint entry; translate it the same
+				// way convertSystemMessages does for the Bifrost->Bedrock egress direction.
+				if cacheControl, ok := m["cache_control"].(map[string]interface{}); ok {
+					result = append(result, BedrockSystemMessage{CachePoint: newBedrockCachePoint(cacheControlTTL(cacheControl))})
+				}
 			}
 		}
 		return result
@@ -811,6 +818,15 @@ func (r *BedrockInvokeRequest) parseSystemMessages() []BedrockSystemMessage {
 		return msgs
 	}
 
+	return nil
+}
+
+// cacheControlTTL extracts the optional "ttl" string from an Anthropic-dialect
+// cache_control map, e.g. {"type": "ephemeral", "ttl": "1h"}.
+func cacheControlTTL(cacheControl map[string]interface{}) *string {
+	if ttl, ok := cacheControl["ttl"].(string); ok {
+		return &ttl
+	}
 	return nil
 }
 
@@ -842,6 +858,15 @@ func (r *BedrockInvokeRequest) convertAnthropicTools() *BedrockToolConfig {
 		}
 
 		bedrockTools = append(bedrockTools, BedrockTool{ToolSpec: spec})
+
+		// Anthropic attaches cache_control to the tool itself; Converse instead expects a
+		// positional cachePoint element in toolConfig.tools. The downstream shared builder
+		// (BedrockConverseRequest.ToBifrostResponsesRequest) already knows how to read this —
+		// see responses.go's tool.CachePoint handling — including the Nova-family exclusion,
+		// so no extra gating is needed here.
+		if cacheControl, ok := toolMap["cache_control"].(map[string]interface{}); ok {
+			bedrockTools = append(bedrockTools, BedrockTool{CachePoint: newBedrockCachePoint(cacheControlTTL(cacheControl))})
+		}
 	}
 
 	if len(bedrockTools) == 0 {
@@ -1101,9 +1126,12 @@ func toBedrockInvokeAnthropicResponse(resp *schemas.BifrostResponsesResponse, mo
 
 	// Usage
 	if resp.Usage != nil {
+		usage := buildBedrockTokenUsage(resp.Usage)
 		result.Usage = &BedrockInvokeMessagesUsage{
-			InputTokens:  resp.Usage.InputTokens,
-			OutputTokens: resp.Usage.OutputTokens,
+			InputTokens:              usage.InputTokens,
+			OutputTokens:             usage.OutputTokens,
+			CacheReadInputTokens:     usage.CacheReadInputTokens,
+			CacheCreationInputTokens: usage.CacheWriteInputTokens,
 		}
 	}
 
@@ -1405,9 +1433,23 @@ func toAnthropicInvokeStreamBytes(resp *schemas.BifrostResponsesStreamResponse) 
 			},
 		}
 		if resp.Response != nil && resp.Response.Usage != nil {
-			messageDelta["usage"] = map[string]interface{}{
-				"output_tokens": resp.Response.Usage.OutputTokens,
+			// Bedrock Converse only reports usage on the terminal event (unlike native
+			// Anthropic, which also puts input_tokens on message_start) — see
+			// buildBedrockTokenUsage's doc comment. This is the only stream event where
+			// Bifrost has input/cache token counts for a Bedrock-backed call, so all of
+			// them are reported here rather than split across message_start/message_delta.
+			usage := buildBedrockTokenUsage(resp.Response.Usage)
+			usageMap := map[string]interface{}{
+				"input_tokens":  usage.InputTokens,
+				"output_tokens": usage.OutputTokens,
 			}
+			if usage.CacheReadInputTokens > 0 {
+				usageMap["cache_read_input_tokens"] = usage.CacheReadInputTokens
+			}
+			if usage.CacheWriteInputTokens > 0 {
+				usageMap["cache_creation_input_tokens"] = usage.CacheWriteInputTokens
+			}
+			messageDelta["usage"] = usageMap
 		}
 
 		// Build message_stop event
--- a/core/providers/bedrock/invoke_test.go
+++ b/core/providers/bedrock/invoke_test.go
@@ -1,0 +1,295 @@
+package bedrock
+
+import (
+	"context"
+	"encoding/json"
+	"testing"
+
+	"github.com/maximhq/bifrost/core/schemas"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// anthropicToolMap builds an Anthropic-native tool map as it would arrive on the wire
+// (i.e. the shape convertAnthropicTools reads out of the untyped r.Tools field).
+func anthropicToolMap(name string, cacheControl map[string]interface{}) map[string]interface{} {
+	m := map[string]interface{}{
+		"name":        name,
+		"description": "a test tool",
+		"input_schema": map[string]interface{}{
+			"type":       "object",
+			"properties": map[string]interface{}{},
+		},
+	}
+	if cacheControl != nil {
+		m["cache_control"] = cacheControl
+	}
+	return m
+}
+
+// TestConvertAnthropicTools_NoCacheControl_Unaffected locks in the pre-existing behavior
+// when no tool carries cache_control: one BedrockTool per input tool, no CachePoint entries.
+func TestConvertAnthropicTools_NoCacheControl_Unaffected(t *testing.T) {
+	req := &BedrockInvokeRequest{
+		Tools: []interface{}{
+			anthropicToolMap("alpha", nil),
+			anthropicToolMap("beta", nil),
+		},
+	}
+
+	toolConfig := req.convertAnthropicTools()
+	require.NotNil(t, toolConfig)
+	require.Len(t, toolConfig.Tools, 2)
+	for _, tool := range toolConfig.Tools {
+		assert.NotNil(t, tool.ToolSpec)
+		assert.Nil(t, tool.CachePoint)
+	}
+}
+
+// TestConvertAnthropicTools_CarriesCacheControl is the regression test for #5629: a
+// cache_control marker on an Anthropic-native tool must survive the invoke->Converse
+// conversion as a positional cachePoint entry appended after the marked tool, the same
+// way the Bifrost->Bedrock egress direction already does (utils.go convertChatTools).
+func TestConvertAnthropicTools_CarriesCacheControl(t *testing.T) {
+	req := &BedrockInvokeRequest{
+		Tools: []interface{}{
+			anthropicToolMap("alpha", nil),
+			anthropicToolMap("beta", map[string]interface{}{"type": "ephemeral"}),
+		},
+	}
+
+	toolConfig := req.convertAnthropicTools()
+	require.NotNil(t, toolConfig)
+	require.Len(t, toolConfig.Tools, 3, "expected an extra cachePoint entry after the marked tool")
+
+	assert.NotNil(t, toolConfig.Tools[0].ToolSpec)
+	assert.Equal(t, "alpha", toolConfig.Tools[0].ToolSpec.Name)
+	assert.Nil(t, toolConfig.Tools[0].CachePoint)
+
+	assert.NotNil(t, toolConfig.Tools[1].ToolSpec)
+	assert.Equal(t, "beta", toolConfig.Tools[1].ToolSpec.Name)
+	assert.Nil(t, toolConfig.Tools[1].CachePoint)
+
+	require.NotNil(t, toolConfig.Tools[2].CachePoint)
+	assert.Nil(t, toolConfig.Tools[2].ToolSpec)
+	assert.Equal(t, BedrockCachePointTypeDefault, toolConfig.Tools[2].CachePoint.Type)
+	assert.Nil(t, toolConfig.Tools[2].CachePoint.TTL)
+}
+
+// TestConvertAnthropicTools_CacheControlTTL confirms newBedrockCachePoint's existing
+// TTL allow-list ("5m" | "1h") is honored via this new code path — an unsupported TTL
+// (e.g. Anthropic's own "1m") is dropped to the Bedrock default rather than forwarded.
+func TestConvertAnthropicTools_CacheControlTTL(t *testing.T) {
+	t.Run("supported TTL is forwarded", func(t *testing.T) {
+		req := &BedrockInvokeRequest{
+			Tools: []interface{}{
+				anthropicToolMap("alpha", map[string]interface{}{"type": "ephemeral", "ttl": "1h"}),
+			},
+		}
+		toolConfig := req.convertAnthropicTools()
+		require.NotNil(t, toolConfig)
+		require.Len(t, toolConfig.Tools, 2)
+		require.NotNil(t, toolConfig.Tools[1].CachePoint)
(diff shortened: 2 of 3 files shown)


akshaydeo:
* **#5833** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/5833?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5828** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/5828?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5824** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/5824?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5821** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/5821?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5819** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/5819?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5814** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/5814?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5811** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/5811?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/maximhq/bifrost/5811?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

core/providers/bedrock/invoke.go
```diff
@@ -1405,9 +1433,23 @@ func toAnthropicInvokeStreamBytes(resp *schemas.BifrostResponsesStreamResponse)
...
-			messageDelta["usage"] = map[string]interface{}{
-				"output_tokens": resp.Response.Usage.OutputTokens,
+			// Bedrock Converse only reports usage on the terminal event (unlike native
+			// Anthropic, which also puts input_tokens on message_start) — see
+			// buildBedrockTokenUsage's doc comment. This is the only stream event where
+			// Bifrost has input/cache token counts for a Bedrock-backed call, so all of
+			// them are reported here rather than split across message_start/message_delta.
+			usage := buildBedrockTokenUsage(resp.Response.Usage)
+			usageMap := map[string]interface{}{
+				"input_tokens":  usage.InputTokens,
```
Pratham-Mishra04, line 1443: can you just cross check if invoke endpoint counts cache tokens in input tokens or not to maintain parity with openai standards

akshaydeo: yep sent seprately https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_TokenUsage.html

core/providers/bedrock/invoke.go
```diff
@@ -801,6 +801,13 @@ func (r *BedrockInvokeRequest) parseSystemMessages() []BedrockSystemMessage {
 					continue
 				}
 				result = append(result, msg)
+
+				// Anthropic-dialect blocks mark a breakpoint with a per-block "cache_control"
+				// instead of Converse's positional cachePoint entry; translate it the same
+				// way convertSystemMessages does for the Bifrost->Bedrock egress direction.
+				if cacheControl, ok := m["cache_control"].(map[string]interface{}); ok {
```
TejasGhatte, line 808: here cache control is handled for system messages and tools, but it should be handled for normal content blocks as well

akshaydeo:
### Merge activity

* **Aug 4, 9:07 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/5811).
* **Aug 4, 9:08 AM UTC**: [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/5811) couldn't merge this PR because it failed for an unknown reason.
* **Aug 4, 9:11 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/5811).
* **Aug 4, 9:11 AM UTC**: [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/5811) couldn't merge this PR because it failed for an unknown reason.
* **Aug 4, 9:13 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/5811).
* **Aug 4, 9:13 AM UTC**: @akshaydeo merged this pull request with [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/5811).

PR #5815, opened 2026-08-03
perf(schemas): make reservedKeys a set instead of scanning a slice

itsHabib:
## Summary

`SetValue`, `ClearValue` and `SetValues` each call `slices.Contains(reservedKeys, key)` on every write once `BlockRestrictedWrites()` is set — a linear scan over a 19-entry `[]any` (so interface comparisons), holding the values mutex. This runs on **every** guarded context write, not just list appends.

## Changes

**1. `reservedKeys` becomes a `map[BifrostContextKey]struct{}`.** Nothing ever iterated it or depended on its order; membership is the only question ever asked of it. Keying on `BifrostContextKey` rather than `any` also drops interface boxing on every lookup.

`isReservedKey` narrows to `BifrostContextKey` before indexing. `SetValue` takes `key any`, so a caller can pass a non-comparable value (slice, map, func) — and indexing a map with one of those **panics**, whereas `slices.Contains` did not, because `==` on `[]any` only compares values whose dynamic types match. The type assertion restores the old tolerance and turns out to be *faster* than the unguarded `map[any]` lookup. Thanks @coderabbitai for catching this — it was a real regression in the first push.

(`SetValue` still panics further down at `bc.userValues[key] = value` for such keys. That predates this PR, happens whether or not restricted writes are blocked, and is untouched here. Confirmed against unmodified `dev`.)

**2. `AppendRoutingEngineLog` appends inline** instead of via `AppendToContextList`:

- That helper's dedup scan **can never hit** here — `RoutingEngineLogEntry` carries a `Timestamp int64`, so no two entries are ever `==`.
- Going through the generic costs **~100 extra allocations per 50 appends** versus a concrete `[]RoutingEngineLogEntry` assertion. `[]T` behind a type parameter doesn't specialise the way the concrete slice type does. I only found this by benchmarking both side by side — removing the dedup scan alone changed nothing.

Routing logs are the highest-volume context list (53 call sites across `core/bifrost.go`, `plugins/governance`, `plugins/modelcatalogresolver`), so both effects land here and essentially nowhere else. Order and duplicates are preserved.

`AppendToContextList` itself is **unchanged** — its set semantics are load-bearing for `routing_engines_used`, which becomes Prometheus labels.

### Known limitation, not introduced here

@coderabbitai also flagged that the routing-log append is a read-modify-write without a lock, so concurrent appends can drop an entry. That's true, and `AppendToContextList` on `dev` has the identical shape — unlocked `Value()` read, then unlocked `SetValue()`. I've deliberately left it: fixing it properly means a locked append path on `BifrostContext` (including delegation to the root `valueDelegate`), which is a correctness change with a wider blast radius than this perf PR should carry. Happy to do it as a follow-up, or here if you'd prefer it bundled.

## Type of change

- [x] Refactor (performance; no intended behaviour change beyond restoring non-comparable-key tolerance)

## Affected areas

- [x] Core (Go) — `core/schemas/context.go` only

## How to test

```sh
cd core
go test ./schemas/
```

Correctness tests (all new):

```sh
go test ./schemas/ -run='ReservedKey|IsReservedKey|NonComparable|RoutingEngineLogOrder' -v
```

Benchmarks — both lookup strategies in one process so they face identical conditions, worst case (key absent, so the scan compares every entry):

```sh
go test ./schemas/ -run='^$' -bench='AppendRouting' -benchtime=50000x -count=3
```

Results (go1.26.4, Intel Ultra 7 265H, minimum of 5 runs at `-benchtime=2000000x`):

| | ns/op |
|---|---|
| slice scan | 35.1 |
| map lookup + type assertion | **3.8** (9x) |

`AppendRoutingEngineLog` ×50: **211 → 111 allocs/op** (−47%), 14654 → 13052 B/op.

I've deliberately not quoted wall-clock for the full `SetValue` path or for `AppendRoutingEngineLog`: both allocate and were too sensitive to background load on my machine to give figures I'd trust to reproduce. Allocation counts were stable across every run. Happy to re-run on quieter hardware.

Full `core` module passes except three `internal/mcptests` tests that need a live MCP server subprocess and fail identically on unmodified `dev` in my environment.

## Screenshots/Recordings

N/A — no UI changes.

## Breaking changes

- [x] No

## Related issues

None linked. Found while profiling gateway request latency.

## Security considerations

One worth naming: `reservedKeys` is the allowlist that drops writes to protected keys, so a change here touches a security-relevant path. Mitigations — the map replaces the slice outright rather than deriving from it, so the two can't drift; `TestReservedKeyWritesStillBlocked` asserts reserved writes are still dropped and non-reserved writes still land; `TestIsReservedKey` asserts the predicate reports exactly `reservedKeys` membership. The `blockRestrictedWrites.Load()` short-circuit is unchanged and still evaluated first.

## Checklist

- [x] I added/updated tests where appropriate — new `core/schemas/contextlist_bench_test.go`
- [ ] I updated documentation where needed — N/A, no public API or docs change
- [x] I verified builds succeed (Go) — `go build ./...` clean in `core`; no UI change
- [ ] I verified the CI pipeline passes locally if applicable — package tests and `core` module run locally, full CI not reproduced


itsHabib:
@coderabbitai thanks — both findings were worth having, one fixed and one deliberately left.

**Non-comparable keys (fixed).** You were right and this was a real regression, not a theoretical one. `slices.Contains` over `[]any` never panicked because `==` only compares values whose dynamic types match, so a `[]string` key just never matched anything. Map indexing hashes unconditionally and panics with `hash of unhashable type`. I confirmed both behaviours in isolation before changing anything.

Fixed by narrowing to `BifrostContextKey` before the lookup and keying the map on that type instead of `any`. That restores the old tolerance and is *faster* than the unguarded `map[any]` version — 3.8 ns/op vs 35.1 ns/op for the slice scan (9x), because it also drops the interface boxing. Added `TestNonComparableKeyDoesNotPanic` covering slice, map and func keys.

One scoping note: `SetValue` still panics on such a key further down at `bc.userValues[key] = value`. My first version of the test asserted `SetValue` was panic-free end to end and failed for that reason — I checked and unmodified `dev` panics there too, with restricted writes off so the reserved-key check never runs. So that one predates this PR and I've left it, with the test narrowed to the guarantee this change is actually responsible for and a comment saying so.

**Atomic appends (not fixing here).** Also correct — unlocked `Value()` read then unlocked `SetValue()`, so concurrent appends can lose an entry. But `AppendToContextList` on `dev` has the identical shape, so this is pre-existing rather than introduced. Fixing it properly means a locked append path on `BifrostContext` including `valueDelegate` delegation, which is a correctness change with a wider blast radius than a perf PR should carry, and it'd want its own concurrent test. I'd rather not smuggle it in here — happy to open it as a follow-up, or add it to this PR if the maintainers prefer it bundled.

Also restructured the description to the repo template, which I'd skipped.

PR #5816, opened 2026-08-03
chore: ignore built Go binaries alongside the two already listed

itsHabib:
## Summary

Seven built Go binaries aren't covered by `.gitignore`. The `# binaries` section already lists two — `tests/cmd/e2eseed/e2eseed` and `tests/cmd/seedvks/seedvks` — which look like they were added *after* each got committed by accident. This finishes the list so the next one doesn't repeat the cycle.

One already has: **`tests/scripts/1millogs/1millogs` (64.6 MB)** landed on `dev` in cbf658b14 (*"upgraded go to 1.26.5"*, #5501) and was removed from the tree later, so it's permanent in history. That directory's `Makefile` only ever invokes `go run`, never `go build` — so the binary wasn't even a by-product of the documented workflow. Someone built it locally and it got swept into the commit.

## Changes

- Seven entries added to the existing `# binaries` section, following the style of the two already there.
- `/bifrost-server` is anchored with a leading slash. Unlike every other entry it has no path separator, so without the anchor it would match a file of that name at **any** depth (thanks @coderabbitai for catching this). It's built at the repository root, which is where it was committed.

| Size | Path | Reached `dev`? |
|---|---|---|
| 64.6 MB | `tests/scripts/1millogs/1millogs` | **yes**, in #5501 |
| 71.7 MB | `/bifrost-server` | side branches only |
| 32.9 MB | `transports/bifrost-http/bifrost-http-transport` | side branches only |
| 17.7 MB | `plugins/maxim/cmd/maxim-plugin` | side branches only |
| 16.7 MB | `plugins/mocker/cmd/mocker-plugin` | side branches only |
| 17.1 MB | `sdk/examples/plugin-host/plugin-host` | side branches only |
| 16.7 MB | `sdk/examples/simple-plugin/simple-plugin` | side branches only |

Six of the seven are pre-emptive — worth being clear about that rather than overstating the problem.

## Type of change

- [x] Chore/CI

## Affected areas

None of the listed areas — `.gitignore` only, no code paths touched.

## How to test

Each rule matches its target:

```sh
git check-ignore -v tests/scripts/1millogs/1millogs bifrost-server \
  transports/bifrost-http/bifrost-http-transport \
  plugins/maxim/cmd/maxim-plugin plugins/mocker/cmd/mocker-plugin \
  sdk/examples/plugin-host/plugin-host sdk/examples/simple-plugin/simple-plugin
# expect: one .gitignore:<line> match per path
```

No currently tracked file becomes ignored — this is the check that matters, since shadowing a tracked file is the only way this could break a build:

```sh
git ls-files | git check-ignore --stdin -v
# expect: no output
```

The anchor works as intended:

```sh
git check-ignore -v bifrost-server          # matches
git check-ignore -v some/dir/bifrost-server # does NOT match
```

## Screenshots/Recordings

N/A — no UI changes.

## Breaking changes

- [x] No

## Related issues

None linked. Found while profiling gateway request latency.

## Security considerations

None. If anything mildly positive — the rules make it less likely a locally built binary is committed unintentionally.

## Checklist

- [ ] I added/updated tests where appropriate — N/A, `.gitignore` has no test surface; verification commands are above
- [ ] I updated documentation where needed — N/A
- [x] I verified builds succeed (Go and UI) — no code changed; confirmed no tracked file becomes ignored
- [ ] I verified the CI pipeline passes locally if applicable — not run locally; nothing here affects build or test behaviour

## Scope

This only prevents future commits. It does **not** shrink the existing clone — the repo is ~866 MB packed (`gh api repos/maximhq/bifrost --jq .size`), ~1.5 GB checked out, largely from build artifacts and repeatedly regenerated `docs/openapi/openapi.json` snapshots already in history. Reclaiming that needs a history rewrite, which is your call and not something a PR should do.

Happy to add or drop entries if any of these are intentionally committed for a reason I've missed.


itsHabib:
@coderabbitai good catch on the anchor — applied as `/bifrost-server`.

You were right that it was the odd one out: every other entry in that section contains a path separator, which anchors it to the repo root by gitignore semantics, but `bifrost-server` has none and so matched at any depth. Verified the difference in a scratch repo rather than taking it on trust — the unanchored form matched both `bifrost-server` and `some/nested/dir/bifrost-server`, the anchored form only the root one. Also confirmed the binary is in fact built at the repository root, which is where it was committed on the side branch, so anchoring loses nothing.

Post-fix checks: the root path is still ignored, a nested path of the same name is no longer matched, and `git ls-files | git check-ignore --stdin -v` is still empty so no tracked file gets shadowed.

Also rewrote the description against `.github/pull_request_template.md`, which I'd skipped — that was a fair flag on the pre-merge check.

PR #5836, opened 2026-08-04
stream cancel fix

akshaydeo:
## Summary

Client-disconnect detection during SSE streaming was purely reactive: `cancel()` only fired when a downstream `SendEvent` write failed, and a write only happened when the producer loop attempted one. For providers like Vertex AI's `streamGenerateContent` that deliver fewer, larger chunks than direct Gemini for the same prompt, the upstream could finish delivering its entire response before the producer ever attempted another write after a client disconnect — causing the stream to log as a false success with no disconnect recorded.

This PR adds a periodic heartbeat goroutine that sends a no-op SSE comment frame (`: heartbeat\n\n`) every 100ms independently of real data. Because SSE comment lines are ignored by every spec-compliant client, the heartbeat is invisible to application code but forces a write attempt during otherwise-idle gaps, closing the detection window.

## Changes

- Added `SendHeartbeat()` to `SSEStreamReader`, which writes an SSE comment frame (`: heartbeat\n\n`) and returns `false` if the reader is already closed, matching the contract of the other `Send*` wrappers.
- Added a `streamHeartbeatInterval` constant (100ms) and a heartbeat goroutine inside `handleStreamingResponse` that ticks at that interval, calls `SendHeartbeat()`, and invokes `cancel()` if a disconnect is detected. The goroutine is stopped via a `heartbeatDone` channel that is closed in the producer's `defer`.
- Added tests verifying the heartbeat frame starts with `:` and ends with `\n\n` (valid SSE comment), and that `SendHeartbeat` returns `false` after `Close`.

## Type of change

- [x] Bug fix

## Affected areas

- [x] Transports (HTTP)

## How to test

```sh
go test ./transports/bifrost-http/...
```

To validate end-to-end: send a streaming request through a Vertex AI provider, disconnect the client mid-stream (e.g. close the browser tab or kill the curl process), and confirm the request is logged as a client disconnect rather than a successful completion.

## Breaking changes

- [x] No

## Security considerations

The heartbeat frame is a standard SSE comment and is never forwarded to upstream providers — it only touches the downstream client connection. No auth, secrets, or PII are involved.

## Checklist

### DIFF
--- a/transports/bifrost-http/handlers/inference.go
+++ b/transports/bifrost-http/handlers/inference.go
@@ -1897,6 +1897,12 @@ func (h *CompletionHandler) handleStreamingTranscriptionRequest(ctx *fasthttp.Re
 	h.handleStreamingResponse(ctx, bifrostCtx, schemas.TranscriptionStreamRequest, getStream, cancel)
 }
 
+// streamHeartbeatInterval is how often handleStreamingResponse's heartbeat goroutine
+// probes the downstream connection with a no-op SSE comment. See that goroutine's
+// comment for why this exists: disconnect detection is otherwise purely reactive to
+// real-data write failures, which a fast/bursty upstream can outrun entirely.
+const streamHeartbeatInterval = 100 * time.Millisecond
+
 // handleStreamingResponse is a generic function to handle streaming responses using Server-Sent Events (SSE)
 // The cancel function is called ONLY when client disconnects are detected via write errors.
 // Bifrost handles cleanup internally for normal completion and errors, so we only cancel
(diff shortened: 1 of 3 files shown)


akshaydeo:
* **#5836** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/5836?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/maximhq/bifrost/5836?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

akshaydeo:
### Merge activity

* **Aug 4, 9:30 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/5836).
* **Aug 4, 9:31 AM UTC**: @akshaydeo merged this pull request with [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/5836).

PR #5866, opened 2026-08-05
docs

### LINKED ISSUE #5010: [Feature]: Server-side SSE keepalive (comment heartbeat) to keep long-idle streams alive through intermediaries
## Prerequisites

## Problem to solve

When Bifrost proxies a streaming response (`text/event-stream`), the only bytes written to the
client are the events the upstream provider produces. If the upstream goes quiet for a while —
a slow first token under load, a long tool/thinking step, or a provider that batches before
flushing — **no bytes cross the connection during that gap**. Two independent timers then work
against a stream that is actually healthy:

1. **Bifrost's own stream idle timeout.** Every streamed response is wrapped in an idle-timeout
   reader that defaults to **120 s** (`core/schemas/provider.go:21`,
   `DefaultStreamIdleTimeoutInSeconds = 120`; applied in `NetworkConfig` at `:562-563`). If the
   upstream emits no chunk within the window, Bifrost force-closes the stream. Operators can
   raise `stream_idle_timeout_in_seconds`, but that only moves the cliff — it does not keep the
   connection warm.

2. **Any intermediary between Bifrost and the client.** Reverse proxies, API gateways, and cloud
   load balancers all impose their own **connection idle timeouts** and reset them only on
   *observed traffic*. Common defaults sit well under a couple of minutes (e.g. AWS ELB/ALB
   default 60 s idle; nginx `proxy_read_timeout` default 60 s). During a silent generation gap
   these intermediaries can drop the socket even though Bifrost and the provider are both fine —
   and the client sees a truncated or reset stream, not a clean error.

The client cannot fix this: it is *reading* a response, not writing to it, so it has no way to
keep the intermediary's idle timer alive. The only component positioned to emit liveness bytes
is the server writing the stream — **Bifrost**. Today it emits nothing during idle gaps.

Evidence this bites Bifrost users in practice (idle-gap-driven stream failures):

- #2910 — *Custom providers that don't send `[DONE]` SSE marker cause ~60s stream hang* — the
  stream sits silent "until the idle timeout fires."
- #4542 — *Bedrock streaming responses buffered server-side — TTFB matches total generation
  time* — long silent gap before any byte reaches the client, exactly the window during which
  an intermediary idle timeout can fire.
- #2416 — *Make request timeout configurable* and #3478 — *provider ignores
  `stream_idle_timeout_in_seconds`* — show the idle/timeout surface is already a recurring pain
  point, but the knobs only lengthen the silence; none keeps the connection alive.

Inspecting the write path confirms there is no keepalive mechanism today: the SSE writer
(`transports/bifrost-http/lib/streamreader.go`, `SSEStreamReader`) is a pull-based `io.Reader`
with `Send`/`SendEvent`/`SendError`/`SendDone`/`Done` and **no timer** — it writes only when the
producer goroutine hands it a provider event. The producer in
`transports/bifrost-http/handlers/inference.go` (SSE headers set at `:1687`) likewise has no
periodic tick. A silent upstream ⇒ a silent socket.

**Who benefits:** everyone running Bifrost behind a load balancer / reverse proxy / API gateway
(i.e. essentially every production deployment) and anyone streaming from providers or models
with high or bursty first-token latency.

## Proposed solution

Add an **opt-in, config-driven SSE keepalive** emitted by the transport while a stream is
otherwise idle.

- **Wire-level shape — an SSE comment line.** The SSE spec defines any line beginning with a
  colon (`:`) as a comment that clients **ignore** (WHATWG HTML "Server-sent events"; MDN
  *Using server-sent events*). So a heartbeat is literally:

  ```
  : keepalive\n\n
  ```

  It resets idle timers on every hop (it is real traffic on the socket) without appearing in the
  parsed event stream, so no client change is required and no provider payload is altered.

- **When to emit.** Start a timer when the SSE response begins; every time a real event is
  written, reset it. If the interval elapses with no real event, write one comment frame. This
  covers both the pre-first-token gap and inter-event gaps. `SSEStreamReader` is the natural home
  for the timer (it already owns `closeCh`/`Done` lifecycle), or the producer goroutine can drive
  a `time.Ticker` and call `Send`.

- **Configuration.** A network-config field, disabled by default to preserve current behavior,
  e.g.:

  ```jsonc
  "network_config": {
    "stream_idle_timeout_in_seconds": 300,
    "stream_keepalive_interval_in_seconds": 15   // 0 or unset = disabled (today's behavior)
  }
  ```

  Guidance: set the keepalive interval **comfortably below the smallest idle timeout on the
  path** (Bifrost's own `stream_idle_timeout_in_seconds` and every intermediary's connection
  idle timeout).

- **Scope.** Applies to all `text/event-stream` responses the HTTP transport produces
  (`/v1/*`, `/anthropic/*`, provider-native SSE, Bedrock EventStream framing, etc.). For the
  Bedrock binary EventStream path the equivalent is a no-op/ping event rather than an SSE
  comment; the config knob and timer logic are shared, only the on-wire keepalive token differs
  per protocol.

## Alternatives considered

- **Only raise `stream_idle_timeout_in_seconds`.** Stops Bifrost from closing early, but does
  nothing about intermediary idle timeouts — the socket can still be dropped mid-gap. Moves the
  cliff; doesn't remove it.
- **Tune every intermediary's idle timeout instead.** Often impossible (managed load balancers
  cap or fix the value) and fragile (one un-tuned hop reintroduces the failure). Also pushes a
  Bifrost streaming concern onto unrelated infrastructure.
- **Client-side mitigation (raise read timeout / retry).** Clients can lengthen their own read
  timeout, but a reader cannot keep an *intermediary* timer alive — only the writer can. And
  binary/SDK clients (e.g. botocore) often ignore `Retry-After` and re-send, which is worse than
  a kept-alive stream.
- **TCP keepalive.** Operates at the socket layer between adjacent hops and is frequently not
  honored as "activity" by L7 idle-timeout logic; an application-level SSE comment is the portable
  signal that resets L7 idle timers end to end.

## Area(s)

- Transports (HTTP)
- Core (Go)

## Additional context

- SSE comment lines as keepalive are the standard, spec-blessed mechanism — WHATWG HTML
  *Server-sent events* (comment = line starting with `:`), MDN *Using server-sent events*
  (explicitly notes comment lines can be sent as a keep-alive/"ping"). Many SSE servers and
  gateways ship this (e.g. periodic `:keep-alive`/`:ping` comments).
- Idle-timeout references for common intermediaries: AWS ELB/ALB connection idle timeout
  (default 60 s, resets on traffic), nginx `proxy_read_timeout` (default 60 s). These reset on
  *observed* bytes, which is exactly what the heartbeat provides.
- Suggested defaults are illustrative; the key ask is (a) the capability exists and (b) it is a
  first-class `network_config` knob, off by default.

### Source anchors verified against Bifrost 1.6.3 (latest)

| Claim | Location |
|---|---|
| Idle-timeout default 120 s | `core/schemas/provider.go:21` (`DefaultStreamIdleTimeoutInSeconds = 120`) |
| Default applied when unset/≤0 | `core/schemas/provider.go:562-563` |
| SSE writer has no timer/keepalive | `transports/bifrost-http/lib/streamreader.go` (`SSEStreamReader`) |
| SSE producer has no periodic tick | `transports/bifrost-http/handlers/inference.go:1687` onward |
| Prior idle-gap symptoms | issues #2910, #4542, #3478, #2416 |

### LINKED ISSUE #5186: [Bug]: Anthropic-surface replay of OpenAI encrypted reasoning mints a fresh item id, OpenAI 400s with "Encrypted content item_id did not match the target item id"
## Summary

Serve an OpenAI reasoning model through `/anthropic/v1/messages` with `thinking` enabled and Bifrost returns the model's encrypted reasoning as an Anthropic `redacted_thinking` block. Replay that block on a later turn, which the Anthropic protocol requires, and the request dies:

```
HTTP 400
The encrypted content for item rs_<50 hex chars> could not be verified.
Reason: Encrypted content item_id did not match the target item id.
```

OpenAI checks the encrypted payload against the reasoning item id it was created under. That is what the 400 is. Bifrost drops the original id on egress (Anthropic's `redacted_thinking` block only has `type` and `data`, there is nowhere to put an id) and mints a fresh random one on replay. So every replayed redacted block fails. Plain multi-turn and tool-loop shapes both fail.

With a fallback chain configured the 400 is masked: traffic silently degrades to the fallback provider and the caller thinks the primary model answered. Without one the client gets the 400 directly.

Affected: `transports/v1.6.3` (`de46d5e8200a`), `transports/v2.0.0-prerelease1` (`3fc6bad4c071`), current `main` (`e864b4f6bbd0`).

## Root cause

Two causal sites in `core/providers/anthropic/responses.go`, the redacted-thinking replay cases in the grouped and ungrouped converters:

- https://github.com/maximhq/bifrost/blob/e864b4f6bbd0/core/providers/anthropic/responses.go#L4914-L4926
- https://github.com/maximhq/bifrost/blob/e864b4f6bbd0/core/providers/anthropic/responses.go#L5233-L5242

```go
case AnthropicContentBlockTypeRedactedThinking:
    // Handle redacted thinking (encrypted content)
    if block.Data != nil {
        bifrostMsg := schemas.ResponsesMessage{
            ID:   schemas.Ptr("rs_" + providerUtils.GetRandomString(50)), // fresh id
            Type: schemas.Ptr(schemas.ResponsesMessageTypeReasoning),
            ResponsesReasoning: &schemas.ResponsesReasoning{
                Summary:          []schemas.ResponsesReasoningSummary{},
                EncryptedContent: block.Data,                              // payload created under the ORIGINAL id
            },
        }
```

On egress the OpenAI item id is dropped when the reasoning item becomes a `redacted_thinking` block. Only `encrypted_content` survives, as `data`: https://github.com/maximhq/bifrost/blob/e864b4f6bbd0/core/providers/anthropic/responses.go#L5715-L5721, plus the streaming converter's encrypted assignments.

`GetRandomString` uses a hex charset, so the minted ids look exactly like real OpenAI ids. Captured request logs settle it: the failing request's reasoning input item carried id `rs_9c9df5cb...` and OpenAI's error named that exact id. The id in the error is the one Bifrost minted.

This is the mirror of #4943 / #5093, fixed in core v1.6.4. Those covered Anthropic-origin `redacted_thinking` on the OpenAI-format surfaces. That direction survives a synthetic id because Bifrost's id never gets serialized into Anthropic's block, so Anthropic never validates it. OpenAI does. #3802 is the same ingress direction but a different bug (reasoning dropped, not corrupted). Not a duplicate.

## Reproduction (no fallbacks, deterministic)

Config: one OpenAI provider key serving an OpenAI reasoning model (I reproduced with a GPT-5.x reasoning model). No routing rules, no fallbacks, no governance.

Turn 1, capture a redacted block. Streaming; the block arrives complete in `content_block_start`:

```json
POST /anthropic/v1/messages
{
  "model": "<openai-reasoning-model>",
  "max_tokens": 6000,
  "stream": true,
  "thinking": {"type": "adaptive"},
  "messages": [{"role": "user", "content": "Think hard: which two of Alpha, Beta, Gamma, Delta are adjacent Greek letters whose English spellings share no vowels? Just name them."}]
}
```

Response content: `[redacted_thinking{data: <932-byte payload>}, text]`.

Turn 2, replay the block byte-identical in a tool-loop shape. Plain multi-turn replay fails the same way:

```json
POST /anthropic/v1/messages
{
  "model": "<openai-reasoning-model>",
  "max_tokens": 6000,
  "thinking": {"type": "adaptive"},
  "tools": [{"name": "get_sales", "description": "weekly sales for a store",
             "input_schema": {"type": "object", "properties": {"store": {"type": "string"}}, "required": ["store"]}}],
  "messages": [
    {"role": "user", "content": "Get sales for the correct store."},
    {"role": "assistant", "content": [
      {"type": "redacted_thinking", "data": "<data from turn 1, byte-identical>"},
      {"type": "tool_use", "id": "toolu_repro_0001", "name": "get_sales", "input": {"store": "Beta"}}
    ]},
    {"role": "user", "content": [{"type": "tool_result", "tool_use_id": "toolu_repro_0001", "content": "$45,210"}]}
  ]
}
```

Result, every time:

```json
{"type":"error","error":{"type":"invalid_request_error","message":"The encrypted content for item rs_2ab8265e069b0eceff581cfdd975a663f6d73ea8189255cba4 could not be verified. Reason: Encrypted content item_id did not match the target item id."}}
```

## Proposed fix

I can send a PR. The shape I would propose: a Bifrost-private envelope inside `data` (encoding, not encryption), decoded only at the final destination-provider conversion.

- Egress: when the actual source provider (from `ExtraFields.Provider`, not the requested model, since routing and fallbacks can change it) is OpenAI and both the item id and encrypted content are present, emit `data = base64(json{_bifrost, v, provider, item_id, payload})`. Anthropic-origin `data` stays byte-identical.
- Replay: recognize the envelope, restore the original `ResponsesMessage.ID`, keep the envelope in the neutral `EncryptedContent`. Unwrap to the raw payload only in the OpenAI provider's final request serialization.
- Fallback safety: if a request carrying an OpenAI-bound envelope gets routed to Anthropic or another provider, drop the hidden item with a trace. Never forward foreign ciphertext.
- Legacy blocks (unwrapped `data` minted by older versions) keep today's behavior. The original id is gone, nothing can restore it statelessly.

Deferred decoding is the important part. Unwrapping eagerly at the Anthropic ingress loses provenance, and an OpenAI failure followed by an Anthropic fallback would then forward OpenAI ciphertext to Claude.

If you would rather carry the id some other way that works too. The invariant that matters is `(item_id, encrypted_content)` surviving the Anthropic round trip as a pair.

akshaydeo:
## Summary

Adds changelog documentation for Bifrost OSS `v1.6.8` and Enterprise `v1.5.8` releases, and registers both pages in the docs navigation.

## Changes

- Added `docs/changelogs/v1.6.8.mdx` covering OSS transport features (WebSocket proxy support, configurable SCIM buffer sizes), bug fixes (proactive SSE disconnect detection, closed-channel panic on stream shutdown, Bedrock header signing denylist, deterministic Bedrock tool ordering, `cache_control` translation, adaptive thinking fixes, encrypted reasoning content mismatch, Bedrock invoke content retention, document message placeholder, VK provider bulk replace), and dependency bumps across core, framework, governance, and all plugins.
- Added `docs/changelogs/ent-v1.5.8.mdx` covering enterprise-only features (inference JIT provisioning for first-time SSO users, `order` role-resolution strategy for SCIM/OIDC, AWS ALB as a selectable IAP option, configurable outbound HTTP buffer sizes, AP-managed VK quota visibility without access-profile permission), additional enterprise bug fixes (calendar-aligned budget reset propagation, access-profile propagation on large profiles, soft-deleted user reactivation via proxy/inference JWT, SCIM vs. periodic role sync conflict, cluster peer liveness isolation from bulk sync backlog), and the full plugin dependency manifest for compiling against this release.
- Registered both new changelog pages at the top of their respective navigation groups (`Open Source` and `Enterprise`) in `docs/docs.json`.

## Type of change

- [ ] Bug fix
- [ ] Feature
- [ ] Refactor
- [x] Documentation
- [ ] Chore/CI

## Affected areas

- [ ] Core (Go)
- [ ] Transports (HTTP)
- [ ] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [x] Docs

## How to test

Navigate to the docs site and confirm:
- The `v1.6.8` OSS changelog page renders correctly under the Open Source changelogs section.
- The `ent-v1.5.8` Enterprise changelog page renders correctly under the Enterprise changelogs section.
- Both pages appear at the top of their respective navigation groups.

## Breaking changes

- [x] No

## Related issues

- Closes #5010
- Closes #5186

## Security considerations

None. This PR contains only documentation changes.

## Checklist

### DIFF
--- a/docs/changelogs/ent-v1.5.8.mdx
+++ b/docs/changelogs/ent-v1.5.8.mdx
@@ -1,0 +1,490 @@
+---
+title: "v1.5.8"
+description: "v1.5.8 changelog - 2026-08-05"
+---
+
+<Update label="Bifrost Enterprise" description="v1.5.8">
+
+## Changelog
+
+This release adds JIT provisioning for first-time SSO users hitting inference endpoints directly, an `order` role-resolution strategy for SCIM/OIDC role mapping conflicts, and AWS ALB as a selectable Identity-Aware Proxy option in the SCIM wizard. It also fixes access-profile budget propagation gaps (calendar-aligned resets, large profiles), a soft-deleted-user reactivation inconsistency between the proxy and OAuth login paths, and isolates cluster peer-liveness tracking from bulk sync backlog.
+
+## ✨ Features
+
+- **Inference JIT Provisioning for First-Time SSO Users** - A first-time SSO user hitting an inference endpoint directly (no prior interactive login) is now JIT-provisioned from their IdP bearer JWT instead of receiving a flat 401, with claims enriched from the provider's userinfo endpoint so group/role mappings resolve the same way they would through the login flow. Resurrection of a deliberately deprovisioned identity is blocked; a user with no matching role mapping is denied with a 403.
+- **`order` Role-Resolution Strategy for SCIM/OIDC** - When a user matches multiple attribute-role mappings at once, admins can now choose `order` (first-listed mapping wins, by list position) as an alternative to the default `highestPermissionCount`. Role sync from SCIM group and attribute pushes is also unified into a single `RecomputeUserRole` path, so a role granted by one signal (e.g. a group) is no longer cleared by an unrelated push that only carries a different signal (e.g. an attribute).
+- **AWS ALB as a Selectable IAP Option** - AWS ALB is now exposed as a selectable Identity-Aware Proxy provider in the SCIM setup wizard, alongside Cloudflare Access and Generic OIDC.
+- **Configurable Outbound HTTP Buffer Sizes** - `ClientReadBufferSize` and `ClientWriteBufferSize` in the `config.json` server block now control the outbound HTTP client's fasthttp buffer sizes, for workloads that need to tune beyond the 64KB default.
+- **AP-Managed VK Quota Visibility Without Access-Profile Permission** - The managed-virtual-key notice now renders for users without access-profile view permissions, using a dedicated `isManagedByProfile` signal instead of depending on the full profile object. The underlying VK quota endpoint also now surfaces the access profile's rate limit alongside its budgets, so AP-managed keys no longer show stale or zeroed-out rate limit data.
+
+## 🌎 Open Source Features
+
+- **WebSocket Proxy Support** - Realtime and Responses WebSocket connections now route through the configured provider-level proxy (HTTP, SOCKS5, env-based) instead of always dialing direct.
+- **Configurable SCIM Buffer Sizes** - `WithFasthttpBufferSizes` on the HTTP client factory lets IdP token endpoints returning headers larger than the 4KB default work without failing SCIM/OAuth clients.
+
+## 🐞 Fixed
+
+- **Calendar-Aligned Budget Resets Not Reaching Existing Users** - Toggling `calendar_aligned` on an access profile now propagates to user copies that already exist, not just newly assigned users. Previously, existing users' budgets kept resetting on a rolling 30-day window instead of the calendar boundary, permanently, with no self-healing restart or re-save.
+- **Access-Profile Propagation Failing on Large Profiles** - Propagating changes from an access profile template to its user copies no longer fails when the profile is large.
+- **Soft-Deleted User Reactivation via Proxy/Inference JWT** - The auth-proxy and inference bearer-JWT paths now reactivate a soft-deleted user when their claims resolve to a valid role, matching the existing behavior of the interactive OAuth login flow instead of permanently blocking with a deprovisioned error.
+- **SCIM Provisioning vs. Periodic Role/Team Sync** - The periodic OIDC pull-based role/team sync now skips itself when SCIM provisioning is configured, since the IdP is already pushing changes in real time; the Okta API token field (only used by the pull sync) is hidden from the wizard in that mode.
+- **Cluster Peer Liveness Isolated from Bulk Sync Backlog** - Peer liveness pings now travel on a dedicated gRPC stream, separate from the bulk config-sync stream, so a slow or backed-up sync no longer delays or masks whether a peer is actually reachable.
+- **Role Conflict Resolution Copy** - Clarified the "order" strategy's label and hint text in the SCIM role-mapping UI.
+- **Proactive SSE Disconnect Detection** - SSE heartbeat handling now detects a client disconnect during streaming proactively instead of only when a producer loop attempts a write, fixing false-success logging on fast/bursty upstreams like Vertex.
+- **Closed Channel Panic on Stream Shutdown** - Fixed a race where a heartbeat goroutine mid-send on the event channel at shutdown could panic with "send on closed channel".
+- **Budget Pruning Crash with `config.json` Source of Truth** - Budget/config pruning now tolerates a not-found error on cascade-deleted rows, fixing a startup crash for API-created model configs absent from `config.json`.
+- **Bedrock Header Signing Denylist** - Caller headers stored for Anthropic OAuth passthrough are no longer forwarded to other providers, preventing SigV4 signature mismatches on Bedrock.
+- **Deterministic Bedrock Tool Ordering** - Fixed non-deterministic tool ordering in `toolConfig` caused by map iteration, which was breaking Bedrock prompt-cache hits.
+- **Bedrock `cache_control` Translation** - `cache_control` markers on Anthropic-format content blocks, system blocks, and tools are now correctly translated through the Bedrock invoke and Converse paths instead of being silently dropped.
+- **Bedrock Adaptive Thinking Fixes** - Reasoning/thinking `max_tokens` validation errors now return HTTP 400 instead of 500; `tool.defer_loading` is gated on its own beta header; Nova2 web search and code execution tools are handled correctly.
+- **Encrypted Reasoning Content Mismatch** - Replaying OpenAI Responses API reasoning items through the Anthropic surface no longer mints a fresh item id while forwarding the original encrypted content, which OpenAI was rejecting.
+- **Bedrock Invoke Content Retention** - Bedrock's InvokeModel route now correctly decodes Anthropic's type-discriminated image/tool_use/tool_result blocks instead of silently dropping them.
+- **Bedrock Document Message Placeholder** - Messages containing a document block without accompanying text no longer get rejected by Bedrock's Converse API.
+- **VK Provider Bulk Replace** - Virtual key provider config replacement is now a single bulk operation instead of per-provider round trips, fixing a hot-path slowdown at scale.
+
+## 🗄️ Database Migrations
+
+No new database migrations in this release, in either the OSS transports layer or the enterprise repo.
+
+## 🐙 Closed OSS Issues
+
+- [#5010](https://github.com/maximhq/bifrost/issues/5010) - Server-side SSE keepalive (comment heartbeat) to keep long-idle streams alive through intermediaries
+- [#5186](https://github.com/maximhq/bifrost/issues/5186) - Anthropic-surface replay of OpenAI encrypted reasoning mints a fresh item id, OpenAI 400s with "Encrypted content item_id did not match the target item id"
+
+## 📀 Base OSS version
+
+`transports/v1.6.8`
+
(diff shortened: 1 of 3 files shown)


akshaydeo:
* **#5866** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/5866?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/maximhq/bifrost/5866?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

akshaydeo:
### Merge activity

* **Aug 5, 1:03 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/5866).
* **Aug 5, 1:03 PM UTC**: @akshaydeo merged this pull request with [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/5866).

PR #5872, opened 2026-08-05
Fix bedrock mantle streaming 5871

### LINKED ISSUE #5871: [Bug]: AWS Bedrock Mantle streaming is broken
### Prerequisites

### Description

Streaming chat completions through the bedrock_mantle provider never finish. Every chunk is delivered, including the terminal chunk with finish_reason and usage, and then the response stalls: data: \[DONE\] is never sent and the HTTP body is never terminated. The client hangs until it or a proxy in front of it gives up. Each affected request leaks one goroutine and one upstream socket.

Two defects combine, and both need fixing:

1. ReleaseStreamingResponse in core/providers/utils/utils.go drains the upstream body with an unbounded io.Copy. Nothing bounds it: BuildStreamingClient sets ReadTimeout = 0, and the drain runs after stopIdleTimeout() has already fired, so it sits outside the idle-timeout reader that guards the SSE loop. The Bedrock Mantle endpoint holds the chunked body open after the terminal event, so the copy blocks forever. That defer runs before the one calling CloseStream(responseChan), so the response channel never closes, the transport's SSE producer loop never exits, and data: \[DONE\] is never written.
2. ProviderSendsDoneMarker does not list schemas.BedrockMantle, so the SSE loop waits for a marker Mantle never sends. The switch matches schemas.Bedrock only, even though its comment already says "Bedrock mantle don't send \[DONE\] marker"; that only ever covered the legacy mantle routing under the bedrock provider key.

Related: maximhq/bifrost#2910 reports defect 2 for custom providers. This report adds a built-in provider key missing from the same switch, plus defect 1, which maximhq/bifrost#2910 does not cover: its hang self-heals at the stream idle timeout, this one never recovers.

### Steps to reproduce

Reproducible with no AWS access by pointing the provider at a mock upstream.

1. Run a TLS server with a network alias of [bedrock-mantle.us-east-1.api.aws](<http://bedrock-mantle.us-east-1.api.aws>) that answers POST /v1/chat/completions with a chunked SSE stream: a few delta.content frames, then a terminal frame `{"choices":[{"index":0,"delta":{},"finish_reason":"stop"}],"usage":{"prompt_tokens":10,"completion_tokens":2,"total_tokens":12}}`, then `data: [DONE]`, and then never send the terminating 0\\r\\n\\r\\n chunk
2. Configure the provider so it uses Bearer auth and accepts the mock:

```json
"bedrock_mantle": {
  "keys": [{
    "id": "a1b2c3d4-e5f6-7890-abcd-ef1234567890",
    "name": "mantle-mock",
    "value": "mock-bearer-key",
    "models": ["openai.gpt-oss-20b"],
    "weight": 1.0,
    "enabled": true,
    "bedrock_mantle_key_config": { "region": "us-east-1" }
  }],
  "network_config": { "insecure_skip_verify": true, "allow_private_network": true }
}
```

3. Stream through the gateway:

```bash
curl -sN http://localhost:8080/v1/chat/completions \
  -d '{"model":"bedrock_mantle/openai.gpt-oss-20b","messages":[{"role":"user","content":"Say pong"}],"stream":true}'
```

Observed against the live endpoint identically w/ `bedrock_mantle/openai.gpt-oss-20b`

Matrix over upstream framings, same request, \[DONE\] = client received the terminal marker:<br>|-|-|-|<br>| upstream framing │ \[DONE\] │ wall clock │<br>│ \[DONE\] + terminating chunk + close            │ yes    │ 0.08s      │<br>│ \[DONE\] + terminating chunk + keep-alive       │ yes    │ 0.07s      │<br>│ no \[DONE\] + terminating chunk                 │ yes    │ 0.07s      │<br>│ \[DONE\], body never terminated (live endpoint) │ no     │ hangs      │<br>│ no \[DONE\], body never terminated              │ no     │ hangs      │

Both hanging rows return yes / \~2s once the drain is bounded and BedrockMantle is added to ProviderSendsDoneMarker.

### Expected behavior

After the terminal chunk, the stream emits data: \[DONE\] and the response body is terminated, regardless of whether the upstream terminates its own body. An upstream that fails to close should be bounded and its connection dropped rather than reused - never allowed to block the defer chain that closes the provider's response channel.

### Actual behavior

All content chunks plus the synthesised terminal chunk (finish_reason: "stop", populated usage) are delivered, then the response stalls indefinitely with no data: \[DONE\] and no body termination. The provider goroutine is parked in the deferred drain; the response channel is never closed and the transport's SSE producer never returns. One goroutine and one upstream connection are leaked per request.

### Affected area(s)

Core (Go)

### Version

v1.7.6

### Environment

```text
Environment

- Bifrost: core v1.7.4, v1.7.5, and v1.7.6, transports v1.6.6, framework v1.5.4 (embedded as a library in a thin Go wrapper, -mod=vendor)
- Also verified present in core v1.7.5 and on main (ProviderSendsDoneMarker and ReleaseStreamingResponse unchanged in both) — not a regression, bedrock_mantle streaming has never worked
- Provider: bedrock_mantle, model openai.gpt-oss-20b, region us-east-1, SigV4 via the AWS default credential chain
- Go 1.26.4, alpine 3.23 container, linux/arm64
- Non-streaming bedrock_mantle chat completions work correctly; openai streaming through the same gateway works correctly
- Affected area: Core (Go) · Severity: High (major functionality broken)
```

### Relevant logs/output

```shell
Client — every chunk arrives, then nothing (live endpoint, trimmed):

data: {...,"delta":{"reasoning":"The user says: \"Count from 1 to 5...\""},...,"chunk_index":0}
data: {...,"delta":{"content":"1\n2\n3\n4\n5"},...,"chunk_index":4}
data: {...,"finish_reason":"stop","delta":{},"usage":{"prompt_tokens":80,"completion_tokens":154,"total_tokens":234},...,"chunk_index":5}
curl: (18) Transferred a partial file

SIGQUIT goroutine dump of the hung gateway — the provider goroutine is parked in the deferred drain:

goroutine 5368 [IO wait, 3 minutes]:
internal/poll.(*pollDesc).wait(...)
crypto/tls.(*Conn).Read(...)
bufio.(*Reader).ReadByte(...)
github.com/valyala/fasthttp.readHexInt(...)
github.com/valyala/fasthttp.parseChunkSize(.../fasthttp/http.go:2748)
github.com/valyala/fasthttp.(*requestStream).Read(.../fasthttp/streaming.go:32)
io.discard.ReadFrom(...)
io.copyBuffer(...)
io.Copy(...)
github.com/maximhq/bifrost/core/providers/utils.ReleaseStreamingResponse(.../providers/utils/utils.go:2904)
github.com/maximhq/bifrost/core/providers/openai.HandleOpenAIChatCompletionStreaming.func2.deferwrap2(.../providers/openai/openai.go:1144)
github.com/maximhq/bifrost/core/providers/openai.HandleOpenAIChatCompletionStreaming.func2(.../providers/openai/openai.go:1447)
created by ...HandleOpenAIChatCompletionStreaming in goroutine 4213
      .../providers/openai/openai.go:1132

openai.go:1144 is defer providerUtils.ReleaseStreamingResponse(ctx, resp), registered after — and so running before — the defer at openai.go:1134 that calls CloseStream(ctx, responseChan).

Gateway logs the request as complete while the client is still hanging, and the fronting proxy never logs a completion at all:


{"level":"info","http.method":"POST","http.target":"/v1/chat/completions","http.status_code":200,"http.request_duration_ms":4912,...,"message":"request completed"}
```

### Regression?

Have not tried in previous versions

### Severity

Low (minor issue or cosmetic)

zbloss:
## Summary

bedrock_mantle streaming chat completions never finish: every chunk is delivered, including the terminal chunk with finish_reason and usage, and then the response stalls - `data: [DONE]` is never sent and the HTTP body is never terminated, leaking one goroutine and one upstream socket per request. Two defects combine to cause this, and this PR fixes both.

## Changes

- core/providers/utils/utils.go - `ReleaseStreamingResponse` drained the upstream body with an unbounded `io.Copy`. `BuildStreamingClient` sets `ReadTimeout = 0`, and the drain runs after the SSE loop's own idle-timeout reader has already been stopped, so nothing bounds it - an upstream that holds its chunked body open past the terminal event (as bedrock_mantle does) blocks the copy forever, parking the provider goroutine in this defer before the defer that closes the response channel runs. The drain now runs in its own goroutine and races it against `GetStreamIdleTimeout(ctx);` on timeout the connection is force-closed with `ErrStreamIdleTimeout` (reusing the same sentinel NewIdleTimeoutReader's timer already uses for this failure mode, rather than adding a duplicate) so fasthttp drops it instead of returning it to the idle pool.
- core/providers/utils/utils.go - ProviderSendsDoneMarker didn't list `schemas.BedrockMantle`, so the SSE loop waited indefinitely for a [DONE] marker Mantle never sends. Added it alongside the existing schemas.Bedrock case.
- Design decision: on the forced-close timeout path, `resp.CloseBodyStream()`/`fasthttp.ReleaseResponse(resp)` are deliberately not called — the drain goroutine may still be mid-Read on the body stream, and a second concurrent close/release would double-Put fasthttp's pooled reader/requestStream, corrupting a later, unrelated request. resp is intentionally leaked to GC in that rare path, the same trade-off already used elsewhere in this function when close ownership is uncertain.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [x] Core (Go)
- [ ] Transports (HTTP)
- [ ] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

```
go version
cd core
go test ./providers/utils/... -race -v -run 'TestReleaseStreamingResponse|TestProviderSendsDoneMarker'
go test ./...
```

Expected: 1TestReleaseStreamingResponse_BoundsDrainAndDropsConnection1 completes in well under the 2s bound instead of hanging (reproduces the reported hang against a body stream that never terminates), `TestReleaseStreamingResponse_NormalDrainReleasesWithoutForcingClose` and `TestProviderSendsDoneMarker` (including the new bedrock_mantle case) pass, and the full `go test ./... run` shows no new failures (providers/bedrock and internal/mcptests have pre-existing, unrelated failures on dev).

Manual repro against a mock upstream (from the linked issue): run a TLS server aliased to bedrock-mantle.us-east-1.api.aws that sends a few delta.content frames, a terminal frame with finish_reason, data: [DONE], and then never sends the closing 0\r\n\r\n chunk. Configure a bedrock_mantle key against it and stream a request through /v1/chat/completions. Before this change the client hangs indefinitely; after, data: [DONE] and the stream end arrive within the configured stream_idle_timeout_in_seconds.

## Breaking changes

- [ ] Yes
- [x] No

## Related issues

Closes maximhq/bifrost#5871
Relates to maximhq/bifrost#2910 (same missing-[DONE]-marker defect, reported for custom providers; that report's hang self-heals at the stream idle timeout, this one didn't because of the unbounded drain)

Security considerations

None. The change only affects timing/cleanup of an already-authenticated provider connection: it bounds how long a stalled upstream can hold a goroutine and socket open, and ensures a connection that failed to terminate cleanly is dropped rather than returned to the pool for reuse (avoiding stale-connection reuse, not introducing it). No auth, secrets, or PII handling is touched.
## Checklist


zbloss:
@akshaydeo @TejasGhatte do you have an estimate on when you will review these changes? I'm wondering if I need to fork this branch and patch my local deployments.

TejasGhatte (dismissed):

TejasGhatte:
> @akshaydeo @TejasGhatte do you have an estimate on when you will review these changes? I'm wondering if I need to fork this branch and patch my local deployments.

Hey @zbloss approved it. @akshaydeo will merge soon

zbloss:
@akshaydeo Do I need to fix these merge conflicts?

zbloss:
Looks like this was addressed in 1.7.11? https://github.com/maximhq/bifrost/releases#release-core/v1.7.11

zbloss:
@TejasGhatte @akshaydeo I think we still need either this PR or another that changes https://github.com/maximhq/bifrost/blob/5bb5c653b7a0c3125c66c167fe542ac820afb122/core/providers/utils/utils.go#L3183-L3185

Bedrock Mantle is not listed here even though it's in the code comments, which is causing our bedrock mantle streams to hang open waiting on a `[DONE]` marker that never arrives

I think this should be

```
case schemas.Cerebras, schemas.Perplexity, schemas.HuggingFace, schemas.Bedrock, schemas.BedrockMantle:
		// Cerebras, Perplexity, HuggingFace, and Bedrock mantle don't send [DONE] marker, ends stream after finish_reason
		return false

zbloss:
This looks related: https://github.com/maximhq/bifrost/pull/6021

PR #5882, opened 2026-08-05
[fix]: Sanitize NUL bytes in PostgreSQL logs

james-norm:
## Summary

PostgreSQL rejects U+0000 in TEXT and VARCHAR binds, which caused valid provider responses to drop their log rows when decoded output flowed into `content_summary`.

## Changes

- Sanitize NUL bytes from PostgreSQL-bound model text on create and update, including map and pointer updates.
- Keep parsed request/response fields untouched; JSON payload columns retain escaped U+0000 content.
- Add real PostgreSQL regressions for create and logging-plugin-style update paths.

## Type of change

- [x] Bug fix

## Affected areas

- [x] Framework

## How to test

```sh
cd framework
go test ./logstore -count=1
cd ../plugins/logging
go test ./... -count=1
```

## Breaking changes

- [x] No

## Security considerations

No auth, secret, or PII handling changes.

## Checklist

- [x] I read the contribution guidelines and followed them
- [x] I updated the framework changelog
- [x] I verified the affected Go tests and lint checks locally


framework/logstore/postgressanitize.go
```diff
@@ -0,0 +1,93 @@
...
+
+import (
+	"reflect"
+	"strings"
+)
+
+// sanitizePostgresModelTextFields removes NUL bytes from persisted string
+// fields. PostgreSQL rejects U+0000 in text and varchar values, while JSON
+// payload fields remain intact because their NUL characters are escaped.
+func sanitizePostgresModelTextFields(model any) {
```
akshaydeo, line 11: this seems like an expensive operation - let me find a better way

PR #5896, opened 2026-08-06
semantic router: incremental embedding + embed usage accounting

kohlivrinda:
## Summary

This PR fixes several correctness and cost bugs across the semantic classifier, vector store backends, embedding usage accounting, and the complexity router UI. The most significant addition is an in-process embedding cache that reuses vectors for unchanged exemplar phrases across warmup runs, so editing a tier list only pays the embedding provider for genuinely new text. A collection of integer overflow, NaN, and token-accounting bugs are also addressed.

## Changes

- Added `semanticEmbeddingCache`, a thread-safe in-process map from exemplar phrase to embedding vector. The cache identity is derived from the provider and embedding model; changing either drops every entry, because a vector from a different model is not stale — it is wrong. After a successful warmup, `cache.retain` prunes the map to only the phrases that were actually used, preventing accumulation of vectors for removed phrases.
- `warmSemanticExemplars` now accepts a `*semanticEmbeddingCache`, checks which phrases already have a cached vector before calling the provider, and writes vectors to the cache only after the dimension check passes. Error returns in the embedding loop no longer carry partial-progress counts.
- `SemanticStatusInfo` gains a `CachedPhrases` field reporting how many vectors the gateway currently holds. Because the cache is in-process only, a restart empties it while the saved phrases look unchanged; the UI uses this field to distinguish a warm cache from a cold one rather than inferring reuse from the persisted config.
- Fixed an `int64` overflow in `ChromemStore.GetAll`: forming `offset + limit` with an unbounded caller-supplied limit wraps negative, passes the `end > total` guard, and panics on the slice. The fix compares `limit` against the remaining count instead.
- Fixed `ChromemStore.GetNearest` preallocating a slice of size `limit` rather than the actual result count, which panics for extreme limit values.
- Introduced `boundedPageLimit` to safely convert a caller-supplied `int64` limit to the `uint32` the Qdrant and Pinecone clients require. A direct cast wraps values above `MaxUint32`, and an exact multiple of `2^32` wraps to zero, which the Qdrant scroll loop reads as a filled page and hands back a cursor that never advances.
- Fixed `ComplexitySemanticConfig.Validate` to explicitly reject `NaN` for `MinSimilarity`. Every comparison against `NaN` is false, so the existing range check silently accepted it and disabled the similarity floor.
[...]

PR #5865, opened 2026-08-05
semantic router ui revamp

kohlivrinda:
## Summary

Simplifies the Complexity Router to be semantic-only, removing the lexical/semantic mode toggle and making embedding-based classification the single path. The page now treats phrase lists as reference phrases for the semantic classifier rather than dual-purpose keyword/exemplar lists, and the backend merges editable phrases onto the built-in keyword vocabulary instead of replacing it.

## Changes

- **Removed lexical/semantic mode toggle**: The page no longer has a `ClassificationMode` state or tab switcher. Semantic classification is the only offered mode; the lexical scorer still runs internally but is no longer user-facing.
- **Editable phrase lists are now exemplars-only**: `DefaultEditableKeywordConfig` returns only the semantic exemplars from `configstore`, not the combined keyword+exemplar list. The lexical scorer's built-in vocabulary is kept out of the administrator-facing lists.
- **`mergeEditableKeywordsOntoDefaults` now adds to built-ins**: Instead of replacing a tier's keyword list when the editable list is non-empty, it appends the editable phrases onto the built-in vocabulary, deduplicating by normalized key. This prevents the lexical scorer from silently losing its short-word signals when only sentence-length exemplars are present.
- **`sharedTierDefaults` deduplicates**: The helper now tracks seen entries (lowercased, trimmed) and skips duplicates when merging extra phrases onto a base list.
- **Tier boundary controls removed from the UI**: Boundaries are carried through the form untouched and validated only for structural correctness. An out-of-range persisted value falls back to defaults rather than blocking every save.
- **Fallback option removed**: The `fallback: "lexical"` option is no longer offered. Configs carrying it are normalized to `"none"` on the next save. The `external` vector store option is similarly hidden; persisted values are presented and re-saved as `"auto"`.
- **`SemanticStatusPanel` updated**: Reflects the new single-mode design — shows "Not configured" when no provider/model is set, and consolidates the failed-state messaging to a single callout regardless of fallback setting.
- **`TagInput` collapse behavior changed from count-based to height-based**: `collapsedTagLimit` replaced by `collapsedMaxHeight` (px). A `ResizeObserver` measures actual rendered height so columns of different-length phrases collapse to the same visual size. The expand/collapse toggle is a single button that switches label rather than two separate buttons.
[...]

PR #5791, opened 2026-08-03
server: export ObserveWarmupRoutingEmbedding for enterprise

kohlivrinda:
## Summary

`observeWarmupRoutingEmbedding` was unexported, making it unreachable from embedder packages that reimplement `Bootstrap` rather than delegating to it. Those embedders could not wire the warmup embedding usage observer, causing warmup embed token usage to be silently dropped with no recovery path.

## Changes

- Exported `observeWarmupRoutingEmbedding` → `ObserveWarmupRoutingEmbedding` so embedders in external packages can reference it directly through the embedded server when setting up their own `Bootstrap` logic.
- Updated both call sites (`Bootstrap` and `ReloadPlugin`) to reference the exported name.
- Expanded the method comment to explain why it is exported and the consequence of missing this wiring.

## Type of change

- [ ] Bug fix
- [ ] Feature
- [x] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [ ] Core (Go)
- [x] Transports (HTTP)
- [ ] Providers/Integrations
- [x] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

Verify that an embedder package that embeds `BifrostHTTPServer` and calls `SetWarmupEmbedUsageObserver` with `s.ObserveWarmupRoutingEmbedding` compiles successfully, and that warmup embedding token usage is recorded in Prometheus routing overhead counters after a governance plugin reload or server bootstrap.

```sh
go test ./...
```

## Breaking changes

- [ ] Yes
- [x] No

## Security considerations

None. This is a visibility change on an internal telemetry observer; no auth, secrets, or PII are involved.

## Checklist

- [ ] I read `docs/contributing/README.md` and followed the guidelines
- [ ] I added/updated tests where appropriate
- [ ] I updated documentation where needed
- [ ] I verified builds succeed (Go and UI)
- [ ] I verified the CI pipeline passes locally if applicable

PR #5790, opened 2026-08-03
weaviate + redis adapter fixes

kohlivrinda:
## Summary

The Redis vector store's dimension-mismatch guard was silently never running. The typed `FTInfo` helper used to read the existing index's declared dimension refuses to decode when the client speaks RESP3 without `UnstableResp3`, which is exactly how this store is configured. As a result, creating a namespace with a mismatched dimension would silently reuse the old index instead of returning an error. The fix reads the dimension directly from the raw `FT.INFO` reply, handling both the RESP3 map shape and the RESP2 flat key/value sequence shape.

Additionally, the Weaviate `GetChunk` implementation was returning a raw client error on 404 rather than the sentinel `ErrNotFound` that callers expect, and `GetChunks` was propagating a 404 as a hard failure even though other backends (Qdrant, Pinecone, Redis) simply omit missing IDs from the result set.

## Changes

- Replaced the `FTInfo` typed helper call in `CreateNamespace` with `vectorDimensionFromFTInfo`, which parses the raw `FT.INFO` reply directly and handles both RESP2 and RESP3 wire shapes.
- Added `vectorDimensionFromFTInfo` and `ftInfoField` helpers that walk the raw reply across both protocol shapes to extract the declared vector dimension.
- Added `isWeaviateNotFound` helper that unwraps a `*fault.WeaviateClientError` and checks for HTTP 404.
- `GetChunk` now wraps both the 404 client error and the empty-result case as `fmt.Errorf("%w: %s", ErrNotFound, id)`, making the sentinel checkable via `errors.Is`.
- `GetChunks` now skips missing IDs (404) instead of returning an error, aligning its behaviour with other vector store implementations.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [ ] Core (Go)
- [ ] Transports (HTTP)
- [x] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

```sh
go test ./framework/vectorstore/...
```

- Verify `TestVectorDimensionFromFTInfo` passes across all RESP2 and RESP3 reply shapes.
- Verify `TestRedisStore_CreateNamespaceRejectsDimensionChange` fails when recreating an existing index with a different dimension and succeeds when the dimension matches.
- Call `GetChunk` with a non-existent ID and verify the returned error satisfies `errors.Is(err, ErrNotFound)`.
[...]

PR #5901, opened 2026-08-06
[DO-NOT-MERGE-YET] session routing config and sticky targets

kohlivrinda:
## Summary

Adds session-aware routing to the complexity analyzer, giving the router a memory of which tier a conversation was assigned to. Within a session, target selection is deterministic (hash-based) rather than random, so repeated turns in the same conversation consistently resolve to the same provider and preserve the provider-side prompt cache. A new `ComplexitySessionConfig` block controls the behaviour, with two modes (`pinned` and `cache_aware`) and an explicit `off` value so the UI can disable session behaviour without discarding tuned settings.

## Changes

- **`ComplexitySessionConfig`**: New config struct with `mode`, `ttl`, `identity_sources`, `release_after_failures`, `switch_min_similarity`, `downgrade_after_n_turns`, `min_cached_tokens_to_hold`, `max_switches_per_session`, and `always_allow_escalation`. TTL accepts both duration strings (`"30m"`) and millisecond numbers, matching the existing semantic timeout encoding. A nil block and an explicit `"off"` mode are equivalent at runtime; `Enabled()` is the single correct test for either.
- **Identity sources**: Three resolution steps — `header` (x-bf-session-id), `harness` (native harness ID), and `fingerprint` (prompt-content hash). Fingerprinting is excluded from defaults and must be opted into. Sources are normalised into resolution-ladder order regardless of input order; unknown values are preserved through normalisation so `Validate` can name them rather than silently dropping a typo.
- **Hysteresis enforcement**: `SwitchMinSimilarity` must be at least `Semantic.MinSimilarity`. A classification too weak to be believed must not be strong enough to move a session. Zero is exempt — it disables the switch gate entirely rather than setting a threshold of zero.
- **`selectWeightedTarget` sticky selection**: When a `SessionID` is present, the selection point is derived from `hash(sessionID, ruleID)` instead of `rand.Float64()`. The target slice is sorted by a stable identity key (provider, model, key ID) before the cumulative walk, so stickiness does not silently degrade when the DB returns rows in a different order. The sort operates on a copy to avoid mutating the shared cached config slice. Keying on the rule as well as the session decorrelates draws across rules, preserving aggregate weight distribution. The all-zero-weight uniform branch is also made sticky.
- **`sessionSelectionPoint`**: Maps session and rule to a uniform `[0,1)` point using 53 bits of SHA-256, matching the exact-integer range of a float64 so no value is ever 1.0 and every representable value is equally likely.
- **`SessionSettings` hash**: Added to `ComplexityAnalyzerConfigHashes` and computed in `GenerateComplexityAnalyzerConfigHashes`. The hash-based merge follows the same "absence means no opinion" rule as the semantic section — a config.json without a session block leaves DB session state untouched.
- **`SessionID` in `RoutingContext`**: Populated from the `x-bf-session-id` header in the governance plugin's `applyRoutingRules`. The full resolution ladder (harness-native IDs, prefix fingerprinting) will replace this single read without changing the field.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [x] Core (Go)
- [ ] Transports (HTTP)
- [ ] Providers/Integrations
- [x] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

```sh
go test ./framework/configstore/... ./plugins/governance/...
```

Key scenarios covered by the new tests:

- TTL decoding from duration strings, millisecond numbers, absent fields, and invalid inputs
- Marshal/unmarshal round-trip preserves TTL (nanosecond default encoding would misread by six orders of magnitude)
- Normalisation applies all defaults; fingerprint is never defaulted
- `Enabled()` returns false for nil, empty mode, and explicit `"off"`
- Disabling session mode preserves tuned TTL and downgrade settings
- Identity source de-duplication and resolution-order sorting
- Unknown identity sources are rejected by `Validate` with the offending value named
- `SwitchMinSimilarity` below `Semantic.MinSimilarity` is rejected; zero is exempt
- Session config round-trips through encode/decode
- Hash-based merge keeps session state when file omits the block; applies it when the hash changes
- Sticky selection is consistent within a session, independent of target slice order, does not mutate the caller's slice, matches weight distribution across sessions, decorrelates across rules, handles all-zero weights, and leaves no-session behaviour unchanged
- `sessionSelectionPoint` output is uniform across `[0,1)`
- `routingTargetIdentity` separates fields correctly and treats nil and empty as equivalent

## Breaking changes

- [ ] Yes
- [x] No

## Security considerations

`sessionSelectionPoint` uses SHA-256 keyed on session and rule IDs. Session IDs sourced from the `x-bf-session-id` header are caller-supplied; no secrets are derived from them and the hash is used only for target selection, not authentication.

## Checklist

### DIFF
--- a/framework/configstore/complexityconfig.go
+++ b/framework/configstore/complexityconfig.go
@@ -327,6 +327,289 @@ func (c *ComplexitySemanticConfig) Validate() error {
 	return nil
 }
 
+// Complexity session modes.
+//
+// "off" is a first-class value rather than only an absent block so the UI's
+// three-way toggle can disable session behaviour without discarding the TTL and
+// threshold settings an admin already tuned. A nil *ComplexitySessionConfig and
+// an explicit "off" are equivalent at runtime; use Enabled to test either.
+//
+// "pinned" classifies once per session and reuses the tier until the pin is
+// released. "cache_aware" classifies every request and gates tier changes; it
+// cannot skip classification because it needs the proposal to decide.
+const (
+	ComplexitySessionModeOff        = "off"
+	ComplexitySessionModePinned     = "pinned"
+	ComplexitySessionModeCacheAware = "cache_aware"
+)
+
+// Complexity session identity sources, listed in resolution order: an explicit
+// x-bf-session-id header, then a harness-native ID.
+const (
+	ComplexitySessionIdentityHeader  = "header"
+	ComplexitySessionIdentityHarness = "harness"
+)
+
+// Complexity session defaults, applied by normalized when a field is unset.
+// SwitchMinSimilarity is deliberately absent: a sensible absolute value depends
+// on the exemplar set's score distribution, so it ships unset (no gating).
+const (
+	DefaultComplexitySessionTTL                   = time.Hour
+	DefaultComplexitySessionReleaseAfterFailures  = 3
+	DefaultComplexitySessionDowngradeAfterNTurns  = 2
+	DefaultComplexitySessionMinCachedTokensToHold = 1024
+)
+
+// DefaultComplexitySessionIdentitySources is the identity ladder applied when
+// none is configured.
+func DefaultComplexitySessionIdentitySources() []string {
+	return []string{ComplexitySessionIdentityHeader, ComplexitySessionIdentityHarness}
+}
+
+// ComplexitySessionConfig gives the complexity router a memory of the
+// conversation a request belongs to. A nil value, like Mode "off", disables it.
+//
+// Session state is a memoised classifier result with a TTL, not an override of
+// routing: it replaces the lazy complexity computation, so every routing rule
+// still evaluates normally and only the classification is skipped.
+type ComplexitySessionConfig struct {
+	Mode string `json:"mode"`
+	// TTL bounds how long a session record survives without traffic. It slides
+	// on every read, so it measures idle time rather than total session age.
+	TTL time.Duration `json:"ttl,omitempty"`
+	// IdentitySources selects which of the resolution steps may establish a
+	// session ID, tried in the constant order above regardless of the order
+	// given here.
+	IdentitySources []string `json:"identity_sources,omitempty"`
+	// ReleaseAfterFailures releases a pin after this many consecutive upstream
+	// failures, on the assumption that the pinned tier is implicated.
+	ReleaseAfterFailures int `json:"release_after_failures,omitempty"`
+
+	// SwitchMinSimilarity is the confidence a classification must reach before
+	// it may change session state, forming hysteresis with the semantic
+	// classifier's MinSimilarity: a low bar to classify, a higher bar to switch.
+	// It must be at least Semantic.MinSimilarity. Zero means no similarity
+	// gating at all — escalations pass and the sustained-confidence downgrade
+	// check is skipped — rather than a threshold of zero.
+	//
+	// Like MinSimilarity this is compared against the VectorStore backend's own
+	// score, and those scales differ (chromem, Qdrant, Pinecone, and Redis
+	// report raw cosine; Weaviate reports certainty). Retune on a backend
+	// switch. A lexical score is a different scale entirely and is never
+	// comparable against this value.
+	SwitchMinSimilarity float64 `json:"switch_min_similarity,omitempty"`
+	// DowngradeAfterNTurns is how many consecutive turns a lower tier must be
+	// proposed before a downgrade is taken. Long agentic sessions are full of
+	// turns like "yes" or "run the tests" that classify as confidently simple;
+	// those are exactly the turns not to downgrade on, because the classifier is
+	// right about the turn and wrong about the session.
+	DowngradeAfterNTurns int `json:"downgrade_after_n_turns,omitempty"`
+	// MinCachedTokensToHold is the observed cached-token count below which the
+	// current tier's cache is considered too small to be worth protecting.
+	MinCachedTokensToHold int `json:"min_cached_tokens_to_hold,omitempty"`
+	// MaxSwitchesPerSession caps total tier moves as an oscillation backstop.
+	// Zero means unlimited.
+	MaxSwitchesPerSession int `json:"max_switches_per_session,omitempty"`
+	// AlwaysAllowEscalation lets an escalation bypass the SwitchMinSimilarity
+	// gate. Escalations already ignore cache cost: holding a session on an
+	// undersized model to protect cache spend produces bad answers, which is a
+	// worse failure than overpaying.
+	AlwaysAllowEscalation bool `json:"always_allow_escalation,omitempty"`
+}
+
+// UnmarshalJSON accepts TTL as a duration string ("30m") or a JSON number
+// (milliseconds), matching ComplexitySemanticConfig.Timeout. All other fields
+// decode through the default path via an alias.
+func (c *ComplexitySessionConfig) UnmarshalJSON(data []byte) error {
+	// alias suppresses ComplexitySessionConfig's UnmarshalJSON to avoid
+	// infinite recursion. The outer TTL (json.RawMessage) shadows alias.TTL
+	// because the json package picks the shallower field.
+	type alias ComplexitySessionConfig
+	aux := &struct {
+		TTL json.RawMessage `json:"ttl,omitempty"`
+		*alias
+	}{alias: (*alias)(c)}
+	if err := json.Unmarshal(data, aux); err != nil {
+		return err
+	}
+
+	if len(aux.TTL) == 0 || string(aux.TTL) == "null" {
+		return nil
+	}
+
+	var s string
+	if err := json.Unmarshal(aux.TTL, &s); err == nil {
+		d, err := time.ParseDuration(s)
+		if err != nil {
+			return fmt.Errorf("failed to parse session ttl duration string %q: %w", s, err)
+		}
+		c.TTL = d
+	} else {
+		var ms float64
+		if err := json.Unmarshal(aux.TTL, &ms); err != nil {
+			return fmt.Errorf("unsupported session ttl value: %s", string(aux.TTL))
+		}
+		c.TTL = time.Duration(ms * float64(time.Millisecond))
+	}
+	if c.TTL < 0 {
+		return fmt.Errorf("session ttl must be non-negative, got %v", c.TTL)
+	}
+	return nil
+}
+
+// MarshalJSON writes TTL as a duration string so persisted configs decode back
+// to the same value (the default int encoding is nanoseconds, which the
+// millisecond-number decode path would misread).
+func (c ComplexitySessionConfig) MarshalJSON() ([]byte, error) {
+	type alias ComplexitySessionConfig
+	var ttl string
+	if c.TTL != 0 {
+		ttl = c.TTL.String()
+	}
+	return json.Marshal(struct {
+		TTL string `json:"ttl,omitempty"`
+		alias
+	}{
+		TTL:   ttl,
+		alias: alias(c),
+	})
+}
+
+// Enabled reports whether session behaviour should engage. It is the only
+// correct way to test this: a nil block, an empty mode, and an explicit "off"
+// all mean disabled.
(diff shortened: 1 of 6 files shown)


kohlivrinda:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/maximhq/bifrost/5901?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#6074** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/6074?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6061** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/6061?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6045** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/6045?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6043** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/6043?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6042** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/6042?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6029** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/6029?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6028** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/6028?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6018** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/6018?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6015** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/6015?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6009** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/6009?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5901** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/5901?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/maximhq/bifrost/5901?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#5896** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/5896?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#5924](https://github.com/maximhq/bifrost/pull/5924) <a href="https://app.graphite.com/github/pr/maximhq/bifrost/5924?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#5865** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/5865?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5842** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/5842?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5810** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/5810?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5791** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/5791?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5790** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/5790?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5743** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/5743?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5748** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/5748?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5742** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/5742?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5686** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/5686?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5680** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/5680?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5656** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/5656?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5655** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/5655?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5628** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/5628?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5602** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/5602?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5530** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/5530?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5529** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/5529?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5499** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/5499?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5497** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/5497?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5494** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/5494?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `dev`


<h2></h2>

harryzsh:
Really glad to see session-aware routing landing in core — we hit this exact problem class in production and ended up building an external plugin for it, so I wanted to share some field notes that may be useful for this PR and the follow-up fingerprinting work.

**Our setup**: Bifrost in front of Amazon Bedrock with multiple API keys (= multiple AWS accounts). Bedrock's prompt cache is isolated per account, so weighted-random key selection meant every key switch dropped the entire provider-side cache and re-billed the full prefix at full price. We built a cache-affinity plugin that pins "same conversation / same long prefix" to the same key: https://github.com/harryzsh/bifrost-cache-affinity-plugin (verified against real Bedrock: first turn cache write ≈38k tokens, every subsequent turn reads it back with only ~17 tokens of new write).

Three lessons from running this that map directly onto this PR's design:

1. **Prefix fingerprinting must strip cache markers before hashing.** The PR description mentions prefix fingerprinting as a future identity source. One trap we hit: Anthropic's own recommended multi-turn pattern moves the `cache_control` breakpoint to the latest message every turn, so a naive hash of the prefix drifts on every turn and the session is never recognized. We ended up anchoring identity on (tools + leading system/developer messages + first non-system message) with all cache marker fields stripped before hashing — stable from turn 1 to turn N.

2. **Zero-benefit protection needs the "provider actually cached nothing" case.** `min_cached_tokens_to_hold` gates on observed cached tokens, which matches what we converged on. One extra case worth covering: when the response shows read + write = 0 (e.g. the caller's cache markers are placed in a way the provider silently ignores), holding the session takes on affinity risk with zero cache benefit. We skip recording entirely in that case.

3. **Assert the pin was honored.** Between "plugin declares key X" and "core actually used key Y" there's room for silent contract drift across versions. We log an error-level `pin_ignored` event whenever the observed selection differs from the pinned one — it has caught real regressions. A similar assertion inside core (session says target T, selection returned T') would make failures loud instead of showing up as mysterious cache misses.

One clarifying note on scope, since users may conflate the two: this PR pins the **target (provider/model)**, while provider-side prompt cache on Bedrock is scoped **per key/account** — so with multiple keys under one provider, a sticky target can still rotate keys and lose the cache. That's the gap our plugin covers (via `RoutingPinnedAPIKeyID`), and the two compose cleanly: core handles tier/target stickiness, the plugin handles key-level affinity.

Happy to share our test scripts (load / chaos / dual-node Redis / e2e against real Bedrock) or run this branch against our setup if useful.

framework/configstore/complexityconfig.go
```diff
@@ -327,6 +327,289 @@ func (c *ComplexitySemanticConfig) Validate() error {
...
+// still evaluates normally and only the classification is skipped.
+type ComplexitySessionConfig struct {
+	Mode string `json:"mode"`
+	// TTL bounds how long a session record survives without traffic. It slides
+	// on every read, so it measures idle time rather than total session age.
+	TTL time.Duration `json:"ttl,omitempty"`
+	// IdentitySources selects which of the resolution steps may establish a
+	// session ID, tried in the constant order above regardless of the order
+	// given here.
+	IdentitySources []string `json:"identity_sources,omitempty"`
```
harryzsh, line 383: When the fingerprint source lands in the resolution ladder: consider hashing only (tools + leading system messages + first non-system message) rather than the full prefix, and strip `cache_control` / `cachePoint` / `prompt_cache_breakpoint` fields before hashing. Anthropic's recommended multi-turn pattern slides the cache breakpoint every turn, so any hash that includes marker fields (or the growing history) drifts and the session is never re-identified. We shipped exactly this bug first in our external cache-affinity plugin, and the marker-stripping + "birth certificate" anchor fixed it.

framework/configstore/complexityconfig.go
```diff
@@ -327,6 +327,289 @@ func (c *ComplexitySemanticConfig) Validate() error {
...
+	SwitchMinSimilarity float64 `json:"switch_min_similarity,omitempty"`
+	// DowngradeAfterNTurns is how many consecutive turns a lower tier must be
+	// proposed before a downgrade is taken. Long agentic sessions are full of
+	// turns like "yes" or "run the tests" that classify as confidently simple;
+	// those are exactly the turns not to downgrade on, because the classifier is
+	// right about the turn and wrong about the session.
+	DowngradeAfterNTurns int `json:"downgrade_after_n_turns,omitempty"`
+	// MinCachedTokensToHold is the observed cached-token count below which the
+	// current tier's cache is considered too small to be worth protecting.
+	MinCachedTokensToHold int `json:"min_cached_tokens_to_hold,omitempty"`
```
harryzsh, line 409: Worth an explicit branch for observed `cache_read + cache_write == 0`: it means the provider built no cache at all for this shape of request (commonly: markers placed where the provider silently ignores them), so holding the session carries affinity risk with zero benefit. We treat it as "skip recording" rather than comparing against the threshold — a threshold comparison happens to cover it while the default is >0, but the intent is different, and an explicit case survives config changes like setting this to 0.

plugins/governance/routing.go
```diff
@@ -306,13 +314,21 @@ func (re *RoutingEngine) EvaluateRoutingRules(ctx *schemas.BifrostContext, routi
...
-func selectWeightedTarget(targets []configstoreTables.TableRoutingTarget) (configstoreTables.TableRoutingTarget, bool) {
+//
+// When sessionID is non-empty the selection point is derived from
+// hash(sessionID, ruleID) instead of rand, so every request in a session picks
+// the same target for a given rule. That stickiness is what makes tier state
+// worth having: a tier resolving to two weighted targets would otherwise re-roll
+// every turn and discard the provider-side prompt cache anyway, one layer below
+// the tier. The hash is stateless, so it agrees across nodes and survives
+// restarts without storage, and it stays proportional as weights change.
+func selectWeightedTarget(targets []configstoreTables.TableRoutingTarget, sessionID, ruleID string) (configstoreTables.TableRoutingTarget, bool) {
```
harryzsh, line 331: Since this pins the target but key selection within the provider can still rotate: on providers where prompt cache is scoped per key/account (Bedrock), a sticky target alone won't preserve the cache across turns. Might be worth a doc note so users don't attribute cache misses to this feature. Also, from running key-level pinning in production: an assertion that the final selection matches the session's pinned target (error log on mismatch) has been our single most valuable observability signal — contract drift shows up as loud errors instead of silent cache misses.

PR #5929, opened 2026-08-07
fix(bedrock): keep the cache breakpoint when inlining a mid-conversation system reminder

anduril-rvanderzee:
## Summary

A mid-conversation `role:"system"` message loses its `cache_control` when inlined for Bedrock, so the client's conversation-level cache breakpoint never reaches the provider: **3 `cache_control` breakpoints in, 2 `cachePoint`s out.** With that anchor gone the cacheable prefix is pinned at the system/tools floor and the entire conversation body is re-read uncached on every turn.

`role:"system"` inside `messages` is a supported Anthropic API feature (mid-conversation system messages, Opus 4.8+ — `SupportsMidConversationSystem` returns `false` for Bedrock, correctly). Claude Code uses it for `<system-reminder>` turns, so this fires on ordinary Claude Code traffic through Bedrock.

## Reproduction

`claude-opus-5` on Bedrock via `/anthropic/v1/messages`. Both arms send identical content and an identical third breakpoint; the only difference is whether the tail breakpoint rides on a `role:"user"` or `role:"system"` turn. Each arm is called twice — once to warm the prefix, once to measure.

| Tail breakpoint rides on | uncached `input_tokens` | `cache_read_input_tokens` | Real hit rate |
|---|---:|---:|---:|
| `role:"user"` (control) | 2 | 36,115 | **100.0%** |
| `role:"system"` | 18,116 | 18,018 | **49.9%** |

Roughly half the input goes uncached on an otherwise identical, already-warm prefix.

Wire-level, same pair of calls — both received `cache_control` × 3:

| Arm | `cachePoint` emitted | Placement |
|---|---:|---|
| control | 3 | 2 in `system` + 1 in `messages` |
| affected | 2 | 2 in `system`, **none in `messages`** |

Also reproduces via `/openai/v1/responses` with `claude-sonnet-5` (100.0% → 66.5%), so it is in the Bedrock converter rather than one inbound translation layer.

The reproduction is self-contained: any Claude-Code-shaped payload with two `system` breakpoints plus a third on a trailing `role:"system"` turn shows it, and the arms differ by ~18K cached tokens — far outside noise. Sampling large real requests and inspecting the emitted Converse body separated cleanly on the same marker: affected requests emit 2 `cachePoint`s and end in a `role:system` item, unaffected ones emit 3.

A representative sequence of three consecutive requests over one growing conversation, seconds apart, same model and key:

| # | cache read ÷ prompt tokens | cache write |
|---|---:|---:|
| 1 | 99.9% | small |
| 2 | **8.0%** | **0** |
| 3 | 99.4% | small |

Request 2 reads 8% of a prefix that was warm one second earlier and writes nothing — not a cold start, not TTL expiry, not a routing change. Its only distinguishing feature is a trailing `role:system` item carrying the breakpoint.

## On the previous behavior

`TestSystemReminderDoesNotCarryCachePoint` pinned the omission deliberately, reasoning that a breakpoint at the moving conversation tail shifts every turn and defeats prefix caching. I think the premise is right and the conclusion inverted:

1. A breakpoint that advances each turn is how incremental conversation caching is meant to work — it extends the cached prefix by one turn for the cost of one write.
2. Dropping it doesn't fall back to a safe state; it removes the *only* conversation-level breakpoint, which is the collapse `inlineSystemReminders` exists to prevent.
3. Measured: preserving it is 100% vs 49.9%.

The rationale would hold if the reminder were appended *after* the last breakpoint. Here the reminder **carries** the breakpoint. That test is replaced by `TestSystemReminderCarriesCachePoint`; happy to revisit if there's a case I'm not seeing.

## Changes

- `convertBifrostSystemReminderToBedrockUserMessage` carries the block's `CacheControl` through and appends a `cachePoint` **after** the text block it terminates (Converse semantics).
- Only the **last** breakpoint within a single reminder is emitted, so a pathological input can't exceed Bedrock's per-request checkpoint budget.
- Reminders with no `cache_control` are unchanged — no invented `cachePoint`.
- The `1h` TTL is preserved rather than silently downgraded to 5m.

## Tests

Six cases in `cache_points_test.go` covering the contract: breakpoint preserved, ordering after text, user-role control unchanged, multiple breakpoints collapse to one, absent `cache_control` stays absent, 1h TTL preserved. Verified to **fail against unpatched `dev`** (3 of 6) and pass with the change. Full `core/providers/...` and `core/schemas/...` suites pass; `gofmt` and `go vet` clean.

### DIFF
--- a/core/providers/bedrock/bedrock_test.go
+++ b/core/providers/bedrock/bedrock_test.go
@@ -6690,11 +6690,22 @@ func TestSystemReminderBetweenToolCallAndResult(t *testing.T) {
 	assert.True(t, hasResult, "user message after tool_use must contain the matching tool_result")
 }
 
-// TestSystemReminderDoesNotCarryCachePoint pins the deliberate omission flagged in review: an
-// inlined mid-conversation reminder must NOT emit a CachePoint, even if its block carries
-// CacheControl. A breakpoint at the moving conversation tail would shift every turn and defeat
-// the prefix caching this whole change exists to preserve.
-func TestSystemReminderDoesNotCarryCachePoint(t *testing.T) {
+// TestSystemReminderCarriesCachePoint reverses the omission previously pinned here. The old
+// behavior — dropping CacheControl when inlining a mid-conversation reminder — was chosen on the
+// grounds that a breakpoint at the moving conversation tail shifts every turn and defeats prefix
+// caching. Measured against Bedrock, the opposite holds: that breakpoint is the client's
+// conversation-level anchor, and dropping it pins the cacheable prefix at the system/tools floor so
+// the entire conversation body is re-read uncached every turn.
+//
+// claude-opus-5, identical warm prefix, third breakpoint on the tail:
+//
+//	tail on role:user    uncached=2      read=36,115  -> 100.0% hit
+//	tail on role:system  uncached=18,116 read=18,018  ->  49.9% hit
+//
+// A breakpoint that advances each turn is how incremental conversation caching is meant to work: it
+// extends the cached prefix by one turn for the cost of one write. See the
+// TestInlinedSystemReminder_* cases in cache_points_test.go for the full contract.
+func TestSystemReminderCarriesCachePoint(t *testing.T) {
 	reminder := systemReminderTextMsg("Reminder that happens to carry a breakpoint.")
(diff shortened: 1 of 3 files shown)


akshaydeo:
its covered in broader change #5931

PR #5576, opened 2026-07-27
fix: IncludeServerSideToolInvocations field in genai api

TejasGhatte:
## Summary

Adds `include_server_side_tool_invocations` as a first-class parameter on both the Chat and Responses schemas. Gemini 3+ rejects requests that mix function declarations with built-in tools like Google Search unless this flag is explicitly set; without it, function declarations are silently dropped to preserve existing behaviour. This PR also fixes a bug where grounding annotations were only emitting one citation per segment (the first chunk index), losing all additional sources.

## Changes

- Added `IncludeServerSideToolInvocations *bool` to `ChatParameters` and `ResponsesParameters` schemas, registered in the HTTP handler's known-fields maps for both chat and responses paths.
- Added `applyServerSideToolInvocations` helper in Gemini utils that sets `toolConfig.includeServerSideToolInvocations = true` and promotes `AUTO` (or unset) function-calling mode to `VALIDATED`, as Gemini rejects `AUTO` in combination mode.
- Updated `convertResponsesToolsToGemini` to accept the flag and, when set, emit function declarations and Google Search as separate `Tool` entries (matching Google's documented tool-combination shape) rather than collapsing them into one or dropping declarations.
- Wired the flag through both the chat (`ToGeminiChatCompletionRequest`) and responses (`ToGeminiResponsesRequest`) conversion paths.
- Extended `ToBifrostResponsesRequest` (native genai round-trip) to read `IncludeServerSideToolInvocations` from `ToolConfig`, with a custom `UnmarshalJSON` on `ToolConfig` that accepts both camelCase and snake_case spellings.
- Fixed grounding annotation emission in both the non-streaming (`convertGeminiCandidatesToResponsesOutput`) and streaming (`emitAnnotationsFromGroundingSupports`) paths to iterate over all `GroundingChunkIndices` per support rather than only the first, skipping chunks with no `Web` field or an empty URI, and using per-annotation index values instead of a shared pointer.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [x] Core (Go)
- [x] Transports (HTTP)
- [x] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

```sh
go test ./core/providers/gemini/... -run TestIncludeServerSideToolInvocations
go test ./core/providers/gemini/... -run TestIncludeServerSideToolInvocationsGenAIRoundTrip
[...]

PR #5935, opened 2026-08-07
[fix]: Anthropic - include_server_side_tool_invocations now reaches the Gemini declaration-drop gate

### LINKED ISSUE #5679: [Bug]: Anthropic Messages does not propagate Gemini mixed server/client tool opt-in
### Prerequisites

- [x] I searched existing open and closed issues and pull requests for duplicates.
- [x] I tested against a recent `dev` snapshot.

### Description

When an Anthropic Messages request routed to Gemini 3 contains both an Anthropic server-side web-search tool and a custom function tool, Bifrost drops the function declaration even when the caller passes `include_server_side_tool_invocations: true`.

Bifrost already models this opt-in as `ResponsesParameters.IncludeServerSideToolInvocations`. PR #5576 added the Gemini Chat/Responses/GenAI handling: when enabled, Bifrost sends `functionDeclarations` and `googleSearch` as separate tool entries and sets `toolConfig.includeServerSideToolInvocations`.

The Anthropic conversion path does not populate that typed parameter. `AnthropicMessageRequest.ToBifrostResponsesRequest` constructs `ResponsesParameters` and converts tools/tool choice, but never assigns `IncludeServerSideToolInvocations`. An unknown top-level Anthropic field is captured as an extra parameter, but the Gemini declaration-drop gate reads the typed field instead.

This blocks Claude Agent SDK clients using `/anthropic/v1/messages` from combining Gemini native Google Search with client/MCP tools in one turn.

### Steps to reproduce

1. Run Bifrost with a Vertex provider and a Gemini 3 model.
2. Send the following request to `/anthropic/v1/messages`:

```json
{
  "model": "vertex/gemini-3-flash-preview",
  "max_tokens": 4096,
  "include_server_side_tool_invocations": true,
  "messages": [
    {
      "role": "user",
      "content": "Use Google Search, then call echo exactly once with the result."
    }
  ],
  "tools": [
    {
      "type": "web_search_20250305",
      "name": "web_search",
      "max_uses": 1
    },
    {
      "name": "echo",
      "description": "Echo the supplied value exactly.",
      "input_schema": {
        "type": "object",
        "properties": {
          "value": { "type": "string" }
        },
        "required": ["value"]
      }
    }
  ]
}
```

3. Inspect the raw Vertex request.

Observed raw request characteristics:

- `googleSearch` is present.
- `functionDeclarations` is absent.
- `toolConfig.includeServerSideToolInvocations` is absent.

The same result occurs when the option is supplied under `extra_params` with `x-bf-passthrough-extra-params: true`. A custom-tool-only control correctly sends a plain `echo` function declaration.

### Expected behavior

For the Anthropic integration, an explicit `include_server_side_tool_invocations: true` opt-in should populate `ResponsesParameters.IncludeServerSideToolInvocations`. Gemini 3 should then receive both `functionDeclarations` and `googleSearch`, plus `toolConfig.includeServerSideToolInvocations: true`. Absent or false should preserve the current compatibility behavior.

### Actual behavior

The Anthropic converter leaves the typed opt-in unset, so Gemini conversion intentionally removes all function declarations when web search is present.

### Affected area(s)

- Core (Go)
- Transports (HTTP)

### Version

- Source audit: `dev` commit `c9f142ba85597e9166cc4ce8d005fc77790dddd7`
- Live probe image: `maximhq/bifrost@sha256:0c0e6f498944d2896867fd9a5e51a79bb9cc82b845763fc4ababe0a6d2ab4402`
- The runtime image did not expose an OCI source-revision label, so these are separate version anchors.

### Relevant source

- `core/providers/gemini/responses.go`: `dropFunctionDeclarations := hasWebSearchTool && !includeServerSideToolInvocations`
- `core/providers/anthropic/responses.go`: `AnthropicMessageRequest.ToBifrostResponsesRequest` does not assign the typed option.
- `core/schemas/responses.go`: defines `ResponsesParameters.IncludeServerSideToolInvocations`.

### Suggested fix

Promote the explicit Bifrost extension field from the Anthropic request into `ResponsesParameters.IncludeServerSideToolInvocations`, with tests covering:

1. absent/false retains current declaration-drop behavior;
2. true preserves both web search and function tools for Gemini 3;
3. raw Vertex payload contains both tool entries and the tool-config opt-in;
4. Anthropic streaming/non-streaming responses preserve server-tool blocks, citations, custom tool names, and subsequent tool-result replay.

### Regression?

Not known. This appears to be an integration gap left after #5576 introduced the typed Gemini opt-in. Related earlier search work: #5139.

### Severity

Medium — web-search-only and function-only requests work, but mixed native-search and client/MCP-tool turns are unavailable through the Anthropic endpoint.

AdityaPainuli:
## Summary

On `anthropic/v1/messages`, `include_server_side_tool_invocations: true` was silently ignored: the field is unregistered on the Anthropic schema, so it lands in ExtraParams as raw JSON, while the Gemini declaration-drop gate (added in #5576) reads the typed `Params.IncludeServerSideToolInvocations`. Combining a server-side tool with function tools on a Gemini-routed request therefore dropped the function declarations even with the opt-in set. 

## Changes

- `ToBifrostResponsesRequest` lifts `include_server_side_tool_invocations` from ExtraParams onto the typed parameter, decoding the `json.RawMessage` capture (a plain type assertion never matches, which is why the flag was invisible)
- Regression test unmarshals a raw Anthropic body carrying the flag and asserts the typed param is set after conversion; fails on current dev

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [x] Core (Go)
- [ ] Transports (HTTP)
- [ ] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

`cd core && go test ./providers/anthropic/ -run TestAnthropicIngressLiftsServerSideToolOptIn -v`

## Breaking changes

- [x] No

## Related issues

Closes #5679 

## Security considerations

None.

## Checklist


core/providers/anthropic/responses.go
```diff
@@ -3486,6 +3486,22 @@ func (req *AnthropicMessageRequest) ToBifrostResponsesRequest(ctx *schemas.Bifro
...
+	// so leaving it in ExtraParams silently drops function tools when a request
+	// combines them with a server-side tool (issue #5679). Unregistered fields
+	// are captured as json.RawMessage, so decode rather than type-assert.
+	if raw, exists := req.ExtraParams["include_server_side_tool_invocations"]; exists {
+		switch v := raw.(type) {
+		case bool:
+			params.IncludeServerSideToolInvocations = schemas.Ptr(v)
+		case json.RawMessage:
+			var optIn bool
+			if sonic.Unmarshal(v, &optIn) == nil {
```
akshaydeo: why do we need to marshal if its a bool? we can do gjson and compare string right?

AdityaPainuli: Yup, no decode needed. Since the captured value is already the bare token (not a document to pluck from), went with a direct string comparison. Same idea as gjson with one less call.

akshaydeo (approved):

PR #6083, opened 2026-08-11
feat: backfill upscale output resolution from `target`/`factor` params and `metrics.resolution_target` for accurate resolution-tiered pricing

Pratham-Mishra04:
## Summary

Upscale-style image models on Replicate (e.g. `prunaai/p-image-upscale`) accept a `target` (desired output megapixels) or `factor` (input dimension multiplier) parameter instead of a plain size string. Because neither the request's `Size` param nor the provider's response carries resolution info in these cases, resolution-tiered image pricing silently fell back to the base per-image rate regardless of actual output size. This PR backfills the output resolution on the `BifrostImageGenerationResponse` so cost calculation uses the correct tier.

## Changes

- Added `applyUpscaleOutputResolution`, which backfills `ImageGenerationResponseParameters.Size` on upscale-style responses when it isn't already set. It is a no-op when neither a `target` param nor a metrics band is present, and it never overwrites an existing `Size` value.
- Added `resolveUpscaleOutputPixels`, which determines the output pixel count in priority order: first from the request's `target` param (when `upscale_mode` is `"target"` or omitted, since that is the default), then from the `metrics.resolution_target` band reported by Replicate post-hoc (used for `factor` mode, where output size depends on the input image and isn't knowable from the request alone).
- Added `resolveUpscaleOutputPixelsFromMetrics`, which parses megapixel band strings like `"8-16MP"` or `"16MP"` from `metrics.resolution_target`, using the upper bound of the band as the billable pixel estimate to avoid under-billing at tier boundaries.
- Added `formatSquarePixelSize`, which converts a total pixel count to a `"WxH"` string using a square approximation with ceiling rounding, ensuring a value sitting exactly on a pricing tier threshold is never miscategorized into the tier below it.
- Added `ResolutionTarget` to `ReplicateMetrics` to capture the `resolution_target` field from Replicate's prediction metrics response.
- Wired `applyUpscaleOutputResolution` into the `ImageGeneration` response path, called after the prediction is converted to a Bifrost response and before extra fields are set.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [x] Core (Go)
- [ ] Transports (HTTP)
- [x] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

```sh
go test ./core/providers/replicate/...
```

The new test file `images_upscale_resolution_test.go` covers:
- `target` mode with explicit and omitted `upscale_mode`
- `factor` mode ignoring the `target` param and falling back to the metrics band
- Metrics band parsing for range bands (`"8-16MP"`), single-value bands (`"16MP"`), and lowercase variants
- `formatSquarePixelSize` ceiling rounding at non-perfect-square boundaries (e.g. 8,000,000 pixels) and exact round-trip for perfect squares (e.g. 16,000,000 → `"4000x4000"`)
- `applyUpscaleOutputResolution` not overwriting an existing `Size` and leaving `Size` empty when no signal is present

## Breaking changes

- [ ] Yes
- [x] No

## Security considerations

None. This change only affects how output resolution is inferred for cost calculation purposes and does not touch auth, secrets, PII, or sandboxing.

## Checklist


Pratham-Mishra04:
* **#6083** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/6083?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/maximhq/bifrost/6083?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#6082** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/6082?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `dev`


<h2></h2>

core/providers/replicate/replicate.go
```diff
@@ -1833,6 +1833,11 @@ func (provider *ReplicateProvider) ImageGeneration(ctx *schemas.BifrostContext,
 		return nil, providerUtils.EnrichError(ctx, err, jsonData, nil, provider.sendBackRawRequest, provider.sendBackRawResponse, latency)
 	}
 
+	// Backfill output resolution for upscale-style models (target/factor
+	// input, no plain size param) so resolution-tiered cost calculation
+	// doesn't silently fall back to the base per-image rate.
+	applyUpscaleOutputResolution(request, prediction, bifrostResponse)
+
```
TejasGhatte, line 1840: this should be handled in image gen stream and edit as well right

Pratham-Mishra04: Good catch, you were right, and the gap was a bit wider than just those two. All four image paths now backfill.

What was missing: `cost.go` reads the output size from `ImageGenerationResponseParameters.Size` for the non-stream paths and from `BifrostImageGenerationStreamResponse.Size` for the streams, and only `ImageGeneration` was populating it. So `ImageEdit`, `ImageGenerationStream` and `ImageEditStream` all fell back to the base per-image rate. `BackfillParams` does not cover it either, since `getNumInputImagesSizeQualityAndAspectRatioFromRequest` only ever reads `Params.Size`.

Changes:

- `ImageEdit` now calls a new `applyUpscaleEditOutputResolution`. The edit path needed its own resolver rather than reusing the generation one, because an upscale arrives there through the first-class `TargetMegapixels` / `UpscaleFactor` params rather than through Replicate's raw `target` / `upscale_mode` input names. Those two are mutually exclusive, so `UpscaleFactor` being set means the output follows the input image and we fall back to the reported metrics band, mirroring how `upscale_mode: "factor"` behaves on the generation path. Raw `target` in `ExtraParams` still resolves on the edit path too.
- Both streams set `Size` on their completion chunk via `applyUpscaleStreamOutputResolution`.

One limitation worth flagging: the streams can only use the request-side target signal. The SSE loop never re-reads the finished prediction, so `metrics.resolution_target` is not available there and a factor-mode stream still ends up without a size. That is the same as before this change rather than a regression, and it is called out in a comment at both stream call sites. Making it work would mean fetching the terminal prediction after the stream completes, which seemed like more than this PR should take on. Happy to do it here if you would rather it be complete.

Tests cover the edit resolver (target_megapixels, upscale_factor falling back to metrics, extra-params target, no signal), both appliers, and the existing-size-preserved cases.

core/providers/replicate/images.go
```diff
@@ -181,6 +183,166 @@ func ToBifrostImageGenerationResponse(
...
+// string, leaving a size the provider itself reported untouched.
+func setUpscaleOutputSize(pixels int, response *schemas.BifrostImageGenerationResponse) {
+	if pixels <= 0 {
+		return
+	}
+	if response.ImageGenerationResponseParameters == nil {
+		response.ImageGenerationResponseParameters = &schemas.ImageGenerationResponseParameters{}
+	}
+	if response.ImageGenerationResponseParameters.Size == "" {
+		response.ImageGenerationResponseParameters.Size = formatSquarePixelSize(pixels)
```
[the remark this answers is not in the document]
Pratham-Mishra04: Declining this one.

The selector is legal to shorten, `ImageGenerationResponseParameters` is an embedded pointer on `BifrostImageGenerationResponse`, so `response.Size` compiles. But two things argue against changing it:

- Nothing enforces QF1008 here. There is no `.golangci.yml` or `.golangci.yaml` anywhere in the repo, and no workflow in `.github/workflows` runs golangci-lint or staticcheck. The reported violation comes from CodeRabbit running its own linter with its own defaults, not from a gate this repo has opted into.
- The explicit form is the convention across the provider tree. `core/providers/bedrock/images.go:169-171` and `core/providers/gemini/images.go:721-722` both spell out `...ImageGenerationResponseParameters.<field>`. Shortening only this one file would make it the odd one out.

There is also a readability argument for keeping it: the surrounding code nil-guards and initializes `response.ImageGenerationResponseParameters` on the line immediately above, so naming the same field on the write makes it obvious which struct is being populated.

If QF1008 is wanted, the right move is adding a golangci config and applying it repo-wide rather than spot-fixing the file that happened to be reviewed.

akshaydeo (dismissed):

akshaydeo:
### Merge activity

* **Aug 19, 1:42 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/6083).
* **Aug 19, 1:44 PM UTC**: @akshaydeo merged this pull request with [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/6083).

PR #6085, opened 2026-08-12
fix: add per-alias override to decouple canonical-model resolution from adaptive-thinking preference

### LINKED ISSUE #6084: Alias model_name for cachePoint/pricing gating silently also opts adaptive-capable (non-adaptive-only) models into adaptive thinking, with no override
## Summary

Attaching `model_name` to a Bedrock (or any provider) alias — purely to unlock unrelated canonical-model-keyed features like Bedrock cachePoint eligibility, pricing/logging attribution, etc. — silently and unavoidably also opts that alias into Anthropic's adaptive-thinking wire shape whenever the resolved canonical model happens to be adaptive-*capable* (not necessarily adaptive-*only*), with no way to opt back out short of reverting `model_name` entirely and losing the unrelated feature with it.

## Background

`ResolveCanonicalModel()` (`core/schemas/account.go`) walks `AliasConfig.ModelName → AliasConfig.ModelID → fallbackModel` and returns a single resolved string ("capModel") that is then fed into many *unrelated* capability gates across the Anthropic and Bedrock providers:

- `BedrockModelSupportsCachePoints(capModel)` — cachePoint eligibility
- `anthropic.SupportsAdaptiveThinking(capModel)` — reasoning wire-shape selection (`thinking:{type:"enabled",budget_tokens:N}` vs `thinking:{type:"adaptive"}`+`output_config.effort`)
- `anthropic.IsAdaptiveOnlyThinkingModel(capModel)`, computer-use tool generation, native-effort support, etc.

This is intentional per #4520 ("fix: resolve model names for feature gating") — before that PR, an alias with an opaque `ModelID` (e.g. a Bedrock Application Inference Profile ID, which does not textually contain a model family/version string) would silently fail every one of these capability checks. #4520 fixed that by letting an admin attach `model_name` (the real canonical model string) to the alias so capability gating works correctly even behind an opaque wire ID.

## The gap

The problem is that `SupportsAdaptiveThinking(capModel)` is used as an unconditional preference switch: *if* the model supports adaptive thinking at all, adaptive is always used over the classic `budget_tokens` shape — for **every** adaptive-*capable* model, not just the adaptive-*only* ones (Opus 4.7+, Sonnet 5+, Fable/Mythos — see `IsAdaptiveOnlyThinkingModel`, which is a strict subset). Concretely, Claude Sonnet 4.6 is adaptive-*capable* but not adaptive-*only*: it still accepts the legacy `budget_tokens` shape (see the "Adaptive-capable but NOT adaptive-only (budget_tokens still accepted)" test cases in `core/providers/anthropic/utils_test.go`'s `TestIsAdaptiveOnlyThinkingModel`).

So an operator who adds `model_name` to a Sonnet-4.6-class alias *purely* to unlock cachePoint gets adaptive thinking flipped on as an unrequested side effect, because both signals are read off the same `capModel` string with no independent control.

This matters operationally: adaptive thinking removes the `budget_tokens` cap that `GetBudgetTokensFromReasoningEffort()` previously derived from `max_output_tokens` and the requested `effort` (e.g. `high` → `budget_tokens ≈ 0.80 × max_output_tokens`, always leaving ~20% of `max_output_tokens` reserved for the final answer). Under adaptive thinking, Anthropic's model decides its own thinking-token spend for a given `effort`, with no such reservation — for existing callers tuned against the old `budget_tokens` behavior (e.g. a fixed `max_output_tokens` sized around the ~20% reserve), this can materially increase generation length/latency and, if the caller's `max_output_tokens` doesn't leave enough room, increase the chance of the final answer being truncated by the overall token cap.

We hit this in production: adding `model_name` to a Sonnet-4.6 alias to fix a cachePoint issue caused median output tokens for an unrelated long-form-writing workload to roughly triple and completion latency to roughly double, purely from the reasoning-mode side effect — cachePoint itself worked exactly as intended.

## Expected behavior

An operator should be able to attach `model_name` to an alias to unlock canonical-model-keyed features (cachePoint, pricing, logging, etc.) *without* also being forced to opt into the adaptive-thinking wire-shape preference for models that are adaptive-capable-but-not-adaptive-only. There's currently no alias-level (or otherwise) escape hatch for this.

## Suggested fix

Add an opt-in, per-alias override (e.g. `AliasConfig.ForceFixedBudgetThinking *bool`) that pins reasoning requests to the classic `budget_tokens` shape even when `SupportsAdaptiveThinking(capModel)` is true, while remaining a no-op for adaptive-*only* models (`IsAdaptiveOnlyThinkingModel`), since those reject `budget_tokens` outright regardless of any override. Default (field unset) preserves all existing behavior exactly.

I have a minimal, additive patch implementing this (new alias field + a small helper that gates the four `SupportsAdaptiveThinking(capModel)` preference-branch call sites in `core/providers/bedrock/{utils,responses}.go` and `core/providers/anthropic/{chat,responses}.go`) — happy to open a PR if this framing looks right to maintainers.

jitokim:
## Summary

Fixes #6084. Adds an opt-in, per-alias override that decouples "give this alias a canonical `model_name` for capability gating (cachePoint, pricing, logging, ...)" from "opt this alias into Anthropic's adaptive-thinking wire shape."

## Problem

`ResolveCanonicalModel()` resolves a single `capModel` string (`AliasConfig.ModelName → ModelID → fallback`) that many *unrelated* capability gates read off of — including both `BedrockModelSupportsCachePoints(capModel)` and `anthropic.SupportsAdaptiveThinking(capModel)`. Attaching `model_name` to an alias purely to fix cachePoint eligibility for an opaque wire ID (e.g. a Bedrock Application Inference Profile ID with no model string in it) has the side effect of also flipping reasoning requests from the classic `thinking:{type:"enabled",budget_tokens:N}` shape to `thinking:{type:"adaptive"}`+`output_config.effort` for any model where `SupportsAdaptiveThinking(capModel)` is true — even for models that are adaptive-*capable* but not adaptive-*only* (e.g. Claude Sonnet 4.6, which still accepts `budget_tokens` — see `IsAdaptiveOnlyThinkingModel`'s test cases), with no way to opt out.

This changes real generation-length/latency characteristics for existing callers tuned against the `budget_tokens` behavior (adaptive thinking has no equivalent to the ~20%-of-`max_output_tokens` reserve that `GetBudgetTokensFromReasoningEffort()` guarantees for the final answer under a fixed budget). See #6084 for the full writeup and a production incident this caused.

## Fix

- `core/schemas/account.go`: new `AliasConfig.ForceFixedBudgetThinking *bool` (opt-in, `omitempty`, default unset = 100% unchanged behavior), plus a `ShouldUseAdaptiveThinking(ctx, supportsAdaptive, adaptiveOnly bool) bool` helper that layers the override on top of the existing capability checks. The override is a no-op when `adaptiveOnly` is true (Opus 4.7+, Sonnet 5+, Fable/Mythos reject `budget_tokens` outright, so there is no "fixed budget" mode to fall back to for those).
- Swapped the 4 call sites that use `SupportsAdaptiveThinking(capModel)` as the adaptive-vs-fixed-budget preference branch to go through the new helper instead:
  - `core/providers/bedrock/utils.go`
  - `core/providers/bedrock/responses.go`
  - `core/providers/anthropic/chat.go`
  - `core/providers/anthropic/responses.go`
  
  (Every other `SupportsAdaptiveThinking`/`IsAdaptiveOnlyThinkingModel` call — e.g. the `display` field default, Fable-family special-casing — is untouched.)
- `isLegacyShape()` updated to account for the new field so an alias with only `ForceFixedBudgetThinking` set doesn't get mis-marshaled as the legacy string-valued shape.

## Example

```json
{
  "aliases": {
    "my-sonnet-4-6": {
      "model_id": "<opaque-wire-id>",
      "model_name": "claude-sonnet-4-6",
      "force_fixed_budget_thinking": true
    }
  }
}
```

`cachePoint`/pricing/logging keep resolving off `model_name` as before; reasoning requests for this alias keep using `thinking:{type:"enabled",budget_tokens:N}` instead of switching to adaptive.

## Testing

- `go build`, `go vet`, and `gofmt -l` are clean on all touched packages.
- New unit tests cover: default (unset override) behavior is unchanged for adaptive-capable models (regression); `ForceFixedBudgetThinking: true` forces `budget_tokens` for an adaptive-capable-but-not-adaptive-only model; the override is ignored (adaptive still used) for an adaptive-only model; and the helper directly, in isolation.
- **Local Docker verification** (built the real release image via `transports/Dockerfile.local`, booted it with a `config.json` containing a Bedrock alias with `force_fixed_budget_thinking: true` alongside a control alias without it, both pointed at the same non-production Bedrock inference-profile id):
  - `GET /api/providers/bedrock/keys` confirms the field round-trips correctly through the JSON config → config-store → admin API path.
  - Sent live, SigV4-signed requests to AWS Bedrock (`reasoning.effort: "high"`) through both aliases and inspected the raw outgoing request body (via `send_back_raw_request: true`) that bifrost actually put on the wire:
    - override alias → `"additionalModelRequestFields":{"thinking":{"budget_tokens":1804,"type":"enabled"}}` (`budget_tokens` matches `GetBudgetTokensFromReasoningEffort`'s formula for `effort=high`, `max_output_tokens=2000`: `1024 + 0.80×(2000−1024) = 1804`)
    - control alias → `"additionalModelRequestFields":{"thinking":{"type":"adaptive"},"output_config":{"effort":"high"}}`
  - Both requests reached Bedrock and were authenticated successfully (no credential/signing errors); Bedrock itself rejected the specific inference-profile id with `ValidationException: The provided model identifier is invalid`, which is an AWS-side resource/permission-scoping issue for the ad-hoc test principal used, unrelated to this change — the two `additionalModelRequestFields.thinking` payloads above are the literal bytes bifrost constructed and signed for delivery, which is exactly what this PR changes.

## Breaking changes

None. The new field is optional and unset by default; all existing aliases and callers keep their current behavior byte-for-byte.

### DIFF
--- a/core/providers/anthropic/chat.go
+++ b/core/providers/anthropic/chat.go
@@ -697,7 +697,9 @@ func ToAnthropicChatRequest(ctx *schemas.BifrostContext, bifrostReq *schemas.Bif
 				}
 			} else if reasoningParams.Effort != nil && *reasoningParams.Effort != "none" {
 				effort := MapBifrostEffortToAnthropic(*reasoningParams.Effort)
-				if caps.SupportsAdaptiveThinking(DefaultSupportsAdaptiveThinking(caps.Model())) {
+				if schemas.ShouldUseAdaptiveThinking(ctx,
+					caps.SupportsAdaptiveThinking(DefaultSupportsAdaptiveThinking(caps.Model())),
+					caps.AdaptiveOnlyThinking(DefaultAdaptiveOnlyThinking(caps.Model()))) {
 					// Opus 4.6+ and Opus 4.7+: adaptive thinking + native effort
 					anthropicReq.Thinking = &AnthropicThinking{Type: "adaptive"}
 					setEffortOnOutputConfig(anthropicReq, effort)
--- a/core/providers/anthropic/chat_test.go
+++ b/core/providers/anthropic/chat_test.go
@@ -1173,7 +1173,7 @@ func TestToAnthropicChatRequest_NonOpus47_ReasoningMaxTokens_EnabledWithBudget(t
 // The wire model id is opaque (mirroring the incident); capModel is resolved
 // from the alias ModelName.
 func TestToAnthropicChatRequest_ForceFixedBudgetThinkingOverride(t *testing.T) {
-	const wireModelID = "v9xyf4o111s5" // opaque resource id, like the incident alias's model_id
+	const wireModelID = "a1b2c3d4e5f6" // opaque resource id, like the incident alias's model_id
 
 	tests := []struct {
 		name         string
--- a/core/providers/anthropic/chat_test.go
+++ b/core/providers/anthropic/chat_test.go
@@ -1168,6 +1168,96 @@ func TestToAnthropicChatRequest_NonOpus47_ReasoningMaxTokens_EnabledWithBudget(t
 	}
 }
 
+// TestToAnthropicChatRequest_ForceFixedBudgetThinkingOverride verifies the
+// per-alias ForceFixedBudgetThinking override on the direct Anthropic chat path.
+// The wire model id is opaque (mirroring the incident); capModel is resolved
+// from the alias ModelName.
+func TestToAnthropicChatRequest_ForceFixedBudgetThinkingOverride(t *testing.T) {
+	const wireModelID = "v9xyf4o111s5" // opaque resource id, like the incident alias's model_id
+
+	tests := []struct {
+		name         string
+		modelName    string
+		force        *bool
+		expectedType string
+		expectBudget bool
+		expectEffort bool
+	}{
+		{
+			name:         "Sonnet4.6_NoOverride_UsesAdaptive_Regression",
+			modelName:    "claude-sonnet-4-6",
+			force:        nil,
+			expectedType: "adaptive",
+			expectBudget: false,
+			expectEffort: true,
+		},
+		{
+			name:         "Sonnet4.6_OverrideTrue_UsesBudgetTokens_NewBehavior",
+			modelName:    "claude-sonnet-4-6",
+			force:        schemas.Ptr(true),
+			expectedType: "enabled",
+			expectBudget: true,
+			expectEffort: false,
+		},
+		{
+			name:         "Sonnet5_AdaptiveOnly_OverrideTrue_IgnoredStaysAdaptive",
+			modelName:    "claude-sonnet-5",
+			force:        schemas.Ptr(true),
+			expectedType: "adaptive",
+			expectBudget: false,
+			expectEffort: true,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			bifrostReq := &schemas.BifrostChatRequest{
+				Provider: schemas.Anthropic,
+				Model:    wireModelID,
+				Input: []schemas.ChatMessage{
+					{Role: schemas.ChatMessageRoleUser, Content: &schemas.ChatMessageContent{ContentStr: schemas.Ptr("think")}},
+				},
+				Params: &schemas.ChatParameters{
+					MaxCompletionTokens: schemas.Ptr(8192),
+					Reasoning:           &schemas.ChatReasoning{Effort: schemas.Ptr("high")},
+				},
+			}
+
+			ctx, cancel := schemas.NewBifrostContextWithCancel(context.Background())
+			defer cancel()
+			ctx.SetValue(schemas.BifrostContextKeyResolvedAlias, &schemas.ResolvedAlias{
+				Key: "best-claude",
+				Config: &schemas.AliasConfig{
+					ModelID:                  wireModelID,
+					ModelName:                schemas.Ptr(tt.modelName),
+					ForceFixedBudgetThinking: tt.force,
+				},
+			})
+
+			result, err := ToAnthropicChatRequest(ctx, bifrostReq)
+			if err != nil {
+				t.Fatalf("unexpected error: %v", err)
+			}
+			if result.Thinking == nil {
+				t.Fatal("expected Thinking to be set")
+			}
+			if result.Thinking.Type != tt.expectedType {
+				t.Errorf("thinking.type: got %q, want %q", result.Thinking.Type, tt.expectedType)
+			}
+			if tt.expectBudget && result.Thinking.BudgetTokens == nil {
+				t.Error("expected BudgetTokens to be set for budget_tokens wire shape")
+			}
+			if !tt.expectBudget && result.Thinking.BudgetTokens != nil {
+				t.Errorf("expected no BudgetTokens for adaptive wire shape, got %v", *result.Thinking.BudgetTokens)
+			}
+			hasEffort := result.OutputConfig != nil && result.OutputConfig.Effort != nil
+			if hasEffort != tt.expectEffort {
+				t.Errorf("output_config.effort presence: got %v, want %v", hasEffort, tt.expectEffort)
+			}
+		})
+	}
+}
+
 func TestToAnthropicChatRequest_Opus47_ReasoningEffort_AdaptiveWithEffort(t *testing.T) {
 	effort := "high"
 
--- a/core/providers/anthropic/responses.go
+++ b/core/providers/anthropic/responses.go
@@ -4219,7 +4219,9 @@ func ToAnthropicResponsesRequest(ctx *schemas.BifrostContext, bifrostReq *schema
 					if *bifrostReq.Params.Reasoning.Effort != "none" {
 						effort := MapBifrostEffortToAnthropic(*bifrostReq.Params.Reasoning.Effort)
 
-						if caps.SupportsAdaptiveThinking(DefaultSupportsAdaptiveThinking(caps.Model())) {
+						if schemas.ShouldUseAdaptiveThinking(ctx,
+							caps.SupportsAdaptiveThinking(DefaultSupportsAdaptiveThinking(caps.Model())),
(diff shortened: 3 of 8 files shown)


jitokim:
Hi @akshaydeo @Pratham-Mishra04 — following up here since this fixes a production-impacting issue on our end.

**Context**: We run Bifrost's Bedrock provider for Claude with `cachePoint` enabled and adaptive thinking intentionally disabled (fixed `budget_tokens` reasoning). Attaching `model_name` to our alias to unlock `cachePoint` eligibility (per #4520) had the side effect of silently switching reasoning to the adaptive wire shape for any adaptive-*capable* model — not just adaptive-*only* ones — since both capability gates read off the same resolved `capModel`. In production this roughly tripled output tokens and doubled latency for one workload, so we had to fully roll back `cachePoint` to recover, even though `cachePoint` itself worked exactly as intended. Full writeup in #6084.

This PR adds an opt-in, per-alias `ForceFixedBudgetThinking` override that decouples the two concerns — additive, defaults to existing behavior when unset, and a no-op for adaptive-only models that reject `budget_tokens` regardless.

I know the review queue is long — I'd really appreciate a look whenever you get a chance, since this is what's blocking us from safely re-enabling `cachePoint` in production. Happy to address feedback quickly.

Thanks for maintaining Bifrost!

jitokim:
Rebased onto current `dev`. The branch was 232 commits behind and `CONFLICTING`; all four conflicts were the same call site, moved by 42617c4b6 from the free-function capability predicates to the `ModelCaps` methods. Resolved by keeping the override wrapper and re-assembling its arguments against the new API:

```go
if schemas.ShouldUseAdaptiveThinking(ctx,
    caps.SupportsAdaptiveThinking(DefaultSupportsAdaptiveThinking(caps.Model())),
    caps.AdaptiveOnlyThinking(DefaultAdaptiveOnlyThinking(caps.Model()))) {
```

Now `MERGEABLE`, CodeRabbit approved, checks green. `go build`, `go vet` and the `schemas` / `providers/anthropic` / `providers/bedrock` packages all pass, matching a `dev` baseline run in a separate worktree — no test changes were needed.

The coupling this PR targets is still present on `dev`:

```go
// core/providers/anthropic/chat.go
capModel := schemas.ResolveCanonicalModel(ctx, bifrostReq.Model)
caps := schemas.ResolveModelCaps(bifrostReq.Provider, capModel)
```

`caps.SupportsAdaptiveThinking(...)` still keys off nothing but the canonical model, so attaching `model_name` to an alias to unlock cachePoint/pricing eligibility still flips adaptive thinking on for adaptive-capable (non-adaptive-only) models, with no way to opt out.

One data point that this alias shape is in real production use: #6342 reported a wedged multi-turn tool-use session on `bedrock/us.anthropic.claude-opus-4-8` routed through `{model_id, model_name, model_family: anthropic, region}`, and listed #6084/#6085 under "Possibly related" with the note *"we route through exactly such an alias"*.

To be clear I am not claiming that issue was caused by this coupling — it was closed by #6346, which fixed a separate and real defect (content-less `tool_result` blocks being dropped and interleaved thinking blocks being reordered). That diff does not touch the adaptive-thinking decision at all, so the override proposed here is still missing.

@akshaydeo @Pratham-Mishra04 — happy to reshape the override if you would prefer a different surface (different config key, or gating it on `model_family` instead of a per-alias bool). The behaviour is the part I care about; the shape is yours to pick.

jitokim:
Adding production evidence for the tradeoff this PR removes, measured today on a deployed gateway (Bedrock, Application Inference Profile IDs as `model_id`).

Two aliases pointing at the same backing model, differing only in whether `model_name` is set. Same request (a ~1.6k-token system block carrying `cache_control: {"type":"ephemeral"}`), reading the second call of a pair:

| alias | `model_name` | `cached_read_tokens` | thinking shape on the wire |
|---|---|---|---|
| A | unset (opaque profile id) | **0** — no `prompt_tokens_details` at all | `enabled` + `budget_tokens` |
| B | `claude-sonnet-4-6` | **1623** | `adaptive` — forced, no opt-out |

Both gates read the same `ResolveCanonicalModel()` string: `BedrockModelSupportsCachePoints()` needs it to look like an Anthropic model, and `SupportsAdaptiveThinking()` reads that same string. An opaque Application Inference Profile ID satisfies neither, so setting `model_name` is the only way to get prompt caching — and it silently also flips the wire shape.

Sonnet 4.6 is adaptive-*capable* but not adaptive-*only*: `budget_tokens` still works, and the ~20%-of-`max_output_tokens` answer reserve that `GetBudgetTokensFromReasoningEffort()` guarantees is what existing generation-length tuning is calibrated against. Dropping caching to keep that reserve is a real cost regression; keeping caching means losing the reserve.

There is currently no third option. I checked all three escape hatches on `dev`:

- no request-level parameter reaches this branch — the decision is `SupportsAdaptiveThinking(capModel)` alone, with no seam for the caller;
- `datasheet.Override` is scope/pricing-only, so `record.SupportsAdaptiveThinking` cannot be set locally (the datasheet is fetched at runtime from `getbifrost.ai/datasheet/model-parameters`);
- removing `model_name` is exactly what row A measures.

Worth noting the two reachable states are the diagonal — caching and adaptive are welded on together and off together. Neither off-diagonal combination is expressible today.

That is exactly the opt-in this PR adds. It is unset by default (so existing behaviour is byte-identical) and a no-op for adaptive-only models (Opus 4.7+, Sonnet 5+, Fable/Mythos), where there is no fixed-budget mode to fall back to.

Rebased onto `dev` and green since Aug 31, CodeRabbit approved. Happy to reshape the API however you prefer — the behaviour is the part I care about; the shape is yours to pick.

@akshaydeo @Pratham-Mishra04

PR #6125, opened 2026-08-13
fix: return informative message when model list is disabled for provider

gokul-scalent:
**fix: respect list_models disabled setting in /v1/models fan-out**

## Summary

When a provider had `list_models` disabled via `AllowedRequests`, the `GET /v1/models` endpoint would still fan out to that provider and include its models in the response (or return an error). Two separate code paths were affected:

1. **Fan-out (no provider specified)** — `GetConfiguredProviders()` returned all providers regardless of their `AllowedRequests` config, causing disabled providers to be included in the aggregate model list.
2. **Single-provider query** — `GET /v1/models?provider=<name>` did not check `AllowedRequests` before making the request, resulting in an `unsupported_operation` error being surfaced to the caller.

## Changes

- **bifrost.go** — `GetConfiguredProviders()` now filters out providers where `list_models` is explicitly disabled in their `CustomProviderConfig.AllowedRequests`. Providers with no `AllowedRequests` config (i.e. all operations allowed) are unaffected.
- **models.go** — Added an optional `message` field to `BifrostListModelsResponse` to convey informational messages (e.g. disabled notice) without breaking the response shape.
- **inference.go** — Added a pre-flight check in `listModels`: if a specific provider is requested and its `list_models` is disabled, return an empty model list with a human-readable message (`"The model_list request is disabled for this provider."`) instead of forwarding the request.
- **bifrost_test.go** and **`inference_list_models_test.go`** — Added unit tests covering both fan-out filtering and single-provider disabled behavior.
## Type of change

-  Bug fix

## Affected areas

- Core (Go)
- Transports (HTTP)

## How to test

image
image
## Screenshots/Recordings

Previously, requests to providers with model_list disabled resulted in error logs. This change ensures those providers are excluded from model_list fan-out, eliminating the unnecessary errors.

image

## Breaking changes

- [ ] No

## Related issues

## Security considerations

## Checklist


core/bifrost.go
```diff
@@ -3602,13 +3602,39 @@ func reorderAtomicSlice[T pluginWithName](ptr *atomic.Pointer[[]T], pos map[stri
...
-		return nil, fmt.Errorf("no providers configured")
+	providerKeys, err := bifrost.account.GetConfiguredProviders()
+	if err != nil {
+		return nil, err
 	}
-	modelProviders := make([]schemas.ModelProvider, len(*providers))
-	for i, provider := range *providers {
-		modelProviders[i] = provider.GetProviderKey()
+
+	initializedProviders := make(map[schemas.ModelProvider]struct{})
```
akshaydeo, line 3688: not sure why are we doing this here? could you explain a bit?

transports/bifrost-http/handlers/inference.go
```diff
@@ -816,6 +816,30 @@ func (h *CompletionHandler) RegisterRoutes(r *router.Router, middlewares ...sche
 
 // listModels handles GET /v1/models - Process list models requests
 // If provider is not specified, lists all models from all configured providers
+func shouldReturnEmptyListModelsResponse(bifrostErr *schemas.BifrostError) bool {
```
akshaydeo, line 851: why separate function if its only used in one place?

akshaydeo (changes requested):

transports/bifrost-http/handlers/inference.go
gokul-scalent: This is a specific ListModels exception: when a provider returns unsupported_operation, we intentionally return an empty list instead of surfacing an error. 
The helper makes that behavior explicit and testable, even though it’s used in one place.

core/bifrost.go
gokul-scalent: My initial intent was to minimize the list of providers that support the model_list operation during the initial check while fetching GetConfiguredProviders. However, I now see that GetConfiguredProviders is referenced in the providers list implementation as well.

So, I’m thinking of moving this filtering logic inside the ListAllModels() goroutine, where the model_list operation is called for each individual provider.

Please let me know if this would be the correct approach. If so, can I make the change and push one more commit?

PR #6210, opened 2026-08-17
add dashboard session sweeper

### LINKED ISSUE #6209: Dashboard sessions table can grow without bound
## Problem

Each successful dashboard login creates a new opaque UUID token and inserts a row into the `sessions` table with a 30-day expiry. Creating a new session does not invalidate or delete earlier session rows.

The request path correctly rejects a token after `expires_at`, but an expired token that is never presented again remains stored. Logout only removes the token used for that specific logout request. A process restart does not invalidate dashboard sessions or remove their rows.

Consequently, deployments without a database cleanup job retain one row for every historical login, causing unbounded table growth over time. Earlier session tokens also remain usable until their own expiry, so a new login does not act as a single-active-session rotation.

## Impact

- The `sessions` table grows with every login, including repeated logins from the same browser or administrator.
- Expired encrypted session tokens remain in the configuration database.
- Operators cannot list or revoke individual dashboard login sessions through the dashboard today.

Shaik-Sirajuddin:
## Summary

Adds a dashboard session sweeper with the same 30-day orphan-retention expectation used by the OAuth2 session sweeper.

## Changes

- Adds `SessionSweepWorker`.
- Runs cleanup once at startup, then every 24 hours.
- Configures a 30-day orphan-retention period.
- Deletes only `sessions` records whose `expires_at` is older than the retention cutoff.
- Wires the worker into config initialization and graceful teardown.
- Extends `ConfigStore` and its mock with `DeleteOrphanedSessions`.
- Adds SQLite and PostgreSQL coverage.

## Retention behavior

A session remains valid until its normal expiry. After expiry, it is retained for 30 additional days before the sweeper removes it.

Closes #6209.


Shaik-Sirajuddin:
sessions maybe retained 30days in addition once expired under orphanPeriod setting
@Pratham-Mishra04  Does this need a different way of handling for cleanup , open to address this

Shaik-Sirajuddin:
fixed naming conventions

PR #6251, opened 2026-08-18
fix: enforce RFC 8414 issuer validation and MCP server discovery

### LINKED ISSUE #4371: [Bug]: MCP OAuth cannot connect servers requiring RFC 8707 resource indicators or OIDC-path AS metadata discovery (e.g. Replit MCP)
### Prerequisites

### Description

Bifrost's MCP OAuth client cannot complete the authorization flow against MCP
servers that follow the current MCP authorization spec's two hard requirements:
**RFC 8707 Resource Indicators** and **path-aware authorization-server metadata
discovery**. Reproduced end-to-end against Replit's hosted MCP server
(`https://replit-mcp.com/server/mcp`), but both gaps are generic — any server
that mandates resource indicators or whose issuer URL has a path will hit them.

There are two independent failures.

#### 1. Authorization-server metadata discovery never tries the OIDC `<issuer>/.well-known/openid-configuration` form.

For the resource `https://replit-mcp.com/server/mcp`, the protected-resource
document is served and correct:

`GET https://replit-mcp.com/.well-known/oauth-protected-resource/server/mcp`

```json
{
  "resource": "https://replit-mcp.com/server/mcp",
  "authorization_servers": ["https://replit.com/oidc"],
  "scopes_supported": ["apps:read", "apps:write", "offline_access"]
}
```

The issuer (`https://replit.com/oidc`) has a path component. RFC 8414 §3.1
and OIDC Discovery define two well-known URL constructions for such an issuer,
and an MCP client is expected to try both:

- RFC 8414 (path-insertion): `https://replit.com/.well-known/oauth-authorization-server/oidc` → HTTP 403
- OIDC (path-append): `https://replit.com/oidc/.well-known/openid-configuration` → HTTP 200, valid metadata:

```json
{
  "issuer": "https://replit.com/oidc",
  "authorization_endpoint": "https://replit.com/oidc/auth",
  "token_endpoint": "https://replit.com/oidc/token",
  "registration_endpoint": "https://replit.com/oidc/reg",
  "scopes_supported": ["openid","profile","email","offline_access","apps:read","apps:write"],
  "code_challenge_methods_supported": ["S256"],
  "grant_types_supported": ["authorization_code","refresh_token"]
}
```

Bifrost aborts the whole flow at this step:

```
Error: Failed to initiate OAuth flow: OAuth discovery failed: failed to fetch authorization server metadata: failed to fetch metadata from any authorization server. Please provide authorize_url, token_url, and registration_url manually
```

Since the OIDC-path document is reachable and valid, this indicates Bifrost only
attempts the RFC 8414 path-insertion form (and/or the bare-origin form) and never
the OIDC `<issuer>/.well-known/openid-configuration` form that is required when
the issuer URL carries a path.

#### 2. No RFC 8707 resource indicator — token exchange is rejected even after manual endpoint config.

Working around (1) by entering `authorize_url` / `token_url` / `registration_url`
+ scopes `apps:read apps:write offline_access` manually lets the flow start, but
the authorization/token request is rejected by the AS:

```json
{"error":"invalid_target","error_description":"apps scopes require a valid MCP resource"}
```

`invalid_target` is the RFC 8707 error for a missing/invalid `resource`
parameter. The MCP authorization spec requires clients to send `resource`
(the canonical MCP server URI) on both the authorization request and the token
request. Bifrost does not:

- The UI OAuth dialog and the documented config schema expose only `client_id`,
  `client_secret`, `authorize_url`, `token_url`, `scopes`, `registration_url`,
  `server_url`, `oauth_config_id` — there is no `resource` / `audience` field.
- In `framework/oauth2`, `ResourceMetadata` is parsed for discovery
  (`authorization_servers`, `scopes`) but there is no field or mechanism that
  emits the `resource` parameter on the authorize/token requests.

The resource value is already known from the protected-resource document
(`resource: https://replit-mcp.com/server/mcp`), so Bifrost has everything it
needs — it just never forwards it. Net effect: any MCP server that mandates
RFC 8707 cannot be connected through Bifrost, regardless of UI or config-file
settings.

Related but distinct: #1043 (add OAuth2 for HTTP MCP clients) and #2806
(per_user_oauth tokens not forwarded to backend).

### Steps to reproduce

1. Bifrost → MCP Server Library → add the Replit server, URL
   `https://replit-mcp.com/server/mcp`, auth type OAuth 2.0.
2. Click Continue / Connect → auto-discovery fails with the "failed to fetch
   metadata from any authorization server" error above (failure 1).
3. Work around it: in OAuth Client Advanced Settings set
   - `authorize_url` `https://replit.com/oidc/auth`
   - `token_url` `https://replit.com/oidc/token`
   - `registration_url` `https://replit.com/oidc/reg`
   - `scopes` `apps:read apps:write offline_access`

   then run the OAuth flow.
4. The token request fails with
   `{"error":"invalid_target","error_description":"apps scopes require a valid MCP resource"}`
   (failure 2).

### Expected behavior

- During discovery, when an authorization server's issuer URL contains a path,
  Bifrost tries the OIDC `<issuer>/.well-known/openid-configuration` construction
  in addition to the RFC 8414 path-insertion form, and succeeds when either
  returns valid metadata.
- Bifrost sends the RFC 8707 `resource` parameter (the canonical MCP server URI
  from the protected-resource document, e.g. `https://replit-mcp.com/server/mcp`)
  on both the authorization and token requests, so servers requiring resource
  indicators issue tokens. Optionally expose a `resource` override field for
  manual configs.

### Actual behavior

- Discovery aborts with "failed to fetch metadata from any authorization server"
  even though the issuer's OIDC discovery document is reachable and valid at
  `<issuer>/.well-known/openid-configuration`.
- After manual endpoint configuration, the token exchange is rejected with
  `invalid_target` ("apps scopes require a valid MCP resource") because no
  `resource` parameter is sent, and there is no UI/config field to supply one.

### Affected area(s)

Core (Go), Framework — MCP OAuth (`framework/oauth2`), MCP gateway

### Version

v1.5.13

### Environment

- Bifrost: HTTP v1.5.13 (self-hosted Docker)
- Deployment: Docker on Linux
- MCP server under test: `https://replit-mcp.com/server/mcp` (Replit hosted MCP; AS = `https://replit.com/oidc`)

### Relevant logs/output

```
# Failure 1 — discovery
Error: Failed to initiate OAuth flow: OAuth discovery failed: failed to fetch authorization server metadata: failed to fetch metadata from any authorization server. Please provide authorize_url, token_url, and registration_url manually

# Failure 2 — token exchange after manual endpoint config
{"error":"invalid_target","error_description":"apps scopes require a valid MCP resource"}
```

### Regression?

No response

### Severity

Shaik-Sirajuddin:
## Summary

- implement the MCP authorization-server discovery order for path-bearing issuers
- reject metadata whose issuer does not match the discovered issuer
- preserve direct-issuer discovery as a documented legacy fallback

## Why

Path-bearing authorization-server issuers can publish valid OIDC metadata at `<issuer>/.well-known/openid-configuration`. The discovery client now tries the RFC 8414 and OIDC locations required by the MCP authorization specification and rejects issuer-mismatched metadata.

Fixes #4371.

## Note for maintainers
Written partially by self ,  assisted by ai 

## Testing

- `go test ./oauth2`

### DIFF
--- a/framework/oauth2/discovery.go
+++ b/framework/oauth2/discovery.go
@@ -321,9 +321,11 @@ func fetchSingleAuthServerMetadata(ctx context.Context, issuer string) (*OAuthMe
 
 			if err := json.Unmarshal(bodyBytes, &metadata); err == nil {
 				// RFC 8414 §3.3 and OIDC Discovery §4.3 require the returned
-				// issuer to exactly match the issuer used for discovery. The direct
-				// issuer URL is a legacy fallback, however, and may be a guessed
-				// MCP server origin rather than an authorization-server issuer.
+				// issuer to exactly match the issuer used for discovery. Deliberately
+				// compare the serialized values without URI normalization, including
+				// scheme and host case. The direct issuer URL is a legacy fallback,
+				// however, and may be a guessed MCP server origin rather than an
+				// authorization-server issuer.
 				if metadata.Issuer != canonicalIssuer && candidateURL != canonicalIssuer {
 					issuerMismatchErr = fmt.Errorf("authorization-server metadata issuer mismatch at %s: got %q, want %q", candidateURL, metadata.Issuer, canonicalIssuer)
 					logger.Warn("[OAuth Discovery] %v", issuerMismatchErr)
--- a/framework/oauth2/discovery.go
+++ b/framework/oauth2/discovery.go
@@ -6,6 +6,7 @@ import (
 	"crypto/sha256"
 	"encoding/base64"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
@@ -238,6 +239,7 @@ func attemptWellKnownDiscovery(ctx context.Context, serverURL string) ([]string,
 // fetchAuthorizationServerMetadata fetches OAuth endpoints from authorization server(s)
 // Tries multiple authorization servers until one succeeds
 func fetchAuthorizationServerMetadata(ctx context.Context, authServers []string) (*OAuthMetadata, error) {
+	var discoveryErrs []error
 	for _, issuer := range authServers {
 		logger.Debug(fmt.Sprintf("[OAuth Discovery] Fetching metadata from authorization server: %s", issuer))
 		metadata, err := fetchSingleAuthServerMetadata(ctx, issuer)
@@ -245,7 +247,13 @@ func fetchAuthorizationServerMetadata(ctx context.Context, authServers []string)
 			logger.Debug(fmt.Sprintf("[OAuth Discovery] Successfully fetched metadata from: %s", issuer))
 			return metadata, nil
 		}
+		if err != nil {
+			discoveryErrs = append(discoveryErrs, err)
+		}
 		logger.Debug(fmt.Sprintf("[OAuth Discovery] Failed to fetch from %s: %v", issuer, err))
+	}
+	if len(discoveryErrs) > 0 {
+		return nil, fmt.Errorf("failed to fetch metadata from any authorization server: %w", errors.Join(discoveryErrs...))
 	}
 	return nil, fmt.Errorf("failed to fetch metadata from any authorization server")
 }
@@ -288,6 +296,7 @@ func fetchSingleAuthServerMetadata(ctx context.Context, issuer string) (*OAuthMe
 	client := &http.Client{
 		Timeout: 10 * time.Second,
 	}
+	var issuerMismatchErr error
 
 	for _, candidateURL := range candidateURLs {
 		logger.Debug(fmt.Sprintf("[OAuth Discovery] Trying metadata endpoint: %s", candidateURL))
@@ -316,7 +325,8 @@ func fetchSingleAuthServerMetadata(ctx context.Context, issuer string) (*OAuthMe
 				// issuer URL is a legacy fallback, however, and may be a guessed
 				// MCP server origin rather than an authorization-server issuer.
 				if metadata.Issuer != canonicalIssuer && candidateURL != canonicalIssuer {
-					logger.Debug(fmt.Sprintf("[OAuth Discovery] Metadata issuer mismatch at %s: got %q, want %q", candidateURL, metadata.Issuer, canonicalIssuer))
+					issuerMismatchErr = fmt.Errorf("authorization-server metadata issuer mismatch at %s: got %q, want %q", candidateURL, metadata.Issuer, canonicalIssuer)
+					logger.Warn("[OAuth Discovery] %v", issuerMismatchErr)
 					continue
 				}
 
@@ -329,6 +339,9 @@ func fetchSingleAuthServerMetadata(ctx context.Context, issuer string) (*OAuthMe
 		} else {
 			resp.Body.Close()
 		}
+	}
+	if issuerMismatchErr != nil {
+		return nil, issuerMismatchErr
 	}
 
 	return nil, fmt.Errorf("no valid metadata found for issuer: %s", issuer)
--- a/framework/oauth2/discovery.go
+++ b/framework/oauth2/discovery.go
@@ -23,6 +23,8 @@ type OAuthMetadata struct {
 	RegistrationURL  *string  `json:"registration_endpoint,omitempty"`
 	ScopesSupported  []string `json:"scopes_supported,omitempty"`
 	Resource         string   `json:"resource,omitempty"`
+	// Required by RFC 8414 and OIDC Discovery metadata; omitempty is retained
+	// for legacy metadata documents that omit it.
 	Issuer           string   `json:"issuer,omitempty"`
 	ResponseTypes    []string `json:"response_types_supported,omitempty"`
 	GrantTypes       []string `json:"grant_types_supported,omitempty"`
--- a/framework/oauth2/discovery.go
+++ b/framework/oauth2/discovery.go
@@ -253,7 +253,11 @@ func fetchAuthorizationServerMetadata(ctx context.Context, authServers []string)
 // fetchSingleAuthServerMetadata tries multiple well-known endpoints for a single authorization server
 // Implements RFC 8414 discovery
 func fetchSingleAuthServerMetadata(ctx context.Context, issuer string) (*OAuthMetadata, error) {
-	base, path := splitURL(issuer)
+	// RFC 8414 §3.1 requires removal of a terminating slash before the
+	// well-known path is inserted. Use the same normalized issuer when
+	// validating the metadata issuer (RFC 8414 §3.3).
+	canonicalIssuer := strings.TrimSuffix(issuer, "/")
+	base, path := splitURL(canonicalIssuer)
 	if base == "" {
 		return nil, fmt.Errorf("invalid issuer URL: %s", issuer)
 	}
@@ -268,7 +272,7 @@ func fetchSingleAuthServerMetadata(ctx context.Context, issuer string) (*OAuthMe
 		candidateURLs = append(candidateURLs,
 			fmt.Sprintf("%s/.well-known/oauth-authorization-server/%s", base, path),
 			fmt.Sprintf("%s/.well-known/openid-configuration/%s", base, path),
-			strings.TrimSuffix(issuer, "/")+"/.well-known/openid-configuration",
+			canonicalIssuer+"/.well-known/openid-configuration",
 		)
 	} else {
 		candidateURLs = append(candidateURLs,
@@ -279,7 +283,7 @@ func fetchSingleAuthServerMetadata(ctx context.Context, issuer string) (*OAuthMe
 
 	// Legacy compatibility fallback for authorization servers that publish
 	// metadata directly at the issuer URL. This is not part of MCP discovery.
-	candidateURLs = append(candidateURLs, strings.TrimSuffix(issuer, "/"))
+	candidateURLs = append(candidateURLs, canonicalIssuer)
 
 	client := &http.Client{
 		Timeout: 10 * time.Second,
@@ -311,8 +315,8 @@ func fetchSingleAuthServerMetadata(ctx context.Context, issuer string) (*OAuthMe
 				// issuer to exactly match the issuer used for discovery. Without
 				// this, metadata published by another issuer on the same host could
 				// be accepted for a path-bearing issuer.
-				if metadata.Issuer != issuer {
-					logger.Debug(fmt.Sprintf("[OAuth Discovery] Metadata issuer mismatch at %s: got %q, want %q", candidateURL, metadata.Issuer, issuer))
+				if metadata.Issuer != canonicalIssuer {
+					logger.Debug(fmt.Sprintf("[OAuth Discovery] Metadata issuer mismatch at %s: got %q, want %q", candidateURL, metadata.Issuer, canonicalIssuer))
 					continue
 				}
 
--- a/framework/oauth2/discovery.go
+++ b/framework/oauth2/discovery.go
@@ -258,19 +258,28 @@ func fetchSingleAuthServerMetadata(ctx context.Context, issuer string) (*OAuthMe
 		return nil, fmt.Errorf("invalid issuer URL: %s", issuer)
 	}
 
-	// Try different well-known endpoint patterns
+	// MCP's required discovery order is documented at:
+	// https://modelcontextprotocol.io/specification/2026-07-28/basic/authorization/authorization-server-discovery#authorization-server-metadata-discovery
+	// RFC 8414 inserts the well-known path before a path-bearing issuer,
+	// while OIDC Discovery appends it to the issuer. Host-level URLs for a
+	// path-bearing issuer describe a different issuer and are not candidates.
 	var candidateURLs []string
 	if path != "" {
 		candidateURLs = append(candidateURLs,
 			fmt.Sprintf("%s/.well-known/oauth-authorization-server/%s", base, path),
 			fmt.Sprintf("%s/.well-known/openid-configuration/%s", base, path),
+			strings.TrimSuffix(issuer, "/")+"/.well-known/openid-configuration",
+		)
+	} else {
+		candidateURLs = append(candidateURLs,
+			fmt.Sprintf("%s/.well-known/oauth-authorization-server", base),
+			fmt.Sprintf("%s/.well-known/openid-configuration", base),
 		)
 	}
-	candidateURLs = append(candidateURLs,
-		fmt.Sprintf("%s/.well-known/oauth-authorization-server", base),
-		fmt.Sprintf("%s/.well-known/openid-configuration", base),
-		strings.TrimSuffix(issuer, "/"), // Try the issuer URL itself
-	)
+
+	// Legacy compatibility fallback for authorization servers that publish
+	// metadata directly at the issuer URL. This is not part of MCP discovery.
+	candidateURLs = append(candidateURLs, strings.TrimSuffix(issuer, "/"))
 
 	client := &http.Client{
 		Timeout: 10 * time.Second,
@@ -298,7 +307,16 @@ func fetchSingleAuthServerMetadata(ctx context.Context, issuer string) (*OAuthMe
 			}
 
 			if err := json.Unmarshal(bodyBytes, &metadata); err == nil {
-				// Validate that we got at least authorization_endpoint
+				// RFC 8414 §3.3 and OIDC Discovery §4.3 require the returned
+				// issuer to exactly match the issuer used for discovery. Without
+				// this, metadata published by another issuer on the same host could
+				// be accepted for a path-bearing issuer.
+				if metadata.Issuer != issuer {
+					logger.Debug(fmt.Sprintf("[OAuth Discovery] Metadata issuer mismatch at %s: got %q, want %q", candidateURL, metadata.Issuer, issuer))
+					continue
+				}
+
+				// Validate that we got at least authorization_endpoint.
 				if metadata.AuthorizationURL != "" {
 					logger.Debug(fmt.Sprintf("[OAuth Discovery] Valid metadata found at: %s", candidateURL))
 					return &metadata, nil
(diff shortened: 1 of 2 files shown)


Shaik-Sirajuddin:
@akshaydeo 
my apologies for pinging again

Could you take a look , inform me if screenshots or related needed to keep it easy for review

framework/oauth2/discovery.go
```diff
@@ -298,7 +307,16 @@ func fetchSingleAuthServerMetadata(ctx context.Context, issuer string) (*OAuthMe
 			}
 
 			if err := json.Unmarshal(bodyBytes, &metadata); err == nil {
-				// Validate that we got at least authorization_endpoint
+				// RFC 8414 §3.3 and OIDC Discovery §4.3 require the returned
+				// issuer to exactly match the issuer used for discovery. Without
+				// this, metadata published by another issuer on the same host could
+				// be accepted for a path-bearing issuer.
+				if metadata.Issuer != issuer {
```
akshaydeo: **Findings 1, 2, 3 and 4 (blocking).** Three separate problems land on this one comparison.

**Trailing slash.** RFC 8414 section 3.1: "any terminating `/` MUST be removed before inserting `/.well-known/`". `splitURL` only strips the leading slash, so issuer `https://as.example/tenant1/` builds `https://as.example/.well-known/oauth-authorization-server/tenant1/` and then compares the untrimmed issuer against the spec-correct `https://as.example/tenant1` the server returns. Mismatch, `continue`, hard fail. The path-less form breaks the same way: issuer `https://as.example/` versus metadata issuer `https://as.example`.

**Absent `issuer`.** `OAuthMetadata.Issuer` is `json:"issuer,omitempty"` at discovery.go:25, so this codebase already treats the member as optional. An authorization server publishing minimal metadata at its issuer URL now fails, which silently defeats the legacy fallback the comment at lines 280-281 promises to preserve.

**Inferred issuers.** When there is no protected-resource metadata, `attemptWellKnownDiscovery` returns `[]string{base}` under the comment "assuming server base is auth server" (discovery.go:234). That is a local guess, not an issuer identifier the client obtained, so exact-matching it is stricter than RFC 8414 section 3.3 asks.

Suggested shape:

```go
normalizedIssuer := strings.TrimSuffix(issuer, "/")
// ...every candidate built from normalizedIssuer...
if strings.TrimSuffix(metadata.Issuer, "/") != normalizedIssuer {
    issuerMismatch = fmt.Sprintf("metadata at %s declares issuer %q, want %q",
        candidateURL, metadata.Issuer, normalizedIssuer)
    logger.Debug("[OAuth Discovery] " + issuerMismatch)
    continue
}
```

Please also return `issuerMismatch` in the terminal error at discovery.go:330. Right now the mismatch is `Debug`-only and the user sees the generic "no valid metadata found", which oauth2/main.go:673 turns into advice to configure the URLs by hand.

framework/oauth2/discovery.go
```diff
@@ -258,19 +258,28 @@ func fetchSingleAuthServerMetadata(ctx context.Context, issuer string) (*OAuthMe
...
-	// Try different well-known endpoint patterns
+	// MCP's required discovery order is documented at:
+	// https://modelcontextprotocol.io/specification/2026-07-28/basic/authorization/authorization-server-discovery#authorization-server-metadata-discovery
+	// RFC 8414 inserts the well-known path before a path-bearing issuer,
+	// while OIDC Discovery appends it to the issuer. Host-level URLs for a
+	// path-bearing issuer describe a different issuer and are not candidates.
 	var candidateURLs []string
 	if path != "" {
 		candidateURLs = append(candidateURLs,
 			fmt.Sprintf("%s/.well-known/oauth-authorization-server/%s", base, path),
```
akshaydeo, line 283: **Finding 1 (blocking), construction side.** `path` comes from `splitURL`, which only strips the *leading* slash (`strings.TrimPrefix(parsedURL.Path, "/")`). For issuer `https://as.example/tenant1/` that yields `path == "tenant1/"` and this candidate becomes `https://as.example/.well-known/oauth-authorization-server/tenant1/`.

RFC 8414 section 3.1 requires the opposite: "any terminating `/` MUST be removed before inserting `/.well-known/` and the well-known URI suffix between the host component and the path component." Strict servers 404 the trailing-slash form.

Cleanest fix is to normalize once at the top of the function and derive everything from it:

```go
normalizedIssuer := strings.TrimSuffix(issuer, "/")
base, path := splitURL(normalizedIssuer)
```

That also fixes the comparison at line 314 and lets line 271 and line 282 drop their own `TrimSuffix` calls.

framework/oauth2/discovery.go
```diff
@@ -258,19 +258,28 @@ func fetchSingleAuthServerMetadata(ctx context.Context, issuer string) (*OAuthMe
...
 	}
-	candidateURLs = append(candidateURLs,
-		fmt.Sprintf("%s/.well-known/oauth-authorization-server", base),
-		fmt.Sprintf("%s/.well-known/openid-configuration", base),
-		strings.TrimSuffix(issuer, "/"), // Try the issuer URL itself
-	)
+
+	// Legacy compatibility fallback for authorization servers that publish
+	// metadata directly at the issuer URL. This is not part of MCP discovery.
+	candidateURLs = append(candidateURLs, strings.TrimSuffix(issuer, "/"))
```
akshaydeo: **Findings 2 and 7.** Two notes on this fallback.

First, as written it is close to unreachable for the servers it exists to serve: the issuer check at line 314 requires an exact `issuer` member, so a server publishing minimal metadata directly at its issuer URL is rejected here. The servers that do pass are the fully RFC 8414 compliant ones, which already matched a spec'd candidate above. Consider accepting an *absent* `issuer` from this candidate only, while still rejecting it from every `.well-known` candidate. That keeps the property that matters - a `.well-known` document cannot claim to speak for a different issuer - and leaves the fallback actually usable.

Second, the comment is right that this is not part of MCP discovery, and it costs one extra request on every failed discovery. A config flag to disable it would let strict deployments opt out. A follow-up PR is fine for that part.

framework/oauth2/discovery_test.go
```diff
@@ -0,0 +1,85 @@
...
+		}
+		w.Header().Set("Content-Type", "application/json")
+		_, _ = w.Write([]byte(`{"issuer":"https://issuer.example/other-tenant","authorization_endpoint":"https://issuer.example/authorize","token_endpoint":"https://issuer.example/token"}`))
+	}))
+	defer server.Close()
+
+	metadata, err := fetchSingleAuthServerMetadata(context.Background(), server.URL+"/tenant/prod/oidc")
+	require.Error(t, err)
+	assert.Nil(t, metadata)
+}
```
akshaydeo, line 91: **Finding 5 (nit).** Nice, focused tests - asserting the exact request paths is the right way to pin discovery order, and it is what let me verify the ordering against the MCP spec quickly. Three gaps worth closing while you are here, since they are the branches most likely to regress:

1. The path-less `else` branch added at discovery.go:273-278 has no test asserting it requests `/.well-known/oauth-authorization-server` then `/.well-known/openid-configuration` and nothing else.
2. A trailing-slash issuer (`server.URL + "/tenant/prod/oidc/"`) whose metadata declares the RFC-normalized issuer without the slash. That case currently fails.
3. Metadata that omits `issuer` entirely, served from the legacy direct-issuer candidate, asserting whichever behavior you settle on for finding 2.

akshaydeo (changes requested):
Thanks for this, @Shaik-Sirajuddin - the discovery-order rewrite and the issuer check are the right change, and citing the MCP spec and RFC 8414 inline in the code made this quick to verify. I checked both claims against the upstream sources:

- [MCP Authorization Server Discovery (2026-07-28)](https://modelcontextprotocol.io/specification/2026-07-28/basic/authorization/authorization-server-discovery): for path-bearing issuers clients **MUST** try, in this order, `/.well-known/oauth-authorization-server/{path}`, `/.well-known/openid-configuration/{path}`, then `{issuer}/.well-known/openid-configuration`; for path-less issuers, the two host-level URLs. The new candidate lists match that exactly, including dropping the host-level URLs for path-bearing issuers.
- Same page: "the `issuer` value in the document **MUST** be identical to the issuer identifier used to construct the well-known URL. If they differ, the client **MUST NOT** use the metadata." Mirrored by [RFC 8414 section 3.3](https://datatracker.ietf.org/doc/html/rfc8414#section-3.3): "The `issuer` value returned MUST be identical to the authorization server's issuer identifier value into which the well-known URI string was inserted [...] If these values are not identical, the data contained in the response MUST NOT be used."

So the direction is correct, and exact-match rather than prefix-match is the right reading of both specs. What holds this back is trailing-slash normalization and the blast radius of the new rejection on configs that work today.

## Findings

|---|---|---|---|---|
| 1 | High | [discovery.go:314](https://github.com/maximhq/bifrost/blob/404826f4eeac401aba4512094b23a06af6cf0876/framework/oauth2/discovery.go#L314) | An issuer with a terminating `/` now fails discovery outright. [RFC 8414 section 3.1](https://datatracker.ietf.org/doc/html/rfc8414#section-3.1): "any terminating `/` MUST be removed before inserting `/.well-known/`". `splitURL` only strips the leading slash, so `https://as.example/tenant1/` builds `https://as.example/.well-known/oauth-authorization-server/tenant1/` (wrong URL, 404 on strict servers) and then compares the untrimmed `https://as.example/tenant1/` against the spec-correct `https://as.example/tenant1` the server returns. Mismatch, `continue`, hard fail. The path-less form breaks the same way: issuer `https://as.example/` versus metadata issuer `https://as.example`. | CONFIRMED |
| 2 | High | [discovery.go:314](https://github.com/maximhq/bifrost/blob/404826f4eeac401aba4512094b23a06af6cf0876/framework/oauth2/discovery.go#L314) | Metadata that omits `issuer` is now rejected, which silently defeats the "legacy compatibility fallback" this PR says it preserves. `OAuthMetadata.Issuer` is declared `json:"issuer,omitempty"` at discovery.go:25, so the codebase already treats it as optional. An authorization server publishing minimal metadata at its issuer URL with only `authorization_endpoint` and `token_endpoint` worked before this PR and hard-fails after it. The fallback added at line 282 can now only succeed for servers that were already fully RFC 8414 compliant, which are exactly the servers that never needed a fallback. | CONFIRMED |
| 3 | Medium | [discovery.go:314](https://github.com/maximhq/bifrost/blob/404826f4eeac401aba4512094b23a06af6cf0876/framework/oauth2/discovery.go#L314) | The strict check is also applied to a *guessed* issuer. When there is no protected-resource metadata, `attemptWellKnownDiscovery` returns `[]string{base}` under the comment "assuming server base is auth server" (discovery.go:234), where `base` is the MCP server's `scheme://host`. That string is not an issuer identifier the client obtained from anywhere. RFC 8414 section 3.3 is specifically about the identifier "into which the well-known URI string was inserted", so exact-matching a local guess is stricter than the spec asks and converts a previously working fallback into a hard failure whenever the real AS issuer differs from the MCP host root. | CONFIRMED |
| 4 | Medium | [discovery.go:314](https://github.com/maximhq/bifrost/blob/404826f4eeac401aba4512094b23a06af6cf0876/framework/oauth2/discovery.go#L314) | The rejection is invisible to operators. The mismatch is logged at `Debug` only, and the returned error is the generic `no valid metadata found for issuer: %s` (discovery.go:330), which `oauth2/main.go:673` wraps into "Please provide authorize_url, token_url, and registration_url manually". Given findings 1 to 3, users will hit this on upgrade with no way to distinguish an issuer mismatch from a 404 without enabling debug logging. | CONFIRMED |
| 5 | Low | [discovery_test.go](https://github.com/maximhq/bifrost/blob/404826f4eeac401aba4512094b23a06af6cf0876/framework/oauth2/discovery_test.go) | Coverage gaps on the new branches: no test for the path-less `else` branch added at lines 273-278, none for a trailing-slash issuer, and none for metadata that omits `issuer`. The three cases most likely to regress are the untested ones. | CONFIRMED |
| 6 | Low | [discovery.go:314](https://github.com/maximhq/bifrost/blob/404826f4eeac401aba4512094b23a06af6cf0876/framework/oauth2/discovery.go#L314) | Byte-exact comparison also rejects case differences in scheme and host, which RFC 3986 section 6.2.2.1 defines as case-insensitive. RFC 8414 does say "identical", so this is spec-literal and defensible. Flagging it only so the choice is deliberate rather than accidental. | PLAUSIBLE |
| 7 | Low | [discovery.go:282](https://github.com/maximhq/bifrost/blob/404826f4eeac401aba4512094b23a06af6cf0876/framework/oauth2/discovery.go#L282) | The legacy direct-issuer probe is not in the MCP MUST list, so every failed discovery now costs one extra request beyond the spec'd candidates. Fine to keep for compatibility given the comment, but worth a config gate so strict deployments can turn it off. | CONFIRMED |

## Suggested fix for 1, 2 and 4

Normalize once, and make the failure legible:

```go
normalizedIssuer := strings.TrimSuffix(issuer, "/")
base, path := splitURL(normalizedIssuer)
// ...build every candidate from normalizedIssuer...

// RFC 8414 section 3.1 removes the terminating "/" before insertion, so compare
// the normalized forms rather than the raw config strings.
if strings.TrimSuffix(metadata.Issuer, "/") != normalizedIssuer {
    issuerMismatch = fmt.Sprintf("metadata at %s declares issuer %q, want %q",
        candidateURL, metadata.Issuer, normalizedIssuer)
    logger.Debug("[OAuth Discovery] " + issuerMismatch)
    continue
}
```

and return the last recorded `issuerMismatch` in the terminal error at discovery.go:330 instead of the generic message.

For finding 2, please make an explicit call either way rather than leaving it implicit. Two defensible options:

1. Keep rejecting an empty `issuer` (spec-strict, RFC 8414 makes the member REQUIRED), but say so in the error text and flag it in the release notes, since it is a breaking change for existing working configs.
2. Accept metadata with an absent `issuer` only from the legacy direct-issuer candidate at line 282, and reject it from every `.well-known` candidate. That keeps the security property that actually matters - a `.well-known` document cannot claim to speak for a different issuer - while leaving the documented fallback usable.

I lean toward option 2, because it preserves what the comment at lines 280-281 promises.

For finding 3, thread a flag out of `attemptWellKnownDiscovery` marking the issuer as inferred rather than discovered, and skip the exact match on that path only. The spec's threat model is a metadata document claiming another party's issuer identity, and a locally guessed base was never an identity claim to begin with.

## Merge recommendation

**Not yet - request changes.** Findings 1 and 2 are confirmed regressions that will hard-fail MCP OAuth configurations working on `dev` today, in a code path whose failure mode is a total loss of connectivity to the server. The security direction is right and I would be glad to see this land as soon as trailing-slash normalization and the empty-issuer decision are settled.

## Followups

1. **In this PR (blocking)** - `framework/oauth2/discovery.go:256,314`: trim the terminating `/` from the issuer before both URL construction and comparison (finding 1).
2. **In this PR (blocking)** - `framework/oauth2/discovery.go:314`: decide and document the behavior for metadata with an absent `issuer`, per the two options above (finding 2).
3. **In this PR (blocking)** - `framework/oauth2/discovery.go:314` plus `discovery.go:234`: do not exact-match against the inferred `base` issuer coming from `attemptWellKnownDiscovery` (finding 3).
4. **In this PR (nit)** - `framework/oauth2/discovery.go:330`: surface the issuer mismatch in the returned error, not only at `Debug` (finding 4).
5. **In this PR (nit)** - `framework/oauth2/discovery_test.go`: add cases for the path-less branch, a trailing-slash issuer, and metadata with no `issuer` (finding 5).
6. **Follow-up PR** - `framework/oauth2/discovery.go:282`: config gate for the legacy direct-issuer probe (finding 7).

## Checked and cleared

- **Discovery order wrong for path-bearing issuers** - REFUTED. The three candidates at lines 268-272 match the MCP spec's required order exactly, including the `{issuer}/.well-known/openid-configuration` append form as third.
- **Dropping host-level candidates for path-bearing issuers is a regression** - REFUTED. The spec's path-bearing list contains no host-level URL, and accepting one would mean accepting metadata for a different issuer. This is the security fix, correctly done.
- **Prefix matching should be permitted for issuers** - REFUTED. Both RFC 8414 section 3.3 and the MCP spec require "identical", not a prefix relationship.
- **Duplicate candidate URL between the legacy fallback and the spec'd candidates** - REFUTED. `strings.TrimSuffix(issuer, "/")` never equals any well-known URL in either branch.
- **Response body leak on the new `continue`** - REFUTED. `resp.Body.Close()` at line 303 runs before the issuer check at line 314.
- **`discovery_test.go` violates the repo Go filename convention** - REFUTED. AGENTS.md line 733 permits the `_test.go` suffix explicitly.
- **Data race on `requestedPaths` between the `httptest` handler and the assertions** - REFUTED. Each handler write happens-before the client finishes reading that response, and the assertions run after `fetchSingleAuthServerMetadata` returns.

Thanks again for putting the spec citations directly in the code comments. That is genuinely useful for the next person who reads this function.

framework/oauth2/discovery.go
Shaik-Sirajuddin: RFC marks 'issuer' as required and is used for verification , 
https://datatracker.ietf.org/doc/html/rfc8414#section-2

keeping it allowed for legacy url constrcted paths

framework/oauth2/discovery.go
```diff
@@ -298,7 +322,19 @@ func fetchSingleAuthServerMetadata(ctx context.Context, issuer string) (*OAuthMe
...
 
 			if err := json.Unmarshal(bodyBytes, &metadata); err == nil {
-				// Validate that we got at least authorization_endpoint
+				// RFC 8414 §3.3 and OIDC Discovery §4.3 require the returned
+				// issuer to exactly match the issuer used for discovery. Deliberately
+				// compare the serialized values without URI normalization, including
+				// scheme and host case. The direct issuer URL is a legacy fallback,
+				// however, and may be a guessed MCP server origin rather than an
+				// authorization-server issuer.
+				if metadata.Issuer != canonicalIssuer && candidateURL != canonicalIssuer {
```
Shaik-Sirajuddin, line 331: the gate was previous missing for legacy compatbility

Shaik-Sirajuddin:
Thanks's for feedback @akshaydeo 

error propogation , test coverage , covered as per review 
enhanced docs clarifying the legacy path

PR #6254, opened 2026-08-18
feat: honor context deadlines over provider defaults

Shaik-Sirajuddin:
## Summary

Apply the remaining `BifrostContext` deadline as the per-request FastHTTP timeout for unary requests.

This lets an explicit context timeout override the provider’s configured default timeout for shorter than default timeouts 
longer than default as upstream tracked at https://github.com/valyala/fasthttp/issues/2370

## Note for maintainers
written by hand by self , assisted by ai
## Testing

- `go test ./providers/utils`
- Added TCP integration coverage for a longer context deadline overriding a shorter FastHTTP timeout.

### DIFF
--- a/core/providers/utils/makerequest_test.go
+++ b/core/providers/utils/makerequest_test.go
@@ -101,14 +101,86 @@ func TestMakeRequestWithContext_DeadlineExceededReturnsTimeoutError(t *testing.T
 	wait()
 	elapsed := time.Since(start)
 
-	// The wait should have taken roughly the remaining server delay (~490ms)
-	if elapsed < 200*time.Millisecond {
-		t.Fatalf("wait() returned too quickly (%v), expected it to block until goroutine finishes", elapsed)
+	// The request inherits the context deadline, so the client goroutine has
+	// already finished by the time the context cancellation is observed.
+	if elapsed > 200*time.Millisecond {
+		t.Fatalf("wait() blocked for %v after context timeout", elapsed)
 	}
 
 	// Now safe to release
 	fasthttp.ReleaseRequest(req)
 	fasthttp.ReleaseResponse(resp)
+}
+
+func TestMakeRequestWithContext_ContextDeadlineExtendsClientReadTimeout(t *testing.T) {
+	// Use a real TCP listener to exercise fasthttp's normal client path. The
+	// upstream accepts the request but deliberately never sends a response.
+	listener, err := net.Listen("tcp", "127.0.0.1:0")
+	if err != nil {
+		t.Fatalf("listen: %v", err)
+	}
+
+	releaseUpstream := make(chan struct{})
+	upstreamDone := make(chan struct{})
+	go func() {
+		defer close(upstreamDone)
+
+		conn, acceptErr := listener.Accept()
+		if acceptErr != nil {
+			return
+		}
+		defer conn.Close()
+
+		<-releaseUpstream
+	}()
+	defer func() {
+		close(releaseUpstream)
+		_ = listener.Close()
+		<-upstreamDone
+	}()
+
+	// This is a scaled-down version of a 30-second FastHTTP timeout with a
+	// 60-second context deadline. The request-level deadline must replace the
+	// shorter client default, not be capped by it.
+	const clientReadTimeout = 100 * time.Millisecond
+	client := &fasthttp.Client{
+		ReadTimeout:  clientReadTimeout,
+		WriteTimeout: clientReadTimeout,
+	}
+
+	req := fasthttp.AcquireRequest()
+	resp := fasthttp.AcquireResponse()
+	defer fasthttp.ReleaseRequest(req)
+	defer fasthttp.ReleaseResponse(resp)
+	req.SetRequestURI("http://" + listener.Addr().String() + "/")
+
+	const contextTimeout = 500 * time.Millisecond
+	ctx, cancel := context.WithTimeout(context.Background(), contextTimeout)
+	defer cancel()
+
+	start := time.Now()
+	_, bifrostErr, wait := MakeRequestWithContext(ctx, client, req, resp)
+	elapsed := time.Since(start)
+
+	// Calling wait before releasing the pooled request and response mirrors the
+	// provider path. It must return promptly because the request's context
+	// deadline also terminates the underlying FastHTTP request.
+	waitStart := time.Now()
+	wait()
+	waitElapsed := time.Since(waitStart)
+
+	if bifrostErr == nil || bifrostErr.Error.Type == nil || *bifrostErr.Error.Type != schemas.RequestTimedOut {
+		t.Fatalf("expected context timeout error, got: %#v", bifrostErr)
+	}
+	if elapsed < 400*time.Millisecond {
+		t.Fatalf("request ended after %v; expected context timeout (%v) to extend client read timeout (%v)", elapsed, contextTimeout, clientReadTimeout)
+	}
+	if elapsed > time.Second {
+		t.Fatalf("request took %v; expected it to time out near the context deadline (%v)", elapsed, contextTimeout)
+	}
+	if waitElapsed > time.Second {
+		t.Fatalf("wait blocked for %v after the context timeout; request timeout was not applied", waitElapsed)
+	}
 }
 
 func TestMakeRequestWithContext_ContextCancelReturnsCancelledError(t *testing.T) {
--- a/core/providers/utils/utils.go
+++ b/core/providers/utils/utils.go
@@ -432,6 +432,10 @@ func makeRequestWithDoFunc(ctx context.Context, do func() error) (time.Duration,
 // path it blocks until the background client.Do goroutine finishes, preventing a data race
 // between the still-running goroutine and the caller's release of req/resp.
 func MakeRequestWithContext(ctx context.Context, client *fasthttp.Client, req *fasthttp.Request, resp *fasthttp.Response) (time.Duration, *schemas.BifrostError, func()) {
+	if deadline, ok := ctx.Deadline(); ok {
+		req.SetTimeout(time.Until(deadline))
+	}
+
 	latency, bifrostErr, wait := makeRequestWithDoFunc(ctx, func() error { return client.Do(req, resp) })
 	return latency, bifrostErr, wait
 }
@@ -439,6 +443,10 @@ func MakeRequestWithContext(ctx context.Context, client *fasthttp.Client, req *f
 // MakeRequestWithContextFollowRedirects is like MakeRequestWithContext but follows up to
 // maxRedirects HTTP redirects automatically (equivalent to curl's -L flag).
 func MakeRequestWithContextFollowRedirects(ctx context.Context, client *fasthttp.Client, req *fasthttp.Request, resp *fasthttp.Response, maxRedirects int) (time.Duration, *schemas.BifrostError, func()) {
+	if deadline, ok := ctx.Deadline(); ok {
+		req.SetTimeout(time.Until(deadline))
+	}
+
 	latency, bifrostErr, wait := makeRequestWithDoFunc(ctx, func() error { return client.DoRedirects(req, resp, maxRedirects) })
 	return latency, bifrostErr, wait
 }


core/providers/utils/utils.go
```diff
@@ -432,13 +432,21 @@ func makeRequestWithDoFunc(ctx context.Context, do func() error) (time.Duration,
 // path it blocks until the background client.Do goroutine finishes, preventing a data race
 // between the still-running goroutine and the caller's release of req/resp.
 func MakeRequestWithContext(ctx context.Context, client *fasthttp.Client, req *fasthttp.Request, resp *fasthttp.Response) (time.Duration, *schemas.BifrostError, func()) {
+	if deadline, ok := ctx.Deadline(); ok {
```
akshaydeo, line 435: Three issues land on this block (findings 1, 3, 4).

**It cannot extend a shorter provider timeout.** In fasthttp v1.71.0 `transport.RoundTrip` computes `readDeadline` from `req.timeout`, then narrows it: `if tmpReadDeadline.Before(readDeadline) { readDeadline = tmpReadDeadline }` when `hc.ReadTimeout > 0`. Measured against the pinned version with `Client{ReadTimeout: 100ms}` and a POST (the shape of every provider call):

```
POST no SetTimeout      elapsed=100ms
POST SetTimeout(500ms)  elapsed=100ms   <-- not extended
POST SetTimeout(50ms)   elapsed=50ms    <-- shortening works
```

**It re-arms a read deadline on clients that intentionally have none.** `BuildLargeResponseClient` (`core/providers/utils/largeresponse.go:81`) sets `ReadTimeout = 0`, `WriteTimeout = 0`, `StreamResponseBody = true`, and `PrepareResponseStreaming` sets `resp.StreamBody = true`. Those clients reach this function from `core/providers/cohere/cohere.go:190`, `core/providers/gemini/gemini.go:140/768/1239/2015/2106/4185`, and `core/providers/azure/azure.go:3545`. With `ReadTimeout == 0` the `RoundTrip` branch is `readDeadline = deadline`, so the connection now carries an absolute wall clock that also covers the body drained lazily after this function returns. Repro with body trickled over 1s: no `SetTimeout` succeeds at 1.02s, `SetTimeout(200ms)` fails with `i/o timeout` at 200ms. `AGENTS.md` records the invariant: "`ReadTimeout=0`; the per-chunk `NewIdleTimeoutReader` is the only governor".

**An expired deadline is treated as unbounded.** `HostClient.Do` does `timeout := req.timeout; if timeout > 0 { ... }`, so a negative `time.Until(deadline)` means no per-request timeout at all, and the deferred `wait()` can still block for the full provider default.

Suggested shape covering all three:

```go
// A streamed body is drained after this call returns, and its client
// deliberately runs with ReadTimeout=0 so NewIdleTimeoutReader is the only
// governor. A request deadline would put a wall clock back on that transfer.
if deadline, ok := ctx.Deadline(); ok && !resp.StreamBody && client.ReadTimeout > 0 {
    remaining := time.Until(deadline)
    if remaining <= 0 {
        // fasthttp reads any value <= 0 as "no timeout"; keep it expired instead.
        remaining = time.Nanosecond
    }
    req.SetTimeout(remaining)
}
```

Extracting that into `applyContextDeadline(ctx, client, req, resp)` would also remove the copy at line 446.

core/providers/utils/makerequest_test.go
```diff
@@ -101,16 +101,88 @@ func TestMakeRequestWithContext_DeadlineExceededReturnsTimeoutError(t *testing.T
...
+	if elapsed > 200*time.Millisecond {
+		t.Fatalf("wait() blocked for %v after context timeout", elapsed)
 	}
 
 	// Now safe to release
 	fasthttp.ReleaseRequest(req)
 	fasthttp.ReleaseResponse(resp)
 }
 
+func TestMakeRequestWithContext_ContextDeadlineExtendsClientReadTimeout(t *testing.T) {
```
akshaydeo, line 115: This test passes identically with and without the two `utils.go` hunks, so it does not currently protect the change (finding 2).

The request has no method set, so it is a GET, and `isIdempotent(GET)` is true. `HostClient.Do` therefore retries up to `DefaultMaxIdemponentCallAttempts` (5), and each attempt burns the full 100ms `clientReadTimeout`. Measured against fasthttp v1.71.0 with the same listener shape:

```
GET no SetTimeout (baseline)  elapsed=510ms
GET SetTimeout(500ms)         elapsed=500ms
```

Both land inside the `elapsed >= 400ms` / `elapsed <= 1s` window, so the assertions cannot tell the two builds apart. The 510ms baseline is also uncomfortably close to the 1s ceiling, which makes the test flaky on a loaded runner (finding 5).

The name and the comment ("The request-level deadline must replace the shorter client default, not be capped by it") assert a property fasthttp does not provide: `transport.RoundTrip` takes the minimum of the request deadline and `hc.ReadTimeout`.

Suggested reshape, which is red on `dev` and green with the change:

```go
func TestMakeRequestWithContext_ContextDeadlineShortensClientReadTimeout(t *testing.T) {
    // ... same listener that accepts and never responds ...

    // Client default is deliberately the LONGER of the two.
    client := &fasthttp.Client{
        ReadTimeout:  2 * time.Second,
        WriteTimeout: 2 * time.Second,
    }

    req := fasthttp.AcquireRequest()
    resp := fasthttp.AcquireResponse()
    defer fasthttp.ReleaseRequest(req)
    defer fasthttp.ReleaseResponse(resp)
    req.SetRequestURI("http://" + listener.Addr().String() + "/")
    // POST matches every provider call and is non-idempotent, so fasthttp's
    // retry loop cannot mask the deadline being applied.
    req.Header.SetMethod(fasthttp.MethodPost)
    req.SetBody([]byte("{}"))

    ctx, cancel := context.WithTimeout(context.Background(), 200*time.Millisecond)
    defer cancel()

    _, bifrostErr, wait := MakeRequestWithContext(ctx, client, req, resp)
    waitStart := time.Now()
    wait()
    waitElapsed := time.Since(waitStart)

    if bifrostErr == nil || bifrostErr.Error.Type == nil || *bifrostErr.Error.Type != schemas.RequestTimedOut {
        t.Fatalf("expected context timeout error, got: %#v", bifrostErr)
    }
    // Without the request deadline this blocks for the full 2s client ReadTimeout.
    if waitElapsed > 500*time.Millisecond {
        t.Fatalf("wait() blocked for %v; context deadline was not applied to the request", waitElapsed)
    }
}
```

Worth adding a sibling case that asserts a streamed-body response (`resp.StreamBody = true` with a `ReadTimeout: 0` client) does *not* pick up a request deadline, to lock in the finding 3 carve-out.

core/providers/utils/makerequest_test.go
```diff
@@ -101,16 +101,88 @@ func TestMakeRequestWithContext_DeadlineExceededReturnsTimeoutError(t *testing.T
 	wait()
 	elapsed := time.Since(start)
 
-	// The wait should have taken roughly the remaining server delay (~490ms)
-	if elapsed < 200*time.Millisecond {
-		t.Fatalf("wait() returned too quickly (%v), expected it to block until goroutine finishes", elapsed)
+	// The request inherits the context deadline, so the client goroutine has
+	// already finished by the time the context cancellation is observed.
+	if elapsed > 200*time.Millisecond {
```
akshaydeo, line 106: Adapting this assertion to the new behavior is reasonable, but inverting it drops the property it existed to protect (finding 6).

The original `elapsed >= 200ms` check verified the contract documented on `MakeRequestWithContext`: on the context path, `wait()` blocks until the background `client.Do` goroutine finishes, so the subsequent `ReleaseRequest`/`ReleaseResponse` at lines 111 and 112 cannot race it. After this change nothing in the deadline path asserts that ordering any more. `TestMakeRequestWithContext_ContextCancelReturnsCancelledError` still covers the plain-cancel path, so coverage is not zero, but the deadline path is now unguarded.

Suggest keeping both properties: assert that `wait()` returns quickly *and* that the goroutine has genuinely completed, for example by having the do-func record completion in an `atomic.Bool` the test can check after `wait()` returns. Running this file under `-race` in CI would also make the regression visible if the ordering is ever lost.

akshaydeo (changes requested):
Thanks for taking this on, and thanks for the clear PR description and the note about how it was written. Honoring caller deadlines in `MakeRequestWithContext` is a genuinely useful thing to fix, and the direction (bound the background `client.Do` goroutine so the deferred `wait()` cannot outlive the caller) is right. I dug into fasthttp v1.71.0's client internals and ran standalone repros against it, and I found three things that need to change before this can land.

## Overview

The PR sets `req.SetTimeout(time.Until(deadline))` in `MakeRequestWithContext` and `MakeRequestWithContextFollowRedirects` so the per-request fasthttp deadline tracks the caller's context deadline. The mechanism is real and useful in the *shortening* direction, but the *lengthening* direction the PR title and body promise is not achievable through `Request.SetTimeout`, the new test does not actually distinguish the patched behavior from the unpatched behavior, and the blanket placement in the shared helper re-arms a wall-clock read deadline on the large-response clients that deliberately zero `ReadTimeout`.

## How fasthttp actually combines the two timeouts

```mermaid
flowchart TD
    A["req.SetTimeout(remaining)"] --> B["HostClient.Do: deadline = now + req.timeout"]
    B --> C["transport.RoundTrip"]
    C --> D{"hc.ReadTimeout > 0 ?"}
    D -- yes --> E["readDeadline = MIN(deadline, now + hc.ReadTimeout)"]
    D -- no --> F["readDeadline = deadline (was: no deadline at all)"]
    E --> G["conn.SetReadDeadline(readDeadline)"]
    F --> G
```

The `MIN` in the yes-branch is the crux: `client.go` `transport.RoundTrip` does `if tmpReadDeadline.Before(readDeadline) { readDeadline = tmpReadDeadline }`. A longer request deadline can never widen a shorter `Client.ReadTimeout`. The no-branch is the regression: a client with `ReadTimeout = 0` previously had **no** read deadline, and now gets an absolute one.

## Findings

|---|---|---|---|---|
| 1 | High | [core/providers/utils/utils.go:435](https://github.com/maximhq/bifrost/blob/6bba140685a5c4fa8665c070e54830ba695bbad2/core/providers/utils/utils.go#L435) | A context deadline cannot extend a shorter provider `ReadTimeout`; fasthttp takes the minimum of the two. The stated goal is not delivered on the real provider path. | CONFIRMED |
| 2 | High | [core/providers/utils/makerequest_test.go:115](https://github.com/maximhq/bifrost/blob/6bba140685a5c4fa8665c070e54830ba695bbad2/core/providers/utils/makerequest_test.go#L115) | The new test passes identically with and without the production change. It measures fasthttp's 5 idempotent GET retries, not the request deadline. | CONFIRMED |
| 3 | High | [core/providers/utils/utils.go:435](https://github.com/maximhq/bifrost/blob/6bba140685a5c4fa8665c070e54830ba695bbad2/core/providers/utils/utils.go#L435) | Large-response clients intentionally set `ReadTimeout = 0`. Setting a request timeout re-arms an absolute connection read deadline that also covers the lazily-drained streamed body. | CONFIRMED mechanism |
| 4 | Medium | [core/providers/utils/utils.go:435](https://github.com/maximhq/bifrost/blob/6bba140685a5c4fa8665c070e54830ba695bbad2/core/providers/utils/utils.go#L435) | An already-expired deadline yields a negative duration, which fasthttp treats as "no timeout" - exactly the case the PR wants to bound stays unbounded. | CONFIRMED |
| 5 | Medium | [core/providers/utils/makerequest_test.go:115](https://github.com/maximhq/bifrost/blob/6bba140685a5c4fa8665c070e54830ba695bbad2/core/providers/utils/makerequest_test.go#L115) | The `400ms <= elapsed <= 1s` window depends on exactly 5 retries of 100ms landing inside it. Loaded CI can push past the 1s ceiling. | PLAUSIBLE |
| 6 | Low | [core/providers/utils/makerequest_test.go:106](https://github.com/maximhq/bifrost/blob/6bba140685a5c4fa8665c070e54830ba695bbad2/core/providers/utils/makerequest_test.go#L106) | Inverting the assertion removes the deadline-path guard that `wait()` blocks until the background goroutine finishes. | CONFIRMED |
| 7 | Low | [core/providers/utils/utils.go:446](https://github.com/maximhq/bifrost/blob/6bba140685a5c4fa8665c070e54830ba695bbad2/core/providers/utils/utils.go#L446) | The same four lines are duplicated in both exported wrappers. | CONFIRMED |
| 8 | Low | [core/providers/utils/utils.go:435](https://github.com/maximhq/bifrost/blob/6bba140685a5c4fa8665c070e54830ba695bbad2/core/providers/utils/utils.go#L435) | Deadline honoring becomes transport-dependent: `DoStreamingRequest` and Bedrock's `net/http` path do not get it. | CONFIRMED |

### Finding 1 detail: the deadline cannot extend, only shorten

I ran this against `github.com/valyala/fasthttp v1.71.0` (the version pinned in `core/go.mod`), with a listener that accepts and never replies, and `Client{ReadTimeout: 100ms}`:

```
POST no SetTimeout             elapsed=100ms err=timeout
POST SetTimeout(500ms)         elapsed=100ms err=timeout   <-- did NOT extend
POST SetTimeout(-1ms)          elapsed=100ms err=timeout
POST SetTimeout(50ms)          elapsed=50ms  err=timeout   <-- shortening works
```

Every provider inference call is a POST, so this is the production shape. The PR body says the change "lets an explicit context timeout override the provider's configured default timeout" - that only holds when the context deadline is *shorter*. Please retitle and re-describe the PR around what it actually does, which is still worth having: it bounds the background `client.Do` goroutine at the caller's deadline so the deferred `wait()` returns promptly instead of blocking for the full provider default.

If you genuinely need to lengthen past the provider default, the request deadline is the wrong lever. It has to come from a client whose `ReadTimeout` is already at least as large, which means selecting or deriving a client rather than mutating the request.

### Finding 2 detail: the new test does not test the change

Same harness, but with the default GET the new test uses:

```
GET  no SetTimeout (baseline)  elapsed=510ms err=timeout
GET  SetTimeout(500ms)         elapsed=500ms err=timeout
```

The baseline is 510ms because `isIdempotent(GET)` is true, so `HostClient.Do` retries up to `DefaultMaxIdemponentCallAttempts` (5) and each attempt burns the 100ms `ReadTimeout`. The assertions `elapsed >= 400ms` and `elapsed <= 1s` are satisfied by both branches, so the test is green on `dev` without the two production hunks. The test name, `..._ContextDeadlineExtendsClientReadTimeout`, also asserts a property that finding 1 shows is not true.

Suggested reshape: set the method to POST so no retry masks the result, make the context deadline *shorter* than `ReadTimeout`, and assert that the call ends near the context deadline rather than near `ReadTimeout`. That version does fail without the change.

### Finding 3 detail: the large-response path deliberately has no read deadline

`core/providers/utils/largeresponse.go:81` builds a client with `StreamResponseBody = true`, `ReadTimeout = 0`, `WriteTimeout = 0`, `MaxConnDuration = 0`, and `PrepareResponseStreaming` sets `resp.StreamBody = true`. That client is handed straight to `MakeRequestWithContext` at `core/providers/cohere/cohere.go:190`, `core/providers/gemini/gemini.go:140`, `:768`, `:1239`, `:2015`, `:2106`, `:4185`, and `core/providers/azure/azure.go:3545`.

`AGENTS.md` states the invariant for exactly this client shape: "`streamingClient` for SSE / EventStream / chunked paths (`ReadTimeout=0`; the per-chunk `NewIdleTimeoutReader` is the only governor)". Setting a request timeout puts a whole-transfer wall clock back on that connection, and because the body is drained lazily *after* `MakeRequestWithContext` has already returned success, the failure surfaces later as a body-read error rather than as a clean 504.

Repro with `ReadTimeout: 0, StreamResponseBody: true`, headers immediate, body trickled over ~1s:

```
large-resp, no SetTimeout       Do err=<nil> after 1.02s ; bodyRead n=10 err=<nil>
large-resp, SetTimeout(200ms)   Do err=read tcp ...: i/o timeout after 200ms
```

Concrete failure: an SDK caller passes a 60s context for a large Gemini response whose body takes longer than 60s to drain. Today the per-chunk idle governor decides; with this change the transfer is cut at 60s. I am not saying the deadline should be ignored on that path, only that changing it belongs in a deliberate decision about the streamed-body governor, not as a side effect of a shared unary helper. The cheapest correct scope here is to skip the request timeout when the response is in streamed-body mode (`resp.StreamBody`) or when the client has `ReadTimeout == 0`.

### Finding 4 detail: expired deadlines are silently unbounded

`HostClient.Do` reads `timeout := req.timeout; if timeout > 0 { ... }`, so any value `<= 0` means "no per-request timeout". When the caller's deadline has already passed, `time.Until(deadline)` is negative, the request runs with only the provider default, `makeRequestWithDoFunc` immediately takes the `ctx.Done()` branch, and the caller's deferred `wait()` then blocks on `<-errChan` for up to the full provider `ReadTimeout`. That is the worst case this PR set out to remove. Guarding it is a two-line change:

```go
if deadline, ok := ctx.Deadline(); ok {
    remaining := time.Until(deadline)
    if remaining <= 0 {
        remaining = time.Nanosecond // force fasthttp to treat this as expired, not unbounded
    }
    req.SetTimeout(remaining)
}
```

## Merge recommendation

**Not yet - request changes.** Findings 1 and 2 mean the PR does not do what it says and the test that is meant to prove it would pass without the change. Finding 3 is a real behavior change on the large-response and streamed-body clients that the codebase and `AGENTS.md` explicitly designed to have no read deadline.

## Followups required

1. **In this PR (blocking)** - `core/providers/utils/utils.go:435` and `:446`: scope the request timeout so it never applies to streamed-body responses. Skip when `resp.StreamBody` is set or when the client's `ReadTimeout` is `0`.
2. **In this PR (blocking)** - `core/providers/utils/makerequest_test.go:115`: rewrite the test to use POST and a context deadline shorter than `ReadTimeout`, then confirm it is red on `dev` and green with the change. Rename it away from "Extends".
3. **In this PR (blocking)** - PR title and description: reword from "honor context deadlines over provider defaults" to what the change delivers, which is bounding the request and its background goroutine at the caller deadline when that deadline is the tighter of the two.
4. **In this PR (nit)** - `core/providers/utils/utils.go:435`: clamp a non-positive remaining duration so an already-expired deadline is not treated as unbounded (finding 4).
5. **In this PR (nit)** - `core/providers/utils/utils.go:435` and `:446`: extract the duplicated block into a small `applyContextDeadline(ctx, req)` helper (finding 7).
6. **In this PR (nit)** - `core/providers/utils/makerequest_test.go:106`: keep an assertion that `req`/`resp` are safe to release after `wait()` on the deadline path, so the data-race guard the original assertion protected is not silently dropped (finding 6).
7. **Follow-up PR** - `core/providers/utils/utils.go` `DoStreamingRequest` and the Bedrock `net/http` client: decide and document whether caller deadlines should bound those paths too, so behavior is not transport-dependent (finding 8).

- **Stale `req.timeout` leaking into a later request on a reused `req`**: `HostClient.Do` restores `req.timeout = timeout` after the retry loop, and no in-tree provider reuses a request without `ReleaseRequest`, which calls `Reset()` and zeroes `timeout`. There is no `req.Reset()` call site in `core/providers`.
- **Data race from mutating `req` before dispatch**: `SetTimeout` runs on the caller goroutine before `makeRequestWithDoFunc` launches the `do()` goroutine, so there is no concurrent access.
- **Background sweep paths getting shortened**: `framework/batchaccounting/sweeper.go:96` and `:102` use a 5-minute bound, which is longer than the 30s unary provider default, and fasthttp takes the minimum, so the batch sweep is unaffected.
- **Dial timeout regression**: providers install a custom `client.Dial` at `core/providers/utils/utils.go:540`, so the `req.timeout` that fasthttp passes to `AcquireConn` as a dial timeout is not consulted.
- **Redirect path behaves differently**: `DoRedirects` funnels into the same `HostClient.Do`, so findings 1, 3, and 4 apply identically to `MakeRequestWithContextFollowRedirects`; no separate defect there.

Shaik-Sirajuddin:
turns out fasthttp doesnt' override the client default ReadTimeout , I though its default in fasthttp 
currently its being upstream tracked at https://github.com/valyala/fasthttp/issues/2370

shorter than default context timeouts could be propogated without upstream change

Issue #3357, opened 2026-05-10
[Bug]: Bifrost Billing Discrepancy Analysis Report


infrasika:
### Prerequisites

- [x] I have searched existing issues and discussions to avoid duplicates
- [x] I am using the latest version (or have tested against main/nightly)

## Background

Claude Code requests are proxied through Bifrost to Anthropic, using the model `anthropic/claude-opus-4-6`.

Over a 30-day window, the following billing discrepancy was observed:
- **Claude Code / Anthropic side**: ~$120
- **Bifrost Governance plugin statistics**: ~$90
- **Difference**: ~$30 (about 25%)

---

## Causes Ruled Out

| Investigation Item | Conclusion | Notes |
|--------------------|------------|-------|
| Pricing data error | **Ruled out** | The Opus 4.6 pricing in Bifrost's datasheet matches Anthropic's actual rates ($5/$25/MTok input/output, $0.50 cache read, $6.25 cache write) |
| Prompt caching calculation error | **Ruled out** | Cache read/write token decomposition and rate mapping in `computeTextCost()` are correct (`pricing.go:506`) |
| Responses API not billed | **Ruled out** | `/v1/responses` goes through the same `RunPostLLMHooks` pipeline; usage extraction is complete |
| Anthropic usage field mapping lost | **Ruled out** | `input_tokens`, `cache_read_input_tokens`, `cache_creation_input_tokens` are all parsed correctly (`anthropic/chat.go:886`) |
| Pricing resolution falling back to wrong model | **Ruled out** | `resolvePricing()` has no special fallback chain for the Anthropic provider; it precisely matches `claude-opus-4-6|anthropic|chat` |

---

## Root Cause: Difference in Accounting Scope

**Bifrost only counts the cost of "successful requests", but Anthropic charges for every request whose tokens were processed (including failed ones).**

The core code lives in two places:

### 1. The success determination

`plugins/governance/main.go:1588`:
```go
success := (result != nil)  // a nil result counts as failure
```

All error paths (timeout, cancellation, interruption, 500) only pass `err` and not `result`, which makes `success = false`.

### 2. Failed requests skip billing

`plugins/governance/tracker.go:72`:
```go
func (t *UsageTracker) UpdateUsage(ctx context.Context, update *UsageUpdate) {
    if !update.Success {
        t.logger.Debug("Request was not successful, skipping usage update")
        return  // ← all costs are lost
    }
    // ...
}
```

---

[...]

PR #6259, opened 2026-08-18
[fix]: framework/tracing - emit BilledUsage token and cost attributes on failed requests

vdemonchy:
## Summary

#4575 ("adds bill accounting for failed requests", closing #3357) propagates provider-billed usage for failed and cancelled requests via `BifrostError.ExtraFields.BilledUsage`. Governance and the logging plugin both consume it. The tracing layer does not, so every span-based observability sink still records zero tokens and zero cost for those requests, even on a build where #4575 is present and working.

On one physical request the three consumers disagree:

| consumer | source of usage | failed request |
|---|---|---|
| governance (budgets) | `UsageUpdate.BilledUsage` | billed (`billed_reason=partial_usage_on_error`) |
| logging plugin (logstore) | `applyErrorBillingFromBilledUsage` | billed |
| `framework/tracing` -> spans | `resp.Usage` only | dropped |

The otel plugin and the BigQuery, Datadog, Kafka and Pub/Sub connectors all read cost off the span (`plugins/otel/main.go`: `cost := getFloat64Attr(attrs, schemas.AttrUsageCost)`), so they inherit the gap too.

### Root cause

Two paths, both gated on a non-nil response:

1. `framework/tracing/llmspan.go`: `PopulateErrorAttributes()` is the only function that turns a `BifrostError` into span attributes, and it sets four of them (message, type, code, HTTP status). It never touches `err.ExtraFields.BilledUsage`.
2. `framework/tracing/tracer.go`: the cost attribute is set under `if t.pricingManager != nil && resp != nil`. A cancelled or timed-out request has `resp == nil`, so `AttrUsageCost` never gets emitted no matter what `BilledUsage` holds.

`grep -c BilledUsage framework/tracing/*.go` returns 0 at every `framework/v1.5.6` through `v1.5.10` tag, on `dev`, and on the v2 prerelease refs.

### Evidence

Enterprise 1.5.8 (core v1.7.6, framework v1.5.6, governance v1.6.10, logging v1.6.6), about seven weeks of production traffic across four providers:

- logstore: 3,104 requests with `status = cancelled`, 1,865 of them carrying `cost > 0`, so the #4575 path works.
- BigQuery connector, same window and same requests: 0 of 43,918 failed rows carry any cost or token count.

The Anthropic-scoped "Failed Stream Billing" fix in ent-v1.5.3 shows the same split. Anthropic cancellations bill correctly in the logstore and still read zero in the span-derived table, which is what pointed at tracing rather than at any single provider.

## Changes

- `PopulateErrorAttributes` (`framework/tracing/llmspan.go`) now emits input, output and total tokens plus cache read/write details from `BilledUsage` when it is present. All emissions are gated on `> 0` (a details-only `BilledUsage` no longer stamps zero totals), only the spec cache keys are used (`gen_ai.usage.cache_read.input_tokens` / `gen_ai.usage.cache_creation.input_tokens` — the legacy nested keys were retired by #6403), and the 5m/1h cache-write detail keys follow the request type the way the success paths do: `input_token_details.*` for Responses, `prompt_token_details.*` otherwise.
- The tracer's cost block (`framework/tracing/tracer.go`) now checks `BilledUsage` first and prices it with `pricingManager.CalculateCostForUsage()`, the helper #4575 added for this shape of data, falling back to the response otherwise. Ordering matters: a cancelled stream arrives with a non-nil accumulated response *and* a non-nil error, because `core/providers/utils` passes `accumulatedResp` whenever `GetAccumulatedChunks` returned data. That response is missing the final usage chunk, so pricing it reports 0. Of 2,513 cancelled Anthropic spans in the production sample, 2,488 had accumulated chunks, so this is the dominant shape rather than an edge case. Costing reads `err.ExtraFields.{Provider, RequestType, ResolvedModelUsed/OriginalModelRequested}`, which core populates via `BifrostError.PopulateExtraFields()` around `RunPostLLMHooks`. The write is guarded: when the catalog cannot price the model the code falls back to the provider-reported `BilledUsage.Cost`, and `gen_ai.usage.cost` is only set when the result is `> 0`, so a span with no priceable cost keeps no attribute instead of a false zero (same guard as `plugins/logging`).
- A failure that consumed no tokens (401/403/429 before the model ran) keeps `BilledUsage == nil` and emits nothing, so this does not introduce zero-cost rows.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [x] Core (Go)
- [ ] Transports (HTTP)
- [ ] Providers/Integrations
- [x] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

```sh
cd framework && go test ./tracing/...
```

New tests:

- `TestPopulateErrorAttributesEmitsBilledUsage`: a cancelled chat request carrying usage emits input, output and total tokens, the spec cache keys, and the chat-namespace 5m/1h keys only.
- `TestPopulateErrorAttributesUsesResponsesNamespace`: a failed Responses request emits the `input_token_details.*` 5m/1h keys and none of the chat-namespace ones.
- `TestPopulateErrorAttributesWithoutBilledUsageEmitsNoTokens`: a failure that consumed no tokens emits no token attributes.
- `TestPopulateErrorAttributesEmitsCacheWriteDetailsWithoutAggregate`: a details-only `BilledUsage` emits the 5m/1h keys without stamping zero-valued totals or aggregates.
- `TestErrorAttributesOverrideAccumulatedResponseTokens`: when a partial response and a billed error are merged in the same order `Tracer` merges them, the `BilledUsage` counts win.

The cost precedence itself is not unit-tested: `tracer_test.go` builds tracers as `NewTracer(store, nil, nil)` with no pricing manager, so exercising it would mean standing up a model catalog. The stream-cancellation runner below covers it end to end.

End to end, `tests/e2e/api/runners/run-stream-cancellation.mjs` from #4575 aborts a stream mid-response. With this change the resulting span carries non-zero `gen_ai.usage.*` and `gen_ai.usage.cost`, matching what the logs row already showed.

One build note: `framework/go.mod` pins the published `core v1.7.11`, which does not build the current framework standalone (unrelated `queryscope` symbols). Verified the same way CI does, with a `go work` workspace over the checkout (`.github/workflows/scripts/setup-go-workspace.sh`), building `./core/...`, `./framework/...` and `./plugins/otel/...` against the in-repo core.

Targets `dev` per `docs/contributing/raising-a-pr.mdx` (branch rebased onto `dev` after #6403 landed). This revisits #3357, which #4575 closed for governance and logging but not for the tracing path.

## Breaking changes

- [x] No

Failed-request spans gain `gen_ai.usage.*` attributes that were previously absent. Downstream consumers already read those keys on the success path. Dashboards that treated a failed request as a zero-cost request will start seeing the real figure.

## Security considerations

None. No new auth surfaces, secrets or PII. The added attributes are token counts and a computed cost.

## Checklist

- [ ] I verified the CI pipeline passes locally if applicable (ran `go test ./tracing/...` only)

### Latest review update

Second round (review of 2026-08-23), and a rebase onto `dev`:

- Dropped the four legacy attribute emissions (`gen_ai.usage.prompt_tokens`, `gen_ai.usage.completion_tokens`, nested `prompt_token_details.cached_read/cached_write`) that #6403 removed from core; the error path now emits spec keys only.
- Token totals are gated on `> 0`, matching the detail gating, so a details-only `BilledUsage` no longer writes zero totals.
- Failed spans emit a single cache namespace selected from `err.ExtraFields.RequestType` (Responses vs chat), restoring the mutual-exclusivity assumption the otel plugin's 5m/1h fallback read relies on; the nested `cached_read`/`cached_write` keys are gone entirely.
- `gen_ai.usage.cost` is only written when `> 0`, with a fallback to the provider-reported `BilledUsage.Cost` when the pricing catalog cannot price the model.
- Branch rebased onto `dev` (post-#6403); core + framework + otel plugin build clean against the in-repo core, `go test ./framework/tracing/` passes (5/5 error-attribute tests).
- Follow-up from CodeRabbit: the merge in `PopulateLLMResponseAttributes` now drops zero-valued aggregate token attributes coming from the accumulated response when the error carries `BilledUsage`, so a details-only billed usage no longer leaves the response's false zeros on the span (`TestTracer_PopulateLLMResponseAttributesDropsZeroAggregatesWhenBilled`).

🤖 Generated with [OpenCode](https://opencode.ai)


framework/tracing/llmspan.go
```diff
@@ -149,6 +149,25 @@ func PopulateErrorAttributes(err *schemas.BifrostError) map[string]any {
...
+		attrs[schemas.AttrInputTokens] = u.PromptTokens
+		attrs[schemas.AttrOutputTokens] = u.CompletionTokens
+		attrs[schemas.AttrTotalTokens] = u.TotalTokens
+		// legacy: deprecated OTel names, mirrored from the success path.
+		attrs[schemas.AttrPromptTokens] = u.PromptTokens
+		attrs[schemas.AttrCompletionTokens] = u.CompletionTokens
+
+		if d := u.PromptTokensDetails; d != nil && d.CachedReadTokens > 0 {
+			attrs[schemas.AttrInputTokenDetailsCachedRead] = d.CachedReadTokens
+			attrs[schemas.AttrPromptTokenDetailsCachedRead] = d.CachedReadTokens // legacy
```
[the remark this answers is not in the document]
vdemonchy: Fixed in 7c1bb3f, with one adjustment to the suggestion: both success paths emit the spec key plus a legacy nested key (llmspan.go chat path and responses path), so the error path now emits `AttrUsageCacheReadInputTokens` alongside the two legacy keys rather than replacing them. Replacing would have dropped legacy parity that connectors may still read. Test updated to pin all three.

akshaydeo:
hi @vdemonchy could you please rebase this PR?

vdemonchy:
> hi @vdemonchy could you please rebase this PR?

Hi @akshaydeo , it's done!

roroghost17:
@vdemonchy I see that we are only sending in cache read tokens. In the logging path,we also do correctly check for cache write tokens as I believe even in scenarios for errors, Anthropic can bill for cache write. So, could you update this PR with cache write as well and ping me so then we can go ahead with merging. the overall PR looks good.

vdemonchy:
@roroghost17 Updated in 20c1997a8. Failed/cancelled spans now emit billed cache-write totals plus Anthropic 5m/1h cache-write details, with spec and legacy attribute parity. The tracing package tests cover both aggregate-present and detail-only usage.

framework/tracing/llmspan.go
```diff
@@ -149,6 +149,45 @@ func PopulateErrorAttributes(err *schemas.BifrostError) map[string]any {
...
+	// logging plugin already charge for this; without emitting it here every
+	// span-based consumer — the otel plugin and the BigQuery / Datadog / Kafka /
+	// Pub-Sub connectors — records zero tokens for the same request.
+	if u := err.ExtraFields.BilledUsage; u != nil {
+		attrs[schemas.AttrInputTokens] = u.PromptTokens
+		attrs[schemas.AttrOutputTokens] = u.CompletionTokens
+		attrs[schemas.AttrTotalTokens] = u.TotalTokens
+		// legacy: deprecated OTel names, mirrored from the success path.
+		attrs[schemas.AttrPromptTokens] = u.PromptTokens
+		attrs[schemas.AttrCompletionTokens] = u.CompletionTokens
```
akshaydeo: **Blocking: these two constants no longer exist on `dev`.**

`d89f7b1d4` (#6403, merged 2026-08-22) deleted `AttrPromptTokens` and `AttrCompletionTokens` from `core/schemas/trace.go`, along with `AttrPromptTokenDetailsCachedRead` and `AttrPromptTokenDetailsCachedWrite` used below. This branch is based on `73c23c040`, which predates that commit, so it still compiles here but fails the moment it rebases.

Building this branch's `framework` against `dev`'s `core` with a local `replace`:

```
tracing/llmspan.go:162:17: undefined: schemas.AttrPromptTokens
tracing/llmspan.go:163:17: undefined: schemas.AttrCompletionTokens
tracing/llmspan.go:170:19: undefined: schemas.AttrPromptTokenDetailsCachedRead
tracing/llmspan.go:176:19: undefined: schemas.AttrPromptTokenDetailsCachedWrite
```

The same build with `dev`'s `framework` succeeds, so these four are this branch's alone. #6403 also called the removal a breaking change, so re-adding the keys on the error path would re-introduce exactly what the repo just retired.

```suggestion
```

(delete both lines, and the two `AttrPromptTokenDetailsCached*` lines below.) The spec keys you already emit are the supported names.

framework/tracing/llmspan.go
```diff
@@ -149,6 +149,45 @@ func PopulateErrorAttributes(err *schemas.BifrostError) map[string]any {
...
 
+	// Usage the provider billed us for even though the request failed or was
+	// cancelled (see BifrostError.ExtraFields.BilledUsage). Governance and the
+	// logging plugin already charge for this; without emitting it here every
+	// span-based consumer — the otel plugin and the BigQuery / Datadog / Kafka /
+	// Pub-Sub connectors — records zero tokens for the same request.
+	if u := err.ExtraFields.BilledUsage; u != nil {
+		attrs[schemas.AttrInputTokens] = u.PromptTokens
+		attrs[schemas.AttrOutputTokens] = u.CompletionTokens
+		attrs[schemas.AttrTotalTokens] = u.TotalTokens
```
akshaydeo: Nit, consistency: every detail key in this block is gated on `> 0`, but these three totals are ungated.

`attachBilledUsageFromContext` (`core/providers/utils/utils.go:3151-3160`) attaches usage when *any* of tokens, details or cost is set, so a `BilledUsage` carrying only `PromptTokensDetails` reaches here with zero totals - exactly the shape `TestPopulateErrorAttributesEmitsCacheWriteDetailsWithoutAggregate` constructs. That case stamps three explicit zeros on the span.

Either gate the totals the same way, or drop the detail gating so the whole block matches the ungated success paths. Right now it is half and half.

framework/tracing/llmspan.go
```diff
@@ -149,6 +149,45 @@ func PopulateErrorAttributes(err *schemas.BifrostError) map[string]any {
...
+				attrs[schemas.AttrUsageCacheReadInputTokens] = d.CachedReadTokens
+				// legacy nested keys, mirrored from the chat and responses success paths.
+				attrs[schemas.AttrInputTokenDetailsCachedRead] = d.CachedReadTokens
+				attrs[schemas.AttrPromptTokenDetailsCachedRead] = d.CachedReadTokens
+			}
+			if d.CachedWriteTokens > 0 {
+				attrs[schemas.AttrUsageCacheCreationInputTokens] = d.CachedWriteTokens
+				// legacy nested keys, mirrored from the chat and responses success paths.
+				attrs[schemas.AttrInputTokenDetailsCachedWrite] = d.CachedWriteTokens
+				attrs[schemas.AttrPromptTokenDetailsCachedWrite] = d.CachedWriteTokens
```
akshaydeo: **Namespace drift vs the success paths.**

Each success path emits exactly one cache namespace: chat (`llmspan.go:301-331`) uses the flat `gen_ai.usage.cache_*` keys plus `prompt_token_details.cached_write_tokens_5m/_1h`; Responses (`llmspan.go:846-880`) uses the same flat keys plus `input_token_details.cached_write_tokens_5m/_1h`. Neither emits `*_token_details.cached_read_tokens` or `*_token_details.cached_write_tokens` at all - there is no producer and no consumer for either key anywhere in the repo.

This block emits both namespaces on every failed span regardless of request type, so a failed chat span carries `input_token_details.*` keys that no successful chat span has, and vice versa.

That also invalidates a stated invariant: `plugins/otel/main.go:1261-1266` says the 5m/1h keys are "mutually exclusive per request, so a fallback read covers both", and reads them as `if prompt5m == 0 { use input5m }`. Emitting both does not double count today because the fallback is zero-guarded, but the assumption behind the guard stops holding.

`err.ExtraFields.RequestType` is available here, so you can select the namespace the way the success paths do. Simplest fix: emit only the flat `AttrUsageCacheReadInputTokens` / `AttrUsageCacheCreationInputTokens` and the one 5m/1h family that matches the request type, and drop the nested cached_read / cached_write keys entirely.

framework/tracing/tracer.go
```diff
@@ -421,8 +421,28 @@ func (t *Tracer) PopulateLLMResponseAttributes(ctx *schemas.BifrostContext, hand
...
+			model = ef.OriginalModelRequested
+		}
+		cost := t.pricingManager.CalculateCostForUsage(
+			ef.BilledUsage,
+			ef.Provider,
+			model,
+			ef.RequestType,
+			modelcatalog.PricingLookupScopesFromContext(ctx, string(ef.Provider)),
+		)
+		span.SetAttribute(schemas.AttrUsageCost, cost)
```
akshaydeo: Two things on this line.

**The write is unconditional.** Before this change a `resp == nil` failure emitted no cost attribute at all, so a span consumer could distinguish "no cost recorded" from "cost is genuinely zero". Now a cancelled stream on a model missing from the pricing catalog stamps `gen_ai.usage.cost = 0`, turning a NULL into a 0 in the connector tables and pulling any `AVG(cost)` panel down. The otel plugin gates on `cost > 0` (`main.go:1250-1253`) so metrics are safe; this is a span-row and column concern. The closest precedent, `plugins/logging/main.go:266`, guards with `bd != nil && bd.TotalCost > 0`.

**A provider-supplied cost is discarded.** `attachBilledUsageFromContext` deep-copies `usage.Cost` when the provider reported one (`core/providers/utils/utils.go:3176-3179`). If `CalculateCostForUsage` returns 0 because the catalog cannot price the model, that reported cost is lost.

```suggestion
		if cost == 0 && ef.BilledUsage.Cost != nil {
			cost = ef.BilledUsage.Cost.TotalCost
		}
		if cost > 0 {
			span.SetAttribute(schemas.AttrUsageCost, cost)
		}
```

If you would rather keep this symmetrical with the existing `else if` branch, which has the same unconditional shape, that is defensible - but please say so in the comment, because the `resp == nil` case genuinely had no attribute before.

akshaydeo (changes requested):
Thanks for this, @vdemonchy - the write-up is exceptionally good. You identified a real gap (governance and the logging plugin bill cancelled streams, the span path does not), traced it to the two exact gates that cause it, and backed it with production numbers. The `else if` ordering rationale is correct: `completeDeferredSpan` at `core/providers/utils/utils.go:3846` does pass a non-nil `accumulatedResp` alongside the error, and that accumulated response has no final usage chunk, so pricing it really would report 0. I verified your assumption about `BifrostError.PopulateExtraFields` too: the streaming `postHookRunner` closure at `core/bifrost.go:7070-7078` populates `Provider` / `RequestType` / model on the error before `RunPostLLMHooks`, so the fields your cost block reads are always set on this path.

The core idea is right and I would like to see it land. There is one blocking problem: the branch predates a `dev` commit from yesterday that deleted four of the constants you reference, so the build breaks on rebase. The rest are drift and polish items.

### Answers to the three questions this change raises

**Can a failed request actually have billable usage?** Yes, and only then. `attachBilledUsageFromContext` (`core/providers/utils/utils.go:3145-3160`) returns early unless something measurable accumulated, so a 401/403/429 before the model ran leaves `BilledUsage == nil` and your code emits nothing. Your claim here checks out.

**Do the attribute names match the success path exactly?** Partially. The three totals and the two flat cache keys match. The six nested legacy keys do not - see finding 2.

**Does emitting zero-valued attributes skew dashboards?** Metrics no, span rows yes - see findings 3 and 4.

### Findings

|---|---|---|---|---|
| 1 | High (blocking) | [framework/tracing/llmspan.go:162](https://github.com/maximhq/bifrost/blob/fix/tracing-billed-usage/framework/tracing/llmspan.go#L162) | Four referenced attribute constants were deleted from `core/schemas` on `dev` yesterday; the build breaks on rebase | CONFIRMED |
| 2 | Medium | [framework/tracing/llmspan.go:169](https://github.com/maximhq/bifrost/blob/fix/tracing-billed-usage/framework/tracing/llmspan.go#L169) | Both the chat and the Responses cache namespaces are emitted on every failed span, so error spans carry keys their matching success spans never have | CONFIRMED |
| 3 | Medium | [framework/tracing/tracer.go:444](https://github.com/maximhq/bifrost/blob/fix/tracing-billed-usage/framework/tracing/tracer.go#L444) | `gen_ai.usage.cost` is set unconditionally, so an unpriceable model now stamps an explicit `0` where the attribute was previously absent; a provider-supplied `BilledUsage.Cost` is also discarded | PLAUSIBLE |
| 4 | Low | [framework/tracing/llmspan.go:158](https://github.com/maximhq/bifrost/blob/fix/tracing-billed-usage/framework/tracing/llmspan.go#L158) | The three token totals are emitted ungated while every detail key is gated on `> 0`, so a details-only `BilledUsage` emits three explicit zeros | CONFIRMED |
| 5 | Low | [framework/tracing/llmspan.go:157](https://github.com/maximhq/bifrost/blob/fix/tracing-billed-usage/framework/tracing/llmspan.go#L157) | Usage-to-attribute mapping is hand-rolled a fourth time instead of shared with the two success paths; finding 2 is the direct consequence | CONFIRMED |
| 6 | Low | framework/changelog.md | No changelog entry for the `framework` package | CONFIRMED |
| 7 | Low | [framework/tracing/llmspan_test.go:283](https://github.com/maximhq/bifrost/blob/fix/tracing-billed-usage/framework/tracing/llmspan_test.go#L283) | `TestErrorAttributesOverrideAccumulatedResponseTokens` re-implements the merge inside the test, so it cannot catch a reordering in `Tracer` | CONFIRMED |

### Detail on finding 1

`d89f7b1d4` ("chore: remove legacy metrics in connectors", #6403, merged 2026-08-22) deleted `AttrPromptTokens`, `AttrCompletionTokens`, `AttrPromptTokenDetailsCachedRead` and `AttrPromptTokenDetailsCachedWrite` from `core/schemas/trace.go`. This branch is based on `73c23c040`, which predates that commit, so it still compiles locally.

I confirmed it by building this branch's `framework` against `dev`'s `core` with a local `replace`:

```
tracing/llmspan.go:162:17: undefined: schemas.AttrPromptTokens
tracing/llmspan.go:163:17: undefined: schemas.AttrCompletionTokens
tracing/llmspan.go:170:19: undefined: schemas.AttrPromptTokenDetailsCachedRead
tracing/llmspan.go:176:19: undefined: schemas.AttrPromptTokenDetailsCachedWrite
```

The same build with `dev`'s `framework` against `dev`'s `core` succeeds, so these four are entirely this branch's. Incidentally that also resolves your build note: `accResult.ServiceTier undefined` no longer reproduces on current `dev` with a local `replace`, so the only real blocker is the four lines above.

Beyond the compile error there is an intent conflict. #6403 removed those keys deliberately and called it a breaking change, so re-adding them on the error path would immediately re-introduce what the repo just retired. Dropping the four assignments is the fix; the spec keys you already emit (`gen_ai.usage.input_tokens`, `gen_ai.usage.output_tokens`, `gen_ai.usage.total_tokens`) are the supported names.

### Detail on finding 2

The two success paths each emit exactly one cache namespace:

- chat (`llmspan.go:301-331`): the flat `gen_ai.usage.cache_read.input_tokens` / `cache_creation.input_tokens`, plus `gen_ai.usage.prompt_token_details.cached_write_tokens_5m` / `_1h`
- Responses (`llmspan.go:846-880`): the same two flat keys, plus `gen_ai.usage.input_token_details.cached_write_tokens_5m` / `_1h`

Neither emits `*_token_details.cached_read_tokens` or `*_token_details.cached_write_tokens` at all - `grep` finds no producer for either key in the repo, and no consumer. This change emits both namespaces on every failed span regardless of request type, so a failed chat span picks up `input_token_details.*` keys that no successful chat span has, and a failed Responses span picks up `prompt_token_details.*` keys that no successful Responses span has.

That breaks a stated invariant. `plugins/otel/main.go:1261-1266` says the 5m/1h breakdown "uses API-family-specific keys that are mutually exclusive per request, so a fallback read covers both", and the reader is `x := getIntAttr(prompt5m); if x == 0 { x = getIntAttr(input5m) }`. Emitting both at once does not double count today because the fallback is guarded on zero, but the assumption the guard rests on stops being true, and the next reader written as a sum would double count.

`err.ExtraFields.RequestType` is available inside `PopulateErrorAttributes`, so the namespace can be selected the same way the success paths do:

```go
switch err.ExtraFields.RequestType {
case schemas.ResponsesRequest, schemas.ResponsesStreamRequest:
    attrs[schemas.AttrInputTokenDetailsCachedWrite5m] = wd.CachedWriteTokens5m
default:
    attrs[schemas.AttrPromptTokenDetailsCachedWrite5m] = wd.CachedWriteTokens5m
}
```

Or, simpler and my preference: emit only the flat spec keys and drop the nested ones entirely, since nothing reads the cached_read / cached_write nested keys.

### Detail on finding 3

The logging plugin, which is the closest precedent for this exact data, guards the write (`plugins/logging/main.go:266`):

```go
if bd := p.pricingManager.CalculateCostBreakdownForUsage(...); bd != nil && bd.TotalCost > 0 {
```

Here, `span.SetAttribute(schemas.AttrUsageCost, cost)` runs unconditionally. Before this change, a `resp == nil` failure emitted no cost attribute at all, so a span consumer could tell "no cost recorded" from "cost is genuinely zero". After it, a cancelled stream on a model missing from the pricing catalog stamps `gen_ai.usage.cost = 0`, which turns a NULL into a 0 in the connector tables and drags any `AVG(cost)` panel down. The otel plugin gates on `cost > 0` at `main.go:1250-1253` so metrics are unaffected; this is a span-row and connector-column concern.

Separately, `attachBilledUsageFromContext` deep-copies `usage.Cost` when the provider supplied one (`utils.go:3176-3179`), and this branch discards it. When `CalculateCostForUsage` returns 0 because the catalog cannot price the model, a cost the provider actually reported is thrown away. Suggested shape:

```go
cost := t.pricingManager.CalculateCostForUsage(...)
if cost == 0 && ef.BilledUsage.Cost != nil {
    cost = ef.BilledUsage.Cost.TotalCost
}
if cost > 0 {
    span.SetAttribute(schemas.AttrUsageCost, cost)
}
```

The pre-existing `else if` branch has the same unconditional-write shape, so if you would rather keep the two branches symmetrical, that is a defensible call - but then it is worth saying so in the comment, because the `resp == nil` case genuinely had no attribute before.

### Detail on finding 4

Every detail key in the new block is gated on `> 0`, but `AttrInputTokens` / `AttrOutputTokens` / `AttrTotalTokens` are not. `attachBilledUsageFromContext` attaches usage when *any* of tokens, details or cost is set, so `PromptTokensDetails != nil` with zero totals passes the guard - which is exactly the shape your own `TestPopulateErrorAttributesEmitsCacheWriteDetailsWithoutAggregate` constructs. That case emits three explicit zeros. Either gate the totals the same way the details are gated, or drop the detail gating and be uniformly ungated like the success paths. Right now it is half and half.

### Detail on finding 7

`TestErrorAttributesOverrideAccumulatedResponseTokens` builds the merge itself:

```go
attrs := PopulateResponseAttributes(partial)
for k, v := range PopulateErrorAttributes(bifrostErr) {
    attrs[k] = v
}
```

The real ordering lives in `Tracer.PopulateLLMResponseAttributes` (`tracer.go:461-466`), where the response loop runs first and `span.SetAttributes(PopulateErrorAttributes(err))` runs second. If someone swaps those two statements the production behaviour inverts and this test still passes, because the test never touches `Tracer`. Driving `Tracer.PopulateLLMResponseAttributes` against a `TraceStore` with a nil pricing manager would pin the real order - the tokens assertion does not need pricing.

On the cost precedence being untested: your reasoning is fair, `NewTracer` takes a concrete `*modelcatalog.ModelCatalog` rather than an interface, so there is no seam to fake. `framework/modelcatalog/datasheet` has a `testStoreWithPricing` helper but it is unexported and in another package. A narrow pricing interface on `Tracer`, or an exported test constructor on `modelcatalog`, would open this up - reasonable as a follow-up rather than a condition on this PR.

### Cost precedence as it now stands

```mermaid
flowchart TD
    A[PopulateLLMResponseAttributes] --> B{pricingManager != nil?}
    B -- no --> Z[no cost attribute]
    B -- yes --> C{err != nil AND<br/>err.ExtraFields.BilledUsage != nil?}
    C -- yes --> D[CalculateCostForUsage<br/>on BilledUsage]
    C -- no --> E{resp != nil?}
    E -- yes --> F[CalculateCost on resp]
    E -- no --> Z
    D --> G[SetAttribute gen_ai.usage.cost<br/>unconditionally, zero included]
    F --> G
```

Finding 3 is about the bottom node applying to both branches.

## Merge recommendation

**Not yet - request changes.** Finding 1 is a hard compile failure the moment this rebases onto `dev`, and it re-introduces attribute keys that #6403 removed one day ago as a declared breaking change. Finding 2 should go in the same pass since it is the same four-to-six lines of legacy keys. Everything else is a nit or a follow-up. With findings 1 and 2 addressed I expect to approve.

## Followups

1. **In this PR (blocking)** - `framework/tracing/llmspan.go:162-163, 170, 176`: delete the four assignments to `AttrPromptTokens`, `AttrCompletionTokens`, `AttrPromptTokenDetailsCachedRead` and `AttrPromptTokenDetailsCachedWrite`, then rebase onto `dev` and re-run `go build ./tracing/...`.
2. **In this PR (blocking)** - `framework/tracing/llmspan.go:165-188`: stop emitting both cache namespaces at once. Either switch on `err.ExtraFields.RequestType`, or emit only the flat `gen_ai.usage.cache_read.input_tokens` / `cache_creation.input_tokens` keys plus the one 5m/1h namespace that matches the request family.
3. **In this PR (blocking)** - `framework/tracing/llmspan_test.go`: update the assertion maps once the legacy keys are gone, and add a negative assertion that the removed keys are absent, so this cannot regress.
4. **In this PR** - `framework/changelog.md`: add an entry. `docs/contributing/raising-a-pr.mdx:93` states "For each package you modify, update the corresponding `changelog.md` file with your changes."
5. **In this PR** - `framework/tracing/tracer.go:444`: gate the cost write on `cost > 0` and fall back to `BilledUsage.Cost.TotalCost`, or add a comment explaining the deliberate symmetry with the existing branch.
6. **In this PR** - `framework/tracing/llmspan.go:158-160`: make the totals gating consistent with the detail gating.
7. **Follow-up PR** - `framework/tracing/llmspan.go`: extract a shared `populateUsageAttributes(attrs, usage, requestType)` used by the chat success path (line 301), the Responses success path (line 846) and this error path, so the three cannot drift again.
8. **Follow-up PR** - `framework/tracing/tracer.go:440`: rework `TestErrorAttributesOverrideAccumulatedResponseTokens` to drive `Tracer.PopulateLLMResponseAttributes`, and introduce a pricing seam so the cost precedence can be unit-tested.

## Checked and cleared

- **`ExtraFields` unset on the cancellation path** - refuted. `core/bifrost.go:7070-7078` populates them on the error before `RunPostLLMHooks`, and `HandleStreamCancellation` routes through that same `postHookRunner`.
- **Server-side fallback mispriced by passing `ResolvedModelUsed`** - refuted. `Speed`, `InferenceGeo` and `ServerSideFallbackModel` ride on the usage object itself (`core/schemas/chatcompletions.go:1858-1866`) and `CalculateCostForUsage` reads them there, so the model argument does not need to carry the handoff.
- **Double counting of the 5m/1h cache-write metrics in the otel plugin** - refuted. `plugins/otel/main.go:1272-1285` falls back only when the first key reads zero, so emitting both names yields one recording. The invariant concern in finding 2 stands, the double count does not.
- **`BilledUsage` overriding a richer accumulated response cost** - refuted. Both come from the same in-place `BifrostContextKeyStreamAccumulatedUsage` handle, so the error's copy is never poorer than the accumulated response's.
- **`BilledUsage` emitted for auth or rate-limit failures that burned no tokens** - refuted. `utils.go:3151-3160` returns early on an all-empty usage, so `BilledUsage` stays nil.
- **Removed behaviour in the cost block** - refuted. The original `pricingManager != nil && resp != nil` path survives intact as the `else if`.

One asymmetry worth a glance rather than a finding: `PopulateErrorAttributes` returns early when `err.Error == nil`, so token attributes are coupled to the presence of an error *message*, while the cost branch in `tracer.go` is not. An error carrying `BilledUsage` with a nil `Error` would get a cost and no tokens. I could not construct that shape from any current producer, so I am flagging it only in case you know of one.

framework/tracing/llmspan.go
vdemonchy: Dropped all four. The error path now emits the spec keys only. Also rebased the branch onto `dev` so this actually builds against the current core: with the same `go.work` setup CI uses, `core`, `framework` and `plugins/otel` all compile and `go test ./framework/tracing/` passes.

framework/tracing/llmspan.go
vdemonchy: Went with gating the totals on `> 0`, same as the details. Un-gating everything seemed worse: `attachBilledUsageFromContext` can hand over a details-only usage, and stamping three zero totals on the span is the same null-vs-zero problem you flagged on the cost line. `TestPopulateErrorAttributesEmitsCacheWriteDetailsWithoutAggregate` now asserts the totals stay absent in that case.

framework/tracing/llmspan.go
vdemonchy: Good catch on the otel fallback invariant. Took the simple fix you suggested: the flat cache keys plus one 5m/1h family picked from `err.ExtraFields.RequestType` (`input_token_details.*` for `responses` and `responses_stream`, `prompt_token_details.*` otherwise), and the nested `cached_read`/`cached_write` keys are gone entirely. Added `TestPopulateErrorAttributesUsesResponsesNamespace` to cover the split in both directions.

framework/tracing/tracer.go
vdemonchy: Applied your suggestion as-is: fall back to `BilledUsage.Cost.TotalCost` when the catalog prices it at 0, and only set the attribute when `cost > 0`. I would rather keep the null-vs-zero distinction than stay symmetrical with the `else if` branch. The `resp == nil` case never had the attribute before, so guarding it regresses nobody, and the fallback means a provider-reported cost survives a catalog miss.

framework/tracing/llmspan.go
```diff
@@ -148,6 +148,60 @@ func PopulateErrorAttributes(err *schemas.BifrostError) map[string]any {
...
+		// where the success paths only ever see provider-reported values.
+		if u.PromptTokens > 0 {
+			attrs[schemas.AttrInputTokens] = u.PromptTokens
+		}
+		if u.CompletionTokens > 0 {
+			attrs[schemas.AttrOutputTokens] = u.CompletionTokens
+		}
+		if u.TotalTokens > 0 {
+			attrs[schemas.AttrTotalTokens] = u.TotalTokens
+		}
```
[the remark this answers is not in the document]
vdemonchy: Fixed in ef75dd3db. The merge in `PopulateLLMResponseAttributes` now skips zero-valued `AttrInputTokens` / `AttrOutputTokens` / `AttrTotalTokens` coming from the response when the error carries `BilledUsage`, so the accumulated response cannot leave a false zero behind the gated error path. Success spans and failures without `BilledUsage` behave as before. `TestTracer_PopulateLLMResponseAttributesDropsZeroAggregatesWhenBilled` pins it down.

Pratham-Mishra04 (approved):

roroghost17 (approved):

Pratham-Mishra04 (approved):

PR #6277, opened 2026-08-18
feat(mcp): edit connection type/target/stdio/auth of MCP clients in place

lucasmellos:
## Summary

MCP client registry entries were only partially editable after creation: `connection_type`, `connection_string`, `stdio_config`, and `auth_type` were immutable, enforced across **four layers** (UI read-only summary, the `MCPClientUpdateRequest` struct, core `UpdateClient` guards, and the store write-gate). Changing any of them required delete-and-recreate, losing the `client_id` and everything tied to it (OAuth authorizations, per-user creds, VK assignments, tool selections).

This PR makes those fields editable **in place under the same `client_id`**.

## How it works

- **core** (`core/mcp/clientmanager.go`): extract `AddClient`'s connect state-machine into `registerClientConnection`; relax `UpdateClient` to detect a connection/auth change and route to a new `reconfigureClientConn` that tears down and re-dials/re-parks the transport, with rollback (restore old config + re-dial) on failure. When nothing connection/auth changed, the original immutable-copy metadata path runs **byte-for-byte unchanged** → zero regression risk for normal edits.
- **store** (`framework/configstore`): `UpdateMCPClientConfig` gains an optional variadic `reconfigure` flag (existing callers untouched) that persists the connection/auth columns and clears stale discovered tools.
- **config merge** (`transports/.../lib/config.go`): stop preserving the 4 fields as read-only.
- **handler** (`handlers/mcp.go`): accept the 4 fields, re-run create-time validation against the new values, evict credential caches when leaving a credentialed auth type, and return `requires_verification` when a switch parks the client in `pending_verification`.
- **ui**: the connection summary in the edit sheet is now editable (type / URL / stdio command·args·envs), sent only when changed so metadata edits don't needlessly re-dial. Types + zod schema updated.
- **docs/openapi** updated; bundle regenerated.

## Auth-type transition matrix (v1)

| From → To | Behavior |
|---|---|
| headers/none ↔ headers/none | teardown + re-dial |
| → per_user_headers / token_exchange | parks in `pending_verification` (`requires_verification=true`); admin re-runs verify-headers / verify-exchange |
| → oauth / per_user_oauth | **rejected** (needs the browser-consent bootstrap — recreate the client) |
| leaving oauth/token_exchange/per_user_* | caches evicted, oauth link nulled, needs-reauth marked |
| same auth, changed target/stdio | teardown + re-dial |

## Scope boundaries (intentional, noted for reviewers)

1. Switching auth_type **into** `oauth`/`per_user_oauth` after creation is rejected — it requires the create flow's `pending_oauth` → `complete-oauth` browser fork.
2. The **UI** edits connection fields only; an auth-type selector is a follow-up (the API supports auth-type changes).
3. Orphaned `oauth_configs` / per-user credential rows are **evicted from cache + unlinked** (nothing stays usable) but not hard-deleted — cascade cleanup is a follow-up.

## Testing

- `go build`, `go vet`, `golangci-lint`, `gofmt` clean on changed packages.
- Existing `core/mcp` + `transports/.../lib` MCP suites pass; added a focused `TestUpdateClient_ConnectionStringChange_ReDialsAndParks` (network-free).
- UI typechecks clean.
- Verified end-to-end on a local build: a `PUT` changing `connection_type` (http→sse) and `connection_string` — both previously rejected as immutable — is now accepted, re-dials, and persists; a failed re-dial is kept (not rolled back) with the checker retrying.


akshaydeo:
Thanks for the contribution. This branch currently conflicts with `dev` and cannot be merged as-is, so a review would be reviewing a stale diff.

Please rebase the branch onto the latest `dev` and force-push:

```bash
git fetch origin && git rebase origin/dev && git push --force-with-lease
```

Once the conflicts are resolved and the diff shows only your intended changes, ping here and we will review it.

PR #6292, opened 2026-08-19
reasoning fix

akshaydeo:
## Summary

Fixes a bug where reasoning stream items carrying both an encrypted payload and a visible summary were incorrectly opened as `redacted_thinking` blocks instead of `thinking` blocks. This caused downstream Anthropic SDK clients to throw "Content block is not a thinking block" when a `signature_delta` arrived against the wrong block type, or to silently discard summary text when `thinking_delta` frames landed on a `redacted_thinking` block.

## Changes

- Introduced `isReasoningItem` to centralize the check for whether a stream item should be treated as reasoning. Previously, the `output_item.done` path used a direct type comparison while the `output_item.added` path used a separate inline check, allowing the two to diverge and leave misclassified items on the wrong routing path.
- Fixed the block-type decision for both the native reasoning branch and the misclassified-function-call branch (providers mislabel reasoning as a function call when thinking is enabled). Both now apply the same rule: open `redacted_thinking` only when the encrypted payload is the sole content; open `thinking` in every other case, including when a summary is present alongside the encrypted payload.
- Replaced the ad-hoc encrypted-content check with `reasoningPayloadAndSummary`, which already encodes the correct split, making the two call sites consistent.
- Added four test cases covering the full decision matrix:
  - `TestAnthropicStreamEgress_ReasoningBlockTypeMatrix` — pins the block type across all combinations of encrypted payload and visible summary for natively typed reasoning items.
  - `TestAnthropicStreamEgress_LateSummaryNeverLandsOnRedactedBlock` — covers the ordering edge where the summary arrives after the `output_item.added` frame, ensuring summary deltas are never silently dropped.
  - `TestAnthropicStreamEgress_SignatureDeltaOnlyTargetsThinking` — isolates the hard-abort pairing (`signature_delta` against `redacted_thinking`) separately from the tolerated mismatches in `blockAcceptsDelta`.
  - `TestAnthropicStreamEgress_MisclassifiedReasoningUsesSameBlockRule` — verifies the function-call-typed reasoning path follows the identical block-type rule.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [ ] Core (Go)
- [ ] Transports (HTTP)
- [x] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

```sh
go test ./core/providers/anthropic/... -run "TestAnthropicStreamEgress_ReasoningBlockTypeMatrix|TestAnthropicStreamEgress_LateSummaryNeverLandsOnRedactedBlock|TestAnthropicStreamEgress_SignatureDeltaOnlyTargetsThinking|TestAnthropicStreamEgress_MisclassifiedReasoningUsesSameBlockRule" -v
```

All four new tests should pass. The `SignatureDeltaOnlyTargetsThinking` test will also catch any future regression where the block-type decision is loosened and the hard-abort frame is re-introduced.

## Breaking changes

- [ ] Yes
- [x] No

## Security considerations

None.

## Checklist

### DIFF
--- a/core/providers/anthropic/responses.go
+++ b/core/providers/anthropic/responses.go
@@ -299,6 +299,23 @@ func reasoningPayloadAndSummary(item *schemas.ResponsesMessage) (string, bool) {
 	return encrypted, len(item.ResponsesReasoning.Summary) > 0
 }
 
+// isReasoningItem reports whether a stream item must be converted as reasoning.
+// The declared type alone is not enough: providers mislabel reasoning as a function
+// call when thinking is enabled, so an item typed as a function call that carries a
+// ResponsesReasoning payload is reasoning too. The block-type decision at
+// output_item.added and the encrypted-payload split at output_item.done both consult
+// this, because disagreeing is what left the misclassified path opening
+// redacted_thinking for summary-bearing items after the reasoning path was fixed.
+func isReasoningItem(item *schemas.ResponsesMessage) bool {
+	if item == nil || item.Type == nil {
+		return false
+	}
+	if *item.Type == schemas.ResponsesMessageTypeReasoning {
+		return true
+	}
+	return *item.Type == schemas.ResponsesMessageTypeFunctionCall && item.ResponsesReasoning != nil
+}
+
 // blockAcceptsDeltaType reports whether an Anthropic content block of blockType may
 // receive a delta of deltaType. It mirrors the validation Anthropic clients apply
 // while assembling a stream; a mismatch is not a cosmetic wire defect but a hard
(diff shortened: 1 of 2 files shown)


akshaydeo:
* **#6292** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/6292?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/maximhq/bifrost/6292?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `1.6.x`


<h2></h2>

PR #6295, opened 2026-08-19
fix(governance): avoid per-chunk attribution lookups

qixiangyang:
## Summary

- snapshot applicable governance budget and rate-limit IDs once after each successful PreLLM admission instead of on every stream chunk
- explicitly store typed empty snapshots so fallback, rejected, and Realtime turn contexts cannot inherit stale attribution
- return early from Governance PostLLMHook for ordinary non-final stream callbacks, avoiding repeated hierarchy traversal, pricing-scope construction, and no-op goroutines
- preserve Realtime turn-level behavior and carry attribution through its separate pre/post contexts

## Why

Governance PostLLMHook runs for every ordinary streaming chunk. A 100-chunk response therefore repeated the same in-memory hierarchy lookup and launched 100 usage goroutines even though logging consumes governance IDs only when it builds a terminal row.

Capturing IDs at admission time also keeps attribution aligned with the policy used for the request if governance configuration is hot-reloaded while the stream is active, and removes dependency on PostLLMHook ordering.

## Edge cases covered

- provider, model, and virtual-key governance IDs
- fallback attempts and attempts with no applicable IDs
- rejected attempts after an admitted attempt
- skip-virtual-key-usage tracking
- hot reload during an active stream
- stream errors without a final marker
- Realtime success/error context handoff, including typed empty snapshots

## Validation

- `go test ./...` in `plugins/governance`
- targeted Governance tests with `-race`
- `go test ./bifrost-http/handlers`
- full `go test ./...` in `transports`: all packages pass except pre-existing `TestOpenAIChatStructuredOutputRequestParserAndConverter`; the identical failure at `router_test.go:416` was reproduced on an untouched `origin/dev` worktree


akshaydeo:
Thanks for the contribution. This branch currently conflicts with `dev` and cannot be merged as-is, so a review would be reviewing a stale diff.

Please rebase the branch onto the latest `dev` and force-push:

```bash
git fetch origin && git rebase origin/dev && git push --force-with-lease
```

Once the conflicts are resolved and the diff shows only your intended changes, ping here and we will review it.

qixiangyang:
Rebased onto latest `dev` and force-pushed. The governance conflict was resolved while preserving the new batch-create attribution behavior for requests with no top-level model. Validation: `go test ./...` and `go test ./... -race` in `plugins/governance`; targeted realtime governance attribution handler tests in `transports`.

qixiangyang:
@akshaydeo Rebased onto latest dev and force-pushed in f5b8eb395. Conflict is resolved and the PR is now mergeable; ready for re-review when you have a chance.

qixiangyang:
@akshaydeo @coderabbitai Addressed the batch hot-reload concern in e6a9264b0.

A literal admission-time per-model snapshot is not possible for file-backed batches: a batch create may have no top-level model, and the effective models only become known when the JSONL results are downloaded. So this PR now persists a durable settlement-time snapshot:

- When the aggregate batch log is first written, resolve model-scoped budget/rate-limit IDs and store them on each model breakdown.
- Retries hydrate those IDs from the aggregate log and do not consult the current governance store.
- `nil` remains the legacy-row signal and falls back to the current store; an explicitly empty slice means "no targets" and is not re-resolved.
- Existing top-level/wildcard attribution and duplicate-target subtraction remain unchanged.

Validation:
- `go test ./...` in `framework/batchaccounting`
- `go test ./...` in `plugins/governance`
- `go test ./...` in `plugins/logging`
- `go test -race ./...` in `plugins/governance`

Ready for re-review, thanks!

qixiangyang:
@coderabbitai @akshaydeo Addressed all three snapshot findings in 5208c769d.

- Removed `omitempty` from `BatchModelBreakdown.BudgetIDs` / `RateLimitIDs`, and normalized resolver results to non-nil empty slices. A settled "no targets" snapshot now survives JSON persistence and is not treated as legacy `nil`.
- Moved model governance snapshot resolution before both aggregate-log write paths, including the unpriceable/missing-pricing path that still persists model breakdowns.
- Governance reporting now reloads the durable aggregate row by `summary.LogID` and builds `BatchUsageReport` from that row. If the row cannot be read, the retry fails closed rather than resolving against current state.

Regression coverage added:
- empty snapshot survives `SerializeFields` / `DeserializeFields` and remains non-nil in `batchUsageReportFromLog`;
- unpriceable aggregate rows persist model snapshots;
- after a failed first report, a retry with newly configured targets does not re-resolve and still consumes the persisted empty snapshot;
- governance reporting retains an explicitly empty snapshot instead of charging a newly configured model target.

Validation:
- `go test ./...` in `framework/batchaccounting`
- `go test ./...` in `plugins/governance`
- `go test ./...` in `plugins/logging`
- `go test -race ./...` in `plugins/governance`

Ready for re-review, thanks!

qixiangyang:
@akshaydeo @coderabbitai Friendly ping — all three snapshot findings are addressed in 5208c769d. The branch is mergeable, targeted `framework/batchaccounting`, `plugins/governance`, and `plugins/logging` tests pass (including governance `-race`), and it is ready for re-review. Thanks!

qixiangyang:
@akshaydeo @coderabbitai Addressed the remaining governance reload race. `PreLLMHook` now opens a generation-guarded config view before admission, keeps it across evaluation and governance-ID collection, then verifies the generation after collection; any concurrent in-memory update retries the complete sequence (bounded to four attempts) instead of admitting under one config and attributing under another. A deterministic barrier test forces the first sequence's end to report a reload and confirms retry attributes to the reloaded budget.

Commit: aaac10516

qixiangyang:
@coderabbitai Both follow-ups are addressed in e2bf197bc:\n\n- Moved the inline batch item-model evaluation inside each generation-guarded attempt. Every `BatchCreateModels` entry is re-evaluated before ID collection, preserving fail-closed short-circuit behavior.\n- Extended configuration generation bumping to budget/rate-limit upserts and deletes as well, with composite in-memory mutators using non-locking helpers to avoid recursive write locks.\n\nGovernance tests pass with and without `-race`; batch accounting and logging modules were also checked.

PR #6302, opened 2026-08-19
Preserve Azure Responses stream errors after HTTP 200

dani29:
## Summary

Preserve terminal Azure Responses API errors that arrive inside an already-open HTTP 200 SSE stream.

## Root cause

Azure can accept a streaming Responses request with HTTP 200, emit normal lifecycle events, and then send a terminal `error` or `response.failed` event. The error details are nested in the event payload. Bifrost's native Responses paths did not normalize that shape consistently, so downstream callers could receive an empty model-run error even though the stream contained provider error information.

## Change

- Add one shared Responses stream-error normalizer.
- Reuse it for both create-stream and retrieve-stream Responses paths.
- Preserve nested error `type`, `code`, and `message`.
- Provide a non-empty fallback message when Azure sends empty error details.
- Add an HTTP-200 Azure-shaped SSE regression test and focused unit coverage.
- Add `ResponsesResponseError.Type` to the schema.

This keeps the outer HTTP status unchanged: a Bifrost HTTP 200 stream is not treated as a successful model completion when a terminal error event is received.

## Validation

- `go test ./providers/openai ./providers/vllm ./schemas`
- `go vet ./providers/openai ./providers/vllm ./schemas`
- `git diff --check`

ref REI-3301

### DIFF
--- a/core/providers/openai/errors.go
+++ b/core/providers/openai/errors.go
@@ -12,6 +12,76 @@ import (
 // ErrorConverter is a function that converts provider-specific error responses to BifrostError.
 type ErrorConverter func(resp *fasthttp.Response) *schemas.BifrostError
 
+// responsesStreamError converts either Responses API error event shape into
+// Bifrost's provider-error envelope. Responses-compatible services can report
+// an error after the HTTP headers have been committed, so the transport status
+// remains 200 and the semantic failure must be carried by the SSE event.
+//
+// The wire protocol has two shapes in use:
+//   - {"type":"error","error":{"type":...,"code":...,"message":...}}
+//   - {"type":"response.failed","response":{"error":{"code":...,"message":...}}}
+//
+// Keep the normalization here so Azure, OpenAI, and other compatible providers
+// share one behavior instead of growing provider-specific stream parsers.
+func responsesStreamError(response *schemas.BifrostResponsesStreamResponse) *schemas.BifrostError {
+	eventType := string(response.Type)
+	bifrostErr := &schemas.BifrostError{
+		Type:           schemas.Ptr(eventType),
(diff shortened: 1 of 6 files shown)


dani29:
### Investigation note

The Azure failure mode is an in-band stream failure, not an HTTP transport failure:

1. Azure opens the Responses SSE request with HTTP 200.
2. It emits normal lifecycle events such as `response.created`.
3. It can then emit a terminal `error` or `response.failed` event whose details are nested under the event payload.
4. If those nested fields are not normalized, downstream code sees the misleading `{"message":""}`.

This PR keeps the outer HTTP 200 unchanged and preserves the terminal semantic error instead: `type`, `code`, `message`, plus a non-empty fallback when Azure sends empty details. The same normalizer is used by both native Responses create-stream and retrieve-stream paths.

The regression test reproduces the wire shape with `response.created` followed by an Azure-style nested `error` event. It does not claim that every such event is an Azure service-side HTTP 429; it ensures Bifrost does not discard whatever provider error Azure delivered inside the already-open stream.

Validation passed locally:

- `go test ./providers/openai ./providers/vllm ./schemas`
- `go vet ./providers/openai ./providers/vllm ./schemas`
- `git diff --check`

dani29:
human note:  hey folks, ran into this issue with Bifrost consuming SSE OpenAI Streaming from Azure,  where we ran out of quota. The stream returns 200 Status Code, but the request hides the quote denied in the payload. This issue should fix it and surface the error properly.

TejasGhatte (approved):

Pratham-Mishra04 (approved):

akshaydeo (approved):

PR #6304, opened 2026-08-19
feat(http): serve aggregate GET /v1/models from the ModelCatalog

geertvanzoest:
## Problem

The aggregate `GET /v1/models` listing fans out live `ListModels` requests to every provider configured on the caller's virtual key, through the same worker queues as inference traffic. On deployments with a low-concurrency provider (e.g. a local llama.cpp custom provider with concurrency 1), a single long-running generation blocks the fan-out for its full duration.

Measured on v1.6.11: `/v1/models` took **738–769 s** (and returned 504 through most proxies) while one chat completion was in flight on the busy provider. Clients that populate a model dropdown from `/v1/models` (pgAdmin, various OpenAI-compatible UIs) effectively see an empty list whenever the gateway is doing work.

Meanwhile `GET /api/models` answers in ~0.16 s from the in-memory ModelCatalog — the data is already there.

## Change

Serve the aggregate listing from the ModelCatalog, applying the same virtual-key scoping the live path enforces (per-provider `allowed_models` incl. `*` wildcard, `blacklisted_models` checked first). Model IDs keep the `provider/model` format the live aggregate path returns.

The live fan-out is fully preserved as a fallback; the catalog path steps aside (returns to the existing code path) for:

- `?provider=` — explicit single-provider listing, unchanged
- `?source=live` — explicit opt-out
- `page_token` / `page_size` — pagination semantics stay live
- any unknown query parameter
- governance store errors, unknown/inactive virtual keys, a nil ConfigStore, or an empty catalog (fail-open to previous behavior)

## Trade-off

Catalog freshness is boot-time discovery + refresh on provider/key changes — identical to what `GET /api/models` already serves, so the dashboard and `/v1/models` now agree.

## Validation

- `go build` / `go vet` / handler tests green on the v1.6.11 tag and current main
- Running in production on our gateway since 2026-08-19: `/v1/models` with a scoped virtual key returns 416 models in 0.3 s during full GPU saturation (previously 738–769 s / 504); scope enforcement verified (in-scope 200 / out-of-scope 403 / no key 401)

### DIFF
--- a/transports/bifrost-http/handlers/list_models_catalog.go
+++ b/transports/bifrost-http/handlers/list_models_catalog.go
@@ -1,0 +1,159 @@
+package handlers
+
+import (
+	"sort"
+	"strings"
+
+	"github.com/maximhq/bifrost/core/schemas"
+	governanceplugin "github.com/maximhq/bifrost/plugins/governance"
+	"github.com/valyala/fasthttp"
+)
+
+// listModelsFromCatalog serves GET /v1/models (without a ?provider= scope) from
+// the in-memory ModelCatalog instead of a live provider fan-out.
+//
+// Rationale: ListAllModels issues a runtime list-models call per provider and
+// waits for all of them. Those calls share the provider's inference worker
+// queue, so on a provider with low concurrency (e.g. a local llama.cpp slot
+// reservation) the listing blocks behind in-flight generations — measured at
+// 700s+ on a busy GPU, which every OpenAI-compatible client experiences as an
+// empty model dropdown. The ModelCatalog already holds the same model set
+// (populated by boot-time discovery and refreshed on provider/key changes),
+// so the aggregate listing can be answered from memory.
+//
+// Returns nil when the catalog cannot answer the request, in which case the
+// caller falls back to the live fan-out. That happens when:
+//   - the catalog is not configured,
+//   - the client asked for live results (?source=live),
+//   - the client uses pagination (page_token/page_size), which the catalog
+//     does not model.
+func (h *CompletionHandler) listModelsFromCatalog(ctx *fasthttp.RequestCtx, bifrostCtx *schemas.BifrostContext) *schemas.BifrostListModelsResponse {
+	if h.config == nil || h.config.ModelCatalog == nil {
+		return nil
+	}
+	if strings.EqualFold(strings.TrimSpace(string(ctx.QueryArgs().Peek("source"))), "live") {
+		return nil
+	}
+	if len(ctx.QueryArgs().Peek("page_token")) > 0 || len(ctx.QueryArgs().Peek("page_size")) > 0 {
+		return nil
+	}
+	// Any other query parameter is a provider-specific list-models feature
+	// (search, ordering, …) that only the live path understands.
+	for k := range ctx.QueryArgs().All() {
+		switch string(k) {
+		case "provider", "source", "page_token", "page_size":
+		default:
+			return nil
+		}
+	}
+
+	providers := h.catalogProvidersForRequest(bifrostCtx)
+	if providers == nil {
+		return nil
+	}
+
+	modelFilter, ok := h.virtualKeyModelFilter(ctx)
+	if !ok {
+		// Virtual-key resolution hit a store error; let the live path handle
+		// (and surface) it instead of answering 200 with a misleading list.
+		return nil
+	}
+
+	data := make([]schemas.Model, 0, 64)
+	for _, provider := range providers {
+		for _, model := range h.config.ModelCatalog.GetModelsForProvider(provider) {
+			if modelFilter != nil && !modelFilter(provider, model) {
+				continue
+			}
+			data = append(data, schemas.Model{ID: string(provider) + "/" + model})
+		}
+	}
+	if len(data) == 0 {
+		// An empty catalog usually means boot-time discovery has not (yet)
+		// populated it; let the live fan-out answer rather than serving an
+		// empty 200 that reads as "no models".
+		return nil
+	}
+	sort.Slice(data, func(i, j int) bool { return data[i].ID < data[j].ID })
+
+	return &schemas.BifrostListModelsResponse{Data: data}
+}
+
+// catalogProvidersForRequest resolves the provider set the catalog listing
+// should cover: the virtual key's providers when the request carries one
+// (stamped on the context by applyListModelsVirtualKeyProviderFilter),
+// otherwise every configured provider. Returns nil when the configured
+// provider set cannot be resolved, signalling the caller to fall back.
+func (h *CompletionHandler) catalogProvidersForRequest(bifrostCtx *schemas.BifrostContext) []schemas.ModelProvider {
+	if bifrostCtx != nil {
+		if raw := bifrostCtx.Value(schemas.BifrostContextKeyAvailableProviders); raw != nil {
+			if available, ok := raw.([]schemas.ModelProvider); ok && len(available) > 0 {
+				return available
+			}
+			return nil
+		}
+	}
+	providers, err := h.client.GetConfiguredProviders()
+	if err != nil {
+		return nil
+	}
+	return providers
+}
+
+// virtualKeyModelFilter returns a predicate applying the virtual key's
+// per-provider allowed_models to catalog entries, mirroring the governance
+// plugin's filtering on the live path (which this handler bypasses). A nil
+// (nil, true) means "no filtering" — the request carries no virtual key at
+// all; (nil, false) means the caller must fall back to the live path (store
+// lookup failed, or the key is unknown/inactive and the live path's
+// governance enforcement should answer).
+// Wildcard semantics follow schemas.WhiteList/BlackList: allowed ["*"] permits
+// every catalog model of that provider, an empty allowed list denies all
+// (deny-by-default), and blacklisted entries are excluded first.
+func (h *CompletionHandler) virtualKeyModelFilter(ctx *fasthttp.RequestCtx) (func(schemas.ModelProvider, string) bool, bool) {
+	vkValue := governanceplugin.ParseVirtualKeyFromFastHTTPRequest(ctx)
+	if vkValue == nil || strings.TrimSpace(*vkValue) == "" {
+		return nil, true
+	}
+	if h.config == nil || h.config.ConfigStore == nil {
+		// A keyed request we cannot verify: fall back to the live path rather
+		// than serving the full catalog unfiltered.
+		return nil, false
+	}
+	vk, err := h.config.ConfigStore.GetVirtualKeyByValue(ctx, strings.TrimSpace(*vkValue))
+	if err != nil {
+		// Store error: signal the caller to fall back to the live path, which
+		// surfaces the failure instead of answering 200 with a misleading list.
+		return nil, false
+	}
+	if vk == nil || vk.IsActive == nil || !*vk.IsActive {
+		// Unknown or inactive key: defer to the live path so its governance
+		// enforcement (per-provider rejection) applies unchanged, instead of
+		// answering with the full unfiltered catalog.
+		return nil, false
+	}
+
+	type modelScope struct {
+		allowed     schemas.WhiteList
+		blacklisted schemas.BlackList
+	}
+	scopeByProvider := make(map[schemas.ModelProvider]modelScope, len(vk.ProviderConfigs))
+	for _, pc := range vk.ProviderConfigs {
+		provider := schemas.ModelProvider(strings.TrimSpace(pc.Provider))
+		if provider == "" {
+			continue
+		}
+		scopeByProvider[provider] = modelScope{allowed: pc.AllowedModels, blacklisted: pc.BlacklistedModels}
+	}
+
+	return func(provider schemas.ModelProvider, model string) bool {
+		scope, ok := scopeByProvider[provider]
+		if !ok {
+			return false
+		}
+		if scope.blacklisted.IsBlocked(model) {
+			return false
+		}
+		return scope.allowed.IsAllowed(model)
+	}, true
+}
--- a/transports/bifrost-http/handlers/inference.go
+++ b/transports/bifrost-http/handlers/inference.go
@@ -892,9 +892,16 @@ func (h *CompletionHandler) listModels(ctx *fasthttp.RequestCtx) {
 		bifrostListModelsReq.ExtraParams = extraParams
 	}
 
-	// If provider is empty, list all models from all providers
+	// If provider is empty, list all models from all providers.
+	// Prefer the in-memory ModelCatalog for the aggregate listing: the live
+	// fan-out shares each provider's inference worker queue and can block for
+	// minutes behind in-flight generations (see list_models_catalog.go).
 	if provider == "" {
-		resp, bifrostErr = h.client.ListAllModels(bifrostCtx, bifrostListModelsReq)
+		if catalogResp := h.listModelsFromCatalog(ctx, bifrostCtx); catalogResp != nil {
+			resp, bifrostErr = catalogResp, nil
+		} else {
+			resp, bifrostErr = h.client.ListAllModels(bifrostCtx, bifrostListModelsReq)
+		}
 	} else {
 		resp, bifrostErr = h.client.ListModelsRequest(bifrostCtx, bifrostListModelsReq)
 	}


transports/bifrost-http/handlers/list_models_catalog.go
```diff
@@ -0,0 +1,159 @@
...
+// all; (nil, false) means the caller must fall back to the live path (store
+// lookup failed, or the key is unknown/inactive and the live path's
+// governance enforcement should answer).
+// Wildcard semantics follow schemas.WhiteList/BlackList: allowed ["*"] permits
+// every catalog model of that provider, an empty allowed list denies all
+// (deny-by-default), and blacklisted entries are excluded first.
+func (h *CompletionHandler) virtualKeyModelFilter(ctx *fasthttp.RequestCtx) (func(schemas.ModelProvider, string) bool, bool) {
+	vkValue := governanceplugin.ParseVirtualKeyFromFastHTTPRequest(ctx)
+	if vkValue == nil || strings.TrimSpace(*vkValue) == "" {
+		return nil, true
```
akshaydeo, line 116: Blocking. Returning `(nil, true)` here means "serve the catalog with no filtering at all", and it is reached in two cases where the live path would not have answered with the full inventory.

1. `lib/ctx.go:437` accepts **any** value in `x-bf-vk` as the virtual key (`handlers/governance.go:4860` states the contract: "x-bf-vk accepts any value; other headers require the sk-bf- prefix"), but `ParseVirtualKeyFromFastHTTPRequest` (`plugins/governance/utils.go:22`) only accepts `x-bf-vk` when it starts with `sk-bf-`. A config-provisioned key like `x-bf-vk: mycorp-prod`, scoped to one provider and one model, therefore parses as nil here and receives every model of every configured provider. Before this change `filterModelsForVirtualKey` in `PostLLMHook` trimmed that same response correctly, because it reads the VK from the context, not from a re-parse of the headers.

2. With `is_vk_mandatory: true` and no key at all, `EvaluateGovernanceRequest` (`plugins/governance/main.go:719`) returns 401 `virtual_key_required` on the live path. Here the request gets a 200 with the full catalog.

Both collapse into one fix: take the identity the pipeline itself uses and fail closed when there is none.

```go
vkValue := bifrost.GetStringFromContext(bifrostCtx, schemas.BifrostContextKeyVirtualKey)
if strings.TrimSpace(vkValue) == "" {
    if hv := governanceplugin.ParseVirtualKeyFromFastHTTPRequest(ctx); hv != nil {
        vkValue = strings.TrimSpace(*hv)
    }
}
if vkValue == "" {
    // No resolvable identity: the live path decides (is_vk_mandatory, user auth).
    return nil, false
}
```

transports/bifrost-http/handlers/list_models_catalog.go
```diff
@@ -0,0 +1,159 @@
...
+		allowed     schemas.WhiteList
+		blacklisted schemas.BlackList
+	}
+	scopeByProvider := make(map[schemas.ModelProvider]modelScope, len(vk.ProviderConfigs))
+	for _, pc := range vk.ProviderConfigs {
+		provider := schemas.ModelProvider(strings.TrimSpace(pc.Provider))
+		if provider == "" {
+			continue
+		}
+		scopeByProvider[provider] = modelScope{allowed: pc.AllowedModels, blacklisted: pc.BlacklistedModels}
```
akshaydeo, line 146: `TableVirtualKeyProviderConfig` (`framework/configstore/tables/virtualkey.go:26`) has no unique constraint on `(virtual_key_id, provider)` and carries a `Weight` consumed by `LoadBalanceProvider`, so two configs for the same provider are representable. This map keeps only the last one.

`filterModelsForVirtualKey` (`plugins/governance/utils.go:158-200`) instead runs two passes over every matching config: any config that blacklists the model blocks it, any config that allows it permits it. With duplicates, this map can both surface a model an earlier config blacklisted and hide a model only an earlier config allowed.

Suggested shape: keep a slice per provider and union the passes.

```go
scopeByProvider := make(map[schemas.ModelProvider][]modelScope, len(vk.ProviderConfigs))
// ...
scopeByProvider[provider] = append(scopeByProvider[provider], modelScope{...})
```

transports/bifrost-http/handlers/list_models_catalog.go
```diff
@@ -0,0 +1,159 @@
...
+
+	return func(provider schemas.ModelProvider, model string) bool {
+		scope, ok := scopeByProvider[provider]
+		if !ok {
+			return false
+		}
+		if scope.blacklisted.IsBlocked(model) {
+			return false
+		}
+		return scope.allowed.IsAllowed(model)
```
akshaydeo, line 157: Reuse: both live enforcement sites call the shared helper and keep `AllowedModels.IsAllowed` only as the degraded fallback when the catalog is missing (`plugins/governance/utils.go:185`, `plugins/governance/resolver.go:398`). This handler already holds `h.config.ModelCatalog`, so the same call is available and removes three divergences for free:

- `allowed_models: ["openai/gpt-4o"]` (provider-prefixed) matches on the live path via `framework/modelcatalog/models.go:327-344` and does not match here, so entitled models silently disappear from the listing.
- A custom provider whose list-models operation is disabled is special-cased at `models.go:313` and not here.
- `WhiteList.Contains` is case-insensitive while the helper's bare-name match is exact, so the two disagree on case-only differences.

```go
providerCfg, ok := configuredProviders[provider]
var providerCfgPtr *configstore.ProviderConfig
if ok { providerCfgPtr = &providerCfg }
return h.config.ModelCatalog.IsModelAllowedForProvider(provider, model, providerCfgPtr, scope.allowed)
```

transports/bifrost-http/handlers/list_models_catalog.go
```diff
@@ -0,0 +1,159 @@
...
+		return nil
+	}
+
+	data := make([]schemas.Model, 0, 64)
+	for _, provider := range providers {
+		for _, model := range h.config.ModelCatalog.GetModelsForProvider(provider) {
+			if modelFilter != nil && !modelFilter(provider, model) {
+				continue
+			}
+			data = append(data, schemas.Model{ID: string(provider) + "/" + model})
```
akshaydeo, line 68: Catalog entries carry only `id`, so `owned_by`, `created` and `alias` (and, on the response, `key_statuses` and `extra_fields.request_type` / `latency`) disappear whenever the catalog answers. `enrichListModelsResponse` restores datasheet-derived fields but none of these.

`schemas.Model` has no `object` field on either path, so OpenAI's per-entry `"object": "model"` is equally absent before and after and this is not an OpenAI-compat regression. It is a within-endpoint shape drift though: the same client hitting the same URL sees `owned_by` appear or vanish depending on which path served it. Either backfill `owned_by` from the provider and `created` from the datasheet entry, or note the reduced shape in the OpenAPI description.

transports/bifrost-http/handlers/list_models_catalog.go
```diff
@@ -0,0 +1,159 @@
...
+	vkValue := governanceplugin.ParseVirtualKeyFromFastHTTPRequest(ctx)
+	if vkValue == nil || strings.TrimSpace(*vkValue) == "" {
+		return nil, true
+	}
+	if h.config == nil || h.config.ConfigStore == nil {
+		// A keyed request we cannot verify: fall back to the live path rather
+		// than serving the full catalog unfiltered.
+		return nil, false
+	}
+	vk, err := h.config.ConfigStore.GetVirtualKeyByValue(ctx, strings.TrimSpace(*vkValue))
```
akshaydeo, line 123: `listModels` calls `applyListModelsVirtualKeyProviderFilter` a few lines earlier, which already performed `GetVirtualKeyByValue` on this exact value (`list_models_vk.go:34`). This is a second store round trip per keyed request.

Returning the resolved `*TableVirtualKey` from that function and passing it into `listModelsFromCatalog` removes the duplicate lookup, and it also removes the oddity that the error branch just below is only reachable through a transient failure between the two calls (the first one already sends a 500 on any non-`ErrNotFound` error).

akshaydeo (changes requested):
Thank you for this contribution, and especially for the measurements and the production soak notes: a 738s aggregate `/v1/models` is a real and painful bug, and serving it from the `ModelCatalog` is the right shape of fix. The fallback design (step aside for `?provider=`, `?source=live`, pagination, unknown params, store errors, empty catalog) is careful and reads well. The concern below is not the caching idea, it is that the new fast path re-implements virtual-key scoping instead of reusing the governance code that enforces it today, and the two disagree in ways that disclose models.

Today the scoping on this endpoint is enforced in two places: `applyListModelsVirtualKeyProviderFilter` narrows the fan-out (an optimization), and `GovernancePlugin.PostLLMHook` calls `filterModelsForVirtualKey` (`plugins/governance/utils.go:141`) which is the actual authorization boundary, plus `EvaluateGovernanceRequest` (`plugins/governance/main.go:701`) which enforces `is_vk_mandatory`. The catalog path skips `handleRequest` entirely, so both of those disappear and `virtualKeyModelFilter` becomes the only gate.

```mermaid
flowchart TD
    A["GET /v1/models"] --> B{provider empty?}
    B -->|no| C["ListModelsRequest (live)"]
    B -->|yes| D["listModelsFromCatalog"]
    D -->|nil| E["ListAllModels (live fan-out)"]
    D -->|non-nil| F["catalog response"]
    E --> G["governance PreHook: is_vk_mandatory + provider allowlist"]
    G --> H["governance PostHook: filterModelsForVirtualKey"]
    H --> I["scoped response"]
    F --> J["no plugin pipeline, no governance"]
```

## Findings

|---|---|---|---|---|
| 1 | High | `transports/bifrost-http/handlers/list_models_catalog.go:116` | A virtual key whose value does not start with `sk-bf-` is invisible to `ParseVirtualKeyFromFastHTTPRequest`, so the filter returns "no filtering" and the caller gets every model of every configured provider | CONFIRMED |
| 2 | High | `transports/bifrost-http/handlers/list_models_catalog.go:116` | With `is_vk_mandatory: true`, a keyless request gets 200 plus the full catalog where the live path returns 401 `virtual_key_required` | CONFIRMED |
| 3 | Medium | `transports/bifrost-http/handlers/list_models_catalog.go:146` | `scopeByProvider` is a map, so a second provider config for the same provider silently overwrites the first; the canonical filter unions across all matching configs | PLAUSIBLE |
| 4 | Medium | `transports/bifrost-http/handlers/list_models_catalog.go:157` | Uses `WhiteList.IsAllowed` rather than `ModelCatalog.IsModelAllowedForProvider`, so provider-prefixed allowlist entries and custom providers with list-models disabled are scoped differently from the live path | CONFIRMED |
| 5 | Medium | `transports/bifrost-http/handlers/list_models_catalog.go:68` | Catalog entries carry only `id`, so `owned_by`, `created`, `alias` and `key_statuses` vanish from the same endpoint depending on which path served it | CONFIRMED |
| 6 | Low | `transports/bifrost-http/handlers/list_models_catalog.go:123` | Second `GetVirtualKeyByValue` per request; `applyListModelsVirtualKeyProviderFilter` already resolved the same row a few lines earlier | CONFIRMED |
| 7 | Low | `transports/bifrost-http/handlers/list_models_catalog.go:31` | The catalog path never enters the plugin pipeline, so `/v1/models` calls stop producing request logs and plugin telemetry | PLAUSIBLE |
| 8 | Low | `docs/openapi/openapi.json` (`/v1/models`) | The new `?source=live` parameter and the catalog-served semantics are undocumented | CONFIRMED |

### 1. Non-prefixed virtual keys are served the whole catalog (High, CONFIRMED)

`lib/ctx.go:437-441` accepts **any** value in `x-bf-vk` as the virtual key, and `handlers/governance.go:4860` states the contract explicitly: "x-bf-vk accepts any value; other headers require the sk-bf- prefix". But `governanceplugin.ParseVirtualKeyFromFastHTTPRequest` (`plugins/governance/utils.go:22`) only accepts `x-bf-vk` when it starts with `sk-bf-`.

Trigger: an operator provisions a virtual key through `config.json` with `value: "mycorp-prod"` (the API generator always emits `sk-bf-`, the config-file path does not enforce it), scoped to `openai` with `allowed_models: ["gpt-4o-mini"]`. Client sends `GET /v1/models` with `x-bf-vk: mycorp-prod`.

- Before: `filterModelsForVirtualKey` reads the VK from the context (set from the raw header, no prefix requirement) and trims the list to `openai/gpt-4o-mini`.
- After: `ParseVirtualKeyFromFastHTTPRequest` returns nil, `virtualKeyModelFilter` returns `(nil, true)`, `catalogProvidersForRequest` falls through to `GetConfiguredProviders()`, and the response is every model of every provider on the gateway (Bedrock ARNs, private deployment names, and so on).

Fix: read the identity the pipeline actually uses, `bifrostCtx.Value(schemas.BifrostContextKeyVirtualKey)`, with the header parse only as a fallback. That is the same string `PostLLMHook` filters on, so the two paths cannot drift.

### 2. `is_vk_mandatory` is not enforced on the catalog path (High, CONFIRMED)

`EvaluateGovernanceRequest` (`plugins/governance/main.go:719-733`) rejects with 401 when there is no valid VK, no direct key auth and no user id, and `is_vk_mandatory` is true. A gateway that uses virtual keys as its authentication (auth config disabled, `is_vk_mandatory: true`) currently returns 401 on a keyless `GET /v1/models`. After this change, `virtualKeyModelFilter` returns `(nil, true)` for the keyless case and the handler answers 200 with the full catalog. Model inventory is not the most sensitive data on a gateway, but it does name every provider, deployment and private model, and it changes an authenticated endpoint into an anonymous one.

The same gap applies to enterprise user-token auth: no `x-bf-vk` header means no filter, even though the pipeline would have scoped the result by the resolved user.

Fix: when there is no resolvable identity, return nil (fall back to live) rather than "no filtering". Fail closed, matching the treatment already given to unknown and inactive keys just below.

### 3. Duplicate provider configs (Medium, PLAUSIBLE)

`TableVirtualKeyProviderConfig` (`framework/configstore/tables/virtualkey.go:26`) has no unique constraint on `(virtual_key_id, provider)`, and it carries a `Weight` used by `LoadBalanceProvider`, so more than one config per provider is representable. `filterModelsForVirtualKey` handles that with two passes over all matching configs: any config that blacklists blocks, any config that allows permits. The map here keeps only the last config seen, which can both surface a model blacklisted by an earlier config and hide a model that only an earlier config allowed. Worth confirming whether the UI can produce duplicates; if it cannot today, the map still bakes in an assumption the schema does not guarantee.

### 4. Reuse: call the shared allowance helper (Medium, CONFIRMED)

Both live enforcement sites (`plugins/governance/utils.go:185` and `plugins/governance/resolver.go:398`) call `ModelCatalog.IsModelAllowedForProvider(provider, model, providerConfig, pc.AllowedModels)` and both keep `AllowedModels.IsAllowed` only as the degraded fallback when the catalog is absent. Since this handler already holds `h.config.ModelCatalog`, calling the same helper removes the divergence for free:

- `allowed_models: ["openai/gpt-4o"]` on a Groq or OpenRouter style provider matches on the live path (`framework/modelcatalog/models.go:327-344`) and does not match here, so entitled models disappear from the dropdown.
- A custom provider with the list-models operation disabled is handled explicitly there (`models.go:313`) and not here.
- `WhiteList.Contains` is case-insensitive while the helper's bare-name match is exact, so the two also disagree on case-only differences.

### 5. Response shape drift within one endpoint (Medium, CONFIRMED)

`schemas.Model{ID: ...}` drops `owned_by`, `created` and `alias`, and the response carries no `key_statuses` or `extra_fields` (`request_type`, `latency`). `enrichListModelsResponse` restores datasheet-derived fields but not those. `schemas.Model` has no `object` field on either path, so OpenAI's per-entry `"object": "model"` is equally absent before and after (not a regression), but a client that reads `owned_by` or `created` will see them appear and disappear depending on whether the catalog answered. Filling `owned_by` from the provider and `created` from the datasheet entry, or documenting the reduced shape, would keep the contract stable.

### 6. Double virtual-key lookup (Low, CONFIRMED)

`listModels` calls `applyListModelsVirtualKeyProviderFilter`, which already performs `GetVirtualKeyByValue`, immediately before the catalog path calls it again for the same value. Returning the resolved `*TableVirtualKey` from the first call and passing it in removes a store round trip on every keyed request. It also makes the store-error branch here reachable-by-design rather than a near-dead path (the first call already 500s on non-`ErrNotFound` errors, so this second one can only fail on a transient error in between).

### 7. Observability (Low, PLAUSIBLE)

The existing `applyListModelsVirtualKeyProviderFilter` doc comment cares about "noisy, expected errors in request logs", which tells us list-models calls show up in the logging plugin today. The catalog path never reaches `handleRequest`, so those entries and any plugin-level metrics for `/v1/models` stop appearing. That may be a welcome side effect, but it should be a stated decision rather than a silent one, and it is worth a line in the PR description.

### 8. Docs, tests and naming (Low)

- `docs/openapi/openapi.json` documents `provider`, `page_size` and `page_token` for `GET /v1/models`. The new `source` parameter and the "catalog by default, `source=live` to force a fan-out" behaviour are not documented. AGENTS.md lists `docs/openapi/openapi.json` as the place where route surface changes land.
- No tests accompany the change. `handlers/list_models_vk_test.go` is a close template; cases worth pinning: keyless request with `is_vk_mandatory`, a non-`sk-bf-` VK value, wildcard allowlist, blacklist wins over allowlist, and each fallback trigger.
- AGENTS.md:733 says: "**Go filenames**: No underscores. The only permitted underscore is the `_test.go` suffix." The new file is `list_models_catalog.go`. The neighbouring `list_models_vk.go` predates the rule, so this is genuinely ambiguous; `listmodelscatalog.go` matches the written rule.
- Nit: `?source=anythingelse` is accepted and silently served from the catalog. Rejecting unknown `source` values with a 400, or treating anything other than `catalog` as live, would be less surprising.

## Merge recommendation

**Not yet - request changes.** Findings 1 and 2 mean the endpoint can return models the caller is not entitled to see, and in the `is_vk_mandatory` case can answer an unauthenticated caller. Both collapse into one change: resolve the identity from `BifrostContextKeyVirtualKey` and fail closed (return nil, fall back to live) whenever there is no resolvable identity. Everything else is a nit or a follow-up.

## Followups

1. **In this PR (blocking)** `list_models_catalog.go:114-118`: resolve the VK from `bifrostCtx.Value(schemas.BifrostContextKeyVirtualKey)` first, header parse as fallback.
2. **In this PR (blocking)** `list_models_catalog.go:116`: no resolvable identity must return `(nil, false)` so the live path enforces `is_vk_mandatory`.
3. **In this PR** `list_models_catalog.go:157`: call `h.config.ModelCatalog.IsModelAllowedForProvider` instead of `AllowedModels.IsAllowed`.
4. **In this PR** `list_models_catalog.go:146`: union across all provider configs for a provider rather than keeping the last one, matching `plugins/governance/utils.go:158-200`.
5. **In this PR** new `list_models_catalog_test.go` (or `listmodelscatalog_test.go`): cover the scoping and fallback matrix above.
6. **In this PR** `docs/openapi/openapi.json`: document `source` and the catalog-served default on `GET /v1/models`.
7. **Follow-up PR** `list_models_catalog.go:123`: thread the already-resolved VK through from `applyListModelsVirtualKeyProviderFilter` to drop the duplicate store lookup.
8. **Follow-up PR** `list_models_catalog.go:68`: decide whether `owned_by` / `created` should be reconstructed from the catalog so the two paths return the same shape.

- **Virtual key with zero provider configs leaks everything**: refuted, `catalogProvidersForRequest` requires `len(available) > 0` and otherwise returns nil, falling back to live, which returns an empty list via `filterProvidersByContext`.
- **Unknown or inactive virtual key served unfiltered**: refuted, `virtualKeyModelFilter` returns `(nil, false)` for both and the live path enforces.
- **Store error answered as a misleading 200**: refuted, `applyListModelsVirtualKeyProviderFilter` already returns 500 on non-`ErrNotFound` errors before this code runs.
- **Pagination silently broken**: refuted, `page_token` and `page_size` both force the live path.
- **Response loses OpenAI `object` fields**: refuted, `schemas.Model` has no `object` field and `BifrostListModelsResponse` has no top-level `object`, so both paths behave identically here.
- **Unknown query parameters bypass provider-specific list features**: refuted, the switch falls back to live for any parameter outside the known four.
- **`ctx.QueryArgs().All()` single-variable range is invalid**: refuted, `inference.go` already ranges over the same iterator two-valued and Go permits the one-variable form.

PR #6367, opened 2026-08-20
[fix]: flatten multi-block system prompts to string content for chat providers

### LINKED ISSUE #6245: [Bug]: Multi-block top-level `system` field is split into multiple system messages, rejected by ollama ("system message must be at the beginning")
## Summary

When an Anthropic-format request carries a **top-level `system` field with more than one content block**, Bifrost appears to emit **more than one system message** toward the provider. Providers that require the system message to be first — ollama among them — reject the request:

```json
{"type":"error","error":{"type":"api_error","message":"system message must be at the beginning"}}
```

**One block succeeds; two or more fail**, with everything else held constant.

This matters because **Claude Code always sends `system` as 3 blocks** (billing header, agent prompt, CWD/date), so every request from it fails against an ollama provider. The same 3-block request sent straight to ollama's own `/v1/messages` succeeds — ollama merges the blocks — so the difference is introduced in Bifrost.

This is **not** the mid-conversation-system family of issues (see "Related" below). There is no `role:"system"` message inside `messages[]` here at all; the input has a single `user` message and the problem is purely the shape of the top-level `system` field.

## Minimal reproduction

Configure an `ollama` provider, then vary only the number of `system` blocks:

```bash
GW=http://<bifrost>:8080/anthropic/v1/messages
KEY=<virtual-key>

# 1 block -> 200 OK
curl -sS "$GW" -H "Authorization: Bearer $KEY" -H 'content-type: application/json' -d '{
  "model": "ollama/<your-model>",
  "max_tokens": 32,
  "system": [{"type":"text","text":"A"}],
  "messages": [{"role":"user","content":"Reply OK"}]
}'
# {"content":[{"type":"text","text":"OK"}], ...}

# 2 blocks -> 400/500
curl -sS "$GW" -H "Authorization: Bearer $KEY" -H 'content-type: application/json' -d '{
  "model": "ollama/<your-model>",
  "max_tokens": 32,
  "system": [{"type":"text","text":"A"},{"type":"text","text":"B"}],
  "messages": [{"role":"user","content":"Reply OK"}]
}'
# {"type":"error","error":{"type":"api_error","message":"system message must be at the beginning"}}
```

| `system` blocks | result |
|---|---|
| 1 | ✅ completion |
| 2 | ❌ `system message must be at the beginning` |
| 3 | ❌ same |

Adding `cache_control` to the blocks makes no difference; nor does `stream`, `tools`, `thinking`, `metadata`, `context_management` or `output_config`. I bisected each of those out of a real Claude Code body and only removing `system` changed the outcome.

### Control: ollama itself accepts the same shape

Sent directly to ollama, bypassing Bifrost, the identical 3-block request returns a completion:

```bash
curl -sS http://<ollama-host>:11434/v1/messages -H 'content-type: application/json' -d '{
  "model": "<your-model>",
  "max_tokens": 32,
  "system": [{"type":"text","text":"A"},{"type":"text","text":"B"},{"type":"text","text":"C"}],
  "messages": [{"role":"user","content":"Reply OK"}]
}'
# {"id":"msg_...","type":"message","role":"assistant","content":[{"type":"text","text":"OK"}], ...}
```

So ollama merges multiple system blocks into a single system message; something between Bifrost's Anthropic ingress and its ollama egress does not.

## Expected behaviour

A multi-block top-level `system` field should reach the provider as **one** system message (blocks concatenated, or joined with newlines), matching how Anthropic and ollama both treat it. `cache_control` on individual blocks should be preserved where the provider supports it.

## Actual behaviour

Providers that enforce "system must be first / only one system message" reject the request. Providers that are lenient presumably accept it silently, which may make this invisible on the more common backends and could also be splitting prompt-cache prefixes there.

## Impact

Claude Code cannot be used against an ollama provider through Bifrost at all — it fails on the first request, every time. Upstream's `ollama launch claude` works precisely because it points `ANTHROPIC_BASE_URL` at ollama directly and skips the gateway, which means users hit this the moment they try to put Bifrost in that path (virtual keys, cost attribution, request logs, fallbacks).

## Possible starting point

I did not confirm the exact split site, so treat this as a hypothesis rather than a diagnosis. `convertAnthropicSystemToBifrostMessages` (`core/providers/anthropic/responses.go`) does look correct in isolation — it builds a **single** `ResponsesMessage` with N content blocks:

```go
if len(contentBlocks) > 0 {
    bifrostMessages = append(bifrostMessages, schemas.ResponsesMessage{
        Role:    schemas.Ptr(schemas.ResponsesInputMessageRoleSystem),
        Content: &schemas.ResponsesMessageContent{ContentBlocks: contentBlocks},
    })
}
```

So the fan-out is more likely downstream, where that one multi-block system message is converted for a chat-style provider (the responses→chat path the ollama provider uses via `ToChatRequest`). A multi-block message becoming one chat message per block would produce exactly the observed `[system, system, user]`.

## Environment

- Bifrost `v1.6.11` (Docker, `maximhq/bifrost`), source read at `22c5da85c`
- Provider: native `ollama` (not a custom Anthropic-base provider), `allow_private_network: true`
- Models: `qwen3.8:27b` and `qwen3.8:27b-mlx` (both reject); `qwen3-coder:30b-a3b-q4_K_M` is lenient and accepts, which is a useful A/B if you want one
- Client: Claude Code 2.1.233, via the `/anthropic` endpoint
- ollama 0.32.13

I have not re-tested against 1.7.x — if this was already addressed there, please close and point me at the change.

## Related issues

These all touch system-message handling for Claude Code but describe the **inverse** direction (a mid-conversation `role:"system"` message inside `messages[]` being *hoisted into* the top-level `system` block), so I don't believe this is a duplicate of any of them. Linking for context since they share the same code area:

- #4276 — [fix]: Anthropic provider - align mid-conversation system handling
- #4592 — Custom Anthropic-base providers hoist mid-conversation system messages, breaking prompt cache
- #4593 — fix(anthropic): keep mid-conversation system inline for custom Anthropic-base providers
- #3879 — Anthropic raw-body passthrough does not propagate `betas[]`, causing Claude Code mid-conversation system messages to fail
- #5544 — [Bug]: Bedrock converter double-wraps already-wrapped `<system-reminder>` blocks (Claude Code)
- #4068 / #4534 (closed) — Bedrock mid-conversation system hoisting

AdityaPainuli:
## Summary

Claude Code can't talk to an ollama provider through Bifrost at all. Every request fails with `system message must be at the beginning`.

The trigger is the shape of the system prompt. Claude Code always sends the top-level `system` field as 3 text blocks. Bifrost keeps those blocks as a content-parts array on the outbound chat-completions payload, and ollama's OpenAI-compat layer expands each part of a content array into its own message with the same role. So one system message with 3 parts becomes `[system, system, system, user]` inside ollama, and its renderer rejects it. One block works because `ToChatMessages` already collapses a single text block to plain string content. Two or more blocks fall off that cliff.

I traced this end to end: Bifrost itself never splits the message (the ingress builds one system message with N blocks, and `ToChatRequest` keeps it as one message). The array shape is the problem, and a plain string system prompt is the only shape every openai-compat server accepts.

## Changes

- `core/schemas/mux.go`: `ToChatMessages` now flattens multi-block all-text system/developer content into a single string prompt (blocks joined with `\n\n`), generalizing the existing single-block collapse.
- Gated on role and all-text blocks only. User messages keep their content arrays, since image/file parts depend on them.
- The fix sits in the generic responses→chat converter rather than the ollama provider: `ToChatRequest` is the fallback path for every chat-style provider (ollama, deepseek, groq, etc.) and none of them benefit from a parts array on a system message. This path already drops per-block `cache_control` today, and a string system prompt has nowhere to carry it anyway, so nothing is lost there.
- `core/providers/anthropic/systemflatten_test.go`: regression test covering the full ingress→chat path with a 3-block Claude Code-shaped request, plus the single-block shape. The main test fails without the converter change.
- `core/schemas/systemflatten_test.go`: the provider-agnostic `ToChatMessages` shape guards (developer-role flatten, empty-block skip, multi-block user message keeps its array with values and order asserted).
- `tests/e2e/api/collections/provider-harness.json`: folder 59 pins the wire shape end to end through deepseek (a real chat-fallback provider in the harness): a 3-block Anthropic `system` must reach the upstream as exactly one leading system message with string content, asserted via `x-bf-send-back-raw-request`.
- Review follow-ups also folded in: blocks whose trimmed text is empty are skipped when joining (no doubled separators), and the `ToChatMessages` doc comment sits back above its declaration.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [x] Core (Go)
- [ ] Transports (HTTP)
- [ ] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

```sh
cd core
go test -run 'TestMultiBlockSystemToChatRequest|TestSystemFlattenShapes' -v ./providers/anthropic/
go test -run TestFlattenSystemTextBlocksShapes -v ./schemas/
go test ./schemas/ ./providers/anthropic/ ./providers/openai/ ./providers/ollama/
```

Harness case (live, paid):

```sh
make run-provider-harness-test PROVIDER=deepseek FEATURE="multi-block system"
```

Or live: point Claude Code at Bifrost's `/anthropic` endpoint with an ollama provider behind it. Before this change the first request fails with `system message must be at the beginning`; after it, requests go through.

## Screenshots/Recordings

N/A

## Breaking changes

- [ ] Yes
- [x] No

## Related issues

Closes #6245

## Security considerations

None.

## Checklist

### DIFF
--- a/core/providers/anthropic/systemflatten_test.go
+++ b/core/providers/anthropic/systemflatten_test.go
@@ -86,8 +86,8 @@ func TestMultiBlockSystemToChatRequest(t *testing.T) {
 	}
 }
 
-// One text block already collapsed to string content before the fix; multi-block user
-// messages must keep their content array (image/file parts depend on it).
+// One text block already collapsed to string content before the fix. Shape coverage
+// that does not involve the Anthropic ingress lives in core/schemas/systemflatten_test.go.
 func TestSystemFlattenShapes(t *testing.T) {
 	buildChat := func(t *testing.T, bodyJSON string) *schemas.BifrostChatRequest {
 		t.Helper()
@@ -114,61 +114,4 @@ func TestSystemFlattenShapes(t *testing.T) {
 		}
 	})
 
-	t.Run("multi-block developer message flattens to string", func(t *testing.T) {
-		devRole := schemas.ResponsesInputMessageRoleDeveloper
-		chatMessages := schemas.ToChatMessages([]schemas.ResponsesMessage{
-			{
-				Role: &devRole,
-				Content: &schemas.ResponsesMessageContent{
-					ContentBlocks: []schemas.ResponsesMessageContentBlock{
-						{Type: schemas.ResponsesInputMessageContentBlockTypeText, Text: schemas.Ptr("dev block A")},
-						{Type: schemas.ResponsesInputMessageContentBlockTypeText, Text: schemas.Ptr("dev block B")},
-					},
-				},
-			},
-		})
-		if len(chatMessages) != 1 {
-			t.Fatalf("expected 1 chat message, got %d", len(chatMessages))
-		}
-		dev := chatMessages[0]
-		if dev.Role != schemas.ChatMessageRoleDeveloper {
-			t.Fatalf("expected developer role, got %s", dev.Role)
-		}
-		if dev.Content == nil || dev.Content.ContentStr == nil {
-			t.Fatalf("expected multi-block developer message to flatten to string content, got: %+v", dev.Content)
-		}
-		if want := "dev block A\n\ndev block B"; *dev.Content.ContentStr != want {
-			t.Errorf("flattened developer prompt mismatch:\ngot:  %q\nwant: %q", *dev.Content.ContentStr, want)
-		}
-	})
-
-	t.Run("multi-block user message keeps content array", func(t *testing.T) {
-		// Flattening applies to system/developer roles only: user messages keep their
-		// content array through ToChatMessages (image/file parts depend on it).
-		userRole := schemas.ResponsesInputMessageRoleUser
-		chatMessages := schemas.ToChatMessages([]schemas.ResponsesMessage{
-			{
-				Role: &userRole,
-				Content: &schemas.ResponsesMessageContent{
-					ContentBlocks: []schemas.ResponsesMessageContentBlock{
-						{Type: schemas.ResponsesInputMessageContentBlockTypeText, Text: schemas.Ptr("part one")},
-						{Type: schemas.ResponsesInputMessageContentBlockTypeText, Text: schemas.Ptr("part two")},
-					},
-				},
-			},
-		})
-		if len(chatMessages) != 1 {
-			t.Fatalf("expected 1 chat message, got %d", len(chatMessages))
-		}
-		user := chatMessages[0]
-		if user.Role != schemas.ChatMessageRoleUser {
-			t.Fatalf("expected user role, got %s", user.Role)
-		}
-		if user.Content == nil || user.Content.ContentBlocks == nil {
-			t.Fatalf("expected multi-block user message to keep content array, got: %+v", user.Content)
-		}
-		if len(user.Content.ContentBlocks) != 2 {
-			t.Fatalf("expected 2 user content blocks, got %d", len(user.Content.ContentBlocks))
-		}
-	})
 }
--- a/core/providers/anthropic/systemflatten_test.go
+++ b/core/providers/anthropic/systemflatten_test.go
@@ -114,6 +114,34 @@ func TestSystemFlattenShapes(t *testing.T) {
 		}
 	})
 
+	t.Run("multi-block developer message flattens to string", func(t *testing.T) {
+		devRole := schemas.ResponsesInputMessageRoleDeveloper
+		chatMessages := schemas.ToChatMessages([]schemas.ResponsesMessage{
+			{
+				Role: &devRole,
+				Content: &schemas.ResponsesMessageContent{
+					ContentBlocks: []schemas.ResponsesMessageContentBlock{
+						{Type: schemas.ResponsesInputMessageContentBlockTypeText, Text: schemas.Ptr("dev block A")},
+						{Type: schemas.ResponsesInputMessageContentBlockTypeText, Text: schemas.Ptr("dev block B")},
+					},
+				},
+			},
+		})
+		if len(chatMessages) != 1 {
+			t.Fatalf("expected 1 chat message, got %d", len(chatMessages))
+		}
+		dev := chatMessages[0]
+		if dev.Role != schemas.ChatMessageRoleDeveloper {
+			t.Fatalf("expected developer role, got %s", dev.Role)
+		}
+		if dev.Content == nil || dev.Content.ContentStr == nil {
+			t.Fatalf("expected multi-block developer message to flatten to string content, got: %+v", dev.Content)
+		}
+		if want := "dev block A\n\ndev block B"; *dev.Content.ContentStr != want {
+			t.Errorf("flattened developer prompt mismatch:\ngot:  %q\nwant: %q", *dev.Content.ContentStr, want)
+		}
+	})
+
 	t.Run("multi-block user message keeps content array", func(t *testing.T) {
 		// Flattening applies to system/developer roles only: user messages keep their
 		// content array through ToChatMessages (image/file parts depend on it).
--- a/core/providers/anthropic/systemflatten_test.go
+++ b/core/providers/anthropic/systemflatten_test.go
@@ -1,0 +1,146 @@
+package anthropic
+
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/maximhq/bifrost/core/providers/openai"
+	"github.com/maximhq/bifrost/core/schemas"
+)
+
+// Regression test for issue #6245: an Anthropic-format request with a multi-block
+// top-level `system` field (what Claude Code sends) must reach chat-style providers
+// (e.g. ollama, which routes Responses through ToChatRequest) as exactly ONE system
+// message at position 0 with plain string content. OpenAI-compatible servers like
+// ollama expand a content-parts array into one message per part, so a multi-part
+// system message becomes [system, system, system] upstream and is rejected with
+// "system message must be at the beginning".
+func TestMultiBlockSystemToChatRequest(t *testing.T) {
+	rawReq := []byte(`{
+		"model": "ollama/llama3.1",
+		"max_tokens": 512,
+		"system": [
+			{"type": "text", "text": "You are Claude Code, Anthropic's official CLI for Claude."},
+			{"type": "text", "text": "Block B: follow the project conventions.", "cache_control": {"type": "ephemeral"}},
+			{"type": "text", "text": "Block C: answer concisely."}
+		],
+		"messages": [
+			{"role": "user", "content": "hello"}
+		]
+	}`)
+
+	var anthropicReq AnthropicMessageRequest
+	if err := json.Unmarshal(rawReq, &anthropicReq); err != nil {
+		t.Fatalf("failed to unmarshal anthropic request: %v", err)
+	}
+
+	// Ingress conversion (what /anthropic/v1/messages does), then the responses→chat
+	// fallback used by chat-style providers (core/providers/ollama/ollama.go).
+	chatReq := anthropicReq.ToBifrostResponsesRequest(nil).ToChatRequest()
+
+	dumpChat, _ := json.MarshalIndent(chatReq.Input, "", "  ")
+
+	chatSysCount := 0
+	for _, m := range chatReq.Input {
+		if m.Role == schemas.ChatMessageRoleSystem {
+			chatSysCount++
+		}
+	}
+	if chatSysCount != 1 {
+		t.Errorf("expected exactly 1 system message in chat request, got %d:\n%s", chatSysCount, dumpChat)
+	}
+	if len(chatReq.Input) == 0 || chatReq.Input[0].Role != schemas.ChatMessageRoleSystem {
+		t.Fatalf("expected system message at position 0, got messages:\n%s", dumpChat)
+	}
+	if len(chatReq.Input) != 2 {
+		t.Errorf("expected 2 chat messages total (system + user), got %d", len(chatReq.Input))
+	}
+
+	sys := chatReq.Input[0]
+	if sys.Content == nil || sys.Content.ContentStr == nil {
+		t.Fatalf("expected multi-block system prompt to be flattened to string content, got:\n%s", dumpChat)
+	}
+	want := "You are Claude Code, Anthropic's official CLI for Claude.\n\nBlock B: follow the project conventions.\n\nBlock C: answer concisely."
+	if *sys.Content.ContentStr != want {
+		t.Errorf("flattened system prompt mismatch:\ngot:  %q\nwant: %q", *sys.Content.ContentStr, want)
+	}
+
+	// The outbound openai-compatible payload must carry the same single string prompt.
+	wireMessages := openai.ConvertBifrostMessagesToOpenAIMessages(chatReq.Input)
+	dumpWire, err := json.Marshal(wireMessages[0])
+	if err != nil {
+		t.Fatalf("marshal wire message: %v", err)
+	}
+	var wireSys struct {
+		Role    string `json:"role"`
+		Content any    `json:"content"`
+	}
+	if err := json.Unmarshal(dumpWire, &wireSys); err != nil {
+		t.Fatalf("unmarshal wire message: %v", err)
+	}
+	if wireSys.Role != "system" {
+		t.Errorf("expected wire role system, got %q", wireSys.Role)
+	}
+	if _, isString := wireSys.Content.(string); !isString {
+		t.Errorf("expected wire system content to be a string, got %T: %s", wireSys.Content, dumpWire)
+	}
+}
+
+// One text block already collapsed to string content before the fix; multi-block user
+// messages must keep their content array (image/file parts depend on it).
+func TestSystemFlattenShapes(t *testing.T) {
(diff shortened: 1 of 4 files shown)


akshaydeo (commented):
Thanks for this one, @AdityaPainuli. The write-up is unusually good: you traced the failure to the exact conversion hop instead of patching the symptom at the provider, and the diagnosis is correct. I confirmed the upstream mechanism you describe in ollama's OpenAI compatibility layer, where a content array is expanded into one `api.Message` per part with the original role preserved ([ollama/openai/openai.go](https://github.com/ollama/ollama/blob/main/openai/openai.go)), so a 3-block system message really does arrive as three system messages. The role guard is the right call, the tests are well targeted, and I verified them in a clean worktree: `TestMultiBlockSystemToChatRequest` and `TestSystemFlattenShapes` pass (5 subtests), and `./schemas/... ./providers/openai/... ./providers/ollama/...` stays green at 1296 tests.

Nothing here blocks on correctness. What follows is one repo-convention gap, one accidental comment reattachment, and a few edges the guard leaves open.

## Findings

| 1 | Medium | `tests/e2e/api/collections/provider-harness.json` (missing) | Wire-visible `core/` change with no provider-harness case and no stated exemption | CONFIRMED |
| 2 | Medium | [core/schemas/mux.go:718](https://github.com/maximhq/bifrost/blob/fix/flatten-multiblock-system-6245/core/schemas/mux.go#L718) | The new function was inserted between `ToChatMessages`'s doc comment and its declaration, so the exported function is now undocumented and the comment describes the wrong function | CONFIRMED |
| 3 | Medium | `core/providers/openai/utils.go:179` | The same bug still reproduces through the native `/v1/chat/completions` ingress, which never touches `ToChatMessages` | CONFIRMED |
| 4 | Low | [core/schemas/mux.go:733](https://github.com/maximhq/bifrost/blob/fix/flatten-multiblock-system-6245/core/schemas/mux.go#L733) | A system message mixing text with one non-text block bails out entirely and still hits the original failure | CONFIRMED |
| 5 | Low | [core/schemas/mux.go:738](https://github.com/maximhq/bifrost/blob/fix/flatten-multiblock-system-6245/core/schemas/mux.go#L738) | Empty text blocks produce doubled separators: `["A", "", "B"]` joins to `"A\n\n\n\nB"` | CONFIRMED |
| 6 | Nit | [core/providers/anthropic/systemflatten_test.go:117](https://github.com/maximhq/bifrost/blob/fix/flatten-multiblock-system-6245/core/providers/anthropic/systemflatten_test.go#L117) | Two subtests exercise `schemas.ToChatMessages` directly with no Anthropic involvement, but live in the `anthropic` package | CONFIRMED |

### 1. Missing provider-harness case

`AGENTS.md` is explicit here:

> Any change under `core/` that a client can observe on the wire must land together with a case in `tests/e2e/api/collections/provider-harness.json`

and on the exemption:

> The narrow exemptions: changes with no wire-visible effect (comments, internal renames, log lines) and behaviour no HTTP request can reach. If a change is exempt, say so explicitly in the PR rather than leaving the omission unexplained.

This change is wire-visible by construction: the outbound `messages[0].content` flips from an array to a string, and your own test asserts exactly that. `.claude/skills/harness-test-writer/SKILL.md` has the insertion recipe. A case posting a 3-block `system` to `/anthropic/v1/messages` and asserting the string shape via `x-bf-send-back-raw-request: true` would pin it, and it is red before this change and green after.

### 3. The chat-completions ingress is still broken

`ToChatMessages` sits on the Responses to Chat hop only. A client that posts OpenAI chat-completions directly:

```json
{"model":"ollama/llama3.1","messages":[
  {"role":"system","content":[{"type":"text","text":"A"},{"type":"text","text":"B"}]},
  {"role":"user","content":"hi"}]}
```

never reaches your helper. `OllamaProvider.ChatCompletion` hands the request to `openai.HandleOpenAIChatCompletionRequest`, and `ConvertBifrostMessagesToOpenAIMessages` copies `Content: message.Content` verbatim (`core/providers/openai/utils.go:179`). So the array survives to ollama and produces the identical `system message must be at the beginning`.

```mermaid
flowchart LR
  A["/anthropic/v1/messages"] --> B[BifrostResponsesRequest]
  B --> C["ToChatRequest / ToChatMessages"]
  C -->|fixed by this PR| E[ollama]
  D["/v1/chat/completions"] --> F[BifrostChatRequest]
  F -->|content array passes through| E
```

I am flagging this as a follow-up rather than a blocker, since this PR closes the reported Claude Code path and the second entrypoint is a pre-existing gap. But it is worth deciding deliberately: the layer that actually knows about the openai-compat part-expansion constraint is the egress, so a small normalization in the openai-compat send path would cover both ingresses and make the mux-level flatten redundant. If you prefer to keep the fix in the mux, a note in the code saying the chat ingress is knowingly uncovered would help the next reader.

### 4. Mixed system content still falls off the cliff

The all-text requirement means one non-text block disables the whole flatten. I verified this against your branch:

```
mixed system (text + input_image): ContentStr=<nil>, ContentBlocks=2
```

so that request still expands into two system messages upstream. Anthropic's own `system` parameter is text-only, so the Claude Code path is safe, but `/v1/responses` can carry a system or developer message with an image block. Joining the text parts and keeping the non-text ones as a shorter array would degrade instead of falling back to the failing shape. Low priority, but worth a comment either way so the choice reads as deliberate.

### 5. Separator handling

`["A", "", "B"]` produces `"A\n\n\n\nB"`. Filtering blocks whose trimmed text is empty before joining keeps the prompt clean. Separately, `"\n\n"` is a good choice and matches `providers/anthropic/text.go:21` and `providers/bedrock/text.go:22`; note `providers/cohere/responses.go:1562` joins system content with `"\n"`, so the codebase is not consistent today. No change requested, just flagging it in case you want to align.

## Merge recommendation

**Merge after nits.** No confirmed correctness regression: findings 1 and 2 are quick in-PR fixes, 3 and 4 are gaps in coverage rather than breakage, and 5 and 6 are polish.

## Followups required

1. **In this PR (blocking per AGENTS.md)** - add the `tests/e2e/api/collections/provider-harness.json` case for the 3-block system shape, or state the exemption explicitly in the PR description. Splice it in surgically, the file is ~50k lines.
2. **In this PR** - `core/schemas/mux.go:718`, move the two `ToChatMessages` doc comment lines back down to immediately above `func ToChatMessages` at line 741.
3. **In this PR (nit)** - `core/schemas/mux.go:738`, skip blocks whose trimmed text is empty before joining.
4. **In this PR (nit)** - `core/providers/anthropic/systemflatten_test.go:117` and `:145`, move the two provider-agnostic subtests into a `core/schemas` test file and keep the Anthropic ingress test where it is.
5. **Follow-up PR** - `core/providers/openai/utils.go:179`, normalize multi-part system content at the openai-compat egress so the `/v1/chat/completions` ingress is covered too.
6. **Follow-up PR** - `core/schemas/mux.go:733`, handle mixed system content by joining the text parts instead of bailing out.

## Candidates checked and cleared

- **`cache_control` on system blocks is newly dropped** - REFUTED. The pre-existing array branch builds `ChatContentBlock{Type: chatBlockType, Text: block.Text}` and never copies `CacheControl`, so per-block cache control was already lost on this hop before your change. Your PR description says exactly this and it checks out. Restoring it is a separate pre-existing gap.
- **Images, files or audio in a system message get silently dropped by the flatten** - REFUTED. The loop returns `false` unless every block is text, so no non-text payload is ever discarded.
- **Prompt caching breaks for Anthropic-family targets reached via `ToChatRequest`** (Claude on Vertex, `core/providers/vertex/vertex.go:1465`) - REFUTED, same reason as above: nothing survived that hop to lose.
- **Response conversion regresses** - REFUTED. `ToChatMessages` is also called on `responsesResp.Output` at `core/schemas/mux.go:1494`, but output messages carry the assistant role, so the role guard makes the flatten a no-op there.
- **User and tool messages lose their content array** - REFUTED by the role guard and by your own subtest, which I ran.
- **Import cycle from importing `providers/openai` inside package `anthropic`** - REFUTED. `providers/openai` does not import `providers/anthropic`; the package compiles and all 5 tests pass.
- **Developer-role messages double-convert with `normalizeDeveloperRoleForChatFallback`** - REFUTED. That helper only rewrites the role after `ToChatMessages`, and the flatten already covers both roles.

core/schemas/mux.go
```diff
@@ -717,6 +717,27 @@ func (cm *ChatMessage) ToResponsesMessages() []ResponsesMessage {
 
 // ToChatMessages converts a slice of ResponsesMessages back to ChatMessages
 // This handles the aggregation of function_call messages back into assistant messages with tool calls
+// flattenSystemTextBlocks joins multi-block all-text system/developer content into a
```
akshaydeo, line 718: The new function landed between `ToChatMessages`'s doc comment and its declaration. Lines 718-719 (`// ToChatMessages converts a slice of ResponsesMessages back to ChatMessages` and the aggregation note) now read as the doc comment for `flattenSystemTextBlocks`, and the exported `ToChatMessages` at line 741 has none.

Fix: move those two lines back down so they sit immediately above `func ToChatMessages`, and let the flatten helper keep only its own four comment lines.

core/schemas/mux.go
```diff
@@ -717,6 +717,27 @@ func (cm *ChatMessage) ToResponsesMessages() []ResponsesMessage {
...
+func flattenSystemTextBlocks(role ChatMessageRole, blocks []ResponsesMessageContentBlock) (string, bool) {
+	if role != ChatMessageRoleSystem && role != ChatMessageRoleDeveloper {
+		return "", false
+	}
+	if len(blocks) < 2 {
+		return "", false
+	}
+	texts := make([]string, len(blocks))
+	for i, block := range blocks {
+		if (block.Type != ResponsesInputMessageContentBlockTypeText && block.Type != ResponsesOutputMessageContentTypeText) || block.Text == nil {
```
akshaydeo, line 731: One non-text block disables the flatten for the whole message, so that request falls back to the exact shape that fails. I ran this against your branch:

```
system = [ {input_text "A"}, {input_image ...} ]
-> ContentStr = <nil>, ContentBlocks = 2
```

ollama then expands that into two system messages and the original error returns. Anthropic's `system` field is text-only so the Claude Code path is safe, but `/v1/responses` can send a system or developer message with an image block.

A degrade-instead-of-bail version keeps the win:

```go
// join runs of text, keep non-text blocks as a (shorter) array
```

If you would rather keep the strict guard, a one-line comment saying the mixed case is knowingly left alone would make the choice explicit for the next reader.

core/schemas/mux.go
```diff
@@ -717,6 +717,27 @@ func (cm *ChatMessage) ToResponsesMessages() []ResponsesMessage {
...
+			return "", false
+		}
+		texts[i] = *block.Text
+	}
+	return strings.Join(texts, "\n\n"), true
```
akshaydeo, line 739: Empty text blocks produce doubled separators. Verified on your branch: `["A", "", "B"]` joins to `"A\n\n\n\nB"`.

```go
texts := make([]string, 0, len(blocks))
for _, block := range blocks {
	if (block.Type != ResponsesInputMessageContentBlockTypeText && block.Type != ResponsesOutputMessageContentTypeText) || block.Text == nil {
		return "", false
	}
	if strings.TrimSpace(*block.Text) == "" {
		continue
	}
	texts = append(texts, *block.Text)
}
```

Separately, `"\n\n"` is a good pick and matches `providers/anthropic/text.go:21` and `providers/bedrock/text.go:22`. Worth knowing that `providers/cohere/responses.go:1562` joins system content with `"\n"`, so the repo is not consistent today. No change requested on the separator itself.

core/providers/anthropic/systemflatten_test.go
```diff
@@ -0,0 +1,174 @@
...
+		sys := chatReq.Input[0]
+		if sys.Role != schemas.ChatMessageRoleSystem {
+			t.Fatalf("expected system first, got %s", sys.Role)
+		}
+		if sys.Content == nil || sys.Content.ContentStr == nil || *sys.Content.ContentStr != "only block" {
+			t.Fatalf("expected single-block system to be plain string content, got: %+v", sys.Content)
+		}
+	})
+
+	t.Run("multi-block developer message flattens to string", func(t *testing.T) {
```
akshaydeo: This subtest and the `multi-block user message keeps content array` one at line 145 call `schemas.ToChatMessages` directly and construct `ResponsesMessage` values by hand, with no Anthropic types involved. They are really `core/schemas` tests living in the provider package, which makes the flatten behaviour harder to find from `mux.go`.

Suggestion: move these two into a `core/schemas` test file next to the other mux tests, and keep `TestMultiBlockSystemToChatRequest` here since it genuinely exercises the Anthropic ingress into the chat fallback. That split also documents the two layers separately.

AdityaPainuli:
Thanks for the review @akshaydeo, all four in-PR items are in. On honest harness case I used deepseek instead of ollama since its a real chat-fallback provider already wired into the harness, and it landed as folder 59 ( dev took 57 and 58 while this was in review, rebased on top). Agreed on both follow-ups, the `/v1/chat/completions` egress normalization and mixed system content. I can pick those up as separate PRs once this lands.

Issue #6276, opened 2026-08-18
[Bug]: Gemini 3.6 Flash catalog omits function-calling support and compat drops tools


toliv:
### Prerequisites

- [x] I have searched existing issues and discussions to avoid duplicates
- [x] I am using the latest version (or have tested against main/nightly)

### Description

`gemini-3.6-flash` supports function tools with JSON-schema parameters and required/forced tool choice through Vertex, but Bifrost's public model-parameters row does not declare those capabilities. With compat `should_drop_params: true`, Bifrost therefore removes `tools` and `tool_choice` before the request reaches Vertex.

The row currently contains only pricing:

```json
{
  "input_cost_per_token_batches": 3.75e-7,
  "output_cost_per_token_batches": 1.875e-6
}
```

The primary fix is to update the hosted `gemini-3.6-flash` row to include at least:

```json
{
  "supports_function_calling": true,
  "supports_tool_choice": true
}
```

Any other Gemini 3.6 Flash capabilities should be added when independently verified. In particular, function tools with JSON-schema parameters should not be conflated with OpenAI's provider-specific `strict: true` field.

There is also a secondary parser issue: `extractSupportedParams` adds `reasoning_with_tool_calls` when that flag is absent. Because this internal marker makes the otherwise empty result non-empty, `applyModelParameters` stores it and the compat plugin treats it as a complete allowlist.

Workaround: set `should_drop_params: false`.

As defensive hardening, only add the default `reasoning_with_tool_calls` marker after a row has declared at least one real supported request parameter. This preserves the backward-compatible default from #4630 for populated rows while leaving capability-empty rows unknown.

### Steps to reproduce

1. Start Bifrost HTTP v1.6.11 with the default public `model_parameters_url`. A local container synced 9,942 rows.
2. Confirm the downloaded row:

   ```shell
   curl 'http://localhost:8080/api/models/parameters?model=gemini-3.6-flash'
   ```

3. On unmodified `dev`, pass that JSON through `Store.applyModelParameters` and call `GetSupportedParameters("gemini-3.6-flash")`.
4. Seed a compat plugin with the returned list and pass a Vertex chat request containing one function tool and `tool_choice: "required"` through `PreLLMHook`.
5. Repeat step 4 with only `should_drop_params` changed to false.

Local output:

```text
cost-only row supported params = [reasoning_with_tool_calls]

should_drop_params=true  tools_present=false tool_choice_present=false dropped=[tool_choice tools]
[...]

PR #4630, opened 2026-06-23
feat: drop reasoning when tools present but reasoning_with_tool_calls unsupported

sammaji:
## Summary

Some models that support reasoning do not support using reasoning alongside tool calls. This PR introduces a `reasoning_with_tool_calls` capability flag so that reasoning parameters can be selectively dropped when tools are present but the model doesn't support that combination.

## Changes

- Added `SupportsReasoningWithToolCalls` field to `modelParametersParseResult`, allowing datasheets to explicitly declare whether a model supports reasoning when tool calls are also present.
- In `extractSupportedParams`, `reasoning_with_tool_calls` is added as a supported param by default (when the field is `nil`), preserving backward compatibility. It is only excluded when explicitly set to `false`.
- Updated `dropUnsupportedParams` in the compat plugin to drop `reasoning` not only when the model doesn't support reasoning at all, but also when tools are present and the model doesn't support `reasoning_with_tool_calls`.
- Updated a stale comment referencing `buildSupportedOutputsIndex` to correctly reference `extractSupportedParams`.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [x] Core (Go)
- [ ] Transports (HTTP)
- [ ] Providers/Integrations
- [x] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

Send a request with both `reasoning` and `tools` parameters to a model whose datasheet has `supports_reasoning_with_tool_calls` set to `false`. Verify that the `reasoning` parameter is dropped and the request proceeds without error.

Send the same request to a model without `supports_reasoning_with_tool_calls` set (i.e., `nil`) and verify that `reasoning` is preserved, confirming backward-compatible behavior.

```sh
go test ./...
```

## Breaking changes

- [ ] Yes
- [x] No

## Security considerations

None.

## Checklist

- [x] I read `docs/contributing/README.md` and followed the guidelines
- [x] I added/updated tests where appropriate
- [x] I updated documentation where needed
- [x] I verified builds succeed (Go and UI)
- [x] I verified the CI pipeline passes locally if applicable

PR #6368, opened 2026-08-20
[fix]: skip default reasoning_with_tool_calls marker for capability-empty datasheet rows

AdityaPainuli:
## Summary

A datasheet row that declares only pricing silently kills tool calling for that model.

The issue's original row (`gemini-3.6-flash`) carried two cost keys and nothing else. `extractSupportedParams` still added the default `reasoning_with_tool_calls` marker to it, which turned an otherwise empty capability list into a one-element allowlist. `applyModelParameters` stores any non-empty list, the compat plugin treats a non-nil list as authoritative, and with `should_drop_params: true` it stripped `tools`, `tool_choice`, and everything else from requests to that model. The model supports function calling fine; Bifrost just never asked.

That row has since been fixed upstream; `claude-opus-4-7-20260416` (`{"deprecation_date": "2027-04-16"}`, no capability fields) still reproduces on the live feed today, along with four other rows.

This PR fixes the code half of #6276. The hosted row itself still needs its capability flags added, but after this change an incomplete row degrades to "unknown, don't drop" instead of an allowlist of one marker.

## Changes

- `framework/modelcatalog/datasheet/types.go`: the default `reasoning_with_tool_calls` marker (from #4630) is now only added when the row declared a parameter surface at all: any `model_parameters` entry or any non-nil `supports_*` flag that maps to a request parameter, explicit false included. A row saying only `"supports_function_calling": false` is making an authoritative statement and keeps its allowlist; a row saying nothing about parameters (cost-only, deprecation-only, mode-only) degrades to "unknown, don't drop".
- An explicit `supports_reasoning_with_tool_calls` value is honored either way, true or false, including false overriding a marker sourced from a `model_parameters` id.
- No compat-side change needed: capability-empty rows now produce an empty list, `applyModelParameters` only stores non-empty lists, `GetSupportedParameters` returns nil, and compat's existing `!= nil` guard already skips dropping for nil.
- Populated rows lacking the flag keep the backward-compatible default from #4630 unchanged.
- `framework/modelcatalog/datasheet/costonlyparams_test.go`: regression tests for the cost-only row (unit and end-to-end through `applyModelParameters`), the explicit-false-only row keeping an authoritative allowlist, the populated-row default, and both explicit flag values. The first two fail without the fix.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [ ] Core (Go)
- [ ] Transports (HTTP)
- [ ] Providers/Integrations
- [x] Plugins
- [ ] UI (React)
- [ ] Docs

The change itself is in Framework (model catalog); the visible behavior change is in the compat plugin's drop decisions.

## How to test

```sh
cd framework
go test -run TestCostOnlyRowProducesNoParamAllowlist -v ./modelcatalog/datasheet/
go test ./modelcatalog/...
cd .. && go test ./plugins/compat/
```

Or live: with the default public `model_parameters_url` and `should_drop_params: true`, send a chat request to `claude-opus-4-7-20260416` with one function tool. Before this change `tools` and `tool_choice` are dropped before the request leaves Bifrost; after it they pass through. (The issue's original row, `gemini-3.6-flash`, no longer reproduces since its upstream row gained capability flags.)

## Screenshots/Recordings

N/A

## Breaking changes

- [ ] Yes
- [x] No

Cost-only rows also disappear from `/v1/models` `supported_parameters` and the qualified-key suffix index; those entries carried no real data.

## Related issues

Fixes the parser half of #6276. Datasheet row updates for the remaining capability-empty rows (`claude-opus-4-7-20260416` and friends) are maintainer-side.

## Security considerations

None.

## Checklist

### DIFF
--- a/framework/modelcatalog/datasheet/types.go
+++ b/framework/modelcatalog/datasheet/types.go
@@ -515,9 +515,6 @@ func extractSupportedParams(parsed *schemas.ModelCapabilities) []string {
 	if parsed.SupportsReasoning != nil && *parsed.SupportsReasoning {
 		addParam("reasoning")
 	}
-	if parsed.SupportsReasoningWithToolCalls == nil || *parsed.SupportsReasoningWithToolCalls {
-		addParam("reasoning_with_tool_calls")
-	}
 	if parsed.SupportsNoneReasoningEffort != nil && *parsed.SupportsNoneReasoningEffort {
 		addParam("supports_none_reasoning_effort")
 	}
@@ -537,6 +534,20 @@ func extractSupportedParams(parsed *schemas.ModelCapabilities) []string {
 	if parsed.SupportsWebSearch != nil && *parsed.SupportsWebSearch {
 		addParam("web_search")
 		addParam("web_search_options")
+	}
+
+	// reasoning_with_tool_calls defaults on for rows that lack the flag, so models
+	// missing it don't get reasoning stripped when tools are present. Restrict the
+	// default to rows that declared at least one capability: for a capability-empty
+	// (e.g. cost-only) row the marker alone would otherwise become a one-element
+	// allowlist, and compat with should_drop_params would strip every other
+	// parameter (tools, tool_choice, ...) from requests.
+	if parsed.SupportsReasoningWithToolCalls != nil {
+		if *parsed.SupportsReasoningWithToolCalls {
+			addParam("reasoning_with_tool_calls")
+		}
+	} else if len(supported) > 0 {
+		addParam("reasoning_with_tool_calls")
 	}
 
 	return supported
--- a/framework/modelcatalog/datasheet/types.go
+++ b/framework/modelcatalog/datasheet/types.go
@@ -536,12 +536,32 @@ func extractSupportedParams(parsed *schemas.ModelCapabilities) []string {
 		addParam("web_search_options")
 	}
 
+	// declaredParamSurface reports whether the row said anything at all about the
+	// request-parameter surface, including explicit "supports X: false". That is a
+	// different set from "produced a parameter name above": a row carrying only
+	// explicit-false flags produces no names but IS making an authoritative
+	// statement, and must keep a (possibly empty-of-tools) allowlist rather than
+	// degrade to "unknown, do not drop".
+	declaredParamSurface := len(parsed.ModelParameters) > 0 ||
+		parsed.SupportsAssistantPrefill != nil ||
+		parsed.SupportsFunctionCalling != nil ||
+		parsed.SupportsParallelFunctionCalling != nil ||
+		parsed.SupportsToolChoice != nil ||
+		parsed.SupportsReasoning != nil ||
+		parsed.SupportsResponseSchema != nil ||
+		parsed.SupportsNoneReasoningEffort != nil ||
+		parsed.SupportsServiceTier != nil ||
+		parsed.SupportsPromptCaching != nil ||
+		parsed.SupportsWebSearch != nil
+
 	// reasoning_with_tool_calls defaults on for rows that lack the flag, so models
 	// missing it don't get reasoning stripped when tools are present. Restrict the
-	// default to rows that declared at least one capability: for a capability-empty
-	// (e.g. cost-only) row the marker alone would otherwise become a one-element
-	// allowlist, and compat with should_drop_params would strip every other
-	// parameter (tools, tool_choice, ...) from requests.
+	// default to rows that declared a parameter surface: for a row that said
+	// nothing about parameters (cost-only, deprecation-only, mode-only) the marker
+	// alone would otherwise become a one-element allowlist, and compat with
+	// should_drop_params would strip every other parameter (tools, tool_choice,
+	// ...) from requests. This block must stay last in the function: the explicit
+	// false below removes the marker regardless of how it was added above.
 	if parsed.SupportsReasoningWithToolCalls != nil {
 		if *parsed.SupportsReasoningWithToolCalls {
 			addParam("reasoning_with_tool_calls")
@@ -552,7 +572,7 @@ func extractSupportedParams(parsed *schemas.ModelCapabilities) []string {
 				return p == "reasoning_with_tool_calls"
 			})
 		}
-	} else if len(supported) > 0 {
+	} else if declaredParamSurface {
 		addParam("reasoning_with_tool_calls")
 	}
 
--- a/framework/modelcatalog/datasheet/types.go
+++ b/framework/modelcatalog/datasheet/types.go
@@ -545,6 +545,12 @@ func extractSupportedParams(parsed *schemas.ModelCapabilities) []string {
 	if parsed.SupportsReasoningWithToolCalls != nil {
 		if *parsed.SupportsReasoningWithToolCalls {
 			addParam("reasoning_with_tool_calls")
+		} else {
+			// Explicit false wins even when a model_parameters id already added
+			// the marker via the default case above.
+			supported = slices.DeleteFunc(supported, func(p string) bool {
+				return p == "reasoning_with_tool_calls"
+			})
 		}
 	} else if len(supported) > 0 {
 		addParam("reasoning_with_tool_calls")
--- a/framework/modelcatalog/datasheet/costonlyparams_test.go
+++ b/framework/modelcatalog/datasheet/costonlyparams_test.go
@@ -15,8 +15,11 @@ import (
 // (should_drop_params) then stripped tools / tool_choice / temperature / everything
 // else from requests to that model.
 func TestCostOnlyRowProducesNoParamAllowlist(t *testing.T) {
-	// Hosted-datasheet shape reported in the issue for gemini-3.6-flash:
-	// pricing keys only, none of which map to ModelCapabilities fields.
+	// Hosted-datasheet shape reported in the issue: pricing keys only, none of
+	// which map to ModelCapabilities fields. The issue's original row
+	// (gemini-3.6-flash) has since been fixed upstream; claude-opus-4-7-20260416
+	// ({"deprecation_date": "2027-04-16"}) still reproduces on the live feed and
+	// parses to the same empty ModelCapabilities.
 	costOnlyRow := json.RawMessage(`{
 		"input_cost_per_token_batches": 0.000000075,
 		"output_cost_per_token_batches": 0.0000003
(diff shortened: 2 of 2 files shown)


framework/modelcatalog/datasheet/types.go
```diff
@@ -539,6 +536,26 @@ func extractSupportedParams(parsed *schemas.ModelCapabilities) []string {
...
+		if *parsed.SupportsReasoningWithToolCalls {
+			addParam("reasoning_with_tool_calls")
+		} else {
+			// Explicit false wins even when a model_parameters id already added
+			// the marker via the default case above.
+			supported = slices.DeleteFunc(supported, func(p string) bool {
+				return p == "reasoning_with_tool_calls"
+			})
+		}
+	} else if len(supported) > 0 {
```
akshaydeo: ### 1. `supportedParams` is also the qualified-key suffix index

`params.go:378` iterates `s.supportedParams` to resolve a bare model name to provider-qualified datasheet keys, and the comment at `params.go:235` names that second role explicitly: `supportedParams` feeds "the compat parameter allowlist and the qualified-key suffix index". Before this change the unconditional marker put essentially every row into that map. After it, a row that is not capability-empty but produces no parameter name drops out of it.

Reproduced on both sides with a scratch probe on a `{"mode":"chat","supports_cache_point":true,"supports_system_messages":true,"beta_headers":{...}}` row keyed `vertex_ai/curated-model`:

```
base 0356a56: supportedParams keys=[vertex_ai/curated-model gpt-4o]
              candidates for bare "curated-model" = [curated-model vertex_ai/curated-model]
this branch:  supportedParams keys=[gpt-4o]
              candidates for bare "curated-model" = [curated-model]
```

Failure scenario: a caller sends `gen4.5`, `ResolveModelParameters` and `LoadModelCapabilities` walk `modelParameterCandidates`, and `runwayml/gen4.5` is no longer offered, so the row is not found. Impact today is small, since the 18 affected rows carry only `mode`, but the coupling is invisible from the changed line and it gets worse the first time a curated row carries `server_tools`, `beta_headers` or `supports_cache_point` without any of the eleven parameter-mapping flags.

Suggested fix, as a follow-up: give the suffix lookup its own key set populated for every parsed row in `applyModelParameters`, so name resolution and the compat allowlist stop sharing one map.

### 2. Capability absent versus capability explicitly false

The contract downstream is two-state: nil means "unknown, do not drop", non-nil means "authoritative allowlist" (`plugins/compat/main.go:147`). A row that says `{"supports_function_calling": false}` and nothing else is making a statement about the parameter surface. On `dev` it produced a one-element allowlist, so `tools` was dropped. On this branch it produces an empty list, is not indexed, `GetSupportedParameters` returns nil, and compat skips dropping entirely, so `tools` now reaches a model whose row explicitly says it has no function calling. That is the same conflation this PR is fixing, pointed the other way.

Does this flip behaviour for existing datasheet rows? For the public feed, no: all 13 explicit-false rows also declare true flags, so they keep a non-empty allowlist and behave exactly as before. The rows that do flip are the 5 capability-empty ones (the intended fix) and the 18 `mode`-only ones (finding 1). But `model_parameters_url` is user-configurable, and hand-written self-hosted feeds are exactly where a bare "supports X: false" row shows up.

The discriminator that separates all three cases cleanly is not "any capability field" and not "any parameter produced", but "did the row say anything about the request-parameter surface":

```go
declaredParamSurface := len(parsed.ModelParameters) > 0 ||
	parsed.SupportsAssistantPrefill != nil ||
	parsed.SupportsFunctionCalling != nil ||
	parsed.SupportsParallelFunctionCalling != nil ||
	parsed.SupportsToolChoice != nil ||
	parsed.SupportsReasoning != nil ||
	parsed.SupportsResponseSchema != nil ||
	parsed.SupportsNoneReasoningEffort != nil ||
	parsed.SupportsServiceTier != nil ||
	parsed.SupportsPromptCaching != nil ||
	parsed.SupportsWebSearch != nil
```

Then `} else if declaredParamSurface {`. Cost-only, deprecation-only and `mode`-only rows say nothing about parameters, so they stay unknown, which is the fix you want. Explicit-false rows keep their authoritative allowlist. Populated rows keep the `#4630` default. If you take this, finding 3 resolves too, because the comment then describes the code.

### 3 and 4: one-line comment nits

The comment at line 542 promises a capability check and the code at line 555 does a parameter-name check. Either adjust the wording or adopt the predicate above. And since the gate reads the accumulated `supported`, it silently depends on staying the last block in `extractSupportedParams`. A short "keep this block last" note would save the next person a debugging session.

akshaydeo (commented):
Thanks @AdityaPainuli - this is a well-diagnosed fix with a clear write-up and focused regression tests, and the root cause analysis in the description is correct.

## Overview

The PR moves the default `reasoning_with_tool_calls` marker to the end of `extractSupportedParams` and only applies it when the row already produced at least one parameter name, so a capability-empty datasheet row degrades to "unknown, do not drop" instead of a one-element allowlist. The direction is right and the change is small and well-tested. Two things are worth a second look: the guard keys off "produced a parameter name" rather than the "declared a capability" wording in its own comment, and `supportedParams` doubles as the backing set for the qualified-key suffix index, so shrinking it has a second effect the PR does not mention.

## Verification against the live datasheet

I fetched the default feed at `https://getbifrost.ai/datasheet/model-parameters` (9945 rows, fetched 2026-08-23) and diffed `extractSupportedParams` behaviour between the merge base (`0356a56`) and this branch.

- **The bug is real and still live, on a different row than the one in the description.** `gemini-3.6-flash` has since been fixed upstream and now carries `"supports_function_calling": true`. But five rows still parse to an empty `ModelCapabilities`: `claude-opus-4-7-20260416` (`{"deprecation_date": "2027-04-16"}`), `claude-3-7-sonnet-20250219`, `claude-3-opus-20240229`, `claude-opus-4-20250514`, and `gemini-embedding-2`. Note that `deprecation_date`, `is_deprecated` and `max_input_tokens` are not fields on `schemas.ModelCapabilities`, which is why those rows parse to the zero value. On `dev` today, `claude-opus-4-7-20260416` gets `supportedParams = ["reasoning_with_tool_calls"]`, so with `should_drop_params: true` every parameter including `tools` is stripped from requests to it. This PR fixes that. It would be worth updating the description repro to a row that still reproduces.
- **18 rows lose their `supportedParams` entry without being capability-empty** - all provider-qualified image and video rows carrying only `mode` (`runwayml/gen4.5`, `replicate/black-forest-labs/flux-1.1-pro`, `replicate/bytedance/seedream-4.5`, and 15 similar). See finding 1.
- **The `slices.DeleteFunc` branch matches 0 of 9945 rows today.** No row combines `"supports_reasoning_with_tool_calls": false` with a `model_parameters` entry whose id is `reasoning_with_tool_calls`, so that branch is purely defensive right now. 13 rows carry the explicit `false` (the `gpt-5.4` and `gpt-5.5` family) and all of them also declare true flags, so none of them changes behaviour.

`go test ./modelcatalog/...` passes on this branch (459 tests, 4 packages).

## Findings

|---|---|---|---|---|
| 1 | Medium | `framework/modelcatalog/datasheet/types.go:555` | `supportedParams` also backs the qualified-key suffix index in `modelParameterCandidates`, so 18 `mode`-only rows drop out of bare-name resolution | CONFIRMED |
| 2 | Medium | `framework/modelcatalog/datasheet/types.go:555` | `len(supported) > 0` cannot distinguish "said nothing about parameters" from "explicitly said false", so an explicit-false-only row now drops nothing | PLAUSIBLE |
| 3 | Low | `framework/modelcatalog/datasheet/types.go:542` | Comment says "declared at least one capability" but the code tests "produced at least one parameter name" - these are different sets | CONFIRMED |
| 4 | Low | `framework/modelcatalog/datasheet/types.go:555` | The gate is position-dependent: any future `addParam` placed after this block silently changes what the default sees | CONFIRMED |

### 1. `supportedParams` is also the qualified-key suffix index

`params.go:378` iterates `s.supportedParams` to resolve a bare model name to provider-qualified datasheet keys, and the comment at `params.go:235` names that second role explicitly: `supportedParams` feeds "the compat parameter allowlist and the qualified-key suffix index". Before this change the unconditional marker put essentially every row into that map. After it, a row that is not capability-empty but produces no parameter name drops out of it.

Reproduced on both sides with a scratch probe on a `{"mode":"chat","supports_cache_point":true,"supports_system_messages":true,"beta_headers":{...}}` row keyed `vertex_ai/curated-model`:

```
base 0356a56: supportedParams keys=[vertex_ai/curated-model gpt-4o]
              candidates for bare "curated-model" = [curated-model vertex_ai/curated-model]
this branch:  supportedParams keys=[gpt-4o]
              candidates for bare "curated-model" = [curated-model]
```

Failure scenario: a caller sends `gen4.5`, `ResolveModelParameters` and `LoadModelCapabilities` walk `modelParameterCandidates`, and `runwayml/gen4.5` is no longer offered, so the row is not found. Impact today is small, since the 18 affected rows carry only `mode`, but the coupling is invisible from the changed line and it gets worse the first time a curated row carries `server_tools`, `beta_headers` or `supports_cache_point` without any of the eleven parameter-mapping flags.

Suggested fix, as a follow-up: give the suffix lookup its own key set populated for every parsed row in `applyModelParameters`, so name resolution and the compat allowlist stop sharing one map.

### 2. Capability absent versus capability explicitly false

The contract downstream is two-state: nil means "unknown, do not drop", non-nil means "authoritative allowlist" (`plugins/compat/main.go:147`). A row that says `{"supports_function_calling": false}` and nothing else is making a statement about the parameter surface. On `dev` it produced a one-element allowlist, so `tools` was dropped. On this branch it produces an empty list, is not indexed, `GetSupportedParameters` returns nil, and compat skips dropping entirely, so `tools` now reaches a model whose row explicitly says it has no function calling. That is the same conflation this PR is fixing, pointed the other way.

Does this flip behaviour for existing datasheet rows? For the public feed, no: all 13 explicit-false rows also declare true flags, so they keep a non-empty allowlist and behave exactly as before. The rows that do flip are the 5 capability-empty ones (the intended fix) and the 18 `mode`-only ones (finding 1). But `model_parameters_url` is user-configurable, and hand-written self-hosted feeds are exactly where a bare "supports X: false" row shows up.

The discriminator that separates all three cases cleanly is not "any capability field" and not "any parameter produced", but "did the row say anything about the request-parameter surface":

```go
declaredParamSurface := len(parsed.ModelParameters) > 0 ||
	parsed.SupportsAssistantPrefill != nil ||
	parsed.SupportsFunctionCalling != nil ||
	parsed.SupportsParallelFunctionCalling != nil ||
	parsed.SupportsToolChoice != nil ||
	parsed.SupportsReasoning != nil ||
	parsed.SupportsResponseSchema != nil ||
	parsed.SupportsNoneReasoningEffort != nil ||
	parsed.SupportsServiceTier != nil ||
	parsed.SupportsPromptCaching != nil ||
	parsed.SupportsWebSearch != nil
```

Then `} else if declaredParamSurface {`. Cost-only, deprecation-only and `mode`-only rows say nothing about parameters, so they stay unknown, which is the fix you want. Explicit-false rows keep their authoritative allowlist. Populated rows keep the `#4630` default. If you take this, finding 3 resolves too, because the comment then describes the code.

### 3 and 4: one-line comment nits

The comment at line 542 promises a capability check and the code at line 555 does a parameter-name check. Either adjust the wording or adopt the predicate above. And since the gate reads the accumulated `supported`, it silently depends on staying the last block in `extractSupportedParams`. A short "keep this block last" note would save the next person a debugging session.

## Merge recommendation

**Merge after nits.** Nothing here is a confirmed regression that should block: the change fixes a live production bug on `claude-opus-4-7-20260416` and four other rows, and findings 1 and 2 affect 18 low-impact rows and 0 rows respectively in today's feed. Findings 3 and 4 are comment-only. I would take finding 2's predicate in this PR if it is cheap, since it is a two-line swap that also resolves finding 3.

## Followups required

1. **Follow-up PR** - `framework/modelcatalog/datasheet/params.go:378`: stop deriving the qualified-key suffix index from `supportedParams`; populate a dedicated key set for every parsed row in `applyModelParameters` so the compat allowlist and the name-resolution index can diverge safely. (finding 1)
2. **In this PR (non-blocking, recommended)** - `framework/modelcatalog/datasheet/types.go:555`: replace `len(supported) > 0` with the `declaredParamSurface` predicate above so an explicitly-false row keeps its authoritative allowlist. (finding 2)
3. **In this PR (non-blocking)** - `framework/modelcatalog/datasheet/types.go:542`: reword the comment to match whichever predicate lands, and note that the block must stay last in the function. (findings 3, 4)
4. **In this PR (non-blocking)** - PR description and `costonlyparams_test.go:24`: `gemini-3.6-flash` now carries `supports_function_calling: true` upstream, so the stated repro no longer reproduces. `claude-opus-4-7-20260416` does. Worth swapping so a future reader can reproduce it; the synthetic fixture itself is fine as a regression test.

## Checked and cleared

- **Explicit false now deletes a marker sourced from a `model_parameters` id** - real behaviour change, but 0 of 9945 live rows match the shape, it is called out in the description, and it is the more correct reading of an explicit false.
- **The `applied == 0` index-swap guard could be tripped by newly-empty lists** - refuted. `applyModelParameters` counts via `IsEmptyModelCapabilities(&caps)` on the parsed row, not via the parameter list, so the guard is untouched.
- **`/v1/models` `supported_parameters` regression** - refuted as a defect. `modelinfo.go:39` gates on `len(params) > 0`, so nil and empty are already indistinguishable there; the five affected rows only lose a field that held one synthetic marker.
- **Removing the marker strips reasoning for cost-only rows** - refuted. `plugins/compat/dropparams.go:78` only consults `reasoning_with_tool_calls` after `main.go:147` has confirmed a non-nil allowlist, and it is nil for exactly these rows.
- **`supportedResponseTypes` shrinks alongside** - refuted. That index is built from `supported_endpoints` and `mode` independently of `extractSupportedParams`, so the 18 `mode`-only rows keep their response-type entry.
- **Go filename convention** - refuted. AGENTS.md says "No underscores. The only permitted underscore is the `_test.go` suffix", and `costonlyparams_test.go` complies.
- **Ordering of the returned slice changed** - refuted as a defect. Consumers build a set (`isSupported`) or expose the list verbatim; there is no ordering contract and the package tests pass.

AdityaPainuli:
Took your declaredParamSurface predicate as-is, it's the right discriminator and the comment now matches the code, with a "keep this block last" note. Added a regression subset for explicit-false-only row keeping its authoritative allowlist, and swapped the repro in the description and test comments to claude-opus-4-7-20260416 since gemini-3.6-flash got fixed upstream. The suffix-index decoupling (finding 1) makes sense as a follow-up PR, happy to take it.

Issue #5820, opened 2026-08-04
[Bug]: Gemini thoughtSignature is emitted after Anthropic text as redacted_thinking


isac322:
### Prerequisites

- [x] I searched existing issues and found related but distinct items: #5259 covers Gemini `/v1/responses` terminal/reasoning completeness; #5274 covers incoming OpenAI-compatible reasoning replay to Anthropic.
- [x] The behavior below is on the Anthropic Messages streaming response path.

### Description

With `transports/v1.6.7` (which pins `core v1.7.5`), a Gemini/Vertex response that emits visible text followed by a non-empty Gemini `thoughtSignature` is serialized to an Anthropic Messages SSE stream in this order:

```text
text content block (index 0)
redacted_thinking content block (index 1, non-empty data)
message_delta(stop_reason=end_turn)
```

The visible text is valid, but the terminal non-visible `redacted_thinking` block makes Claude Code's internal WebFetch secondary summarizer return `No response from model`. Replaying the same preserved HTTP 200 SSE response with only that final `redacted_thinking` block removed returns the exact title text successfully.

This is not a request-side signature-replay issue. The signature must remain opaque and must be preserved for any later turn or tool continuation. The problem is the Gemini-to-Anthropic **streaming output ordering**.

Anthropic documents thinking blocks before text blocks and requires signatures to be replayed unchanged in multi-turn/tool-use conversations: https://platform.claude.com/docs/en/build-with-claude/thinking

### Relevant code

`core/providers/gemini/responses.go`, `processGeminiThoughtSignaturePart`, closes an open text item and then emits a separate `ResponsesReasoning.EncryptedContent` item:

- https://github.com/maximhq/bifrost/blob/core/v1.7.5/core/providers/gemini/responses.go#L1301-L1352

`core/providers/anthropic/responses.go` then projects that encrypted content to `redacted_thinking.data`:

- https://github.com/maximhq/bifrost/blob/core/v1.7.5/core/providers/anthropic/responses.go#L2649-L2670

Together these paths produce `text -> redacted_thinking` for a Gemini part sequence with a trailing signature.

### Minimal wire shape

```text
event: content_block_start
data: {"index":0,"content_block":{"type":"text","text":""}}

event: content_block_delta
data: {"index":0,"delta":{"type":"text_delta","text":"The title of the web page is \"Example Domain\"."}}

event: content_block_stop
data: {"index":0}

event: content_block_start
[...]

Issue #5259, opened 2026-07-15
[Bug]: Gemini streaming /v1/responses: response.completed carries no output array and reasoning items are never completed


fus3r:
### Prerequisites
- [x] I have searched existing issues and discussions to avoid duplicates
- [x] I am using the latest version (or have tested against main/nightly)

### Description
On the Gemini provider (and Vertex Gemini, which routes through the same stream handler), streaming `/v1/responses` never completes its terminal events. Three related gaps, all in `core/providers/gemini/responses.go`:
1. **`response.completed` carries no output array.** `closeGeminiOpenItems` builds the final response with id/usage/model/stop reason but never sets `Output`, and the stream state keeps no item bookkeeping, so the wire shows `"output":null`. Every item type is missing from the final snapshot: the text message, function calls, reasoning, web_search_call. The non-streaming path returns the same content correctly, and the Anthropic stream ingress populates `Output` from its `OutputItems` map precisely so that clients checking `Output` keep working, so this is a Gemini-only gap. Clients that read the terminal snapshot, like openai-python's `stream.get_final_response()`, see a response with no output at all.
2. **Reasoning terminal events are empty shells.** `processGeminiThoughtPart` streams the thought text as `reasoning_summary_text.delta`, but `reasoning_summary_text.done` carries no text and the reasoning `output_item.done` carries `"summary":[]`. Consumers that treat the done events or the completed snapshot as authoritative lose all reasoning. Each thought part also opens its own reasoning item (there is no open-item reuse like the text path has with `TextOutputIndex`/`TextBuffer`), so one turn's thinking fragments into several empty-done items.
3. **A `thoughtSignature` arriving on a thought part is dropped.** The stream dispatch routes any part with `thought:true` and text to `processGeminiThoughtPart`, which never reads `part.ThoughtSignature`, and the non-streaming converter has the same gap in its `part.Thought` case (a thought-flagged part carrying only a signature is dropped there entirely). Signature-only parts and function-call parts keep theirs, but a signature riding a thought part is lost on both paths. Gemini requires returning these signatures to continue thinking across turns with tools.

### Steps to reproduce
1. Configure the Gemini provider and stream a request with reasoning enabled:
```sh
curl -N http://localhost:8080/v1/responses -H 'Content-Type: application/json' -d '{"model":"gemini/gemini-2.5-flash","stream":true,"reasoning":{"effort":"low"},"input":"What is 330 metres in feet? Reply in one short sentence."}'
```
[...]

Issue #1977, opened 2026-03-06
[Bug]: Chat-to-Responses mux does not create "reasoning" output items from assistant reasoning content


colexbruhn:
### Prerequisites

- [x] I have searched existing issues and discussions to avoid duplicates
- [x] I am using the latest version (or have tested against main/nightly)

### Description

When converting chat completions to Responses API format via the mux path, `ToResponsesMessages()` does not create `type: "reasoning"` output items from `ChatAssistantMessage.Reasoning` or `ChatAssistantMessage.ReasoningDetails`. Reasoning content is silently dropped.

Native providers (Anthropic, Bedrock, Gemini, Cohere) all correctly produce reasoning output items, confirming this is the expected format. The types and structs needed (`ResponsesMessageTypeReasoning`, `ResponsesReasoning`, `ResponsesReasoningSummary`) all exist, but the mux code doesn't use them.

The streaming mux (`ToBifrostResponsesStreamResponse`) partially handles reasoning. It emits `ReasoningSummaryTextDelta` events, but the `response.completed` event's `Output` never includes a reasoning item.

Affected code:
- `ToResponsesMessages()` (mux.go) - handles tool calls, refusal, text content, content blocks, but has no code for `ChatAssistantMessage.Reasoning` or `ChatAssistantMessage.ReasoningDetails`
- `ToBifrostResponsesStreamResponse()` (mux.go) - `response.completed` Output omits reasoning items

Proposed fix:
- Non-streaming: In `ToResponsesMessages`, check `ChatAssistantMessage.Reasoning`/`ReasoningDetails`. If non-empty, prepend a `ResponsesMessage{Type: ResponsesMessageTypeReasoning, ResponsesReasoning: &ResponsesReasoning{Summary: [...]}}` before the text message.
- Streaming: Track accumulated reasoning text in `ChatToResponsesStreamState`, include reasoning output item in `response.completed` Output.

### Steps to reproduce

1. Send a Responses API request to a mux-path provider (e.g. vLLM) with a reasoning-capable model
2. Observe the response output array

### Expected behavior

Output array contains a `type: "reasoning"` item with `summary: [{type: "summary_text", text: "..."}]` before the text message

### Actual behavior

Output array only contains the text message, and reasoning content is dropped

### Affected area(s)

Core (Go)

### Version

v1.4.7

### Environment

```text

```

### Relevant logs/output

```shell

```

### Severity

Medium (some functionality impaired)

akshaydeo:
[...]

PR #6369, opened 2026-08-20
[fix]: set id and summary on Gemini non-streaming Responses reasoning items

### LINKED ISSUE #6329: [Bug]: Gemini/Vertex non-streaming /v1/responses emits reasoning items without id and summary
### Prerequisites

### Description

When Bifrost routes a non-streaming `/v1/responses` request to Gemini through Vertex with reasoning and tools enabled, it can return a malformed reasoning output item:

```json
{
  "type": "reasoning",
  "role": "assistant",
  "content": [
    {
      "type": "reasoning_text",
      "text": "<redacted reasoning>"
    }
  ]
}
```

This item is missing both `id` and `summary`, which are required by strict OpenAI Responses clients.

The provider returns HTTP 200 and the response can contain a valid `function_call` after the reasoning item. However, the client rejects the complete response envelope before the function call can be consumed.

With the AI SDK, this produces:

```text
AI_APICallError: Invalid JSON response

Caused by AI_TypeValidationError:
- output[0].id: expected string, received undefined
- output[0].summary: expected array, received undefined
```

This breaks the agent/tool conversation even though Gemini generated a valid tool call. Validation libraries may also include the complete invalid response—including reasoning text—in application error logs.

Related but distinct issues:

- #5259 concerns incomplete Gemini streaming Responses events.
- #1977 concerns reasoning items missing from the Chat-to-Responses mux.
- #5820 concerns Gemini thought-signature ordering on the Anthropic surface.

The current Gemini non-streaming conversion appears to construct unsigned thought items without populating the OpenAI-compatible reasoning fields:

https://github.com/maximhq/bifrost/blob/dev/core/providers/gemini/responses.go#L3213-L3251

### Steps to reproduce

1. Configure Bifrost with a Vertex provider and a reasoning-capable Gemini model.

2. Send a non-streaming Responses request that requires a function call:

```bash
curl "$BIFROST_URL/v1/responses" \
  -d '{
    "model": "vertex/gemini-3.5-flash-lite",
    "input": "Call lookup_value with the value example.",
    "stream": false,
    "store": false,
    "reasoning": {
      "effort": "low",
      "summary": "auto"
    },
    "tools": [
      {
        "type": "function",
        "name": "lookup_value",
        "description": "Look up a value",
        "parameters": {
          "type": "object",
          "properties": {
            "value": {
              "type": "string"
            }
          },
          "required": ["value"],
          "additionalProperties": false
        }
      }
    ],
    "tool_choice": {
      "type": "function",
      "name": "lookup_value"
    }
  }'
```

3. Inspect the returned `output` array.

4. Parse the response using the AI SDK OpenAI Responses provider.

5. Observe that validation fails on `output[0].id` and `output[0].summary`, preventing the following `function_call` from being returned to the application.

### Expected behavior

Bifrost should return a schema-valid OpenAI Responses reasoning item containing an id and summary array, while preserving encrypted_content when a Gemini thought signature exists. The following function_call should remain consumable by strict OpenAI-compatible clients.

### Actual behavior

Bifrost returns a reasoning item with role and content/reasoning_text but without id or summary. The client rejects the complete HTTP 200 response before consuming the valid function_call that follows.

### Affected area(s)

Core (Go)

### Version

v1.6.7

### Environment

```text

```

### Relevant logs/output

```shell

```

### Regression?

_No response_

### Severity

AdityaPainuli:
## Summary

Non-streaming Gemini/Vertex `/v1/responses` returns reasoning items that strict OpenAI Responses clients refuse to parse.

The Gemini→Responses converter built reasoning items with no `id` at all, and set `summary` only when the part carried a thought signature. The Vercel AI SDK validates the whole envelope and rejects it:

```
output[0].id: expected string, received undefined
output[0].summary: expected array, received undefined
```

The response is HTTP 200 and contains a perfectly valid `function_call` after the reasoning item, but the client throws before it ever sees the tool call, so agent loops die on the first reasoning turn. The streaming path already sets `id` on its reasoning items; `summary` is absent there too, but the AI SDK's chunk schema for `output_item.added`/`.done` only validates `id` and `encrypted_content`, so only the non-streaming converter was actually breaking clients.

## Changes

- `core/providers/gemini/responses.go`: all three reasoning-item construction sites now set an id, using the same `"rs_" + GetRandomString(50)` pattern the anthropic, cohere, bedrock, and mux converters already use:
  - the `part.Thought` case in `convertGeminiCandidatesToResponsesOutput` (also sets `summary: []` unconditionally now, instead of only when a signature exists)
  - `reasoningFromThoughtSignature` (server-side tool-call signatures)
  - the standalone `part.ThoughtSignature != nil` case
- Thought signatures still ride in `encrypted_content`, base64-encoded, so the replay round trip is unchanged.
- Kept the fix minimal on purpose: reasoning text stays in the `reasoning_text` content block with an empty `summary` array, matching the existing repo-wide convention (anthropic does the same). Moving text into summaries would be a behavior change beyond this bug.
- Replay side (review follow-ups): `thoughtTextParts` reads reasoning content blocks first with `summary` as fallback, so making `ResponsesReasoning` non-nil doesn't lose thought text on the Responses→Gemini replay. Same content-first rule now applies in `ToGeminiResponsesResponse`, whose reasoning branch previously read `summary` unconditionally and emitted duplicate thought parts for an item carrying text in both places. The inline signature-only item construction was deduped into `reasoningFromThoughtSignature`.
- `core/providers/gemini/reasoningidsummary_test.go`: regression tests for thought parts with and without a signature and for standalone signature parts, asserting id presence, `rs_` prefix, summary array, signature preservation, and that the following `function_call` stays intact; round-trip tests proving thought text and signature survive replay to Gemini; and duplicate-part tests for `ToGeminiResponsesResponse`. All fail without their fix.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [x] Core (Go)
- [ ] Transports (HTTP)
- [x] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

```sh
cd core
go test -run TestReasoningItemIDAndSummary -v ./providers/gemini/
go test ./providers/gemini/ ./providers/vertex/
```

Or live: send a non-streaming `/v1/responses` request to a reasoning-capable Gemini model through Vertex with a forced function tool (repro in #6329), then parse the response with the AI SDK's OpenAI Responses provider. Before this change validation fails on `output[0].id` / `output[0].summary`; after it the reasoning item carries both and the function call is consumable.

## Screenshots/Recordings

N/A

## Breaking changes

- [ ] Yes
- [x] No

## Related issues

Closes #6329

Related but distinct: #5259 (streaming events), #1977 (chat-to-responses mux, already fixed), #5820 (signature ordering on the Anthropic surface).

## Security considerations

None.

## Checklist

### DIFF
--- a/core/providers/gemini/responses.go
+++ b/core/providers/gemini/responses.go
@@ -356,18 +356,36 @@ func thoughtTextParts(msg *schemas.ResponsesMessage) []*Part {
 	if msg == nil {
 		return nil
 	}
-	var parts []*Part
-	if msg.Content != nil {
-		for _, block := range msg.Content.ContentBlocks {
-			if block.Type == schemas.ResponsesOutputMessageContentTypeReasoning &&
-				block.Text != nil && *block.Text != "" {
-				parts = append(parts, &Part{Text: *block.Text, Thought: true})
-			}
-		}
-	}
-	if len(parts) > 0 || msg.ResponsesReasoning == nil {
+	if parts := reasoningBlockThoughtParts(msg); len(parts) > 0 {
 		return parts
 	}
+	return reasoningSummaryThoughtParts(msg)
+}
+
+// reasoningBlockThoughtParts renders the reasoning content blocks of a reasoning
+// item as Gemini thought parts.
+func reasoningBlockThoughtParts(msg *schemas.ResponsesMessage) []*Part {
+	if msg == nil || msg.Content == nil {
+		return nil
+	}
+	var parts []*Part
+	for _, block := range msg.Content.ContentBlocks {
+		if block.Type == schemas.ResponsesOutputMessageContentTypeReasoning &&
+			block.Text != nil && *block.Text != "" {
+			parts = append(parts, &Part{Text: *block.Text, Thought: true})
+		}
+	}
+	return parts
+}
+
+// reasoningSummaryThoughtParts renders a reasoning item's Summary array as Gemini
+// thought parts. Callers that have already emitted the item's content blocks
+// through another path use this directly instead of thoughtTextParts.
+func reasoningSummaryThoughtParts(msg *schemas.ResponsesMessage) []*Part {
+	if msg == nil || msg.ResponsesReasoning == nil {
+		return nil
+	}
+	var parts []*Part
 	for _, summaryBlock := range msg.ResponsesReasoning.Summary {
 		if summaryBlock.Text == "" {
 			continue
@@ -626,16 +644,13 @@ func ToGeminiResponsesResponse(bifrostResp *schemas.BifrostResponsesResponse) *G
 					continue
 				}
 
-				// Reasoning content is in the Summary array
-				if len(msg.ResponsesReasoning.Summary) > 0 {
-					for _, summaryBlock := range msg.ResponsesReasoning.Summary {
-						if summaryBlock.Text != "" {
-							currentParts = append(currentParts, &Part{
-								Text:    summaryBlock.Text,
-								Thought: true,
-							})
-						}
-					}
+				// The content-block loop above already emitted this item's reasoning
+				// content blocks as thought parts, so fall back to the Summary array
+				// only when the item carries no thought text in its content blocks
+				// (the same content-first rule as thoughtTextParts). Text present in
+				// both places must not be emitted twice.
+				if len(reasoningBlockThoughtParts(&msg)) == 0 {
+					currentParts = append(currentParts, reasoningSummaryThoughtParts(&msg)...)
 				}
 				if msg.ResponsesReasoning.EncryptedContent != nil {
 					decodedSig := thoughtSignatureFromEncryptedContent(msg.ResponsesReasoning.EncryptedContent)
@@ -3531,17 +3546,9 @@ func convertGeminiCandidatesToResponsesOutput(candidates []*Candidate) []schemas
 
 			case part.ThoughtSignature != nil:
 				// Handle thought signature
-				thoughtSig := base64.StdEncoding.EncodeToString(part.ThoughtSignature)
-				msg := schemas.ResponsesMessage{
-					ID:   schemas.Ptr("rs_" + schemas.GetRandomString(50)),
-					Role: schemas.Ptr(schemas.ResponsesInputMessageRoleAssistant),
-					Type: schemas.Ptr(schemas.ResponsesMessageTypeReasoning),
-					ResponsesReasoning: &schemas.ResponsesReasoning{
-						Summary:          []schemas.ResponsesReasoningSummary{},
-						EncryptedContent: &thoughtSig,
-					},
+				if msg, ok := reasoningFromThoughtSignature(part); ok {
+					messages = append(messages, msg)
 				}
-				messages = append(messages, msg)
 			}
 		}
 
--- a/core/providers/gemini/responses.go
+++ b/core/providers/gemini/responses.go
@@ -344,16 +344,31 @@ func (response *GenerateContentResponse) ToResponsesBifrostResponsesResponse() *
 	return bifrostResp
 }
 
-// thoughtTextParts renders a reasoning item's summary as Gemini thought parts.
+// thoughtTextParts renders a reasoning item's text as Gemini thought parts.
 //
 // Gemini's thinking guide requires thought blocks to be resent unmodified, so
 // wherever a reasoning item's signature is taken its text has to travel with it.
-func thoughtTextParts(reasoning *schemas.ResponsesReasoning) []*Part {
-	if reasoning == nil {
+// Gemini's own outbound converter stores thought text as reasoning content
+// blocks (with summary left an empty array for OpenAI-compat clients), while
+// OpenAI-ingress replay carries it in summary — so content blocks are read
+// first and summary is the fallback.
+func thoughtTextParts(msg *schemas.ResponsesMessage) []*Part {
+	if msg == nil {
 		return nil
 	}
 	var parts []*Part
-	for _, summaryBlock := range reasoning.Summary {
+	if msg.Content != nil {
+		for _, block := range msg.Content.ContentBlocks {
+			if block.Type == schemas.ResponsesOutputMessageContentTypeReasoning &&
+				block.Text != nil && *block.Text != "" {
+				parts = append(parts, &Part{Text: *block.Text, Thought: true})
+			}
+		}
+	}
+	if len(parts) > 0 || msg.ResponsesReasoning == nil {
+		return parts
+	}
+	for _, summaryBlock := range msg.ResponsesReasoning.Summary {
 		if summaryBlock.Text == "" {
 			continue
 		}
@@ -561,7 +576,7 @@ func ToGeminiResponsesResponse(bifrostResp *schemas.BifrostResponsesResponse) *G
 								// because the normal reasoning branch below also
 								// emits a signature-only part - that path would
 								// send the same signature twice.
-								consumedThoughtText = thoughtTextParts(nextMsg.ResponsesReasoning)
+								consumedThoughtText = thoughtTextParts(&nextMsg)
 							}
 						}
 					}
@@ -4151,7 +4166,7 @@ func convertResponsesMessagesToGeminiContents(messages []schemas.ResponsesMessag
 		// A reasoning message with no text still has nothing to add here, so it
 		// keeps being skipped.
 		if msg.Type != nil && *msg.Type == schemas.ResponsesMessageTypeReasoning && msg.ResponsesReasoning != nil {
-			parts := thoughtTextParts(msg.ResponsesReasoning)
+			parts := thoughtTextParts(&msg)
 
 			// The signature is carried by the PRECEDING function call's
 			// look-ahead - but only when there is one. A standalone signed
--- a/core/providers/gemini/responses.go
+++ b/core/providers/gemini/responses.go
@@ -3088,6 +3088,7 @@ func reasoningFromThoughtSignature(part *Part) (schemas.ResponsesMessage, bool) 
 	}
 	thoughtSig := base64.StdEncoding.EncodeToString(part.ThoughtSignature)
 	return schemas.ResponsesMessage{
+		ID:   schemas.Ptr("rs_" + schemas.GetRandomString(50)),
 		Role: schemas.Ptr(schemas.ResponsesInputMessageRoleAssistant),
 		Type: schemas.Ptr(schemas.ResponsesMessageTypeReasoning),
 		ResponsesReasoning: &schemas.ResponsesReasoning{
@@ -3225,6 +3226,7 @@ func convertGeminiCandidatesToResponsesOutput(candidates []*Candidate) []schemas
 				if part.Text != "" || len(part.ThoughtSignature) > 0 {
 					text := part.Text
 					msg := schemas.ResponsesMessage{
+						ID:   schemas.Ptr("rs_" + schemas.GetRandomString(50)),
 						Role: schemas.Ptr(schemas.ResponsesInputMessageRoleAssistant),
 						Content: &schemas.ResponsesMessageContent{
 							ContentBlocks: []schemas.ResponsesMessageContentBlock{
@@ -3235,6 +3237,13 @@ func convertGeminiCandidatesToResponsesOutput(candidates []*Candidate) []schemas
 							},
 						},
 						Type: schemas.Ptr(schemas.ResponsesMessageTypeReasoning),
+						// Strict OpenAI Responses clients require both id and a summary
+						// array on every reasoning item; set summary unconditionally so
+						// signature-less thoughts serialize as `"summary": []` instead of
+						// omitting the key.
+						ResponsesReasoning: &schemas.ResponsesReasoning{
+							Summary: []schemas.ResponsesReasoningSummary{},
+						},
 					}
 					if len(part.ThoughtSignature) > 0 {
 						// Stored base64-encoded, which is the form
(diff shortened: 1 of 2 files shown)


core/providers/gemini/responses.go
```diff
@@ -344,16 +344,31 @@ func (response *GenerateContentResponse) ToResponsesBifrostResponsesResponse() *
...
 //
 // Gemini's thinking guide requires thought blocks to be resent unmodified, so
 // wherever a reasoning item's signature is taken its text has to travel with it.
-func thoughtTextParts(reasoning *schemas.ResponsesReasoning) []*Part {
-	if reasoning == nil {
+// Gemini's own outbound converter stores thought text as reasoning content
+// blocks (with summary left an empty array for OpenAI-compat clients), while
+// OpenAI-ingress replay carries it in summary — so content blocks are read
+// first and summary is the fallback.
+func thoughtTextParts(msg *schemas.ResponsesMessage) []*Part {
```
akshaydeo, line 355: Nice complement to the id fix - without this, `summary: []` would have stopped thought text from reaching Gemini on replay.

One gap: the *other* replay reader was not updated. `ToGeminiResponsesResponse` (same file, around line 627) still does its own `for _, summaryBlock := range msg.ResponsesReasoning.Summary` loop and never looks at content blocks. So the two readers now disagree, and an item that carries the text in both places yields the part twice:

```
// input: reasoning item with Content[reasoning_text]="Reasoning here." AND Summary=["Reasoning here."]
// ToGeminiResponsesResponse ->
//   parts: [{"thought":true,"text":"Reasoning here."},{"thought":true,"text":"Reasoning here."}]
// convertResponsesMessagesToGeminiContents (this helper) ->
//   parts: [{"thought":true,"text":"Reasoning here."}]
```

Suggested fix in `ToGeminiResponsesResponse`, replacing the inline summary loop:

```go
currentParts = append(currentParts, thoughtTextParts(&msg)...)
```

That also unblocks the natural follow-up of populating `summary` with the thought text, which is currently unsafe for exactly this reason.

core/providers/gemini/responses.go
```diff
@@ -3512,6 +3533,7 @@ func convertGeminiCandidatesToResponsesOutput(candidates []*Candidate) []schemas
 				// Handle thought signature
 				thoughtSig := base64.StdEncoding.EncodeToString(part.ThoughtSignature)
 				msg := schemas.ResponsesMessage{
+					ID:   schemas.Ptr("rs_" + schemas.GetRandomString(50)),
```
akshaydeo: Now that this carries an `ID`, this block is byte-for-byte identical to `reasoningFromThoughtSignature` at line 3100 (same ID, Role, Type, empty Summary, base64 EncryptedContent). Two copies that have to stay in sync is how the missing-id bug spread in the first place.

```go
case part.ThoughtSignature != nil:
    if reasoningMsg, ok := reasoningFromThoughtSignature(part); ok {
        messages = append(messages, reasoningMsg)
    }
```

core/providers/gemini/responses.go
```diff
@@ -3225,6 +3241,7 @@ func convertGeminiCandidatesToResponsesOutput(candidates []*Candidate) []schemas
 				if part.Text != "" || len(part.ThoughtSignature) > 0 {
 					text := part.Text
 					msg := schemas.ResponsesMessage{
+						ID:   schemas.Ptr("rs_" + schemas.GetRandomString(50)),
```
akshaydeo, line 3259: Correct format and consistent with `anthropic`, `cohere`, `bedrock` and `schemas/mux.go`, and I confirmed the id is unique and stable within a response (`ToResponsesBifrostResponsesResponse` is called once per request from `gemini.go:710`, `gemini.go:814`, `vertex/vertex.go:1450`).

Follow-up rather than a change request: `"rs_" + schemas.GetRandomString(50)` is now hand-written at 10+ sites across the repo, and this PR exists because one of them was missing. A single `schemas.NewReasoningItemID()` would make the next provider correct by construction. Worth pairing with aligning the streaming ids (`generateItemID("reasoning", outputIndex)` at line 1630 produces `msg_<responseId>_reasoning_<n>`), so both surfaces agree on the shape.

akshaydeo (commented):
Thanks for this - the bug report is precise, the repro is reproducible from the description alone, and the two commits together are a genuinely well-scoped fix. I verified the claim against the Vercel AI SDK's own schema: in `openaiResponsesResponseSchema` the reasoning item is `{ type: 'reasoning', id: z.string(), encrypted_content: z.string().nullish(), summary: z.array(...) }`, so `id` and `summary` really are required on the non-streaming path ([openai-responses-api.ts](https://github.com/vercel/ai/blob/main/packages/openai/src/responses/openai-responses-api.ts)). The second commit (reading reasoning content blocks in `thoughtTextParts`) is the necessary complement: without it, `summary: []` would have silently stopped thought text from reaching Gemini on replay, which the thinking guide forbids ("You MUST always resend all `thought` blocks exactly as they were received from the model", https://ai.google.dev/gemini-api/docs/thinking).

I ran `go vet ./providers/gemini/` and `go test ./providers/gemini/ ./providers/vertex/ ./schemas/...` on your branch: all green, 1423 tests. Nothing here blocks the merge. What follows is one real asymmetry the second commit leaves behind, plus reuse nits.

### On the two questions a reviewer should ask here

**Is the synthesized `id` stable and unique?** Yes, and it matches the convention. `ToResponsesBifrostResponsesResponse` is called exactly once per request (`providers/gemini/gemini.go:710`, `:814`, `providers/vertex/vertex.go:1450`), so ids are stable within a response and unique across items. `"rs_" + GetRandomString(50)` is what `anthropic`, `cohere`, `bedrock`, and `schemas/mux.go` already emit, and the AI SDK only requires `z.string()` - no prefix check. Confirmed good.

**Was the streaming path left inconsistent?** Two divergences, both non-breaking, one worth a follow-up:

- Item id format differs: streaming emits `msg_<responseId>_reasoning_<n>` via `generateItemID` (`responses.go:1630`, `:1714`), non-streaming now emits `rs_<random>`. Harmless for strict clients, but the two surfaces of the same API now disagree on the shape of an id for the same item type.
- Reasoning *text* placement differs, and this one is user-visible. Streaming sends the thought text as `response.reasoning_summary_text.delta` events; non-streaming keeps it in the `reasoning_text` content block with `summary: []`. The AI SDK's `doGenerate` reads reasoning only from `summary`, and explicitly pushes `{type:'summary_text', text:''}` when `summary.length === 0`. So after this fix a non-streaming AI SDK client gets a reasoning part with **empty text**, while the same prompt with `stream: true` shows the thinking. That is still a large improvement over the response failing validation outright, and it matches what `anthropic` does today (`providers/anthropic/responses.go:6281` builds reasoning content blocks with `summary: []`, and its streaming path also uses summary deltas), so changing it in this PR alone would make Gemini diverge from the rest of the repo. Calling it out as a repo-wide follow-up rather than something to fix here.

### Findings

|---|---|---|---|---|
| 1 | Medium | [core/providers/gemini/responses.go:627](https://github.com/maximhq/bifrost/blob/fix/gemini-responses-reasoning-id-summary-6329/core/providers/gemini/responses.go#L627) | `ToGeminiResponsesResponse`'s reasoning branch still reads thought text from `Summary` only, while `thoughtTextParts` now reads content blocks first. A reasoning item carrying the same text in both places is emitted as two identical `{"thought":true,"text":...}` parts. | PLAUSIBLE (duplication reproduced; today's producers populate only one side) |
| 2 | Low | [core/providers/gemini/responses.go:3536](https://github.com/maximhq/bifrost/blob/fix/gemini-responses-reasoning-id-summary-6329/core/providers/gemini/responses.go#L3536) | After adding `ID`, the inline `case part.ThoughtSignature != nil` block is byte-identical to `reasoningFromThoughtSignature` (`:3100`). Two copies now have to be kept in sync. | CONFIRMED (reuse) |
| 3 | Low | [core/providers/gemini/responses.go:3244](https://github.com/maximhq/bifrost/blob/fix/gemini-responses-reasoning-id-summary-6329/core/providers/gemini/responses.go#L3244) | `"rs_" + GetRandomString(50)` is now hand-written at 10+ sites across gemini, anthropic, cohere, and `schemas/mux.go`. A `schemas.NewReasoningItemID()` helper would make the next provider correct by construction. | CONFIRMED (reuse/altitude) |
| 4 | Low | [core/providers/gemini/responses.go:1630](https://github.com/maximhq/bifrost/blob/fix/gemini-responses-reasoning-id-summary-6329/core/providers/gemini/responses.go#L1630) | Streaming reasoning ids stay `msg_..._reasoning_<n>` while non-streaming becomes `rs_<random>`. No client rejects it, but it is a needless split. | CONFIRMED (consistency nit) |
| 5 | Info | PR description | "The streaming path already sets both fields correctly" is true for `id` but not for `summary`: `output_item.added` on the thought path builds the item with no `ResponsesReasoning` at all, so `summary` is absent there too. It happens not to matter, because the AI SDK's chunk schema for `output_item.added`/`.done` validates only `id` and `encrypted_content`. Worth correcting in the description so the next reader does not trust it. | CONFIRMED |
| 6 | Info | [core/providers/gemini/reasoningidsummary_test.go](https://github.com/maximhq/bifrost/blob/fix/gemini-responses-reasoning-id-summary-6329/core/providers/gemini/reasoningidsummary_test.go) | `TestReasoningItemRoundTripToGeminiContents` covers `convertResponsesMessagesToGeminiContents` but not `ToGeminiResponsesResponse`, which is the reader that finding 1 is about. A third round-trip case there would pin it down. | CONFIRMED (coverage gap) |

Repro for finding 1, run against your branch:

```go
txt := "Reasoning here."
msgs := []schemas.ResponsesMessage{{
    ID: schemas.Ptr("rs_x"), Role: schemas.Ptr(schemas.ResponsesInputMessageRoleAssistant),
    Type: schemas.Ptr(schemas.ResponsesMessageTypeReasoning),
    Content: &schemas.ResponsesMessageContent{ContentBlocks: []schemas.ResponsesMessageContentBlock{
        {Type: schemas.ResponsesOutputMessageContentTypeReasoning, Text: &txt}}},
    ResponsesReasoning: &schemas.ResponsesReasoning{Summary: []schemas.ResponsesReasoningSummary{
        {Type: schemas.ResponsesReasoningContentBlockTypeSummaryText, Text: txt}}},
}}
ToGeminiResponsesResponse(&schemas.BifrostResponsesResponse{Model: "gemini-2.5-pro", Output: msgs})
// parts: [{"thought":true,"text":"Reasoning here."},{"thought":true,"text":"Reasoning here."}]
```

`convertResponsesMessagesToGeminiContents` handles the same input correctly (one part), because it goes through the new `thoughtTextParts`. The fix is to make the other reader go through it too.

### Merge recommendation

**Merge after nits.** No confirmed correctness regression: the fix is right, the second commit is required, and the full gemini/vertex/schemas suite is green. Finding 1 is the only one I would like handled before or immediately after merge, because the PR is precisely what makes the two readers asymmetric; findings 2-6 are cleanups.

### Followups

1. **In this PR (recommended, not blocking)** - `core/providers/gemini/responses.go:627`: replace the inline `for _, summaryBlock := range msg.ResponsesReasoning.Summary` loop with `currentParts = append(currentParts, thoughtTextParts(&msg)...)` so both replay readers share the content-first-then-summary rule.
2. **In this PR (nit)** - `core/providers/gemini/responses.go:3532-3545`: replace the inline signature-only construction with `if m, ok := reasoningFromThoughtSignature(part); ok { messages = append(messages, m) }`.
3. **In this PR (nit)** - update the PR description line about the streaming path setting both fields, per finding 5.
4. **Follow-up PR** - add `schemas.NewReasoningItemID()` and migrate the 10+ `"rs_" + GetRandomString(50)` sites; optionally switch `generateItemID("reasoning", ...)` to the same prefix so streaming and non-streaming agree.
5. **Follow-up PR (repo-wide)** - decide whether non-streaming reasoning text should also be mirrored into `summary` so OpenAI-Responses clients can display it. This is a gemini + anthropic + bedrock decision, not a gemini-only one, and it depends on followup 1 landing first: with both `summary` and content blocks populated, `ToGeminiResponsesResponse` would duplicate the thought part today.

### Checked and cleared (refuted candidates)

- *Missing `summary` on the streaming `output_item.added` breaks strict clients* - refuted: the AI SDK chunk schema for reasoning `output_item.added`/`.done` requires only `id` and `encrypted_content`.
- *Unsigned reasoning text is dropped by `ToGeminiResponsesResponse` now that `Summary` is empty* - refuted: the generic content-block path still emits `{"thought":true,"text":...}`; verified by round-tripping an unsigned thought part.
- *Always-non-nil `ResponsesReasoning` changes `isAssistantPrefillMessage` or the cross-provider strip in `core/encryptedreasoning.go`* - refuted: the former already returns false on a reasoning `Type`, and the latter branches on `EncryptedContent != nil`, which is unchanged.
- *Setting both the content-block `Signature` and `encrypted_content` puts the thought signature on the wire twice* - refuted: `thoughtTextParts` does not read block signatures, and the round trip emits exactly one `thoughtSignature` part.
- *`GetRandomString` is `math/rand/v2`, so ids may collide* - refuted: 50 characters, and the converter runs once per response.
- *The em dash in the new `thoughtTextParts` comment breaks file style* - refuted: the file already contains 11 of them.
- *`reasoningidsummary_test.go` violates the Go filename rule* - refuted: AGENTS.md permits the `_test.go` suffix and forbids only other underscores; the name is compliant and matches `anthropic/reasoningid_test.go`.

Nice work, and thank you for including failing-first regression tests and the upstream issue references.

AdityaPainuli:
@akshaydeo Findings 1, 2 and 6 are in. On the suggested one-linear for finding 1: appending thoughtTextParts there would have duplicated in the other direction, since ToGeminiResponsesResponse's generic content-block path already emits reasoning blocks as thought parts. So I split the helper into block/summary halves and gated the summary read on "no block text", which gives both readers the same content-first rule. Your repro is now a regression test and fails without the fix with exactly the two identical parts. Description corrected per finding 5. NewReasoningItemID and the summary-mirroring decision I'll leave for the follow-up PRs you outlined.

PR #2, opened 2025-03-21
feat: plugin interface and tests added

Pratham-Mishra04:

PR #6378, opened 2026-08-20
fix(anthropic): stop double-counting cached prompt tokens in passthrough stream usage

### LINKED ISSUE #5510: [Bug]: Anthropic passthrough double-counts cache-read tokens for Kimi-compatible streams
## Description

The Anthropic-compatible passthrough endpoint double-counts cache-read tokens for Moonshot/Kimi streaming responses.

The issue is caused by a difference in usage semantics between native Anthropic responses and the Kimi-compatible Anthropic response format.

For native Anthropic responses, `input_tokens` excludes cache-read and cache-creation tokens.

However, the Kimi response uses different semantics across stream events:

- `message_start.message.usage.input_tokens` contains the total prompt input, including cached tokens.
- `message_delta.usage.input_tokens` contains only the non-cached input.
- `message_delta.usage.cache_read_input_tokens` contains the cached input.
- `message_delta.usage.prompt_tokens` contains the total prompt input.

Bifrost currently merges the raw `input_tokens` fields using `max`, then adds the cache fields again according to the native Anthropic convention. This causes cache-read tokens to be counted twice.

## Steps to reproduce

1. Configure a Moonshot/Kimi model, for example `moonshotai/kimi-k3`.
2. Send a streaming request through the Anthropic-compatible endpoint:

   `POST /v1/messages`

3. Inspect the raw upstream SSE response.

### Control case: native Anthropic/Claude Opus

A native Anthropic response uses consistent semantics:

```json
{
  "type": "message_start",
  "message": {
    "model": "claude-opus-4-8",
    "usage": {
      "input_tokens": 2,
      "cache_creation_input_tokens": 733,
      "cache_read_input_tokens": 376156,
      "cache_creation": {
        "ephemeral_5m_input_tokens": 733,
        "ephemeral_1h_input_tokens": 0
      },
      "output_tokens": 4,
      "service_tier": "standard",
      "inference_geo": "global"
    }
  }
}
```

```json
{
  "type": "message_delta",
  "usage": {
    "input_tokens": 2,
    "cache_creation_input_tokens": 733,
    "cache_read_input_tokens": 376156,
    "output_tokens": 302,
    "output_tokens_details": {
      "thinking_tokens": 41
    }
  }
}
```

For this response:

```text
Uncached input tokens: 2
Cache-read tokens:     376156
Cache-creation tokens: 733
Total input tokens:    376891
```

This matches the standard Anthropic usage convention.

### Problematic case: Kimi

The Kimi-compatible response is:

```json
{
  "type": "message_start",
  "message": {
    "id": "chatcmpl-6a5e00ec9e0b67cff98615eb",
    "type": "message",
    "role": "assistant",
    "content": [],
    "model": "kimi-k3",
    "usage": {
      "input_tokens": 173306,
      "cache_creation_input_tokens": 0,
      "cache_read_input_tokens": 0,
      "output_tokens": 0,
      "service_tier": "standard",
      "inference_geo": "not_available",
      "prompt_tokens": 173306,
      "cached_tokens": 0
    }
  }
}
```

```json
{
  "type": "message_delta",
  "usage": {
    "input_tokens": 250,
    "cache_creation_input_tokens": 0,
    "cache_read_input_tokens": 173056,
    "output_tokens": 166,
    "output_tokens_details": {
      "thinking_tokens": 1
    },
    "prompt_tokens": 173306,
    "completion_tokens": 166,
    "total_tokens": 173472,
    "cached_tokens": 173056
  }
}
```

The Kimi values are internally consistent:

```text
250 + 173056 = 173306
173306 + 166 = 173472
```

Therefore, the correct normalized usage is:

```text
Uncached input tokens: 250
Cache-read tokens:     173056
Total input tokens:    173306
Output tokens:         166
Total tokens:          173472
```

However, with the current Bifrost implementation, the stream is effectively merged as follows:

```text
Merged input_tokens:
    max(173306, 250) = 173306

Merged cache_read_input_tokens:
    max(0, 173056) = 173056

PromptTokens:
    173306 + 173056 = 346362

TotalTokens:
    346362 + 166 = 346528
```

Thus, the cache-read tokens are counted twice.

## Expected behavior

Bifrost should normalize the usage into a canonical representation before calculating cost.

For the Kimi response above, the normalized usage should be:

```text
Prompt/Input token total: 173306
Uncached input tokens:    250
Cache-read tokens:        173056
Output tokens:            166
Total tokens:             173472
```

Since Bifrost's internal convention includes cached tokens in the total `PromptTokens` value, the expected internal usage should be equivalent to:

```text
PromptTokens: 173306
CompletionTokens: 166
TotalTokens: 173472
PromptTokensDetails.CachedReadTokens: 173056
```

The cost calculation should charge:

- `250` tokens at the regular input rate
- `173056` tokens at the cache-read rate
- `166` tokens at the output rate

## Actual behavior

The current implementation produces an inflated usage equivalent to:

```text
PromptTokens: 346362
CompletionTokens: 166
TotalTokens: 346528
PromptTokensDetails.CachedReadTokens: 173056
```

The additional `173056` tokens are effectively charged as regular input tokens.

This results in a significant overcharge whenever the cache-read token count is large.

## Likely root cause

There are two related assumptions in the current Anthropic passthrough implementation.

### 1. Raw `input_tokens` values are merged before their semantics are normalized

In:

```text
core/providers/anthropic/passthrough_usage.go
```

the streaming accumulator merges `input_tokens` using a maximum value:

```go
if u.InputTokens > c.InputTokens {
    c.InputTokens = u.InputTokens
}
```

This assumes that `input_tokens` has the same meaning in every stream event.

That assumption is valid for the native Anthropic response shown above, but not for the Kimi response:

```text
message_start.input_tokens = total input tokens
message_delta.input_tokens  = uncached input tokens
```

The accumulator retains `173306` from `message_start` and separately retains `173056` as cache-read tokens from `message_delta`.

### 2. Cache tokens are then added to `input_tokens` again

The passthrough usage conversion currently contains logic equivalent to:

```go
totalInput := au.InputTokens +
    au.CacheReadInputTokens +
    au.CacheCreationInputTokens
```

This is correct when `input_tokens` excludes cache tokens, which is the native Anthropic convention.

It is incorrect when the upstream `input_tokens` already represents the total input, as in the Kimi response.

The result is:

```text
173306 + 173056 = 346362
```

instead of:

```text
173306
```

## Suggested fix

Usage normalization should be provider- or model-specific rather than assuming that every Anthropic-compatible provider follows the native Anthropic usage semantics.

Possible approaches:

### Option 1: Provider/model usage mode

Add an explicit usage mode such as:

```text
anthropic_usage_mode:
  - standard
  - input_includes_cache
```

For native Anthropic:

```text
uncached_input_tokens = input_tokens

prompt_tokens =
    input_tokens
  + cache_read_input_tokens
  + cache_creation_input_tokens
```

For Kimi-compatible responses:

```text
uncached_input_tokens =
    input_tokens
  - cache_read_input_tokens
  - cache_creation_input_tokens

prompt_tokens = input_tokens
```

### Option 2: Normalize each stream event before merging

For providers that expose `prompt_tokens` and `cached_tokens`, normalize the event first:

```text
prompt_tokens = prompt_tokens
cached_tokens = cached_tokens
uncached_input_tokens =
    prompt_tokens - cached_tokens
```

The accumulator should merge canonical fields instead of merging raw `input_tokens` values whose meaning changes between events.

### Option 3: Add a provider-specific compatibility adapter

Add a Kimi/Moonshot usage adapter that recognizes:

```text
prompt_tokens
cached_tokens
```

and prevents Bifrost from adding `cache_read_input_tokens` to an already cache-inclusive prompt total.

A regression test should cover both:

1. Native Anthropic/Claude usage semantics.
2. Kimi-compatible usage semantics shown above.

## Affected area(s)

- Core (Go)
- Framework

## Version

```text
v1.6.5
```

## Environment

```text
- Provider: Moonshot
- Model: moonshotai/kimi-k3
- API style: Anthropic-compatible
- Endpoint: POST /v1/messages
- Response mode: Streaming/SSE
- Bifrost version: v1.6.5
```

## Regression

Unknown.

## Severity

High

wangkanai:
Closes #5510.

## Problem

`AnthropicPassthroughStreamUsage.ObserveEvent` merges streaming `/v1/messages` usage by taking the **max of each usage field independently**, then sums `input_tokens + cache_read_input_tokens + cache_creation_input_tokens` into `PromptTokens`.

That is correct only for an upstream that follows the convention Bifrost's own `AnthropicUsage` documents:

> Unlike OpenAI models, Anthropic (claude) models separately track cache creation and cache read tokens, and its not included in the `input_tokens` field.

Some Anthropic-**dialect** upstreams do not follow it. They report a prompt-scale `input_tokens` in `message_start` with **no cache counters at all**, and send the real cache-aware split only in `message_delta`. Max-merging keeps the cache-less figure *and* the `cache_read`, so the prefix-cached prompt is counted twice — once as fresh input, once as cached.

This is **not vendor-specific**. Confirmed on two unrelated providers: Moonshot/Kimi (as originally reported in #5510) and **QwenCloud**, captured verbatim from `https://token-plan.<region>.maas.aliyuncs.com/apps/anthropic/v1/messages` with Bifrost out of the path:

```jsonc
// message_start — prompt-scale input_tokens, NO cache counters
{"type":"message_start","message":{"id":"msg_ba0b5ef1","model":"qwen3.7-plus","content":[],
 "usage":{"input_tokens":8834,"output_tokens":0}}}

// message_delta — the authoritative, cache-aware split
{"type":"message_delta","delta":{"stop_reason":"end_turn"},
 "usage":{"input_tokens":18,"output_tokens":426,"cache_creation_input_tokens":0,
          "cache_read_input_tokens":8831,"cache_creation":{"ephemeral_5m_input_tokens":0}}}
```

The non-streaming body for the same request agrees with `message_delta`: `18 + 0 + 8831 = 8849`.

```
InputTokens  = max(8834, 18)   = 8834
CacheRead    = max(0,  8831)   = 8831
PromptTokens = 8834 + 8831 + 0 = 17665      // want 8849
```

MiniMax on the same code path is unaffected, because its `message_start` reports `input_tokens: 0` — which is what makes the trigger condition precise: *`message_start` carries a prompt-scale `input_tokens` with no cache counters.*

Impact is not cosmetic: over one week of agent traffic on an affected provider, **42% of every logged prompt token was invented**, which corrupts cost attribution, budgets, rate limiting and dashboards, and inverts the apparent cache-hit rate (a real ~99.7% displayed as ~52%).

## Fix

Decide which event is authoritative for `input_tokens` by **cache evidence**, and take `input_tokens` and the cache counters **from that same event**.

**The test is on the values, not on field presence.** An earlier revision of this PR probed the raw event JSON for the presence of the cache keys. That is wrong for the provider this issue was filed against: Moonshot/Kimi sends `"cache_read_input_tokens": 0` *explicitly* on `message_start` while `input_tokens` already includes the cached prefix, so a presence check trusts that frame and the total is still doubled (346362 instead of 173306). A frame counts as cache-aware when `cache_read_input_tokens`, `cache_creation_input_tokens`, or either half of the `cache_creation` 5m/1h breakdown is non-zero.

**`input_tokens` is latched, not max-merged.** Max-merging `input_tokens`, `cache_read` and `cache_creation` independently can assemble a prompt total out of values that never coexisted in one frame. A `message_start` carrying a prompt-scale `input_tokens` together with a *partially known* `cache_read` is exactly that shape — the frame #5510's hierarchical-cache explanation predicts — and it clears any authority test:

```
message_start  input_tokens 173306, cache_read   5000
message_delta  input_tokens    250, cache_read 173056
max(173306, 250) + max(5000, 173056) = 346362      // want 173306
```

So `input_tokens` is latched from the single authoritative event with the largest cache total. On equal cache totals the **smaller** `input_tokens` wins, because the premise of the whole split is that the prompt-scale value is the untrustworthy one; for a compliant upstream `input_tokens` does not vary across frames, so the tie-break never fires there.

**`cache_creation` is a breakdown, not an addend.** It is documented as the 5m/1h split *of* `cache_creation_input_tokens`, so it contributes to the total only when the top-level counter is absent — otherwise a frame carrying both would be counted twice. This applies to the emitted prompt total and to `PromptTokensDetails`, not just to the authority test.

Two properties are preserved deliberately:

- **Order-independent** — the documented behaviour of this merge. A late-arriving cache-less frame cannot clobber an authoritative one, and every test below asserts both event orders.
- **Monotone in the accumulators** — `looseInput` and `authInput` never decrease, which keeps compliant upstreams (Anthropic itself, Bedrock, MiniMax) provably unaffected. The *emitted* `InputTokens` is deliberately **not** monotone: it drops from the loose figure to the authoritative one when the first cache-aware event arrives. That is safe because `StreamPassthrough` keeps the last non-nil observation and the authoritative `message_delta` is last; a client cancelling between `message_start` and `message_delta` is billed the loose figure, exactly as before this PR.

## Tests

`passthroughusagecachesplit_test.go` is table-driven over SSE frames captured verbatim from the live upstreams on 2026-08-20, plus the derived cases review surfaced. Every case asserts both event orders.

Captured, and unchanged by this PR's later revisions:

- the QwenCloud regression case — fails on `dev` with exactly `17665`, passes at `8849`
- Moonshot/Kimi, both the deviating frame (346362 → 173306) and a compliant one from the same provider
- MiniMax and native-Anthropic controls, which must not change
- a stream that never sends cache counters, which must still report its `input_tokens`

Derived cases added in review, each failing before its fix:

| case | before | after |
|---|---:|---:|
| partially known `cache_read` on `message_start` | 346362 | **173306** |
| breakdown-only cache creation, prompt total | 1000 | **50000** |
| breakdown-only cache creation, `CachedWriteTokenDetails` | `nil` | **49000 @ 5m** |
| tie on cache total, breakdown-only vs top-level | 99000 | **50000** |
| cache-less event with larger `input_tokens` after an authoritative one | — | pinned at **8931** |

The pre-existing `TestAnthropicPassthroughStreamUsage` continues to pass, along with the full `providers/anthropic`, `providers/azure`, `providers/utils` and `schemas` suites.

## Verification in production

Deployed to a live v1.6.11 gateway. Identical bytes, `stream` the only variable — streaming now agrees with non-streaming exactly:

| provider | `prompt_tokens` before | after | non-streaming truth |
|---|---:|---:|---:|
| QwenCloud `qwen3.7-plus` | 17,669 | **8,849** | 8,849 |
| Moonshot `kimi-k3` | 17,862 | **8,932** | 8,932 |
| MiniMax `M3` (control) | 8,992 | **8,992** | 8,992 |

Organic traffic at ~300K-token prompts went from `prompt_tokens ≈ 2 × cached_read` to a residual of 411–1,369 tokens, matching the providers that were always correct.

## Scope

This PR fixes the **passthrough** accumulator only.

There is a second, independent site with the same flaw — `accumulateAnthropicResponsesUsage` in `core/providers/anthropic/anthropic.go`, the native streaming path — tracked as #5354 with #5355 open against it. I verified the two are independent: deploying a fix to only this passthrough site changed nothing for traffic that takes the native path. Both need fixing; I've deliberately left the native half to #5355 rather than compete with it, and posted the same cross-provider evidence there.

Once the rule here settles it is worth extracting into one shared helper so the two paths cannot drift, as suggested in review — good pairing with #5355.

## Review history

The latch and the breakdown handling were originally split out as #6439 on review's suggestion. Both reviewers then asked for them here, so #6439 was folded back into this PR and closed; this branch carries the whole fix.


wangkanai:
Good catch — and it was more than a coverage gap. **Writing the requested Kimi case showed the fix itself was wrong for Kimi.** Fixed in f2a11c4 / 24b7431.

## What the test exposed

My first version decided authority by **field presence**: an event was treated as authoritative for `input_tokens` if its usage object carried cache-counter *keys*. That works for QwenCloud, whose `message_start` omits them entirely. It fails for Moonshot/Kimi — the provider this issue was actually filed for — because Kimi sends them **explicitly as zero**:

```jsonc
"usage":{"input_tokens":173306,"cache_creation_input_tokens":0,"cache_read_input_tokens":0, ...}
```

A presence check trusts that frame, keeps `input_tokens: 173306` (which already includes the cached prefix), and the total is still doubled. Measured against the previous commit:

```
--- FAIL: TestPassthroughStreamUsageKimiDoesNotDoubleCountCachedPrompt
    prompt tokens = 346362, want 173306 (inflated by 173056)
```

So the PR as first pushed did not fix #5510. Thanks for pushing on the coverage.

## The correction

Decide on the **values**, not on presence — an event is authoritative only if its cache counters are actually populated:

```go
if u.CacheReadInputTokens > 0 || u.CacheCreationInputTokens > 0 {
	a.authSeen = true
	if u.InputTokens > a.authInput {
		a.authInput = u.InputTokens
	}
} else if u.InputTokens > a.looseInput {
	a.looseInput = u.InputTokens
}
```

This also drops the raw-JSON probing helper entirely and makes the passthrough accumulator consistent with the equivalent logic in `anthropic.go`. Still order-independent, still monotone.

## Coverage now

| case | frames | result |
|---|---|---|
| Qwen regression | captured 2026-08-20 | 17665 → **8849** |
| **Kimi regression** | as reported in #5510 | 346362 → **173306**, completion 166 |
| **Kimi, cache-aware `message_start`** | captured 2026-08-20 | **8931**, unchanged by the fix |
| MiniMax control | captured 2026-08-20 | 8991, unchanged |
| native Anthropic control | — | 8849, unchanged |
| order-independence | Qwen **and** Kimi | forward == reverse |
| no cache counters at all | synthetic | falls back to `input_tokens` |

Both regression cases fail against unpatched `dev` (17665 and 346362) and pass with the fix. The pre-existing `TestAnthropicPassthroughStreamUsage` and the full `providers/anthropic`, `providers/azure`, `providers/utils` and `schemas` suites still pass.

## One finding worth recording

While capturing fixtures I hit a Kimi response whose `message_start` **was** fully cache-aware (`input_tokens: 0`, `cache_read_input_tokens: 8931`) — on the same endpoint and model that deviated earlier the same day. So the deviation is **per-request, not per-provider**, which matches @qixiangyang's hierarchical-cache explanation in #5510: the final split may simply not be known when `message_start` is emitted.

That argues against any fix keyed on "provider X is non-compliant" — the same connection can produce both shapes minutes apart. I've added that compliant frame as a control so the fix is pinned against both.

core/providers/anthropic/passthrough_usage.go
```diff
@@ -122,8 +141,18 @@ func (a *AnthropicPassthroughStreamUsage) ObserveEvent(event []byte) *schemas.Bi
 
 	a.seen = true
 	c := &a.combined
-	if u.InputTokens > c.InputTokens {
-		c.InputTokens = u.InputTokens
+	if u.CacheReadInputTokens > 0 || u.CacheCreationInputTokens > 0 {
```
akshaydeo: **Finding 1 (medium, confirmed).** The value test admits any event with a non-zero cache counter, and `max` inside the authoritative bucket then keeps whichever `input_tokens` is largest. A `message_start` with partial cache knowledge defeats it. I ran these two frames through the patched accumulator:

```jsonc
{"type":"message_start","message":{"usage":{"input_tokens":173306,"cache_creation_input_tokens":0,"cache_read_input_tokens":5000,"output_tokens":0}}}
{"type":"message_delta","usage":{"input_tokens":250,"cache_creation_input_tokens":0,"cache_read_input_tokens":173056,"output_tokens":166}}
```

Result: `prompt_tokens = 346362`, want 173306. Your own test comment predicts this frame - a hierarchical cache whose split is not final at `message_start` is exactly the thing that would report `cache_read: 5000` there. `dev` is equally wrong here, so this does not block, but it is worth closing.

Suggested rule: latch the whole tuple from the authoritative event with the largest cache total instead of taking three independent maxima.

```go
// pick the authoritative event with the most cache evidence and use ITS input_tokens,
// so input/cache_read/cache_creation always come from one frame.
cacheTotal := u.CacheReadInputTokens + u.CacheCreationInputTokens
if cacheTotal > 0 {
    if !a.authSeen || cacheTotal > a.authCacheTotal ||
        (cacheTotal == a.authCacheTotal && u.InputTokens > a.authInput) {
        a.authSeen, a.authCacheTotal, a.authInput = true, cacheTotal, u.InputTokens
    }
} else if u.InputTokens > a.looseInput {
    a.looseInput = u.InputTokens
}
```

**Finding 3 (low).** Separately, this line ignores the `cache_creation` breakdown. A frame carrying only `{"cache_creation":{"ephemeral_5m_input_tokens":8831}}` with no top-level `cache_creation_input_tokens` is classified as cache-less; I measured 9000 instead of 8849. Your earlier presence-based version did probe `usage.cache_creation`. One line restores it:

```go
if u.CacheReadInputTokens > 0 || u.CacheCreationInputTokens > 0 ||
    u.CacheCreation.Ephemeral5mInputTokens > 0 || u.CacheCreation.Ephemeral1hInputTokens > 0 {
```

core/providers/anthropic/passthrough_usage.go
```diff
@@ -122,8 +141,18 @@ func (a *AnthropicPassthroughStreamUsage) ObserveEvent(event []byte) *schemas.Bi
...
+		a.authSeen = true
+		if u.InputTokens > a.authInput {
+			a.authInput = u.InputTokens
+		}
+	} else if u.InputTokens > a.looseInput {
+		a.looseInput = u.InputTokens
+	}
+	if a.authSeen {
```
akshaydeo, line 196: **Finding 2 (low, plausible).** Once `authSeen` is set, `looseInput` is discarded permanently, so a later cache-less event carrying a larger `input_tokens` is dropped rather than considered. Measured:

```jsonc
{"type":"message_start","message":{"usage":{"input_tokens":0,"cache_read_input_tokens":8931,"output_tokens":0}}}
{"type":"message_delta","usage":{"input_tokens":250,"output_tokens":16}}
```

Patched: 8931. On `dev`: 9181. No captured frame triggers this - Anthropic's minimal `message_delta` (`{"output_tokens": 15}` per the streaming docs) omits `input_tokens` entirely, so the loose bucket stays at 0 - which is why I am flagging it as plausible rather than confirmed, and not as a blocker.

If 8931 is the intended answer, please pin it with a test so a later refactor cannot silently flip it back. If 9181 is intended, the tuple-latch rule suggested on the authority check above resolves it as a side effect, since it selects on cache evidence rather than on a max that can strand a legitimate later `input_tokens`.

core/providers/anthropic/passthrough_usage.go
```diff
@@ -100,6 +100,25 @@ func buildAnthropicPassthroughUsage(au *AnthropicUsage) *schemas.BifrostPassthro
...
+	// prompt-scale input_tokens on message_start while its cache counters are still zero or absent,
+	// and send the real cache-aware split only on message_delta. Max-merging both kept the
+	// cache-less figure and then added cache_read on top of a total that already included it,
+	// roughly doubling prompt_tokens on every cached streaming request.
+	//
+	// An event whose cache counters are actually populated is therefore authoritative for
+	// input_tokens; one without them is only a fallback, used until an authoritative event turns
+	// up. Note the test is on the VALUES, not on whether the fields were present: Kimi sends
+	// "cache_read_input_tokens": 0 explicitly on message_start, so a presence check would wrongly
+	// trust that frame. authSeen keeps the merge order-independent, and nothing here ever lowers a
```
akshaydeo: **Finding 5 (low, confirmed, comment only).** "nothing here ever lowers a value" holds for `looseInput` and `authInput` individually, but not for the value this function publishes. On the Qwen stream the emitted `c.InputTokens` goes 8834 after `message_start`, then 18 after `message_delta` when `authSeen` flips. `PromptTokens` still rises overall there (8834 to 8849) only because `cache_read` climbs at the same time.

This is harmless in practice and I want to be clear about why: `StreamPassthrough` in `core/providers/utils/passthrough_stream.go` keeps the **last** non-nil observation for the final chunk, and `message_delta` is last, so the billed figure is the corrected one. The only observable effect is that a client cancelling between the two frames is billed the loose figure, which is exactly what `dev` does today.

Worth rewording so the next reader does not build on a monotonicity guarantee the emitted field does not offer, for example:

```go
// The two accumulators are individually monotone; the published InputTokens is not, because
// it switches buckets the first time an authoritative event arrives. That is safe: the
// passthrough stream bills the LAST observation, and the authoritative message_delta is last.
```

akshaydeo (commented):
## Review

Thanks for this - it is one of the best-evidenced contributions we have had on the billing path. Captured frames from two unrelated upstreams, a non-streaming ground truth for each, a live before/after table, and a control provider that must not move: that is exactly the shape of evidence a token-accounting change needs. The fix itself is small, order-independent, and I verified it leaves compliant upstreams untouched.

I re-derived the semantics from Anthropic's own docs rather than taking them on trust, and both premises hold:

- `input_tokens` excludes the cache counters. From https://platform.claude.com/docs/en/build-with-claude/prompt-caching : "Number of input tokens which were not read from or used to create a cache (that is, tokens after the last cache breakpoint)", with `total_input_tokens = cache_read_input_tokens + cache_creation_input_tokens + input_tokens`. So `buildAnthropicPassthroughUsage`'s three-way sum is right, and an upstream that puts the full prompt in `input_tokens` genuinely does break it.
- Streaming usage is cumulative, not incremental. From https://platform.claude.com/docs/en/build-with-claude/streaming : "The token counts shown in the `usage` field of the `message_delta` event are *cumulative*." That is what makes a max-merge legitimate in the first place, so the PR is not fighting the event model, only the one field whose meaning varies.

I also checked native Anthropic against the doc's own captured web-search stream (`message_start` `input_tokens: 2679, cache_creation_input_tokens: 0, cache_read_input_tokens: 0`, `message_delta` `input_tokens: 10682, ... :0, :0`). Both frames have zero cache counters, so both land in the `looseInput` bucket and the answer is unchanged at 10682. Native Anthropic is safe.

I ran the new tests plus the pre-existing `TestAnthropicPassthroughStreamUsage` on your branch: all green. I then probed three cases the tests do not cover, by running the real accumulator.

### Findings

|---|---|---|---|---|
| 1 | Medium | `core/providers/anthropic/passthrough_usage.go:144` | A `message_start` that carries a prompt-scale `input_tokens` **and** a partially-known non-zero `cache_read` passes the value test, and the max inside the authoritative bucket then keeps the inflated figure. Measured: still 346362 instead of 173306. | CONFIRMED |
| 2 | Low | `core/providers/anthropic/passthrough_usage.go:152` | Once `authSeen` flips, a later cache-less event with a **larger** `input_tokens` is discarded entirely, which can undercount. Measured: 8931 where `dev` reported 9181. | PLAUSIBLE |
| 3 | Low | `core/providers/anthropic/passthrough_usage.go:144` | The value test ignores `cache_creation.ephemeral_5m/1h_input_tokens`, so a frame carrying only the breakdown is misclassified as non-authoritative. The superseded presence-based version did check `usage.cache_creation`. | CONFIRMED (misclassification), PLAUSIBLE (trigger) |
| 4 | Low | `core/providers/anthropic/passthrough_usage.go:98` | The type doc still says "Taking the max of each field across events ... the same merge the native Anthropic stream does (anthropic.go)". Neither half is true any more: `input_tokens` is no longer a plain max, and `accumulateAnthropicResponsesUsage` (`anthropic.go:681`) still does the unconditional max. | CONFIRMED |
| 5 | Low | `core/providers/anthropic/passthrough_usage.go:117` | "nothing here ever lowers a value" is true of `looseInput`/`authInput` but not of the emitted `c.InputTokens`, which drops from 8834 to 18 on the Qwen stream when `authSeen` flips. | CONFIRMED |
| 6 | Nit | PR description | The **Fix** section still describes the superseded presence-based design ("Presence is read from the raw event JSON"). Commit 4's message also claims the new rule "matches the equivalent accumulator in anthropic.go", which contradicts the **Scope** section and `anthropic.go:681`. | CONFIRMED |
| 7 | Nit | `core/providers/anthropic/passthrough_usage.go:144-156` | #5355 will need this identical rule on the native path. Worth extracting it now as a tiny shared merger so the two sites cannot drift. | Cleanup |

### Detail on #1, the one worth acting on

Your own hypothesis in the test comments is that Kimi deviates **per request**, because a hierarchical cache does not know its final split when `message_start` is emitted. That hypothesis predicts an intermediate frame: partial cache knowledge alongside a prompt-scale `input_tokens`. Feeding exactly that through the patched accumulator:

```jsonc
// message_start: prompt-scale input_tokens, cache_read partially known
{"type":"message_start","message":{"usage":{"input_tokens":173306,"cache_creation_input_tokens":0,"cache_read_input_tokens":5000,"output_tokens":0}}}
// message_delta: the authoritative split
{"type":"message_delta","usage":{"input_tokens":250,"cache_creation_input_tokens":0,"cache_read_input_tokens":173056,"output_tokens":166}}
```

gives `prompt_tokens = 346362`, the original bug, because `message_start` is admitted to the authoritative bucket and `max(173306, 250)` keeps the wrong number.

The root cause is that `input_tokens`, `cache_read` and `cache_creation` are still max-merged **independently**, so the reported prompt total can be a composite of values that never coexisted in one frame. Keeping them as a tuple removes the whole class:

```mermaid
flowchart LR
  A["usage event"] --> B{"cache_read > 0 or cache_creation > 0?"}
  B -->|no| C["looseInput = max(looseInput, input_tokens)"]
  B -->|yes| D{"cache total > best cache total?<br/>(tie: larger input_tokens)"}
  D -->|yes| E["latch tuple:<br/>input, cache_read, cache_creation<br/>from THIS event"]
  D -->|no| F["keep latched tuple"]
  E --> G["prompt = latched input + latched cache_read + latched cache_creation"]
  F --> G
  C --> H["prompt = looseInput"]
```

Picking the authoritative event with the largest `cache_read + cache_creation` (tie-break on the larger `input_tokens`) and taking **that event's** `input_tokens` reproduces every case you already assert - Qwen 8849, Kimi 173306, MiniMax 8991, Kimi cache-aware 8931, stock Anthropic 8849, doc web-search 10682 - and additionally returns 173306 for the partial frame above. It also happens to fix #2, since the tuple is chosen by cache evidence rather than by a max that can strand a legitimate later `input_tokens`.

This is a limitation you are inheriting, not one you introduced: `dev` reports 346362 for that frame too. So it does not block the merge. I raise it because the frame is predicted by your own model of the upstream, and because tightening it now is cheaper than a second regression report.

### Detail on #2

```jsonc
{"type":"message_start","message":{"usage":{"input_tokens":0,"cache_read_input_tokens":8931,"output_tokens":0}}}
{"type":"message_delta","usage":{"input_tokens":250,"output_tokens":16}}
```

Patched: 8931. On `dev`: 9181. No captured frame triggers this today - Anthropic's minimal `message_delta` (`{"output_tokens": 15}` in the docs) has no `input_tokens` at all, so the loose bucket stays at 0 - which is why this is PLAUSIBLE rather than confirmed. A short test pinning whichever answer you consider correct would lock the intent in.

### Detail on #5

To be clear, this costs nothing today: `StreamPassthrough` keeps the **last** non-nil observation (`core/providers/utils/passthrough_stream.go`), and the terminal `message_delta` is last, so the billed figure is right. The only visible effect is that a client cancelling between `message_start` and `message_delta` is billed the loose figure, which is exactly the pre-PR behaviour. Please just soften the comment so a future reader does not rely on a monotonicity that the emitted field does not have.

## Merge recommendation

**Merge after nits.** Nothing here regresses a case anyone has captured, the fix is correct for every frame in evidence, and it is a large improvement on a real billing bug. Findings 4, 5 and 6 are documentation accuracy and should be fixed before merge because they will mislead whoever picks up #5355. Finding 1 is a confirmed gap that the PR narrows but does not close, and finding 3 is a one-line hardening.

## Followups

1. **In this PR (blocking, doc only)** - `core/providers/anthropic/passthrough_usage.go:98`: update the type doc so it no longer claims a plain per-field max, and no longer claims parity with `anthropic.go`.
2. **In this PR (blocking, doc only)** - `core/providers/anthropic/passthrough_usage.go:117`: reword "nothing here ever lowers a value" to say the accumulators are monotone while the emitted `c.InputTokens` is not, and note that the last-non-nil rule in `StreamPassthrough` makes that safe.
3. **In this PR (blocking, doc only)** - PR description: replace the presence-based paragraph in **Fix** with the value-based rule that actually shipped, and drop the "matches `anthropic.go`" claim from commit 4's message.
4. **In this PR (nit, one line)** - `core/providers/anthropic/passthrough_usage.go:144`: add `|| u.CacheCreation.Ephemeral5mInputTokens > 0 || u.CacheCreation.Ephemeral1hInputTokens > 0` so a breakdown-only frame is not misread as cache-less.
5. **Follow-up PR** - `core/providers/anthropic/passthrough_usage.go:144-156`: adopt the tuple-latch rule for finding 1, with the partial-cache `message_start` frame added to the table. Happy for this to be separate if you would rather land the measured win now.
6. **Follow-up PR** - `core/providers/anthropic/passthrough_usage.go:144` and `core/providers/anthropic/anthropic.go:681`: once the rule settles, extract it into one shared helper so the passthrough and native paths cannot diverge. Good pairing with #5355.

## Checked and cleared

- **Regression on native Anthropic** - refuted. The doc's own web-search stream has zero cache counters on both frames, so both stay in the loose bucket and the result is unchanged.
- **Final billed number corrupted by the non-monotone intermediate** - refuted. `StreamPassthrough` keeps the last non-nil observation, and `message_delta` is last.
- **Order dependence** - refuted. `authSeen` plus per-bucket maxima make the final value order-free; your reversed-order test confirms it.
- **Cross-request state leak or a new data race from the three added fields** - refuted. A fresh accumulator is built per stream at `anthropic.go:3202` and `azure.go:3585`, and the new fields are plain ints beside the existing ones.
- **AGENTS.md Go filename rule** - refuted. `passthroughusagecachesplit_test.go` carries no underscore other than the `_test.go` suffix, which is exactly what the rule permits.
- **Test package mismatch with the sibling `passthrough_usage_test.go`** - refuted. 51 of the 53 test files in this package use `package anthropic`; yours follows the majority convention.
- **Azure passthrough left behind** - refuted. `azure.go:3605` reuses the same accumulator type, so it picks the fix up for free.

wangkanai:
Findings 1, 2 and 3 reproduce exactly as measured — I ran all three through the real accumulator on the branch before changing anything: 346362, 8931, and the breakdown-only frame classified as loose.

Pushed for the blocking items:

- **4** — type doc no longer claims a plain per-field max, and no longer claims parity with `anthropic.go`; it now points at #5354 / #5355 and says explicitly that the native accumulator still has the double-count.
- **5** — reworded. The comment now separates the two accumulators (monotone) from the emitted `combined.InputTokens` (not monotone, drops when `authSeen` flips), and records why the last-non-nil rule in `StreamPassthrough` makes that safe, including the cancel-mid-stream case.
- **3** — one-liner applied, plus a test. Worth noting the breakdown was already max-merged into `combined` (`passthrough_usage.go:166-171`), so the misclassification was confined to the authority test and the one line closes it. The test fails at 50000 without the change, passes at 1000 with it, and covers reversed order like the rest of the table.
- **2** — pinned as a test at 8931, with a comment saying no captured frame triggers it and why the intent is being fixed explicitly.
- **6** — PR description updated: the **Fix** section now describes the value-based rule that shipped instead of the superseded presence-based one.

### One correction to the review, measured

> It also happens to fix #2, since the tuple is chosen by cache evidence rather than by a max that can strand a legitimate later `input_tokens`.

It does not. I prototyped the tuple-latch exactly as specified and ran your #2 frames through it:

```
F1 partial-cache message_start:  173306   (was 346362) ✅
F1 reversed order:               173306   order-independent ✅
F2 later loose input_tokens:       8931   unchanged
```

#2 is unchanged because the loose bucket is still discarded once any authoritative event exists — the tuple latch changes *which* authoritative event wins, not whether a cache-less event can contribute. Fixing #2 would need a separate decision about whether a cache-less `input_tokens` may be added to an authoritative cache total, and I do not think it should be: that is exactly the composite-of-values-that-never-coexisted problem #1 is about.

### One caveat on the tie-break, if it helps the follow-up

The tie-break as specified (equal cache totals → prefer the larger `input_tokens`) can select the prompt-scale frame, which is the frame the fix exists to distrust. A breakdown-only `message_start` at `input_tokens: 50000, ephemeral_5m: 49000` against a `message_delta` at `input_tokens: 1000, cache_creation_input_tokens: 49000, ephemeral_5m: 49000` ties at 49000 and latches the 50000, giving 99000. Synthetic — no captured frame does this — but preferring the *smaller* `input_tokens` on a tie is the safer default, since the whole premise is that the prompt-scale value is the untrustworthy one.

### Remaining

- **Followup 5 (tuple-latch)** — opened as #6439, stacked on this branch, with your partial-cache frame in the table. It implements the **opposite tie-break** for the reason above (smaller `input_tokens` on a tie), and documents that `cache_creation` is a breakdown of `cache_creation_input_tokens` rather than an addend. Both new cases fail on this PR's head (346362, 99000) and pass there.
- **Followup 6 (shared helper)** — agreed, and it pairs naturally with #5355.
- **Commit 4's message** still carries the "matches the equivalent accumulator in `anthropic.go`" claim. The branch has a merge from `dev` in it, so correcting that means a history rewrite and a force-push. If you squash-merge, the corrected description covers it. Say the word if you would rather I rewrite the branch.

core/providers/anthropic/passthrough_usage.go
```diff
@@ -122,8 +155,19 @@ func (a *AnthropicPassthroughStreamUsage) ObserveEvent(event []byte) *schemas.Bi
 
 	a.seen = true
 	c := &a.combined
-	if u.InputTokens > c.InputTokens {
-		c.InputTokens = u.InputTokens
+	if u.CacheReadInputTokens > 0 || u.CacheCreationInputTokens > 0 ||
+		u.CacheCreation.Ephemeral5mInputTokens > 0 || u.CacheCreation.Ephemeral1hInputTokens > 0 {
```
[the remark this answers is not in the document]
wangkanai: Correct, and the test I added asserted the wrong answer — fixed in 8e7f51f.

The classification change made the breakdown-only frame authoritative but `buildAnthropicPassthroughUsage` still summed `CacheCreationInputTokens` alone, so the 49000 tokens were dropped from the total, and `PromptTokensDetails` came back `nil` because the details block is gated on the same two top-level counters.

`cacheCreation` is now derived once and used for both the total and the details, guarded on the top-level counter being zero so a frame carrying both is not counted twice — `cache_creation` is the 5m/1h split *of* `cache_creation_input_tokens`, not an addend.

The test now expects 50000 and additionally asserts the 5m breakdown survives into `CachedWriteTokenDetails`. It fails on both counts without the change:

```
prompt tokens = 1000, want 50000
prompt token details = nil, want cache-write details carrying the breakdown
```

wangkanai:
Folded #6439 back into this branch and closed it — both reviews asked for the latch here, so a stacked PR was more friction than value. Two commits added:

**`0ce8682` — latch `input_tokens` instead of max-merging it.** @akshaydeo's finding 1 and CodeRabbit's outside-diff comment are the same defect: `input_tokens`, `cache_read` and `cache_creation` were still max-merged independently, so the prompt total could be assembled from values that never coexisted in one frame. `input_tokens` now comes from the single authoritative event with the largest cache total. The partial-cache `message_start` frame goes 346362 → 173306, and every case already in the table reproduces unchanged.

On the tie-break I went the **opposite** way to the review's suggestion. Preferring the larger `input_tokens` on equal cache totals selects the prompt-scale frame, which is the value this split exists to distrust — a breakdown-only `message_start` at `50000 / ephemeral_5m 49000` against a `message_delta` at `1000 / cache_creation_input_tokens 49000` ties at 49000 and reports 99000. Smaller wins instead, and it is pinned as a test. For a compliant upstream `input_tokens` does not vary across frames, so this never fires there.

**`8e7f51f` — count breakdown-only cache creation.** CodeRabbit caught a real bug in my previous commit, and my own test had asserted the wrong expectation. Recognising the ephemeral breakdown in the *authority* test without also using it in `buildAnthropicPassthroughUsage` meant such a frame was classified correctly and then had its 49000 cache-creation tokens dropped from the total anyway, with `PromptTokensDetails` coming back `nil`. Both now derive from one effective value, guarded so a frame carrying the top-level counter *and* the breakdown is not counted twice.

### On finding 2

Still not fixed, and I do not think it should be. The review expected the latch to fix it as a side effect; measured, it does not — 8931 either way — because the loose bucket is discarded once any authoritative event exists, and the latch only changes *which* authoritative event wins. Letting a cache-less `input_tokens` be added to an authoritative cache total is exactly the composite-of-values-that-never-coexisted problem finding 1 is about. Pinned as a test at 8931 so the intent is explicit.

### Verification

Every derived case fails before its fix and passes after — checked by reverting only the implementation file and keeping the tests, since a test that cannot fail proves nothing:

| case | before | after |
|---|---:|---:|
| partially known `cache_read` on `message_start` | 346362 | 173306 |
| breakdown-only, prompt total | 1000 | 50000 |
| breakdown-only, `CachedWriteTokenDetails` | `nil` | 49000 @ 5m |
| tie on cache total | 99000 | 50000 |

Full `providers/anthropic`, `providers/azure`, `providers/utils` and `schemas` suites pass. PR description updated to describe the latch rather than the two-accumulator split it started as.

Followup 6 (extracting the rule into a shared helper with `anthropic.go`) still makes sense once this settles, and pairs with #5355.

core/providers/anthropic/passthrough_usage.go
wangkanai: 8931 is intended, and it is pinned — `TestPassthroughStreamUsageCacheLessEventAfterAuthoritativeIsIgnored`, added in the round-2 push. Your exact two frames, both orders, with a comment recording that no captured frame triggers it and why 9181 is the wrong answer.

On the second half of your suggestion, measured rather than assumed: **the tuple-latch does not fix #2.** I prototyped it as specified and ran your frames through it before opening the follow-up:

```
F1 partial-cache message_start:  173306   (was 346362) ✅
F1 reversed order:               173306   order-independent ✅
F2 later loose input_tokens:       8931   unchanged
```

The latch changes *which* authoritative event wins, not whether a cache-less event may contribute. `looseInput` is still discarded once `authSeen` flips, so both rules agree on 8931.

That is deliberate, and I think 9181 would be the wrong answer for the same reason #1 is a bug. `input_tokens` is documented as excluding the cache counters, so adding a cache-less frame's 250 to an authoritative frame's 8931 builds a prompt total out of two numbers that never coexisted — the composite you identified in #1, just assembled across buckets instead of within one. If a later frame really does carry a larger legitimate `input_tokens`, the honest signal is that it also carries the cache counters, and then the latch picks it up on cache evidence.

The latch is no longer a follow-up in any case: both you and CodeRabbit asked for it here, so #6439 was folded back into this branch as `0ce8682` and closed. `TestPassthroughStreamUsagePartiallyKnownCacheOnMessageStart` covers your #1 frame at 173306.

wangkanai:
All review conversations are resolved. Head is **645b3e0**.

Finding 2 was the one thread still open. It was answered in the round-2 push but not in its own thread — that is now done: 8931 is the intended answer and is pinned by `TestPassthroughStreamUsageCacheLessEventAfterAuthoritativeIsIgnored`, and the tuple-latch is measured *not* to change it, contrary to the review's expectation. Detail in the thread.

**CodeRabbit's outstanding `CHANGES_REQUESTED` is stale.** Its one outside-diff finding — "keep `input_tokens` and cache counters from one authoritative event", the same gap as finding 1 — was reviewed at 14:13, eight minutes before `0ce8682` landed the latch. It is fixed at head, with the exact frame it names covered by `TestPassthroughStreamUsagePartiallyKnownCacheOnMessageStart` (346362 → 173306). Since it is outside the diff it has no thread to resolve; noting it here so the review state is not read as an open objection.

`645b3e0` itself is documentation only: the value-vs-presence note had ended up appended to the tie-break paragraph at 164 columns while its neighbours wrap at ~97. `gofmt` does not reflow comments, so it survived the finding-4/5 doc pass. Separate point, separate paragraph — that block is what whoever picks up #5355 will read.

Status against the six followups:

| # | Item | State |
|---|---|---|
| 1 | Type doc — no plain per-field max, no `anthropic.go` parity claim | Done |
| 2 | Monotonicity comment reworded | Done |
| 3 | PR description — value-based rule | Done |
| 4 | Ephemeral 5m/1h in the authority test | Done, plus a test |
| 5 | Tuple-latch | Folded in as `0ce8682`; #6439 closed |
| 6 | Shared helper with `anthropic.go:681` | Agreed, follow-up, pairs with #5355 |

Verified on `645b3e0`: `gofmt` clean on both changed files, `go vet ./providers/anthropic/` clean, and all 10 tests in the new file plus the pre-existing `TestAnthropicPassthroughStreamUsage` pass.

One thing still needs your call, unchanged from before: commit 4's message carries the "matches the equivalent accumulator in `anthropic.go`" claim, which the type doc now contradicts. The branch has a `dev` merge in it, so correcting the message means a history rewrite and a force-push. A squash-merge takes the corrected description instead and the problem disappears. Happy to rewrite the branch if you would rather merge it as-is.

wangkanai:
recheck

PR #6383, opened 2026-08-20
move streaming back to net/http to avoid race conditions

### LINKED ISSUE #6143: [Bug]: data race — fasthttp requestStream released to pool while SSE reader is still inside Read (stream cancellation)
### What happened

When a streaming request is cancelled while its SSE body is still being read, two goroutines touch the same pooled `fasthttp.requestStream` concurrently with no synchronization. Go's race detector reports it reliably.

* **Reader goroutine** — the SSE loop is blocked inside `(*requestStream).Read`, reading `rs.chunkLeft` / `rs.reader` / `rs.header`.
* **Closer goroutine** — `SetupStreamCancellation`'s watchdog observes `ctx.Done()` and calls `CloseWithError`, which runs fasthttp's `RoundTrip` cleanup → `releaseRequestStream`, which **zeroes those same fields and returns the struct to** `requestStreamPool`.

The object isn't merely written during a read — it's **recycled into a shared pool while another goroutine is still inside a method on it**. A subsequent request can `Get` the same `*requestStream` and alias it.

### Why the existing guard doesn't cover this

The comment at `core/providers/utils/utils.go:2636-2643` describes a double-`Put` producing a "negative `chunkLeft` slice bound" panic, and adds an atomic claim (`GetAndSetValue`) to stop *two closers* from releasing the same stream.

That fix addresses **closer-vs-closer**. The race reported here is **closer-vs-reader**, which the atomic claim does not synchronize against — the reader never participates in the claim.

(Possibly adjacent to maximhq/bifrost#4617, which was a panic in `closeBodyStream`; this one is a data race on the pooled stream object rather than a panic.)

### Racing stacks (abridged)

```
Write by goroutine N:
  fasthttp.releaseRequestStream()            fasthttp@v1.71.0/streaming.go:100
  fasthttp.(*transport).RoundTrip.func1()    fasthttp@v1.71.0/client.go:3173
  fasthttp.(*closeReader).CloseWithError()   fasthttp@v1.71.0/http.go:367
  utils.SetupStreamCancellation.func1()      core@v1.7.8/providers/utils/utils.go:2653

Previous read by goroutine N-1:
  fasthttp.(*requestStream).Read()           fasthttp@v1.71.0/streaming.go:31
  utils.(*idleTimeoutReader).Read()          core@v1.7.8/providers/utils/utils.go:2863
  bufio.(*Scanner).Scan()
  utils.(*defaultSSEDataReader).nextLine()   core@v1.7.8/providers/utils/sse.go:169
  openai.HandleOpenAIChatCompletionStreaming.func2()
                                             core@v1.7.8/providers/openai/openai.go:1214
```

Closer goroutine created at `core@v1.7.8/providers/utils/utils.go:2632` (`SetupStreamCancellation`); reader goroutine created at `core@v1.7.8/providers/openai/openai.go:1148`.

The pooling is visible at `fasthttp@v1.71.0/streaming.go:93-106` (`releaseRequestStream` + `requestStreamPool`).

### Reproduction

Run any `-race` build that cancels a streaming request mid-body. In our case: a loopback mock provider emits 40 SSE chunks at 50 ms intervals, and the request is cancelled at 250 ms — i.e. cancellation lands while the body is still streaming, which is the trigger condition.

No frame from our own code appears anywhere in either racing stack; both sides are bifrost + fasthttp.

### Versions

| Module | Version |
| -- | -- |
| `github.com/maximhq/bifrost/core` | `v1.7.8` |
| `github.com/valyala/fasthttp` | `v1.71.0` (required by `core`'s `go.mod`) |
| Go toolchain | `go1.26.5 darwin/arm64` |

### Upgrades that do **not** fix it

Both upstream axes were tested independently. This is valid because `dev` HEAD's `core/go.mod` still requires `fasthttp v1.71.0` — bumping bifrost does not bump fasthttp.

| Configuration | `WARNING: DATA RACE` occurrences | Result |
| -- | -- | -- |
| baseline: `core v1.7.8` + `fasthttp v1.71.0` | 57 | FAIL |
| bifrost `dev` HEAD (`d834a012ff7389f769a00695d95436da54f6cddd`) + `fasthttp v1.71.0` | 60 | FAIL |
| `core v1.7.8` + `fasthttp v1.73.0` | 60 | FAIL |

Counts are one report per cancelled stream across a full suite run; 57 → 60 is run-to-run variation in a timing-dependent race, **not** a regression signal.

### Impact

This is the generic OpenAI-compatible streaming handler over the fasthttp client — the same path a real provider request takes. Any client that cancels streams routinely (request timeouts, `SIGINT`, server-side per-request deadlines) hits it. Consequences of a pooled-object aliasing bug range from silent response corruption to the `negative chunkLeft` panic your own comment documents.

### What we could not determine

* Whether a commit **after** `dev` HEAD `d834a01` fixes it — only that commit was tested.
* Whether this has produced observable corruption or a panic in a real run. We have the detector report only; no user-visible failure observed yet.
* Whether non-OpenAI-compatible providers (Anthropic, Bedrock, …) hit the identical path. Only the OpenAI handler appears in our captured stacks. `SetupStreamCancellation` lives in shared `providers/utils`, which suggests they would, but we did not execute those paths.
* Whether Linux/amd64 reproduces at the same rate — only `darwin/arm64` was exercised.

Happy to supply the full race reports or test a candidate patch.

akshaydeo:
## Summary

Fixes a data race (issue #6143) where fasthttp's streaming close callback returns the pooled `*requestStream` and `*bufio.Reader` to their `sync.Pool` **before** closing the connection. Since closing the connection is the only thing that unblocks a reader parked in `(*requestStream).Read`, cancelling a stream mid-body always resumes the reader on an object another request may already own. This ordering is inside fasthttp and cannot be fixed with caller-side locking. The fix routes all streaming responses through `net/http`, whose `Body.Close` is mutex-guarded against concurrent `Read` calls and pools nothing.

## Changes

- Added `core/providers/utils/httpstream.go`, which implements `DoStreamingRequestViaHTTP`. It converts a `*fasthttp.Request` to a `*http.Request`, sends it through a cached `*http.Client` twin (keyed on the long-lived provider `streamingClient` pointer), and injects the resulting `http.Response` body into the `fasthttp.Response` via `SetBodyStream`. All downstream helpers (`ExtractProviderResponseHeaders`, `DecompressStreamBody`, `ReleaseStreamingResponse`) continue to work unchanged because they operate on the `fasthttp.Response` fields, not the underlying transport.
- `DoStreamingRequest` now delegates to `DoStreamingRequestViaHTTP` instead of `client.Do`, making the fix transparent to every provider.
- Added `MakeStreamingRequestWithContext` as the streaming counterpart to `MakeRequestWithContext`, sharing the same cancellation, latency and error-classification path while sending through `net/http`.
- Replaced `PrepareResponseStreaming` (which built a per-request fasthttp client clone) with `PrepareStreamResponseThreshold`, which only sets `resp.StreamBody = true`. The fasthttp clone's `StreamResponseBody`, `MaxResponseBodySize` and zeroed timeouts have no effect on a `net/http` send; the large-response threshold is enforced by Bifrost's own readers.
- Removed per-request `BuildStreamingClient` and `BuildLargeResponseClient` calls from the Anthropic, OpenAI, Gemini, Azure, and Vertex streaming paths. Reusing the provider's long-lived client keeps the `net/http` twin cache bounded by the number of providers rather than growing once per request.
- Normalized `io.ErrUnexpectedEOF` to `io.EOF` in `defaultSSEDataReader.ReadDataLine` and `defaultSSEEventReader.ReadEvent`. `net/http` reports a mid-chunk connection close as `io.ErrUnexpectedEOF` where fasthttp reported plain `io.EOF`, so without this the truncation detection path (which surfaces as a retryable 502) would be bypassed and replaced with a generic stream error.
- Added `httpstream_test.go` covering twin stability, dialer/TLS inheritance, request conversion fidelity, status/header propagation, and `Content-Length` ordering inside `injectHTTPStreamResponse`.
- Added `streamcancelrace_test.go` with a deterministic assertion that the streaming body is never a fasthttp type, plus a `-race`-detectable reproduction of the original race using a real in-memory fasthttp server, concurrent cancelled streams, and the full `idleTimeoutReader` + `SetupStreamCancellation` wrapper stack.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [x] Core (Go)
- [x] Transports (HTTP)
- [x] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

```sh
# Run all tests including the new race-detection coverage
go test -race ./core/providers/utils/... -v -run "TestDoStreamingRequest|TestStreamCancellation|TestStreamingHTTPTwin|TestFastHTTPRequestToHTTP|TestDoStreamingRequestViaHTTP|TestInjectHTTPStreamResponse"

# Full suite
go test -race ./...
```

The `TestStreamCancellation_PooledRequestStreamRace` test should be run with `-race`; without the fix it reports a data race between `releaseRequestStream` (write) and `(*requestStream).Read`. `TestDoStreamingRequest_BodyIsNotPooledFastHTTPStream` fails deterministically if the streaming body is ever a fasthttp type.

## Breaking changes

- [ ] Yes
- [x] No

## Related issues

Closes #6143

## Security considerations

`StreamingHTTPTwin` inherits the fasthttp client's `TLSConfig` (custom CA, `InsecureSkipVerify`) and `Dial` function (which carries SSRF address filtering and proxy configuration). No security policy is weakened or bypassed by the transport switch.

## Checklist

### DIFF
--- a/core/providers/openai/openai.go
+++ b/core/providers/openai/openai.go
@@ -491,13 +491,15 @@ func HandleOpenAITextCompletionStreaming(
 
 	setStreamingRequestBody(ctx, req, jsonBody, providerName)
 
-	// Use streaming-aware client when large payload optimization is active — ensures
-	// MaxResponseBodySize > 0 so ErrBodyTooLarge triggers StreamBody for Content-Length responses.
-	activeClient := providerUtils.PrepareResponseStreaming(ctx, client, resp)
+	// Mark the response for streaming when large payload optimization is active.
+	// The fasthttp MaxResponseBodySize/StreamResponseBody clone this used to build has
+	// no effect now that streaming sends go through net/http; the threshold is enforced
+	// by Bifrost's own readers.
+	providerUtils.PrepareStreamResponseThreshold(ctx, resp)
 
 	startTime := time.Now()
 	// Make the request
-	err := providerUtils.DoStreamingRequest(ctx, activeClient, req, resp)
+	err := providerUtils.DoStreamingRequest(ctx, client, req, resp)
 	if err != nil {
 		defer providerUtils.ReleaseStreamingResponse(ctx, resp)
 		latency := time.Since(startTime)
@@ -1109,13 +1111,15 @@ func HandleOpenAIChatCompletionStreaming(
 
 	setStreamingRequestBody(ctx, req, jsonBody, providerName)
 
-	// Use streaming-aware client when large payload optimization is active — ensures
-	// MaxResponseBodySize > 0 so ErrBodyTooLarge triggers StreamBody for Content-Length responses.
-	activeClient := providerUtils.PrepareResponseStreaming(ctx, client, resp)
+	// Mark the response for streaming when large payload optimization is active.
+	// The fasthttp MaxResponseBodySize/StreamResponseBody clone this used to build has
+	// no effect now that streaming sends go through net/http; the threshold is enforced
+	// by Bifrost's own readers.
+	providerUtils.PrepareStreamResponseThreshold(ctx, resp)
 
 	startTime := time.Now()
 	// Make the request
-	err := providerUtils.DoStreamingRequest(ctx, activeClient, req, resp)
+	err := providerUtils.DoStreamingRequest(ctx, client, req, resp)
 	latency := time.Since(startTime)
 	if err != nil {
 		defer providerUtils.ReleaseStreamingResponse(ctx, resp)
@@ -1823,13 +1827,15 @@ func HandleOpenAIResponsesStreaming(
 
 	setStreamingRequestBody(ctx, req, jsonBody, providerName)
 
-	// Use streaming-aware client when large payload optimization is active — ensures
-	// MaxResponseBodySize > 0 so ErrBodyTooLarge triggers StreamBody for Content-Length responses.
-	activeClient := providerUtils.PrepareResponseStreaming(ctx, client, resp)
+	// Mark the response for streaming when large payload optimization is active.
+	// The fasthttp MaxResponseBodySize/StreamResponseBody clone this used to build has
+	// no effect now that streaming sends go through net/http; the threshold is enforced
+	// by Bifrost's own readers.
+	providerUtils.PrepareStreamResponseThreshold(ctx, resp)
 
 	startTime := time.Now()
 	// Make the request
-	err := providerUtils.DoStreamingRequest(ctx, activeClient, req, resp)
+	err := providerUtils.DoStreamingRequest(ctx, client, req, resp)
 	latency := time.Since(startTime)
 	if err != nil {
 		defer providerUtils.ReleaseStreamingResponse(ctx, resp)
@@ -2431,13 +2437,15 @@ func HandleOpenAISpeechStreamRequest(
 
 	setStreamingRequestBody(ctx, req, jsonBody, providerName)
 
-	// Use streaming-aware client when large payload optimization is active — ensures
-	// MaxResponseBodySize > 0 so ErrBodyTooLarge triggers StreamBody for Content-Length responses.
-	activeClient := providerUtils.PrepareResponseStreaming(ctx, client, resp)
+	// Mark the response for streaming when large payload optimization is active.
+	// The fasthttp MaxResponseBodySize/StreamResponseBody clone this used to build has
+	// no effect now that streaming sends go through net/http; the threshold is enforced
+	// by Bifrost's own readers.
+	providerUtils.PrepareStreamResponseThreshold(ctx, resp)
 
 	startTime := time.Now()
 	// Make the request
-	err := providerUtils.DoStreamingRequest(ctx, activeClient, req, resp)
+	err := providerUtils.DoStreamingRequest(ctx, client, req, resp)
 	latency := time.Since(startTime)
 	if err != nil {
 		defer providerUtils.ReleaseStreamingResponse(ctx, resp)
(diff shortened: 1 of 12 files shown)


akshaydeo:
* **#6383** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/6383?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/maximhq/bifrost/6383?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `dev`


<h2></h2>

theidy3001-ai (approved):
<sub>Approved in [CodeRabbit Change Stack](https://app.coderabbit.ai/change-stack/maximhq/bifrost/pull/6383)</sub>

core/providers/openai/openai.go
```diff
@@ -491,13 +491,15 @@ func HandleOpenAITextCompletionStreaming(
 
 	setStreamingRequestBody(ctx, req, jsonBody, providerName)
 
-	// Use streaming-aware client when large payload optimization is active — ensures
-	// MaxResponseBodySize > 0 so ErrBodyTooLarge triggers StreamBody for Content-Length responses.
-	activeClient := providerUtils.PrepareResponseStreaming(ctx, client, resp)
+	// Mark the response for streaming when large payload optimization is active.
+	// The fasthttp MaxResponseBodySize/StreamResponseBody clone this used to build has
```
theidy3001-ai, line 495: WakeUp

core/providers/openai/openai.go
```diff
@@ -491,13 +491,15 @@ func HandleOpenAITextCompletionStreaming(
...
+	// no effect now that streaming sends go through net/http; the threshold is enforced
+	// by Bifrost's own readers.
+	providerUtils.PrepareStreamResponseThreshold(ctx, resp)
 
 	startTime := time.Now()
```
theidy3001-ai, line 500: [

```suggestion

	startTime := time.Now()

```suggestion

	startTime := time.Now()

```suggestion

- [ ] 	startTime := time.Now()

![7C83B8F4-7772-4E54-BA50-64006DEAE255.jpg](https://app.coderabbit.ai/api/change-stack/comment-images/51ac28fe-0dc0-48c3-89a8-4d089b55ac1d/7C83B8F4-7772-4E54-BA50-64006DEAE255.jpg)

![IMG_1122.png](https://app.coderabbit.ai/api/change-stack/comment-images/3b456bc4-1018-42f5-94c7-8b9b8c3dfdfc/IMG_1122.png)

![IMG_1121.png](https://app.coderabbit.ai/api/change-stack/comment-images/9c40f5ea-c551-4716-b903-b4189e7f431c/IMG_1121.png)
```

```

```

](url)

theidy3001-ai (commented):
<sub>Commented in [CodeRabbit Change Stack](https://app.coderabbit.ai/change-stack/maximhq/bifrost/pull/6383)</sub>

PR #6392, opened 2026-08-21
fix(anthropic): mid-conv system placement — skip roleless assistant-side fragments

is911:
## Problem

`midConvPlacementOK` in `ConvertBifrostMessagesToAnthropicMessages` decides whether a
mid-conversation system message can be forwarded natively as `role:"system"` at array
position. Its "followed by an assistant turn" clause inspected only the **immediately
next** message:

```go
next := bifrostMessages[i+1]
return next.Role != nil && *next.Role == schemas.ResponsesInputMessageRoleAssistant
```

When the adjacent assistant turn arrives as **roleless `reasoning` / `function_call`
items** — which is how every extended-thinking and tool-calling conversation replays
(Anthropic's own wire carries thinking and tool_use *inside* the assistant message, and
the egress below regroups these fragments into it) — the clause false-negatived and the
gate fell back to hoist/inline for exactly the thinking-replay shape that
extended-thinking conversations always produce on replay.

## Fix

Evaluate the clause against the next **role-bearing** message, skipping assistant-side
fragments:

- roleless `reasoning` items are skipped (they regroup into the adjacent assistant message);
- a roleless `function_call` settles the clause affirmatively (its tool results follow it,
  never precede it, confirming the regrouped assistant message is emitted directly after
  the system turn);
- a roleless `function_call_output` with no `function_call` ahead of it is the **user**
  side of a tool exchange (it regroups into a user turn), so it and any other roleless
  shape still fail the clause — guarding the `[user, system, user]` placement Anthropic
  rejects.

## Tests

Three roundtrip cases in `core/providers/anthropic/roundtrip_test.go`:

- **B3** — mid-conv system followed by a thinking-replay assistant turn: forwarded
  natively, no hoist (`user,system,assistant,user`).
- **B4** — mid-conv system followed by a textless tool_use assistant turn (no role-bearing
  item of its own): forwarded natively.
- **B5** — mid-conv system followed by a tool-result user turn: **not** forwarded
  (regression guard for the `[user, system, user]` 400 shape).

Full `go test ./core/providers/anthropic/` passes on `dev` + this commit.

## Provider-harness note

No harness case is added, deliberately: when the gate declines, the request routes to the
inline/hoist fallback — both produce valid Anthropic requests with identical response
envelopes — so this change has no client-observable signature (status code, error body, or
response fields) for a harness case to pin. The behavior it changes is the *egress* message
placement, which is pinned by the roundtrip tests above; route acceptance of these
mid-conversation system shapes is already covered by the "Cross-Cut Round 29:
Mid-Conversation System Message Matrix" cases in the collection.

### DIFF
--- a/core/providers/anthropic/responses.go
+++ b/core/providers/anthropic/responses.go
@@ -4966,8 +4966,45 @@ func ConvertBifrostMessagesToAnthropicMessages(ctx *schemas.BifrostContext, bifr
 		if i == len(bifrostMessages)-1 {
 			return true
 		}
-		next := bifrostMessages[i+1]
-		return next.Role != nil && *next.Role == schemas.ResponsesInputMessageRoleAssistant
+		// The "followed by an assistant turn" clause is evaluated against the next
+		// ROLE-BEARING message. Roleless reasoning and function_call items are
+		// intermediate representations of the adjacent assistant turn — Anthropic's
+		// own wire carries thinking and tool_use INSIDE the assistant message, and
+		// the egress below regroups these fragments into it — so they are skipped.
+		// Failing on them was a false negative that forced the fallback (hoist or
+		// inline) for exactly the thinking-replay shape every extended-thinking
+		// conversation produces on replay. A function_call fragment confirms the
+		// regrouped assistant message is emitted directly after the system turn
+		// (its tool results follow it, never precede it), which settles the clause
+		// then and there. A roleless function_call_output with no function_call
+		// ahead of it is the USER side of a tool exchange (it regroups into a user
+		// turn), so it and any other roleless shape fail the clause instead.
+		// Known theoretical false positive, unreachable from opencode-shaped
+		// traffic (which anchors system entries right after a role-bearing user
+		// message): a hand-crafted [user, function_call_output, system,
+		// function_call] sequence — clause 1 passes (the buffered output flushes
+		// later), the scan sees function_call and returns true, but the egress
+		// flushes the buffered results as a USER message between the system turn
+		// and the regrouped assistant, yielding [user, system, user, assistant].
+		for j := i + 1; j < len(bifrostMessages); j++ {
+			next := bifrostMessages[j]
+			if next.Role != nil {
+				return *next.Role == schemas.ResponsesInputMessageRoleAssistant
+			}
+			if next.Type == nil {
+				return false
+			}
+			switch *next.Type {
+			case schemas.ResponsesMessageTypeFunctionCall:
+				return true
+			case schemas.ResponsesMessageTypeReasoning:
+				continue
+			}
+			return false
+		}
+		// Only reasoning fragments followed the system turn; they regroup into
+		// an assistant message, satisfying the clause.
+		return true
 	}
 
 	// Helper to emit orphaned tool results (no matching tool_use) as a single user
--- a/core/providers/anthropic/responses.go
+++ b/core/providers/anthropic/responses.go
@@ -4973,19 +4973,16 @@ func ConvertBifrostMessagesToAnthropicMessages(ctx *schemas.BifrostContext, bifr
 		// the egress below regroups these fragments into it — so they are skipped.
 		// Failing on them was a false negative that forced the fallback (hoist or
 		// inline) for exactly the thinking-replay shape every extended-thinking
-		// conversation produces on replay. A function_call fragment confirms the
-		// regrouped assistant message is emitted directly after the system turn
-		// (its tool results follow it, never precede it), which settles the clause
-		// then and there. A roleless function_call_output with no function_call
-		// ahead of it is the USER side of a tool exchange (it regroups into a user
-		// turn), so it and any other roleless shape fail the clause instead.
-		// Known theoretical false positive, unreachable from opencode-shaped
-		// traffic (which anchors system entries right after a role-bearing user
-		// message): a hand-crafted [user, function_call_output, system,
-		// function_call] sequence — clause 1 passes (the buffered output flushes
-		// later), the scan sees function_call and returns true, but the egress
-		// flushes the buffered results as a USER message between the system turn
-		// and the regrouped assistant, yielding [user, system, user, assistant].
+		// conversation produces on replay. A function_call fragment settles the
+		// clause affirmatively only when it carries a tool payload — a payload-less
+		// one emits no tool_use (convertBifrostFunctionCallToAnthropicToolUse
+		// returns nil), leaving the system turn followed by user turns, the exact
+		// 400 shape this gate exists to avoid. A roleless function_call_output with
+		// no function_call ahead of it is the USER side of a tool exchange (it
+		// regroups into a user turn), so it — and any other roleless shape,
+		// including the assistant-side item types not handled below (computer_call,
+		// mcp_call, web_search_call, ...) — fails the clause. That fallthrough is
+		// the safe, conservative direction: the caller falls back to inlining.
 		for j := i + 1; j < len(bifrostMessages); j++ {
 			next := bifrostMessages[j]
 			if next.Role != nil {
@@ -4996,7 +4993,7 @@ func ConvertBifrostMessagesToAnthropicMessages(ctx *schemas.BifrostContext, bifr
 			}
 			switch *next.Type {
 			case schemas.ResponsesMessageTypeFunctionCall:
-				return true
+				return next.ResponsesToolMessage != nil
 			case schemas.ResponsesMessageTypeReasoning:
 				continue
 			}
--- a/core/providers/anthropic/roundtrip_test.go
+++ b/core/providers/anthropic/roundtrip_test.go
@@ -11,11 +11,15 @@ package anthropic
 //   F) No top-level system, mid-conv only, unsupported
 //   G) Multiple mid-conv system messages, supported
 //   H) Multiple mid-conv system messages, unsupported
+//   B3) Mid-conv system followed by a thinking-replay assistant turn, supported
+//   B4) Mid-conv system followed by a textless tool_use assistant turn, supported
+//   B5) Mid-conv system followed by a tool-result user turn — NOT forwarded
 //
 // "Supported" means provider=Anthropic + model=claude-opus-4-8 (SupportsMidConversationSystem=true).
 
 import (
 	"context"
+	"encoding/json"
 	"strings"
 	"testing"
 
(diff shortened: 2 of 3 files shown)


core/providers/anthropic/responses.go
```diff
@@ -4966,8 +4966,45 @@ func ConvertBifrostMessagesToAnthropicMessages(ctx *schemas.BifrostContext, bifr
...
+			next := bifrostMessages[j]
+			if next.Role != nil {
+				return *next.Role == schemas.ResponsesInputMessageRoleAssistant
+			}
+			if next.Type == nil {
+				return false
+			}
+			switch *next.Type {
+			case schemas.ResponsesMessageTypeFunctionCall:
+				return true
```
akshaydeo: **Blocking: this can produce the `[user, system, user]` shape Anthropic 400s on.**

`convertBifrostFunctionCallToAnthropicToolUse` returns `nil` when `msg.ResponsesToolMessage == nil` (responses.go:6916), so a `function_call` item with no tool payload contributes nothing to the output. This `return true` promises an assistant turn that never materializes.

Reproduction, input parsed straight off the wire with `caps = (Anthropic, claude-opus-4-8)`:

```json
[{"type":"message","role":"user","content":"hi"},
 {"type":"message","role":"system","content":"be concise"},
 {"type":"function_call"},
 {"type":"message","role":"user","content":"thanks"}]
```

* merge base `78a220fb7`: `roleSeq = user,user,user` (system inlined, safe)
* this head `810c513bb`: `roleSeq = user,system,user`

Upstream: "A `system` message must immediately follow a `user` turn [...] and must precede an `assistant` turn or end the array. [...] Placing it elsewhere returns a 400 error." (https://platform.claude.com/docs/en/build-with-claude/mid-conversation-system-messages)

The doc comment above your change already states the invariant this breaks: "A false negative costs nothing [...] A false positive would be a 400."

Suggested fix, which degrades back to the old conservative behavior for the empty item:

```go
case schemas.ResponsesMessageTypeFunctionCall:
	// Only an item that actually carries a tool payload emits a tool_use;
	// convertBifrostFunctionCallToAnthropicToolUse returns nil otherwise, and
	// promising an assistant turn that never materializes is a 400.
	return next.ResponsesToolMessage != nil
```

Worth a B6 case pinning `[user, system, function_call(no payload), user]` as not forwarded, alongside B5.

core/providers/anthropic/responses.go
```diff
@@ -4966,8 +4966,45 @@ func ConvertBifrostMessagesToAnthropicMessages(ctx *schemas.BifrostContext, bifr
...
+		// the egress below regroups these fragments into it — so they are skipped.
+		// Failing on them was a false negative that forced the fallback (hoist or
+		// inline) for exactly the thinking-replay shape every extended-thinking
+		// conversation produces on replay. A function_call fragment confirms the
+		// regrouped assistant message is emitted directly after the system turn
+		// (its tool results follow it, never precede it), which settles the clause
+		// then and there. A roleless function_call_output with no function_call
+		// ahead of it is the USER side of a tool exchange (it regroups into a user
+		// turn), so it and any other roleless shape fail the clause instead.
+		// Known theoretical false positive, unreachable from opencode-shaped
```
akshaydeo: This paragraph documents a hazard that does not exist, and I would delete it.

`flushPendingToolResults()` runs at the **top** of the `case schemas.ResponsesMessageTypeMessage` branch (responses.go:5118), which is the branch a system message takes. So the buffered tool results are flushed *before* the system turn is appended, never after it. Running that exact sequence:

```
input:  [user, function_call_output, system, function_call]
output: roleSeq = user,user,system,assistant
```

That is a valid request: the system turn immediately follows a user turn and is immediately followed by an assistant turn, and the two leading user turns are fine because "Consecutive `user` or `assistant` turns in your request will be combined into a single turn" (https://platform.claude.com/docs/en/api/messages).

Left as is, the next person to touch this gate will believe there is a latent 400 here. Removing it also brings the comment block closer to the size of the code it explains: 20 lines of prose on 20 lines of logic is heavy for a helper whose contract is already stated in the doc comment above.

While you are here, it is worth noting in the comment that `function_call` and `reasoning` are not the only roleless assistant-side item types. `computer_call`, `custom_tool_call`, `code_interpreter_call`, `local_shell_call`, `mcp_call`, `image_generation_call`, `web_search_call` and `web_fetch_call` all regroup into an assistant turn in the egress below and all hit the `return false` fallthrough, so a computer-use session with extended thinking still pays the fallback (`[user, system, computer_call, user]` gives `roleSeq = user,user,assistant,user`). Safe direction, so a follow-up is fine, but the comment currently reads as complete.

akshaydeo (changes requested):
Thanks for this - the diagnosis is genuinely sharp. You correctly identified that an extended-thinking replay arrives as roleless `reasoning` / `function_call` fragments and that the old one-message lookahead false-negatived on exactly the shape those conversations always produce. The tests are well chosen: I confirmed B3 and B4 are **red on the merge base** (`role seq = "user,user,assistant,user", want "user,system,assistant,user"`) and green on this head, and B5 is a proper regression guard. The full `core/providers/anthropic` package passes on this head.

The verification did surface one new false positive that I was able to reproduce, plus a comment paragraph that documents a hazard that does not actually exist. Details below, with the reproductions I ran.

### Findings

|---|---|---|---|---|
| 1 | High | [`responses.go:4999`](https://github.com/maximhq/bifrost/blob/810c513bb285f7ec7861a197d4fde76f2958791b/core/providers/anthropic/responses.go#L4999) | A `function_call` item with no tool payload emits no `tool_use`, so `return true` green-lights the `[user, system, user]` shape Anthropic 400s on | CONFIRMED |
| 2 | Medium | [`responses.go:4982`](https://github.com/maximhq/bifrost/blob/810c513bb285f7ec7861a197d4fde76f2958791b/core/providers/anthropic/responses.go#L4982) | The "Known theoretical false positive" paragraph is factually wrong: that sequence produces a valid request | CONFIRMED |
| 3 | Medium | PR description, "Provider-harness note" | AGENTS.md requires a harness case here; the stated exemption is refuted by the collection's own `x-bf-send-back-raw-request` pattern | CONFIRMED |
| 4 | Low | [`responses.go:4997-5002`](https://github.com/maximhq/bifrost/blob/810c513bb285f7ec7861a197d4fde76f2958791b/core/providers/anthropic/responses.go#L4997-L5002) | Only `function_call` and `reasoning` are handled; the other roleless assistant-side item types still false-negative | CONFIRMED (safe direction) |
| 5 | Low | [`responses.go:4989`](https://github.com/maximhq/bifrost/blob/810c513bb285f7ec7861a197d4fde76f2958791b/core/providers/anthropic/responses.go#L4989) | Consecutive mid-conv system messages now get split treatment: first inlined, second native | CONFIRMED (valid, but a missed cache win) |

### Finding 1 (blocking): `function_call` with no tool payload

`convertBifrostFunctionCallToAnthropicToolUse` returns `nil` when `msg.ResponsesToolMessage == nil` (responses.go:6916-6917), so such an item contributes nothing to the output. The new `case ResponsesMessageTypeFunctionCall: return true` does not check for that, so the gate promises an assistant turn that never materializes.

Reproduction, driving `ConvertBifrostMessagesToAnthropicMessages` directly with `caps = (Anthropic, claude-opus-4-8)`. Input parsed straight off the wire, no hand-built structs:

```json
[{"type":"message","role":"user","content":"hi"},
 {"type":"message","role":"system","content":"be concise"},
 {"type":"function_call"},
 {"type":"message","role":"user","content":"thanks"}]
```

* merge base `78a220fb7`: `roleSeq = user,user,user` (system inlined, safe)
* this head `810c513bb`: `roleSeq = user,system,user`

Per the upstream placement rules, that second output is a 400:

> "A `system` message must immediately follow a `user` turn (including a `user` turn that carries `tool_result` blocks) or an `assistant` turn ending in a server tool result, and must precede an `assistant` turn or end the array. [...] Placing it elsewhere returns a 400 error."
> - https://platform.claude.com/docs/en/build-with-claude/mid-conversation-system-messages

This matters because the function's own contract says so in the doc comment right above your change: "A false negative costs nothing [...] A false positive would be a 400." The fix is one condition, and it degrades exactly back to the old conservative behavior for the empty item:

```go
case schemas.ResponsesMessageTypeFunctionCall:
	// Only an item that actually carries a tool payload emits a tool_use;
	// convertBifrostFunctionCallToAnthropicToolUse returns nil otherwise, and
	// promising an assistant turn that never materializes is a 400.
	return next.ResponsesToolMessage != nil
```

A B6 case pinning `[user, system, function_call(no payload), user]` as not-forwarded would lock this down alongside B5.

### Finding 2: the documented false positive does not exist

The comment says:

> a hand-crafted `[user, function_call_output, system, function_call]` sequence [...] the egress flushes the buffered results as a USER message between the system turn and the regrouped assistant, yielding `[user, system, user, assistant]`.

That is not what happens. `flushPendingToolResults()` runs at the **top** of the `case schemas.ResponsesMessageTypeMessage` branch (responses.go:5118), which is the branch a system message takes, so the buffered results are flushed *before* the system turn is appended, never after it. I ran that exact sequence:

```
roleSeq = user,user,system,assistant
msgs = [{"role":"user","content":[{"type":"text","text":"hi"}]},
        {"role":"user","content":[{"type":"text","text":"Tool result: 21"}]},
        {"role":"system","content":"be concise"},
        {"role":"assistant","content":[{"type":"tool_use","id":"toolu_y",...}]}]
```

That is a valid request: the system turn immediately follows a user turn and is immediately followed by an assistant turn, and the two leading user turns are fine because "Consecutive `user` or `assistant` turns in your request will be combined into a single turn" (https://platform.claude.com/docs/en/api/messages).

I would delete that paragraph rather than keep it. Left as is, the next person to touch this gate will believe there is a latent 400 here and either write a defensive guard for a non-problem or hesitate to extend the scan. Trimming it also brings the comment block closer to the size of the code it explains - 20 lines of prose on 20 lines of logic is heavy for a helper whose contract is already stated above it.

### Finding 3: the harness exemption

AGENTS.md, Testing section:

> "Any change under `core/` that a client can observe on the wire must land together with a case in `tests/e2e/api/collections/provider-harness.json` [...] The narrow exemptions: changes with no wire-visible effect (comments, internal renames, log lines) and behaviour no HTTP request can reach."

The note argues there is "no client-observable signature (status code, error body, or response fields) for a harness case to pin". That reasoning holds only for the *response*, and the harness does not limit itself to responses: the collection asserts the **outbound** payload in 12 places via `x-bf-send-back-raw-request`, with a case comment that literally says it checks "the OUTBOUND payload via x-bf-send-back-raw-request rather than inferring from model behaviour". The Round 29 mid-conversation system matrix you cite lives in the same file.

So a case is both possible and, by the rule above, required: send a thinking-replay transcript with a mid-conv system message, set `x-bf-send-back-raw-request: true`, and assert that `extra_fields.raw_request.messages` contains an entry with `role == "system"` rather than a `<system-reminder>` user turn. That case is red before this change and green after, which is exactly the shape AGENTS.md asks for. Given finding 1, it would also catch the empty-`function_call` regression at the wire level.

### Finding 4: the skip set is narrower than "assistant-side fragments"

`function_call` and `reasoning` are not the only roleless assistant-side item types. `computer_call`, `custom_tool_call`, `code_interpreter_call`, `local_shell_call`, `mcp_call`, `image_generation_call`, `web_search_call` and `web_fetch_call` all regroup into an assistant turn in the egress below, and all of them hit the `return false` fallthrough. Confirmed for computer use:

```
input:  [user, system, computer_call, user]
output: roleSeq = user,user,assistant,user   (system inlined, not forwarded)
```

This is the safe direction, so it is not blocking. But a Claude computer-use session with extended thinking still pays the fallback on every turn, which is the same cost the PR set out to remove. Since these types all funnel into `pendingToolCalls` in the same way `function_call` does, extending the `case` list (with the same "carries a payload" guard from finding 1) would close it. Fine as a follow-up if you would rather keep this PR tight - just worth naming the remaining gap in the comment so it does not read as complete.

### Finding 5: consecutive system messages split

The scan returns `false` on a role-bearing system/developer message, so in `[user, system1, system2, assistant, user]` the first system is inlined and the second goes native:

```
roleSeq = user,user,system,assistant,user
msgs[1] = <system-reminder>s1</system-reminder>   (inlined)
msgs[2] = {"role":"system","content":"s2"}        (native)
```

That is a legal request, so nothing is broken. But it splits one instruction run across two representations and half of it still loses the cache benefit, and upstream explicitly supports the whole-run form:

> "Consecutive system messages are accepted and treated as a single system section, which follows the same placement rule as a whole."
> - https://platform.claude.com/docs/en/build-with-claude/mid-conversation-system-messages

Skipping over system/developer messages in the scan (the run is evaluated as a unit, so the clause for the run is the clause for its last member) would let both go native. Follow-up, not blocking.

### Merge recommendation

**Not yet - request changes.** Finding 1 is a new client-triggerable 400 on a path that is safe on `dev` today, and it contradicts the gate's own stated invariant; the fix is a single `next.ResponsesToolMessage != nil` guard plus a B6 test. Findings 2 and 3 are cheap to address in the same push. Findings 4 and 5 are genuinely optional and can ship separately. The core insight and the B3/B4/B5 tests are good and should land.

### Followups

1. **In this PR (blocking)** - `core/providers/anthropic/responses.go:4999`: guard the `function_call` case with `next.ResponsesToolMessage != nil` so an item that emits no `tool_use` cannot green-light the gate.
2. **In this PR (blocking)** - `core/providers/anthropic/roundtrip_test.go`: add B6 pinning `[user, system, function_call(no payload), user]` as not forwarded natively.
3. **In this PR** - `core/providers/anthropic/responses.go:4982-4988`: delete the "Known theoretical false positive" paragraph; the flush ordering at responses.go:5118 makes that sequence valid.
4. **In this PR** - `tests/e2e/api/collections/provider-harness.json`: add the raw-request harness case described in finding 3, next to the Round 29 matrix.
5. **Follow-up PR** - `core/providers/anthropic/responses.go:4997-5002`: extend the skip set to the remaining roleless assistant-side item types, or state the remaining gap in the comment.
6. **Follow-up PR** - `core/providers/anthropic/responses.go:4989`: skip over consecutive system/developer messages so an instruction run goes native as a whole.
7. **Follow-up PR** - `core/providers/anthropic/chat.go:810`: the Chat Completions path carries the same one-message-lookahead gate. `ChatMessage` always has a role so it has no roleless-fragment blind spot, but if the two gates are meant to stay in lockstep, a comment pointing at this asymmetry would help. Not a bug today.

* **Trailing reasoning-only after a mid-conv system drops the assistant turn.** `[user, system, reasoning]` yields `[user, system]`. System as the last entry is explicitly legal, so the new `return true` fallthrough is correct. (Separately, and not caused by this PR: `pendingReasoningContentBlocks` has no post-loop flush, so trailing thinking is dropped either way.)
* **An assistant message with empty text after the system turn collapses to `[user, system, user]`.** It does not; the empty assistant message is still emitted. `roleSeq = user,system,assistant,user`.
* **The `function_call` "merge into previous assistant message" branch could rewind past the just-appended system turn.** It cannot: the merge is guarded on `lastMsg.Role == assistant`, and the system message it would have to rewind past has `role: "system"`.
* **The trailing-assistant prefill trim could strip the assistant turn the scan relied on.** It can, but the result is `[user, system]` with the system turn last, which is legal.
* **B5 is not red-before-green.** It is green on both sides, which is correct for a regression guard rather than a defect.

---

Reviewed at head `810c513bb`. Reproductions were run in a throwaway worktree at merge base `78a220fb7` and at this head, driving `ConvertBifrostMessagesToAnthropicMessages` directly.

core/providers/anthropic/responses.go
is911: Fixed in fa3788d42, exactly as you suggested: the `FunctionCall` case now returns `next.ResponsesToolMessage != nil`, so a payload-less item degrades back to the conservative inline instead of promising an assistant turn that never materializes.

Pinned two ways:
- **B6** (`TestRoundTrip_B6_MidConvFollowedByPayloadlessFunctionCall_NotForwarded`) drives `ConvertBifrostMessagesToAnthropicMessages` with your exact wire JSON — red on the previous head (`role seq = "user,system,user"`), green after the guard.
- **Harness**: a Round 29 case sends `[user, system, function_call(no payload), user]` through `/v1/responses` with `x-bf-send-back-raw-request`, asserting 2xx, no `role:"system"` in `raw_request.messages`, and the `<system-reminder>` inline fallback present. It also catches finding 3's ask — the native-forwarding sibling case (mid-conv system before a function_call replay) asserts `role:"system"` lands between the leading user turn and the regrouped assistant turn with no `system-reminder` fallback.

core/providers/anthropic/responses.go
is911: Deleted in fa3788d42 — you're right that the flush ordering at the top of the message branch makes that sequence valid, so the paragraph documented a hazard that can't occur.

The replacement comment now states the payload rule for `function_call` and names the remaining roleless assistant-side item types (`computer_call`, `mcp_call`, `web_search_call`, ...) as a deliberate conservative fallthrough rather than reading as complete. Extending the skip set (your follow-up 5) and consecutive system/developer handling (follow-up 6) are left for follow-up PRs as suggested.

core/providers/anthropic/responses.go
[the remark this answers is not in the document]
is911: Fixed in fa3788d42 — the discriminator now returns `next.ResponsesToolMessage != nil` for the `FunctionCall` case, so a non-emittable item fails the clause instead of being classified as an assistant turn. Pinned by B6 (`TestRoundTrip_B6_MidConvFollowedByPayloadlessFunctionCall_NotForwarded`, red before / green after) and a Round 29 harness case asserting no `role:"system"` and the `<system-reminder>` fallback in the raw request.

core/providers/anthropic/roundtrip_test.go
```diff
@@ -518,3 +522,118 @@ func TestRoundTrip_ContainerUpload_Grouped(t *testing.T) {
...
+	result := "21"
+	messages := []AnthropicMessage{
+		{
+			Role: AnthropicMessageRoleUser,
+			Content: AnthropicContent{ContentBlocks: []AnthropicContentBlock{
+				{Type: AnthropicContentBlockTypeToolResult, ToolUseID: &callID, Content: &AnthropicContent{ContentStr: &result}},
+			}},
+		},
+		anthMsg(AnthropicMessageRoleSystem, "From now on, be concise."),
+		anthMsg(AnthropicMessageRoleUser, "Thanks."),
```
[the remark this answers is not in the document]
is911: Fixed in fa3788d42 — B5 now runs `[user, system, tool_result-user, user]`, so the Bifrost order is `user, system, function_call_output, user` and the scan meets the roleless `function_call_output` fallthrough directly instead of settling at the role check.

PR #6384, opened 2026-08-20
fix: build fix for scimAPI fallback

roroghost17:
## Summary

Improves type safety for the SCIM providers fallback query hook by replacing the generic `unknown[]` return type with a typed `{ enabled: boolean }[]` shape.

## Changes

- The `data` return type of `useGetSCIMProvidersQuery` in the enterprise fallback was changed from `unknown[]` to `{ enabled: boolean }[]`, ensuring consumers can safely access the `enabled` property without additional type assertions or runtime checks.

## Type of change

- [ ] Bug fix
- [ ] Feature
- [x] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [ ] Core (Go)
- [ ] Transports (HTTP)
- [ ] Providers/Integrations
- [ ] Plugins
- [x] UI (React)
- [ ] Docs

## How to test

```sh
cd ui
pnpm i || npm i
pnpm test || npm test
pnpm build || npm run build
```

## Breaking changes

- [ ] Yes
- [x] No

## Security considerations

No security implications.

## Checklist

- [ ] I read `docs/contributing/README.md` and followed the guidelines
- [ ] I added/updated tests where appropriate
- [ ] I updated documentation where needed
- [ ] I verified builds succeed (Go and UI)
- [ ] I verified the CI pipeline passes locally if applicable

PR #6393, opened 2026-08-21
fix(ui): type SCIM provider fallback stub elements

is911:
## Problem

The OSS fallback for `useGetSCIMProvidersQuery`
(`ui/app/_fallbacks/enterprise/lib/store/apis/scimApi.ts`) types its response as
`data: unknown[] | undefined`. Any property access on an element then fails typecheck —
`dev`'s `ui/hooks/useOnboardingChecklist.ts` currently does:

```ts
const ssoGatesDashboard = IS_ENTERPRISE && (scimProviders?.some((provider) => provider.enabled) ?? false);
```

which fails `npm run typecheck` with:

```
hooks/useOnboardingChecklist.ts(73,80): error TS18046: 'provider' is of type 'unknown'.
```

(Reproduced against `dev`'s exact code and lockfile, TS 5.9.2 — no local toolchain drift.)
The other three call sites work around the same root cause with per-site
`as { enabled?: boolean }` casts.

## Fix

Type the stub's elements as `{ enabled: boolean }[]`, mirroring the fields OSS consumers
read off the enterprise response. The un-cast access now typechecks, and the existing
call-site casts remain valid. Single-file change; `npm run typecheck` and
`npm run build` pass.

### DIFF
--- a/ui/app/_fallbacks/enterprise/lib/store/apis/scimApi.ts
+++ b/ui/app/_fallbacks/enterprise/lib/store/apis/scimApi.ts
@@ -19,13 +19,13 @@ export const useGetAuthTypeQuery = (
 // widget's enterprise-only "configure SCIM" step is always considered
 // incomplete (the step itself is hidden in OSS via IS_ENTERPRISE).
 // Element type mirrors the fields OSS consumers read off the enterprise
-// response (e.g. onboarding's `provider.enabled` gate); `unknown` elements
-// force every consumer into a per-site cast.
+// response (onboarding's `provider.enabled` gate, the MCP registry sheets'
+// `name` fallback label) so no consumer needs a per-site cast.
 export const useGetSCIMProvidersQuery = (
 	_args?: undefined,
 	_opts?: { skip?: boolean },
 ): {
-	data: { enabled: boolean }[] | undefined;
+	data: { enabled: boolean; name?: string }[] | undefined;
 	isLoading: boolean;
 	isError: boolean;
 	error: null;
--- a/ui/app/_fallbacks/enterprise/lib/store/apis/scimApi.ts
+++ b/ui/app/_fallbacks/enterprise/lib/store/apis/scimApi.ts
@@ -18,6 +18,9 @@ export const useGetAuthTypeQuery = (
 // OSS stub for SCIM providers — returns an empty list so the onboarding
 // widget's enterprise-only "configure SCIM" step is always considered
 // incomplete (the step itself is hidden in OSS via IS_ENTERPRISE).
+// Element type mirrors the fields OSS consumers read off the enterprise
+// response (e.g. onboarding's `provider.enabled` gate); `unknown` elements
+// force every consumer into a per-site cast.
 export const useGetSCIMProvidersQuery = (
 	_args?: undefined,
 	_opts?: { skip?: boolean },
--- a/ui/app/workspace/mcp-registry/views/mcpClientFormFields.tsx
+++ b/ui/app/workspace/mcp-registry/views/mcpClientFormFields.tsx
@@ -374,7 +374,7 @@ export function MCPClientFormFields({ form, satellites, headersValidationError, 
 	// exchange-client requirement is enforced server-side at create; a missing
 	// tokenExchangeClient section surfaces as the create error.
 	const { data: scimProviders } = useGetSCIMProvidersQuery(undefined, { skip: !IS_ENTERPRISE });
-	const enabledScimProvider = scimProviders?.find((p) => (p as { enabled?: boolean }).enabled) as { name?: string } | undefined;
+	const enabledScimProvider = scimProviders?.find((p) => p.enabled);
 	const idpConfigured = !!enabledScimProvider;
 	// Entra's on-behalf-of grant requires use_idp_credentials — see the
 	// Prerequisites warning in docs/mcp/auth/token-exchange.mdx for why a
--- a/ui/app/workspace/mcp-registry/views/mcpClientSheet.tsx
+++ b/ui/app/workspace/mcp-registry/views/mcpClientSheet.tsx
@@ -195,7 +195,7 @@ export default function MCPClientSheet({
 	// Prerequisites warning in docs/mcp/auth/token-exchange.mdx for why a
 	// dedicated exchange app structurally can't work there.
 	const { data: scimProviders } = useGetSCIMProvidersQuery(undefined, { skip: !IS_ENTERPRISE || !supportsTokenExchangeCredentialUpdate });
-	const enabledScimProvider = scimProviders?.find((p) => (p as { enabled?: boolean }).enabled) as { name?: string } | undefined;
+	const enabledScimProvider = scimProviders?.find((p) => p.enabled);
 	// Matches the create form's idpConfigured gate: without an enabled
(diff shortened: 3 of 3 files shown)


ui/app/_fallbacks/enterprise/lib/store/apis/scimApi.ts
```diff
@@ -18,11 +18,14 @@ export const useGetAuthTypeQuery = (
 // OSS stub for SCIM providers — returns an empty list so the onboarding
 // widget's enterprise-only "configure SCIM" step is always considered
 // incomplete (the step itself is hidden in OSS via IS_ENTERPRISE).
+// Element type mirrors the fields OSS consumers read off the enterprise
+// response (e.g. onboarding's `provider.enabled` gate); `unknown` elements
+// force every consumer into a per-site cast.
```
akshaydeo: This comment is worth keeping, and it is the only part of the PR that survives a rebase: `dev` already carries `data: { enabled: boolean }[] | undefined` from 897f06c69 (#6384), which merged after this PR's base c1b84fdc5.

Small accuracy point while you are here. The comment says `unknown` elements "force every consumer into a per-site cast", but the chosen element type only retires the `as { enabled?: boolean }` casts. Two call sites read `name` off the result and still need a cast:

- `ui/app/workspace/mcp-registry/views/mcpClientFormFields.tsx:377`
- `ui/app/workspace/mcp-registry/views/mcpClientSheet.tsx:198`

Both do `... as { name?: string } | undefined`. Covering that field too would make the comment fully true:

```ts
	data: { enabled: boolean; name?: string }[] | undefined;
```

Optional, and it typechecks either way.

akshaydeo (dismissed):
Thanks for the clean, well-scoped fix and for the unusually good PR description - the repro, the exact TS error code, and the toolchain note made this fast to verify.

## Overview

This types the OSS SCIM fallback stub's array elements as `{ enabled: boolean }[]` instead of `unknown[]`, plus a three-line comment explaining why. The typing is correct: I ran `npx tsc --noEmit` in `ui/` against a tree carrying this exact signature and it exits 0, with all five consumers (`useOnboardingChecklist.ts`, `mcpView.tsx`, `mcpLibraryAddServerSheet.tsx`, `mcpClientFormFields.tsx`, `mcpClientSheet.tsx`) still compiling, including the two `as { name?: string }` assertions that a narrower element type could plausibly have broken. Nothing here is a correctness risk.

One thing you should know before merging, and it is not your fault: the type change itself has already landed on `dev`.

## Findings

| 1 | Info | [ui/app/_fallbacks/enterprise/lib/store/apis/scimApi.ts:25](https://github.com/maximhq/bifrost/blob/dev/ui/app/_fallbacks/enterprise/lib/store/apis/scimApi.ts#L25) | The `unknown[]` to `{ enabled: boolean }[]` change already landed on `dev` in 897f06c69 (#6384), which merged after this PR's base c1b84fdc5. `git show origin/dev:ui/app/_fallbacks/enterprise/lib/store/apis/scimApi.ts` already reads `data: { enabled: boolean }[] | undefined`. After a merge, the net remaining change is the three-line comment. | CONFIRMED |
| 2 | Nit | [ui/app/_fallbacks/enterprise/lib/store/apis/scimApi.ts:25](https://github.com/maximhq/bifrost/blob/dev/ui/app/_fallbacks/enterprise/lib/store/apis/scimApi.ts#L25) | The element type covers `enabled` but not `name`, so two call sites keep a cast the comment says the fix removes: `mcpClientFormFields.tsx:377` and `mcpClientSheet.tsx:198` both do `as { name?: string } \| undefined`. Widening to `{ enabled: boolean; name?: string }` would let those two casts go as well. | CONFIRMED |

## Merge recommendation

**Merge.** Nothing blocks: finding 1 is informational, finding 2 is a nit. The comment you added is genuinely useful documentation for the stub and is worth keeping on `dev`, so approving rather than asking you to close. If the maintainers prefer a zero-code-change PR not to carry a `fix(ui)` title, retitling to `docs(ui)` before merge would describe the post-rebase content more accurately. Please rebase onto current `dev` first so the diff shows only the comment.

## Followups

1. **In this PR (non-blocking)** - `ui/app/_fallbacks/enterprise/lib/store/apis/scimApi.ts:25`: consider `data: { enabled: boolean; name?: string }[] | undefined` so `mcpClientFormFields.tsx:377` and `mcpClientSheet.tsx:198` can drop their `as { name?: string }` assertions too, which is what the new comment promises.
2. **Follow-up PR** - the three surviving `as { enabled?: boolean }` casts (`mcpView.tsx:32`, `mcpLibraryAddServerSheet.tsx:85`, `mcpClientFormFields.tsx:377`, `mcpClientSheet.tsx:198`) are now redundant given the typed stub, but removing them touches files outside this diff and reads better as its own small cleanup.

## Candidates checked and refuted

- *Required `enabled: boolean` (not optional) breaks the existing `as { enabled?: boolean }` casts.* Refuted: `tsc --noEmit` exits 0 with this exact signature in place.
- *Asserting `{ enabled: boolean }` to `{ name?: string }` trips TS2352 insufficient-overlap at `mcpClientFormFields.tsx:377` / `mcpClientSheet.tsx:198`.* Refuted: same clean typecheck run covers both lines.
- *Narrowing from `unknown[]` breaks a consumer that treated elements generically.* Refuted: all five consumers only read `.enabled`, `.name`, or `.length`, and all compile.
- *The stub returns `[]` but the type keeps `| undefined`, making the `scimProviders !== undefined` gate at `useOnboardingChecklist.ts:59` dead code in OSS.* Refuted as a defect: that clause is guarded by `!IS_ENTERPRISE ||` and short-circuits in OSS, and keeping `| undefined` is what preserves signature compatibility with the real enterprise RTK Query hook.
- *AGENTS.md convention violation.* Refuted: no rule in AGENTS.md governs stub element typing or `unknown` usage in the UI fallbacks.

ui/app/_fallbacks/enterprise/lib/store/apis/scimApi.ts
is911: Done in 5da5dc405 (rebased onto dev first). Since #6384 already carries `{ enabled: boolean }[]`, the PR now reduces to your suggested widening: `data: { enabled: boolean; name?: string }[] | undefined`, and both `as { enabled?: boolean }` / `as { name?: string }` casts are gone (`mcpClientFormFields.tsx`, `mcpClientSheet.tsx` — both just `find((p) => p.enabled)` now, with `?.name` downstream typechecking against the element type). The comment is reworded to the accurate claim: it names the two fields consumers read rather than "every consumer". `tsc --noEmit` and the production build pass.

is911:
@akshaydeo re-requesting your review when you get a chance — the accuracy point from your thread is addressed in 5da5dc405 (rebased onto dev after #6384, per the reply in-thread), so the PR now reduces to the widened element type with both per-site casts retired.

is911:
Closing as superseded — dev already carries the widened element type (`data: { enabled: boolean }[] | undefined`) via 897f06c69 in #6384, which merged after this PR's base.

Per @akshaydeo's re-review, the only part of this diff that survives a rebase is the explanatory comment on the fallback stub — that can ride along with the next change touching `scimApi.ts` rather than keeping this PR open.

Thanks for the thorough verification!

Issue #6394, opened 2026-08-21
[Bug]: Responses SSE emits "item": null on events without item, breaking strict OpenAI Responses clients


ReStranger:
### Prerequisites

- [x] I have searched existing issues and discussions to avoid duplicates
- [x] I am using the latest version (or have tested against main/nightly)

### Description

When Bifrost streams OpenAI Responses API events over SSE, some event types that do not carry an item object are serialized with "item": null.
This breaks strict clients that validate the Responses streaming schema, including @opencode-ai/ai/providers/openai, which expects item to be optional but not null.
Observed failing event types included:
- response.created
- response.in_progress
- response.output_text.delta
- response.content_part.added
- response.completed
The item field should only be present on events such as:
- response.output_item.added
- response.output_item.done

### Steps to reproduce

1. Start Bifrost on latest/main with an OpenAI-compatible provider configured.
2. Send a streaming request to POST /v1/responses:
{
  "model": "openai/gpt-4o-mini",
  "input": "Say hello in one short sentence.",
  "stream": true
}
3. Inspect the SSE frames returned by Bifrost.
4. Observe frames like:
{"type":"response.created","sequence_number":0,"item":null,...}
5. Use the same endpoint through a strict Responses client such as @opencode-ai/ai/providers/openai.
6. The client fails to decode the stream with an invalid stream event error.

### Expected behavior

Events that do not have an item payload should omit the item field entirely. Example: {"type":"response.created","sequence_number":0,...}

### Actual behavior

Bifrost serializes: {"type":"response.created","sequence_number":0,"item":null,...} Strict Responses clients reject the stream.

### Affected area(s)

Core (Go)

### Version

1.2

### Environment

```text
- OS: NixOS
- Deployment: self-hosted Bifrost
- Client affected: @opencode-ai/ai/providers/openai
- Protocol: OpenAI Responses API over SSE
```

### Relevant logs/output

```shell
Client-side error:
Invalid bifrost/openai-responses stream event
Example problematic SSE frame:
{"type":"response.output_text.delta","sequence_number":3,"item":null,"delta":"Hello"}
```

### Regression?

Unknown

### Severity

Medium (some functionality impaired)

basil-k-aji-dev:
This looks fixed on `main`, so it may be closeable. Evidence below in case it is useful.

**What changed.** Commit `6fbefaf` (2026-09-15, part of #7173) addressed both halves, and it names this exact client in the code:

```go
[...]

Issue #5959, opened 2026-08-07
[Bug]: /v1/responses streaming bridge to chat-completions backends breaks Vercel AI SDK v5 clients ("text part msg_... not found")


mattezell:
### Summary

When a Vercel AI SDK v5 client streams `POST /v1/responses` through Bifrost to a backend that only speaks chat completions (provider type `vllm`, upstream llama.cpp/llama-swap), the client fails deterministically with:

```
Error: text part msg_1f7wVO5UgNVr0AQ6GcbkRW9zgsu6mXWE not found
```

This is the AI SDK's stream-part registration failure: it registers stream parts keyed by `item_id` when a part is announced, then looks that key up on every delta. The error indicates a `response.output_text.delta` arrived for a message item id that the SDK had not seen announced for a text part.

The same client, request shape, and gateway work correctly when the alias resolves to an Azure backend (native Responses API, passthrough). The failure appears only on the bridged path where Bifrost synthesizes Responses-API SSE from a chat-completions upstream.

### Environment

- Bifrost: `maximhq/bifrost:latest` pulled 2026-08-06, digest `sha256:14f704fcee64de509c139d14458871a4221f71f28a397f2b4d92bbe092d7a63c`, single container, SQLite config store
- Backend: llama-swap (llama.cpp server) exposing an OpenAI-compatible API; model `qwen3.5-35b-a3b` (GGUF); configured as provider `vllm` with `vllm_key_config: {url, model_name}`
- Client: VoltAgent `@voltagent/core` 2.7.7, which drives the Vercel AI SDK v5 `@ai-sdk/openai` provider (Responses API model) with `OPENAI_BASE_URL` pointed at Bifrost
- Auth: virtual key, `enforce_auth_on_inference: true`

### Reproduction

1. Configure a `vllm` provider key against any chat-completions-only OpenAI-compatible server.
2. Drive `POST /v1/responses` with `stream: true` from an AI SDK v5 `openai` provider client whose request carries multiple function tools and a large system prompt. Ours is a VoltAgent agent with 9 MCP-derived function tools; the failure reproduces on 100% of runs (5/5 across two days, fresh conversation each time).
3. The stream completes from Bifrost's perspective (logs show `responses_stream ... success`), but the client aborts parsing with `text part msg_... not found` and surfaces the error as the assistant text.

[...]

Issue #5667, opened 2026-07-29
[Bug]: Bedrock streaming /v1/responses: response.completed carries no output array, reasoning done events are empty shells, thinking signatures never reach terminal events


fus3r:
### Prerequisites
- [x] I have searched existing issues and discussions to avoid duplicates
- [x] I am using the latest version (or have tested against main/nightly)

### Description
Streaming `/v1/responses` on the Bedrock provider never completes its terminal events. Same class as the Anthropic (#5149), Gemini (#5259), and Cohere (#5385) reports; Bedrock is the last one. All of it lives in `core/providers/bedrock/responses.go`:
1. `response.completed` has no output array. `FinalizeBedrockStream` sets id/usage/model/stop_reason but never `Output`, and the stream state keeps no item bookkeeping, so the wire says `"output":null`. The non-streaming path fills `output`, and the Anthropic ingress keeps an `OutputItems` map for exactly this. Clients that read the final snapshot get nothing, and openai-python's `stream.get_final_response()` validates that payload and can fail on it. PR #3188 flagged this symptom in May with a text-only reconstruction that never landed; #3838 later brought its text-delta buffering half, the output array never came.
2. Reasoning done events are empty shells. There is no contentBlockStop case, so reasoning blocks close lazily (next tool use, next text block, or end of stream), and all three close sites emit a hardcoded empty `reasoning_summary_text.done`, an empty `content_part.done`, and an `output_item.done` with an empty summary, no content, no signature. Thinking deltas are never buffered; `TextBuffers` only got text deltas in #3838. The streamed item is not replayable either: `convertBifrostReasoningToBedrockReasoning` reads content blocks or a filled summary, the streamed item has neither, so a replayed conversation rebuilds zero `reasoningContent` blocks.
3. The thinking signature never reaches a terminal event. Converse sends it as its own delta after the thinking text; the converter forwards it in flight and never stores it. Anthropic models on Bedrock need the signed block back verbatim on tool-use turns, so anything built from terminal events silently loses the thinking chain.
4. Citation annotations are dropped from the done events and the snapshot: the text item's `content_part.done` and `output_item.done` hardcode empty annotation arrays. nova_grounding keeps its sources on its own done event, but that item is missing from the snapshot like everything else.
Structured-output streams are the worst case: the interception emits bare `output_text.delta` events with no item, no indices, no item id, so those streams produce no `output_item` events at all.
[...]

PR #6395, opened 2026-08-21
fixes responses sse item null regression

ReStranger:
## Summary

Fixes a Responses SSE compatibility bug where Bifrost serialized `"item": null` on stream event types that do not carry an item payload.

Strict OpenAI Responses clients such as `@opencode-ai/ai/providers/openai` reject these events as invalid stream frames, which breaks streamed `/v1/responses` usage even though the event should simply omit the `item` field.

## Changes

- Changed `core/schemas/responses.go` so `BifrostResponsesStreamResponse.Item` uses `json:"item,omitempty"`
- Added a Go regression test covering both sides of the contract:
  - events without an item must omit the field entirely
  - `response.output_item.added` and `response.output_item.done` must still emit the item object
- Added a provider-harness regression case for streamed `POST /v1/responses` that validates:
  - the response is SSE
  - lifecycle events are present
  - no event outside `output_item.added` / `output_item.done` contains `item: null`

Notable design decision:
- This is intentionally a minimal wire-shape fix in the shared schema layer rather than per-provider normalization. The bug was caused by JSON serialization of a nil pointer, so the smallest correct fix is to omit the field at the schema definition.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [x] Core (Go)
- [ ] Transports (HTTP)
- [ ] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

Validate the schema-level regression test:

```sh
cd core
go test ./schemas/... -count=1
```

Expected outcome:
- `github.com/maximhq/bifrost/core/schemas` passes

Validate downstream streaming helpers still pass:

```sh
cd framework
go test ./streaming/... -count=1
```

Expected outcome:
- `github.com/maximhq/bifrost/framework/streaming` passes

Validate the provider-harness collection remains structurally valid and includes the new regression case:

```sh
node tests/e2e/api/runners/augment-provider-harness.mjs \
  --source tests/e2e/api/collections/provider-harness.json \
  --out /tmp/provider-harness-augmented.json

node tests/e2e/api/runners/filter-collection.mjs \
  --source /tmp/provider-harness-augmented.json \
  --out /tmp/provider-harness-item-null.json \
  --feature "item:null"
```

Expected outcome:
- augment succeeds without collection errors
- filter selects the new regression case

Optional end-to-end validation against a running deployment:

```sh
curl -N -X POST "$BIFROST_BASE_URL/v1/responses" \
  -d '{
    "model": "openai/gpt-4o-mini",
    "input": "Say hello in one short sentence.",
    "stream": true
  }'
```

Expected outcome:
- SSE frames for events like `response.created`, `response.in_progress`, `response.output_text.delta`, and `response.completed` do not contain `"item": null`
- `response.output_item.added` / `response.output_item.done` still include `item`

## Screenshots/Recordings

N/A

## Breaking changes

- [ ] Yes
- [x] No

## Related issues

Related:
- #6394
- #5959
- #5259
- #5667

## Security considerations

No security impact. This change only adjusts Responses SSE serialization for schema correctness.

## Checklist


core/schemas/responses.go
```diff
@@ -3719,8 +3719,11 @@ type BifrostResponsesStreamResponse struct {
 
 	Response *BifrostResponsesResponse `json:"response,omitempty"`
 
-	OutputIndex *int              `json:"output_index,omitempty"`
-	Item        *ResponsesMessage `json:"item"`
+	OutputIndex *int `json:"output_index,omitempty"`
+	// Item is only emitted on output_item.added / output_item.done. omitempty is
+	// required: other event types must not serialize "item": null — strict
+	// Responses clients (opencode open-responses protocol) reject null there.
+	Item *ResponsesMessage `json:"item,omitempty"`
```
akshaydeo, line 3731: The fix is correct here, but the sibling field five lines down has the identical defect and is left behind:

```go
LogProbs []ResponsesOutputMessageContentTextLogProb `json:"logprobs"`
```

Marshalling `WithDefaults()` output on `dev` shows both:

```
response.created  {"type":"response.created","sequence_number":0,"item":null,"logprobs":null,...}
```

OpenAI's `ResponseCreatedEvent` carries only `response`, `sequence_number`, `type`, so `logprobs: null` there is off-spec for exactly the reason `item: null` is: https://github.com/openai/openai-python/blob/main/src/openai/types/responses/response_created_event.py

Important trap: do NOT just add `omitempty`. `WithDefaults` deliberately backfills `[]` for `output_text.delta` / `output_text.done`, and `omitempty` drops empty slices too, which would delete a field OpenAI marks required on that event (https://github.com/openai/openai-python/blob/main/src/openai/types/responses/response_text_delta_event.py). Use a pointer slice instead:

```go
LogProbs *[]ResponsesOutputMessageContentTextLogProb `json:"logprobs,omitempty"`
```

or emit the key only for the event types that define it.

core/schemas/responses_test.go
```diff
@@ -6,6 +6,47 @@ import (
...
+	} {
+		withItem := &BifrostResponsesStreamResponse{
+			Type: typ,
+			Item: &ResponsesMessage{Type: Ptr(ResponsesMessageTypeMessage), ID: Ptr("msg_1")},
+		}
+		encoded, err := MarshalSorted(withItem)
+		if err != nil {
+			t.Fatalf("%s: marshal: %v", typ, err)
+		}
+		if !strings.Contains(string(encoded), `"item":{"id":"msg_1"`) {
```
akshaydeo: Nit: this passes today only because `ID` is the first field declared on `ResponsesMessage` (`core/schemas/responses.go:1462`). `MarshalSorted` is `sonic.ConfigStd`, which sorts map keys but not struct fields, so an unrelated field reordering breaks this test with an error message that blames item loss instead of the real cause.

Order-independent and about the same length:

```go
var decoded map[string]any
if err := Unmarshal(encoded, &decoded); err != nil {
	t.Fatalf("%s: unmarshal: %v", typ, err)
}
item, ok := decoded["item"].(map[string]any)
if !ok || item["id"] != "msg_1" {
	t.Errorf("%s: lost item object:\n%s", typ, encoded)
}
```

tests/e2e/api/collections/provider-harness.json
```diff
@@ -365,6 +365,84 @@
...
+                  "      outputItemEvents++;",
+                  "      if (j.item && typeof j.item === 'object' && !Array.isArray(j.item)) { outputItemObjects++; }",
+                  "      return;",
+                  "    }",
+                  "    if ('item' in j) { nonOutputItemFields++; }",
+                  "  } catch (e) {}",
+                  "});",
+                  "pm.test('Responses stream emits response.created and response.completed', function () {",
+                  "  pm.expect(types['response.created'], 'expected response.created').to.be.defined;",
+                  "  pm.expect(types['response.completed'], 'expected response.completed').to.be.defined;",
```
akshaydeo: This assertion throws rather than asserting. Chai has no `defined` property, and Chai 4+ proxy protection turns an unknown property access into an error. Reproduced with the Chai vendored in this repo:

```
> expect(1).to.be.defined
THREW: Invalid Chai property: defined. Did you mean "undefined"?
```

Failure scenario: the harness runs against a fully correct stream that does contain `response.created` and `response.completed`, and the test `Responses stream emits response.created and response.completed` still fails with `Invalid Chai property: defined`. The regression case reports red for a reason unrelated to the regression it guards. Nothing else in this collection uses `.to.be.defined`.

```suggestion
                  "  pm.expect(types['response.created'], 'expected response.created').to.not.be.undefined;",
                  "  pm.expect(types['response.completed'], 'expected response.completed').to.not.be.undefined;",
```

tests/e2e/api/collections/provider-harness.json
```diff
@@ -365,6 +365,84 @@
               ]
             }
           }
+        },
+        {
+          "name": "POST /v1/responses — streaming (no item:null outside output_item.added/done)",
```
akshaydeo: Placement: this pins the regression to `openai/gpt-4o-mini` under `1. Native Bifrost API`, but the collection already has `8. Criss-Cross ... / 8.2 Text Chat (streaming) / 8.2.B Native /v1/responses streaming x all providers`, with one streaming `/v1/responses` case per provider (openai, anthropic, gemini, vertex, bedrock, bedrock_mantle, azure, deepseek), plus a folder-level `test` script on `8.2` that already runs for all of them.

Since `Item` lives on the shared `BifrostResponsesStreamResponse`, this regression affects every provider that streams Responses. Moving the item-null scan into the `8.2` folder-level script gives you the whole matrix for free, catches a per-provider converter that reintroduces the field, and avoids adding another billed OpenAI call to every harness run.

It also removes duplicated work: the `text/event-stream` content-type assertion here is already covered by the `8.2` folder script and by `streamingTest` in `tests/e2e/api/runners/augment-provider-harness.mjs:43-51`.

The `if (pm.response.code >= 400) { return; }` guard you used matches the existing `8.2` convention, so that part is good as-is.

akshaydeo (changes requested):
Thanks for this one, @ReStranger. The diagnosis is correct, the root cause is identified precisely, and the fix is at the right layer. I reproduced the defect against `dev` by marshalling `BifrostResponsesStreamResponse.WithDefaults()` directly:

```
response.created            {"type":"response.created","sequence_number":0,"item":null,"logprobs":null,"extra_fields":{...}}
response.output_text.delta  {"type":"response.output_text.delta","sequence_number":0,"item":null,"logprobs":[],"extra_fields":{...}}
response.completed          {"type":"response.completed","sequence_number":0,"item":null,"logprobs":null,"extra_fields":{...}}
response.output_item.added  {"type":"response.output_item.added","sequence_number":0,"item":{"id":"msg_1",...},"logprobs":null,...}
```

So `"item": null` really is on the wire for every event, and `omitempty` is the correct minimal fix. I traced the path to confirm the fix lands: `transports/bifrost-http/integrations/openai.go:528-539` calls `resp.WithDefaults()` and `transports/bifrost-http/integrations/router.go:3067` does `sonic.Marshal` on that struct, so the schema tag is what the client sees. I also confirmed the change is safe in-process: every Go reader of `Item` nil-checks it (`framework/streaming/responses.go:61`, `transports/bifrost-http/integrations/cursor.go:118`, the provider files), and absent and null both decode to `nil`.

Four things need attention before this can merge: the PR targets the wrong base branch, the new harness assertion fails deterministically, the sibling `logprobs` field has the identical defect and is left behind, and the harness case duplicates coverage that already exists for every provider.

## Findings

| 1 | Blocking | PR base branch | Targets `main`; contributor PRs must target `dev` | CONFIRMED |
| 2 | High | [provider-harness.json:409](https://github.com/maximhq/bifrost/pull/6395/files#diff-0R409) | `.to.be.defined` is not a Chai property; the assertion throws, so the case always fails | CONFIRMED |
| 3 | Medium | [core/schemas/responses.go:3733](https://github.com/maximhq/bifrost/blob/main/core/schemas/responses.go#L3733) | `logprobs` has the identical nil-serializes-null defect and is not fixed | CONFIRMED |
| 4 | Medium | [provider-harness.json:370](https://github.com/maximhq/bifrost/pull/6395/files#diff-0R370) | New openai-only case duplicates the existing all-provider streaming folder | CONFIRMED |
| 5 | Low | [core/schemas/responses_test.go:44](https://github.com/maximhq/bifrost/pull/6395/files#diff-1R44) | Substring assertion depends on Go struct field declaration order | CONFIRMED |

### 1. Blocking: wrong base branch

`gh pr view 6395 --json baseRefName` returns `main`. Every other open contributor PR in the repo (6436, 6435, 6420, 6414, 6407, 6405, 6402, 6397, 6393, 6392, 6378, ...) targets `dev`, and `docs/contributing/raising-a-pr.mdx:10` states it directly:

> Fork the repository and create a feature branch from `dev` (the default development branch; `main` is reserved for releases). Target `dev` when opening your PR.

Merging into `main` would land the change outside the normal release flow and it would be reverted on the next release merge. Please retarget the PR to `dev` (GitHub's "Edit" button next to the title lets you change the base without reopening) and rebase onto `dev` if needed. Everything else in this review is straightforward to fix, so this is the only structural blocker.

### 2. High: `pm.expect(...).to.be.defined` throws in the Postman sandbox

```js
pm.expect(types['response.created'], 'expected response.created').to.be.defined;
```

Chai has no `defined` property. Chai 4 and later ship proxy protection that turns an unknown property access into a thrown error rather than a silent no-op. Reproduced locally with the Chai in this repo's `ui/node_modules`:

```
> expect(1).to.be.defined
THREW: Invalid Chai property: defined. Did you mean "undefined"?
```

Failure scenario: the harness runs against a perfectly correct stream, `response.created` and `response.completed` are both present, and the test `Responses stream emits response.created and response.completed` still fails with `Invalid Chai property: defined`. The regression case then reports red for a reason unrelated to the regression it guards, which is worse than no case at all. Nothing else in `provider-harness.json` uses `.to.be.defined`; the collection's own convention is explicit boolean or numeric assertions.

Suggested fix:

```js
pm.expect(types['response.created'], 'expected response.created').to.not.be.undefined;
pm.expect(types['response.completed'], 'expected response.completed').to.not.be.undefined;
```

or, matching the style used elsewhere in the collection, `pm.expect(types['response.created'] || 0).to.be.above(0)`.

### 3. Medium: `logprobs` has the same defect and is left behind

`core/schemas/responses.go:3733`:

```go
LogProbs []ResponsesOutputMessageContentTextLogProb `json:"logprobs"`
```

Same mechanism, same wire surface: on `response.created`, `response.completed`, `response.output_item.added` and every other event that is not `output_text.delta` / `output_text.done`, `WithDefaults` leaves the slice nil and the frame carries `"logprobs": null`. That is visible in the probe output at the top of this review.

Per OpenAI's own published types, `ResponseCreatedEvent` carries exactly three fields, `response`, `sequence_number`, `type`, with no `logprobs` and no `item`: https://github.com/openai/openai-python/blob/main/src/openai/types/responses/response_created_event.py . So `logprobs: null` on that event is off-spec for the same reason `item: null` is.

Whether it also trips the opencode validator depends on how that client's per-event schema is written, and neither the PR nor #6394 quotes the client-side error text, so I cannot say for certain that fixing `item` alone unblocks the reported user. If you have the exact validator message from `@opencode-ai/ai/providers/openai`, pasting it into the issue would settle that quickly.

One trap worth calling out so nobody "fixes" this the obvious way: `logprobs` must NOT simply get `omitempty`. `WithDefaults` deliberately backfills `[]` for `output_text.delta` / `output_text.done` (`core/schemas/responses.go:3838-3841`), and `omitempty` drops empty slices too, so it would delete a field that OpenAI marks required on that event (`logprobs: List[Logprob]`, required, https://github.com/openai/openai-python/blob/main/src/openai/types/responses/response_text_delta_event.py ). The correct shape is either a `*[]ResponsesOutputMessageContentTextLogProb` with `omitempty`, or emitting the key only for the event types that define it.

Same observation, lower confidence on intent: `extra_fields` is emitted on every SSE frame on the OpenAI-compatible surface and is not part of the Responses spec either. That one may well be deliberate for Bifrost's own superset surface, so I am flagging it as a question rather than a defect.

Happy to see `logprobs` handled in this PR since it is three lines next to the change you already made, but a follow-up PR is fine too if you would rather keep this one minimal. Please do not leave it unfiled either way.

### 4. Medium: the harness case duplicates all-provider coverage that already exists

The new case is added under `1. Native Bifrost API` and pinned to `openai/gpt-4o-mini`. The collection already has:

- Folder `8. Criss-Cross ... / 8.2 Text Chat (streaming)`, which carries a folder-level `test` script that runs for every case beneath it (content-type check plus stream-terminator check).
- Folder `8.2.B Native /v1/responses streaming x all providers`, with one streaming `/v1/responses` case per provider: openai, anthropic, gemini, vertex, bedrock, bedrock_mantle, azure, deepseek.

Because `Item` lives on the shared `BifrostResponsesStreamResponse`, this regression affects every provider that streams Responses, not just OpenAI. Moving the item-null scan into the `8.2` folder-level script gives you the whole provider matrix for free, catches a per-provider converter that reintroduces the field, and avoids adding one more billed OpenAI call to every harness run. It also removes the duplicated content-type assertion, which is already covered twice: by the `8.2` folder script and by `streamingTest` in `tests/e2e/api/runners/augment-provider-harness.mjs:43-51`.

The `if (pm.response.code >= 400) { return; }` guard you used is correct and matches the existing `8.2` script, so that part is good.

### 5. Low: unit-test assertion depends on struct field order

```go
if !strings.Contains(string(encoded), `"item":{"id":"msg_1"`) {
```

`MarshalSorted` is `sonic.ConfigStd`, which sorts map keys but not struct fields, so this passes only because `ID` happens to be the first field declared on `ResponsesMessage` (`core/schemas/responses.go:1462`). I verified it passes today. It will break on an unrelated field reordering, with an error message that points at item loss rather than at the real cause. Decoding into a `map[string]any` and asserting `m["item"].(map[string]any)["id"] == "msg_1"` would be equally short and order-independent.

## Merge recommendation

**Not yet - request changes.** Finding 1 blocks mechanically: the change cannot land on `main`. Finding 2 makes the new harness case fail on every run, so the regression guard this PR adds would not do its job. Findings 3, 4 and 5 are quality issues, not correctness regressions in the shipped fix.

## Followups required

1. **In this PR (blocking)** - Retarget the PR base from `main` to `dev` and rebase if needed. Reference: `docs/contributing/raising-a-pr.mdx:10`.
2. **In this PR (blocking)** - `tests/e2e/api/collections/provider-harness.json:409-410`: replace `.to.be.defined` with `.to.not.be.undefined` (or `.to.be.above(0)`), so the case can pass.
3. **In this PR (blocking)** - `tests/e2e/api/collections/provider-harness.json:370`: move the item-null scan into the `8.2 Text Chat (streaming)` folder-level test script and drop the standalone openai-only case, so all providers are covered and the content-type assertion is not triplicated.
4. **In this PR, or a follow-up PR if you prefer** - `core/schemas/responses.go:3733`: stop emitting `"logprobs": null` on events that do not define it, using a pointer slice or event-scoped emission rather than plain `omitempty` (which would drop the required `[]` on `output_text.delta` / `output_text.done`).
5. **Follow-up PR** - Decide whether `extra_fields` belongs on the OpenAI-compatible `/v1/responses` SSE surface at all, or only on Bifrost's native superset surface.
6. **Nit, this PR** - `core/schemas/responses_test.go:44`: assert on the decoded JSON rather than a byte-order-dependent substring.

## Checked and cleared (refuted candidates)

- *`omitempty` breaks in-process consumers of `Item`.* Refuted: every reader nil-checks (`framework/streaming/responses.go:61`, `transports/bifrost-http/integrations/cursor.go:118`, `core/providers/gemini/responses.go:919`, and the rest), and on decode an absent key and an explicit null both produce `nil`.
- *A golden fixture or snapshot asserts `"item": null`.* Refuted: a repo-wide search for `"item":null` and `"item": null` across Go, JSON, TS and TSX returns no matches.
- *The new test needs a `strings` import that is not added.* Refuted: `core/schemas/responses_test.go:5` already imports `strings`.
- *`return` at the top level of a Postman `exec` script is invalid.* Refuted: the existing `8.2` folder-level script uses the same `if (pm.response.code >= 400) { return; }` guard.
- *The OpenAI raw-passthrough branch means the fix never runs.* Refuted: `transports/bifrost-http/integrations/openai.go:529-533` short-circuits only when `ExtraFields.RawResponse` is populated; the default path falls through to `WithDefaults()` and marshals the Bifrost struct.
- *The test's `"item"` substring check could false-match `"item_id"`.* Refuted: the needle includes the closing quote, so `"item_id"` does not match.
- *The `--feature "item:null"` selector in the PR description will not match the new case.* Refuted: `filter-collection.mjs:45` lowercases and substring-matches the item name path, and the case name contains `item:null`.

Thanks again for the clean diagnosis, the reproduction steps, and for adding both a unit test and a harness case. Once the base branch is corrected and the Chai assertion is fixed, this is a good change.

core/schemas/responses.go
```diff
@@ -3769,6 +3772,21 @@ type BifrostResponsesStreamResponse struct {
 	Citations     []string       `json:"citations,omitempty"`
 }
 
+// MarshalJSON omits event-scoped fields that are nil so strict Responses
+// clients do not see explicit nulls on unrelated event types. Some event types
+// still intentionally emit empty arrays after WithDefaults populates them.
+func (resp BifrostResponsesStreamResponse) MarshalJSON() ([]byte, error) {
+	type alias BifrostResponsesStreamResponse
+	encoded, err := json.Marshal(alias(resp))
```
akshaydeo: use sonic

akshaydeo (changes requested):

core/schemas/responses.go
ReStranger: done

Pratham-Mishra04 (approved):

akshaydeo (approved):

PR #6402, opened 2026-08-21
[fix]: distinguish "cache not engaged" from "no data" on the local cache gauge

### LINKED ISSUE #6400: [Bug]: Dashboard "Local cache hit rate" gauge shows "No data available" when the cache is enabled but never engaged
# [Bug]: Dashboard "Local cache hit rate" gauge shows "No data available" when the cache is configured but never engaged

## Summary

When the semantic cache plugin is **enabled but never engaged** — i.e. requests carry no
`x-bf-cache-key` and the plugin has no `default_cache_key`, so `PreLLMHook` no-ops — the dashboard
gauge renders the placeholder **"No data available"** instead of a real `0%`.

That makes three very different states indistinguishable in the UI:

1. No requests in the window at all.
2. The cache was never reached (plugin enabled, but no request cleared the gating).
3. The logstore genuinely has no cache telemetry.

Only (1) is "no data". State (2) is a healthy, fully-populated window — in our case 5,170 requests —
and the operator is given no signal that the cache simply isn't being invoked. We spent real time
diagnosing this as a broken panel before tracing it to source.

## Root cause

`SearchStats.DirectCacheHits` / `SemanticCacheHits` are `*int64` with `json:",omitempty"`
(`framework/logstore/tables.go:143`). Both stats paths deliberately leave them **nil** when no row in
the window carries valid `cache_debug` JSON:

- matview path, `framework/logstore/matviews.go:1358`:
  ```go
  if agg.CacheDebugCount > 0 {          // else both stay nil
      stats.DirectCacheHits   = &direct
      stats.SemanticCacheHits = &semantic
  }
  ```
- raw path, `framework/logstore/rdb.go:1444` — the `cacheDebugJSONGuard` predicate matches zero rows,
  so `SUM()` returns SQL NULL and the function returns `nil, nil, nil`.

Nil + `omitempty` means the keys are **absent from the response body**. The UI then does
(`ui/app/workspace/dashboard/components/charts/localCacheTokenMeterChart.tsx:17`):

```ts
const hasCacheCounters = data?.direct_cache_hits != null && data?.semantic_cache_hits != null;
...
const hasData = !!data && hasCacheCounters && totalRequests > 0;
```

In JS `undefined != null` is `false`, so absent keys collapse to `hasData === false` and the
placeholder renders — even though `totalRequests` is 5,170.

The nil contract itself looks intentional and consistent across both paths; the defect is that the UI
maps "counters absent" onto the same visual as "no data".

## Reproduction

Enable `semantic_cache` with no `default_cache_key` (and send no `x-bf-cache-key`), then:

```
$ curl -su USER:PW 'http://127.0.0.1:8080/api/logs/stats'
{'total_requests': 117131, 'cache_hit_rate_total_requests': 117131,
 'direct_cache_hits': 2, 'semantic_cache_hits': 0}          # window reaches old probe rows

$ curl -su USER:PW 'http://127.0.0.1:8080/api/logs/stats?start_time=...&end_time=...'   # last 24h
keys present: ['cache_hit_rate_total_requests']
{'total_requests': 5170, 'cache_hit_rate_total_requests': 5170}   # counters GONE
```

The dashboard sends a time filter, so it always takes the second form. The gauge reads
"No data available" against 5,170 requests.

## Suggested fix

Either is fine; the second is smaller and preserves the existing API contract.

1. **Server:** always assign the counters when the window has terminal rows, so
   "engaged zero times" serialises as an explicit `0` rather than an absent key.
2. **UI (preferred):** distinguish the two cases. When the counters are absent but
   `totalRequests > 0`, render something like *"Cache not engaged — 0 of 5,170 requests"* with a
   pointer to `default_cache_key` / `x-bf-cache-key`, rather than "No data available".

Option 2 is also more actionable: the underlying reason is almost always a missing cache key, which
is a config fix the operator can make immediately once they know it is the cause.

## Environment

- Bifrost v1.6.11, Postgres logstore, `semantic_cache` enabled, vector store Valkey 9.0.3 (`type: redis`).
- Reproduced against both the all-time and filtered `/api/logs/stats` paths.

## Note

Not the same as #3049 / #2906 (CacheDebug lost on streaming). Here `cache_debug` is correctly absent —
the cache genuinely never ran; the issue is purely how the UI reports that.

wangkanai:
## Summary

Fixes #6400.

The dashboard's **Local Cache Hit Rate** gauge renders the `No data available` placeholder for any window where the stats response omits `direct_cache_hits` / `semantic_cache_hits`. Those keys are absent whenever no log row in the window carried `cache_debug` — i.e. the semantic cache is configured but was never invoked. The result is that a window full of traffic is indistinguishable from an empty one, and reads as a broken panel.

On our gateway this meant a 24-hour window with **5,334 requests** displaying `No data available`. We spent real time diagnosing it as a UI bug before tracing it to source.

The stats API already carries the distinction:

| response | meaning |
|---|---|
| counters **absent** | the cache was never engaged |
| counters **present and zero** | the cache ran and missed |

So this PR does **not** touch the server. `matviews.go:1358` and `rdb.go:1444` deliberately return nil (the comment there calls it "aggregateCacheHits' nil contract"), and changing that would turn two `omitempty` fields into always-present ones. Only the UI is at fault — it collapses both cases into one via `data?.direct_cache_hits != null`, and in JS `undefined != null` is `false`.

## Changes

- **`ui/lib/utils/cacheGauge.ts`** (new) — `resolveLocalCacheGauge()` derives one of three states (`no-data` / `not-engaged` / `ready`) plus the clamped percentage. Extracted as a pure function so it is testable: the UI has `vitest` but no jsdom and no `@testing-library`, so components cannot be rendered in a test. This matches the existing pure-logic test files under `lib/`.
- **`ui/lib/utils/cacheGauge.test.ts`** (new) — 13 tests covering the state matrix, the denominator precedence (`cache_hit_rate_total_requests ?? total_requests`), and clamping.
- **`ui/app/workspace/dashboard/components/charts/localCacheTokenMeterChart.tsx`** — renders the new `not-engaged` state. `no-data` and `ready` render exactly as before.

Rendering logic:

```
!data || totalRequests === 0    ->  "No data available"                     (unchanged)
requests > 0, counters absent   ->  gauge + "--"                            (NEW)
                                    "Cache not engaged"
                                    "5.4K requests, no cache activity recorded"
                                    (i) engages only with the plugin enabled
                                        AND x-bf-cache-key / default_cache_key
otherwise                       ->  gauge + "0.0%"                          (unchanged)
```

The `not-engaged` value reads `--` rather than `0.0%` so a measured zero stays visually distinct from "never consulted", and the caption is worded as an observation — the counters are also nil when the cache-hit aggregation query fails, which the payload cannot distinguish.

The new state follows the idiom already used by the sibling `externalCacheTokenMeterChart.tsx` (`Tooltip` / `TooltipTrigger asChild` / `Info` icon button with `data-testid` and `aria-label`, `formatCompactNumber` for the count).

### Design notes

- The tooltip names the actual remedy (`x-bf-cache-key` header, or `default_cache_key` in the plugin config) so the state is actionable rather than merely accurate.
- `clampPercentage(value, max)` replaces the inline `Math.max(0, Math.min(...))` pairs; the segment maths is unchanged and still clamps `semanticPct` to `100 - directPct`, and both segments are gated on `ready` so only measured hits are ever drawn.
- `ChartErrorBoundary`'s `resetKey` now includes the state.

## Review round 2 (maintainer review, `a26cc5a`)

Findings 1, 2, 3, 5 and 6 are addressed. No resolver or test change — every finding was about what the new state *says* and how it is marked up, and the state machine was verified correct in review.

| # | Change |
|---|---|
| 1 | Tooltip leads with plugin enablement, then the cache key. The old copy pointed every plugin-less gateway at an `x-bf-cache-key` header that would change nothing for it. |
| 2 | `not-engaged` renders `--` instead of `0.0%`, so a measured zero stays distinguishable at a glance rather than only by an 11px caption. |
| 3 | Caption softened to "no cache activity recorded", and the resolver doc now records that nil counters also come from a failed aggregation query (`rdb.go:1544-1546`). |
| 5 | The legend's duplicated wrapper `div` + className collapsed to one wrapper with swapped children. |
| 6 | `data-testid="local-cache-meter-summary"` is now unconditional, with `data-state={state}`, so E2E can assert `ready` as well as `not-engaged`. The info-button testid is unchanged. |

Findings 5 and 6 were scoped as a follow-up PR in the review; both touch only lines this PR introduces, so they are folded in rather than deferred. **Finding 4 is not** — `clampPercentage` does not exist on `dev` until this merges, so moving it to `lib/utils/numbers.ts` and adopting it at `externalCacheTokenMeterChart.tsx:29` / `providerGovernanceTable.tsx:99` lands as a follow-up once this is in, exactly as offered.

Re-verified on `a26cc5a`: `vitest` 299 passed (unchanged — no test touched), `tsc --noEmit` clean, `oxlint` and `oxfmt --check` clean on the changed files, `vite build` succeeds. Rendered via `renderToStaticMarkup` against the same live payloads:

```
not-engaged : --    Cache not engaged   5.39K requests, no cache activity recorded
ready       : 0.2%  of requests served from local cache   Direct: 2   Semantic: 0
no-data     : No data available
```

## Review round 3 (CodeRabbit, `6c21d60`)

Both findings reproduced against the round-2 code and are fixed. Still no resolver or test change.

| Severity | Change |
|---|---|
| Major | `directPct` / `semanticPct` are gated on `hasData`, not only on `totalRequests > 0`. `resolveLocalCacheGauge` returns `not-engaged` when **either** counter is nil, so a payload carrying one non-nil counter — the case `cacheGauge.test.ts` covers explicitly — would have drawn it as a coloured hit segment while the summary read `--` and "no cache activity recorded". `not-engaged` now renders the neutral arc alone. The needle was already correct, since `percentage` is 0 outside `ready`. |
| Minor | The tooltip's first sentence no longer asserts that no request reached the cache. That is the same claim round 2's finding 3 removed from the caption, for the same reason: nil counters also cover a failed aggregation query, which the payload cannot distinguish. It now opens "No cache activity was recorded in this window." |

The tooltip's mechanism sentence is kept as the maintainer worded it in round 1 — plugin enablement first, then the cache key. CodeRabbit's suggested replacement ("configure an `x-bf-cache-key` header or `default_cache_key`") is not accurate for the header half: `x-bf-cache-key` is sent per request by the caller, not configured on the gateway, and that distinction is the point of round 1's finding 1.

Re-verified on `6c21d60`: `vitest` 299 passed (unchanged — no test touched), `tsc --noEmit` clean, `oxlint` and `oxfmt --check` clean on the changed file, `vite build` succeeds. No new render probe — a closed Radix tooltip renders no content, and the pie segments do not render at zero container size, so neither change is observable in static markup.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [ ] Core (Go)
- [ ] Transports (HTTP)
- [ ] Providers/Integrations
- [ ] Plugins
- [x] UI (React)
- [ ] Docs

## How to test

```sh
cd ui
npx vitest run            # 19 files, 299 tests
npx tsc --noEmit
npx oxlint  lib/utils/cacheGauge.ts lib/utils/cacheGauge.test.ts app/workspace/dashboard/components/charts/localCacheTokenMeterChart.tsx
npx oxfmt --check lib/utils/cacheGauge.ts lib/utils/cacheGauge.test.ts app/workspace/dashboard/components/charts/localCacheTokenMeterChart.tsx
npx vite build
```

Results on this branch:

- `vitest` — **299 passed** (286 on `dev` + 13 new). `app/workspace/logs/views/columns.test.ts` fails **identically on unmodified `dev`** (a `.ts` test importing a `.tsx` module: *"Failed to parse source for import analysis…"*). Untouched by this PR, flagged in case it is news.
- `tsc --noEmit` — no diagnostics for any file in this PR. The 93 pre-existing errors are all missing `./routeTree.gen` and its knock-on route-literal errors, present on unmodified `dev`.
- `oxlint` — 0 warnings, 0 errors on the changed files.
- `oxfmt --check` — clean on the changed files.
- `vite build` — succeeds.

### Verified against a live gateway

Rather than mock data, the two real payloads from a running instance were fed through the resolver:

```
GET /api/logs/stats?start_time=<24h ago>   ->  { state: "not-engaged", totalRequests: 5334 }
GET /api/logs/stats                        ->  { state: "ready", directHits: 2, percentage: 0.0017 }
same payload with total_requests: 0        ->  { state: "no-data" }
```

The first is the reported bug: 5,334 requests, both counters omitted, previously `No data available`.

## Screenshots/Recordings

Not included — I don't have a browser session against a dashboard build, and I'd rather not paste a mocked-up image as if it were the real thing. The three states are specified exactly in "Changes" above, and the live-payload output above shows which state each real response now resolves to. Happy to add real captures if you'd like them before merging.

## Breaking changes

- [ ] Yes
- [x] No

No API or contract change; a window that previously rendered the gauge still renders the gauge identically.

## Related issues

Closes #6400

## Security considerations

None. No auth, secrets, PII or network behaviour is touched; the change is presentational and reads only fields the dashboard already fetches.

## Checklist

- [x] I verified builds succeed (Go and UI) — UI only; no Go files are touched

---

**Note on the changelog.** This PR no longer touches `transports/changelog.md`. It carried a one-line entry under `🐞 Fixed` until the v2.0.0 release rotation emptied the file on `dev`, which is what made this PR conflict. Rather than re-seed a fresh `## 🐞 Fixed` section into the now-empty file — which would just re-conflict with the next PR that does the same — the entry is dropped and left for the release process. Happy to add it back in whatever form you prefer.


wangkanai:
Follow-up: I said above I hadn't captured the rendered output. I have now, without mocking anything — rendering the component with `renderToStaticMarkup` (no new dependencies; `react-dom` is already present) against the **live payloads from a running gateway**:

```
--- 24h window (5,394 requests, counters omitted) ---
Cache not engaged  5.39K requests, none used the cache

--- all-time (counters present) ---
0.0% of requests served from local cache   Direct: 2   Semantic: 0

--- null data ---
No data available
```

The first line is the bug from #6400: on `dev` that same payload renders `No data available`. The other two are unchanged from `dev`.

---

### Root cause of the `columns.test.ts` failure I flagged

Worth a separate look — it isn't that test's fault, and it isn't caused by this PR.

`ui/vitest.config.ts` declares only `resolve.alias` and `test.globals`. It has **no plugins**, so the React JSX transform never runs under vitest, and any test that imports a `.tsx` module dies at import analysis:

```
Failed to parse source for import analysis because the content contains invalid JS syntax.
If you use tsconfig.json, make sure to not set jsx to preserve.
```

`ui/vite.config.mts` does include `react()`, and running the same test against it passes:

```sh
$ npx vitest run app/workspace/logs/views/columns.test.ts                          # default config
  Test Files  1 failed (1)

$ npx vitest run app/workspace/logs/views/columns.test.ts --config vite.config.mts  # with react()
  Test Files  1 passed (1)
  Tests       5 passed (5)
```

So the effect is that **no component can currently be tested at all**, which is why this PR extracts the state logic into a plain `.ts` module instead of testing the component directly.

I've deliberately left that out of this PR — it's a config change affecting every test, and it belongs on its own. Happy to raise it as a separate issue or PR if useful.

wangkanai:
Pushed `36e53a0` after operator feedback on a live gateway.

**Change:** the gauge is now drawn in the `not-engaged` state too, reading `0.0%` with a `Cache not engaged` caption and the request count, instead of a text-only placeholder.

The first cut suppressed the gauge on the reasoning that a needle asserts a measured value, and in this state nothing was measured. Testing it against a real dashboard showed that argument losing to a simpler one: a text-only card sitting in a row of gauges just reads as a broken panel — the operator's first reaction was "where's my gauge?", which is the same failure mode this PR set out to remove, wearing a different hat. The explicit caption carries the semantics; the arc carries the visual consistency.

`no-data` still renders the original placeholder, and `ready` is untouched.

```
not-engaged : 0.0%  Cache not engaged  5.36K requests, none used the cache
ready       : 0.0%  of requests served from local cache   Direct: 2  Semantic: 0
no-data     : No data available
```

299 tests pass, `tsc` / `oxlint` / `oxfmt` clean on the changed files.

---

### On the `@/lib` alias nitpick — respectfully declining

The suggestion is to import the subject as `@/lib/utils/cacheGauge` rather than `./cacheGauge`. Checking every existing test under `ui/lib`, all eight import their own subject module relatively:

```
lib/constants/logs.test.ts                from "./logs"
lib/contexts/topbarContext.utils.test.ts  from "./topbarContext.utils"
lib/message/variables.test.ts             from "./message"
lib/notifications/preferences.test.ts     from "./preferences"
lib/queryParamsParser.test.ts             from "./queryParamsParser"
lib/utils/redaction.test.ts               from "./redaction"
lib/utils/strings.test.ts                 from "./strings"
lib/utils/validation.test.ts              from "./validation"
lib/utils/versionSkew.test.ts             from "./versionSkew"
```

The alias is used for *cross-directory* imports — `governance.test.ts` pulls `@/lib/budgetOutline`, and this PR's test already uses `@/lib/types/logs` for the `LogStats` type. So the file follows the convention as it exists in the tree; applying the suggestion would make it the only test importing its own neighbour through the alias.

Happy to change it if maintainers prefer the alias everywhere — in that case it's probably worth doing as a sweep rather than on this one file.

ui/app/workspace/dashboard/components/charts/localCacheTokenMeterChart.tsx
```diff
@@ -76,22 +69,50 @@ function LocalCacheTokenMeterChartImpl({ data }: LocalCacheTokenMeterChartProps)
...
+												type="button"
+												data-testid="local-cache-meter-not-engaged-info-btn"
+												className="text-zinc-500 transition-colors hover:text-zinc-300"
+												aria-label="Why the local cache was not engaged"
+											>
+												<Info className="h-3 w-3" />
+											</button>
+										</TooltipTrigger>
+										<TooltipContent side="top">
+											Requests bypass the cache unless they carry an x-bf-cache-key header, or the semantic cache plugin sets a
```
akshaydeo: Finding 1 (Medium): this remedy is wrong for anyone not running the semantic cache plugin.

The counters are nil whenever no in-window row carried valid `cache_debug` JSON (`framework/logstore/rdb.go:1586-1589`), and the plugin only stamps `cache_debug` after `resolveCacheKey` succeeds (`plugins/semanticcache/main.go:351-354`). "Plugin not installed at all" and "plugin installed but no cache key" are therefore indistinguishable here. Since this card is rendered unconditionally (`ui/app/workspace/dashboard/components/overviewTab.tsx:275-277`), every gateway with traffic and no semantic cache plugin now shows "Cache not engaged" and is told to add an `x-bf-cache-key` header, which will change nothing for them.

Leading with plugin enablement covers both populations:

```tsx
<TooltipContent side="top">
	No request in this window reached the local cache. It engages only when the semantic cache plugin is
	enabled and the request carries an x-bf-cache-key header, or the plugin sets a default_cache_key.
</TooltipContent>
```

ui/app/workspace/dashboard/components/charts/localCacheTokenMeterChart.tsx
```diff
@@ -76,22 +69,50 @@ function LocalCacheTokenMeterChartImpl({ data }: LocalCacheTokenMeterChartProps)
 						</>
 					)}
 				</div>
-				{hasData && (
-					<div>
+				{showGauge && (
+					<div data-testid={hasData ? undefined : "local-cache-meter-not-engaged"}>
 						<div className="flex flex-col items-center pt-1 leading-none">
 							<div className="text-muted-foreground text-3xl font-semibold tracking-tight">{percentage.toFixed(1)}%</div>
```
akshaydeo: Finding 2 (Low): `not-engaged` renders the same large `0.0%` as a genuinely measured zero (`ready` with both counters at 0, which is reachable because `stampCacheDebugForMiss` runs on misses). The two states then differ only in the 11px caption below, which is a quieter version of the ambiguity this PR sets out to remove.

A neutral value keeps the separation visible at a glance:

```tsx
<div className="text-muted-foreground text-3xl font-semibold tracking-tight">
	{hasData ? `${percentage.toFixed(1)}%` : "--"}
</div>
```

ui/lib/utils/cacheGauge.ts
```diff
@@ -0,0 +1,66 @@
...
+	const totalRequests = toCount(data?.cache_hit_rate_total_requests ?? data?.total_requests);
+	const base = { percentage: 0, directHits, semanticHits, totalRequests };
+
+	if (!data || totalRequests <= 0) {
+		return { ...base, state: "no-data" };
+	}
+
+	// Either counter being absent means the window carried no cache telemetry at
+	// all; the logstore only populates them as a pair.
+	if (data.direct_cache_hits == null || data.semantic_cache_hits == null) {
```
akshaydeo, line 64: Finding 3 (Low): nil counters have a third cause that the caption does not allow for. `framework/logstore/rdb.go:1544-1546` logs a warning and leaves both fields nil when the cache-hit aggregation query itself fails, so a logstore error resolves to `not-engaged` and the UI states "5.3K requests, none used the cache" as fact.

The resolver cannot tell the two apart, and the server contract does not need to change for this - wording the caption as an observation rather than a conclusion is enough. Suggest "no cache activity recorded" instead of "none used the cache" at `localCacheTokenMeterChart.tsx:113`, and a matching note on this doc comment that nil also covers an aggregation failure.

ui/lib/utils/cacheGauge.ts
```diff
@@ -0,0 +1,66 @@
...
+	state: LocalCacheGaugeState;
+	/** Combined hit rate, clamped to 0-100. Zero unless `state` is `ready`. */
+	percentage: number;
+	directHits: number;
+	semanticHits: number;
+	totalRequests: number;
+}
+
+/** Clamps a raw percentage into the 0-100 range the gauge can render. */
+export function clampPercentage(value: number, max = 100): number {
```
akshaydeo, line 24: Finding 4 (nit, reuse): `clampPercentage` has nothing cache-specific about it. `ui/lib/utils/numbers.ts` is the natural home, beside `formatCompactNumber`, and two existing call sites already inline the same expression:

- `ui/app/workspace/dashboard/components/charts/externalCacheTokenMeterChart.tsx:29`
- `ui/app/workspace/providers/views/providerGovernanceTable.tsx:99`

Happy for this to be a follow-up PR rather than growing this one.

ui/app/workspace/dashboard/components/charts/localCacheTokenMeterChart.tsx
```diff
@@ -76,22 +69,50 @@ function LocalCacheTokenMeterChartImpl({ data }: LocalCacheTokenMeterChartProps)
 						</>
 					)}
 				</div>
-				{hasData && (
-					<div>
+				{showGauge && (
+					<div data-testid={hasData ? undefined : "local-cache-meter-not-engaged"}>
```
akshaydeo: Finding 6 (nit, convention): `data-testid={hasData ? undefined : "local-cache-meter-not-engaged"}` leaves the `ready` state with no hook, so an E2E test can assert the new state but not the old one. AGENTS.md rule 17 says "E2E tests depend on `data-testid` attributes"; an unconditional id plus the state as data makes both assertable:

```tsx
<div data-testid="local-cache-meter-summary" data-state={state}>
```

akshaydeo (commented):
## Review of #6402

Thanks for this - it is a genuinely well-argued PR. The diagnosis is correct and I verified it against the source: `aggregateCacheHits` (`framework/logstore/rdb.go:1586-1589`) returns `nil, nil` when no in-window row carries valid `cache_debug` JSON, the matview path reproduces that contract at `framework/logstore/matviews.go:1421-1425`, and `stampCacheDebugForMiss` (`plugins/semanticcache/main.go:596`) does stamp on a miss, so "counters present and zero" really is reachable and really is a different fact from "counters absent". Extracting the resolver as a pure function was the right call given the UI test harness has no jsdom, and the `ready` / `no-data` paths are behaviourally identical to the old code, so existing dashboards are untouched.

The findings below are all about what the new third state *says*, not about whether it is computed correctly. Nothing here is a correctness regression.

### Findings

|---|---|---|---|---|
| 1 | Medium | [localCacheTokenMeterChart.tsx:93](https://github.com/maximhq/bifrost/blob/fix/local-cache-gauge-not-engaged/ui/app/workspace/dashboard/components/charts/localCacheTokenMeterChart.tsx#L93) | The tooltip's remedy is wrong for the largest group of users: anyone not running the semantic cache plugin at all. Nil counters mean "no row carried `cache_debug`", which covers "plugin never installed" just as much as "plugin installed, no cache key". The card is rendered unconditionally (`overviewTab.tsx:275-277`), so every gateway with traffic and no semantic cache plugin now reads "Cache not engaged" and is told to add an `x-bf-cache-key` header, which will do nothing for them. | CONFIRMED |
| 2 | Low | [localCacheTokenMeterChart.tsx:75](https://github.com/maximhq/bifrost/blob/fix/local-cache-gauge-not-engaged/ui/app/workspace/dashboard/components/charts/localCacheTokenMeterChart.tsx#L75) | `not-engaged` renders the same large `0.0%` as a real measured zero. The two states then differ only in an 11px caption, which is a milder version of the ambiguity this PR is fixing. | CONFIRMED |
| 3 | Low | [cacheGauge.ts:57](https://github.com/maximhq/bifrost/blob/fix/local-cache-gauge-not-engaged/ui/lib/utils/cacheGauge.ts#L57) | Nil counters are also produced by an aggregation *failure*: `rdb.go:1544-1546` logs a warning and leaves the fields nil. A logstore error therefore now renders the confident claim "5.3K requests, none used the cache" instead of a neutral placeholder. | CONFIRMED (mechanism), trigger needs a DB error |
| 4 | Nit | [cacheGauge.ts:24](https://github.com/maximhq/bifrost/blob/fix/local-cache-gauge-not-engaged/ui/lib/utils/cacheGauge.ts#L24) | `clampPercentage` is not cache-specific. `lib/utils/numbers.ts` is the natural home, next to `formatCompactNumber`, and two other call sites already inline the same expression: `externalCacheTokenMeterChart.tsx:29` and `providerGovernanceTable.tsx:99`. | CONFIRMED (reuse) |
| 5 | Nit | [localCacheTokenMeterChart.tsx:100](https://github.com/maximhq/bifrost/blob/fix/local-cache-gauge-not-engaged/ui/app/workspace/dashboard/components/charts/localCacheTokenMeterChart.tsx#L100) | The legend ternary repeats the identical wrapper `div` and its full className in both branches. One wrapper with swapped children is shorter and keeps the two layouts from drifting. | CONFIRMED (simplification) |
| 6 | Nit | [localCacheTokenMeterChart.tsx:73](https://github.com/maximhq/bifrost/blob/fix/local-cache-gauge-not-engaged/ui/app/workspace/dashboard/components/charts/localCacheTokenMeterChart.tsx#L73) | `data-testid={hasData ? undefined : "..."}` leaves the `ready` state with no hook at all. A stable testid plus a `data-state={state}` attribute lets an E2E test assert either state, and matches the spirit of AGENTS.md rule 17 ("E2E tests depend on `data-testid` attributes"). | CONFIRMED (convention) |

### Merge recommendation

**Merge after nits.** No correctness regression survived verification: `ready` and `no-data` behave exactly as before, and the new state is derived correctly. Findings 1 and 2 are copy changes worth making before this lands, because finding 1 gives actively misleading advice to every deployment that does not run the semantic cache plugin.

### Followups

1. **In this PR** - `localCacheTokenMeterChart.tsx:93`: lead the tooltip with plugin enablement, then the cache key. Something like: "No request in this window reached the local cache. It engages only when the semantic cache plugin is enabled and the request carries an `x-bf-cache-key` header (or the plugin sets a `default_cache_key`)."
2. **In this PR** - `localCacheTokenMeterChart.tsx:75`: for `not-engaged`, render a neutral placeholder value instead of `0.0%`, so a measured zero stays visually distinct from "never consulted".
3. **In this PR (optional, one word)** - `localCacheTokenMeterChart.tsx:113`: soften "none used the cache" to "no cache activity recorded", which stays true when the counters are nil because the aggregation query failed (finding 3).
4. **Follow-up PR** - move `clampPercentage` to `ui/lib/utils/numbers.ts` and adopt it at `externalCacheTokenMeterChart.tsx:29` and `providerGovernanceTable.tsx:99`.
5. **Follow-up PR** - collapse the duplicated legend wrapper (finding 5) and give the summary block an unconditional `data-testid` plus `data-state` (finding 6).

On your two maintainer notes: matching the existing `changelog.md` entry shape rather than the doc was the right call, and leaving the "Closed GitHub Issues" section alone is correct - that section is curated at release time.

- **`Tooltip` needs a `TooltipProvider` ancestor.** Refuted: `ui/components/ui/tooltip.tsx:12` wraps every `Tooltip` in its own `TooltipProvider`.
- **The new state changes what existing dashboards render.** Refuted: `state === "ready"` expands to exactly the old `!!data && counters != null && totalRequests > 0`, and `no-data` covers the old `total === 0` branch including a negative denominator.
- **`not-engaged` is unreachable because the plugin stamps `cache_debug` on misses too.** Refuted: `PreLLMHook` returns at `plugins/semanticcache/main.go:351-354` when `resolveCacheKey` fails, before any stamping, so a keyless request leaves no telemetry.
- **The "either counter alone missing" branch is dead, so its test asserts fiction.** Refuted as a defect: `rdb.go:1546-1548` assigns both fields together so the pair is always nil-or-both, but the defensive branch and its test are harmless and cheap.
- **`toCount` mishandles `null`/`NaN`.** Refuted: `Number.isFinite(null)` is `false`, so both fall to `0`, and the `state` decision reads the raw fields rather than the coerced counts.
- **The `cache_hit_rate_total_requests ?? total_requests` denominator changed.** Refuted: identical to the previous expression, including the case where `cache_hit_rate_total_requests` is `0`.

ui/app/workspace/dashboard/components/charts/localCacheTokenMeterChart.tsx
wangkanai: Fixed in `a26cc5a` with your copy, near-verbatim:

> No request in this window reached the local cache. It engages only when the semantic cache plugin is enabled and the request carries an x-bf-cache-key header, or the plugin sets a default_cache_key.

You are right about the population: I wrote the original remedy from a gateway that does run the plugin, so "installed but keyless" was the only case in view. Since the card renders unconditionally from `overviewTab.tsx`, the plugin-less gateway is the common reading, and it was being told to add a header that would do nothing.

ui/app/workspace/dashboard/components/charts/localCacheTokenMeterChart.tsx
wangkanai: Fixed in `a26cc5a` — the value now reads `--` in `not-engaged`, and `0.0%` stays reserved for a measured zero.

```
not-engaged : --    Cache not engaged   5.39K requests, no cache activity recorded
ready       : 0.2%  of requests served from local cache   Direct: 2   Semantic: 0
```

The gauge arc is still drawn in both, per the operator feedback behind `36e53a0` — the arc carries visual consistency with the sibling cards, the value and caption carry the semantics.

ui/lib/utils/cacheGauge.ts
wangkanai: Fixed in `a26cc5a`, both places you named.

Caption is now "no cache activity recorded", and the resolver doc says the second cause explicitly:

> The counters are also left nil when the cache-hit aggregation query itself fails — the logstore logs a warning and returns nil rather than an error — and the payload cannot distinguish that from a window nothing engaged. So `not-engaged` means "no cache activity was reported", and the UI wording stays observational rather than asserting that no request hit the cache.

Agreed the server contract should not change for this. A logstore failure staying indistinguishable from a quiet window is a real limitation, but it belongs in the stats API if anywhere, not in a gauge resolver.

ui/lib/utils/cacheGauge.ts
wangkanai: Taking the follow-up option, for a mechanical reason: `clampPercentage` does not exist on `dev` until this PR merges, so a follow-up branched from `dev` cannot move it, and stacking it on this branch would put the same two files in two open PRs. I will raise it once this lands — move to `lib/utils/numbers.ts`, adopt at `externalCacheTokenMeterChart.tsx:29` and `providerGovernanceTable.tsx:99`.

Note the two existing call sites are not quite the same expression: `providerGovernanceTable.tsx:99` rounds before clamping and has no `!Number.isFinite` guard, so adopting the helper there is a small behaviour change for `NaN` (currently renders `NaN`, would render 0). That is worth its own PR rather than a line in this one.

ui/app/workspace/dashboard/components/charts/localCacheTokenMeterChart.tsx
wangkanai: Fixed in `a26cc5a`:

```tsx
<div data-testid="local-cache-meter-summary" data-state={state}>
```

Both states are now assertable, and `data-state` carries all three values from the resolver rather than a boolean. The `local-cache-meter-not-engaged-info-btn` testid on the info button is unchanged.

This was scoped as a follow-up PR in your list; it touches only a line this PR introduces, so folding it in seemed cleaner than a second PR against the same file. Same for finding 5 — the legend is now one wrapper with swapped children.

wangkanai:
Pushed `a26cc5a`. Findings 1, 2, 3, 5 and 6 are addressed; all five inline threads have a reply and are resolved.

| # | What changed |
|---|---|
| 1 | Tooltip leads with plugin enablement, then the cache key — your wording, near-verbatim. The old copy pointed every plugin-less gateway at a header that would change nothing for it. |
| 2 | `not-engaged` renders `--`; `0.0%` is now reserved for a measured zero. |
| 3 | Caption is "no cache activity recorded", and the resolver doc records that nil counters also come from a failed aggregation query. |
| 5 | Legend collapsed to one wrapper with swapped children. |
| 6 | `data-testid="local-cache-meter-summary"` unconditional, plus `data-state={state}`. |

Nothing in the resolver or its tests moved — every finding was about copy and markup, which matches your read that the state machine itself verified correct.

**Two deliberate divergences from the followups list.**

Findings 5 and 6 were scoped as a follow-up PR. Both touch only lines this PR introduces, so folding them in avoided a second PR against the same file. Say the word if you would rather they came out.

Finding 4 stays a follow-up, for a mechanical reason: `clampPercentage` does not exist on `dev` until this merges, so a follow-up branched from `dev` cannot move it, and stacking would put the same files in two open PRs. I will raise it once this lands. One thing found while checking the call sites — `providerGovernanceTable.tsx:99` rounds before clamping and has no `Number.isFinite` guard, so adopting the shared helper there changes `NaN` from rendering `NaN` to rendering `0`. Small, but it is a behaviour change and deserves its own PR rather than a line in this one.

Re-verified on `a26cc5a`: `vitest` 299 passed (`columns.test.ts` still fails identically on unmodified `dev` — the vitest config has no React plugin, so any test importing a `.tsx` dies at import analysis), `tsc --noEmit` clean, `oxlint` / `oxfmt --check` clean on the changed files, `vite build` succeeds. Rendered via `renderToStaticMarkup` against the same live gateway payloads:

```
not-engaged : --    Cache not engaged   5.39K requests, no cache activity recorded
ready       : 0.2%  of requests served from local cache   Direct: 2   Semantic: 0
no-data     : No data available
```

ui/app/workspace/dashboard/components/charts/localCacheTokenMeterChart.tsx
```diff
@@ -76,21 +69,51 @@ function LocalCacheTokenMeterChartImpl({ data }: LocalCacheTokenMeterChartProps)
...
+												data-testid="local-cache-meter-not-engaged-info-btn"
+												className="text-zinc-500 transition-colors hover:text-zinc-300"
+												aria-label="Why the local cache was not engaged"
+											>
+												<Info className="h-3 w-3" />
+											</button>
+										</TooltipTrigger>
+										<TooltipContent side="top">
+											No request in this window reached the local cache. It engages only when the semantic cache plugin is enabled and the
+											request carries an x-bf-cache-key header, or the plugin sets a default_cache_key.
```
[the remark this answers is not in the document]
wangkanai: Fixed in 6c21d60, though not with the suggested replacement.

The first sentence was wrong for the reason you give, and it is the same reason the maintainer's round-2 finding 3 gave for softening the caption at line 115: nil counters also cover a failed cache-hit aggregation query (`framework/logstore/rdb.go:1544-1546` logs a warning and returns nil), and the payload cannot distinguish that from a window nothing engaged. Round 2 applied that to the caption and left the tooltip asserting the stronger claim. It now opens:

> No cache activity was recorded in this window. The local cache engages only when the semantic cache plugin is enabled and the request carries an x-bf-cache-key header, or the plugin sets a default_cache_key.

I kept the mechanism sentence as the maintainer worded it in round 1 rather than taking your second half. "Configure an x-bf-cache-key header or default_cache_key" is not accurate for the header: `x-bf-cache-key` is sent per request by the caller — `resolveCacheKey` checks the context key before the config default — while `default_cache_key` is gateway configuration. Collapsing the two into "configure" loses exactly the distinction round 1 asked for, which was that a plugin-less gateway cannot be fixed by adding a header.

The major finding from the review body is fixed in the same commit — `directPct` and `semanticPct` are now gated on `hasData`.

wangkanai:
Review round 3 is in **6c21d60**. Both CodeRabbit findings reproduced against the round-2 code and are fixed; the one inline thread is resolved. The major finding was outside the diff, so it has no thread — it is covered here.

**Major — `localCacheTokenMeterChart.tsx:29-30`, measured slices on a state that reports nothing measured.** Confirmed. `resolveLocalCacheGauge` returns `not-engaged` when **either** counter is nil, not only when both are, and `cacheGauge.test.ts:41-43` covers that partial case explicitly:

```ts
expect(resolveLocalCacheGauge(stats({ total_requests: 10, direct_cache_hits: 3 })).state).toBe("not-engaged");
```

That payload reached the chart as `directHits: 3, totalRequests: 10` and drew a 30% cyan segment while the summary read `--` and "no cache activity recorded". Both percentages are now gated on `hasData`, so `not-engaged` renders the neutral arc alone. The needle was already right — `percentage` is 0 outside `ready`.

The input is defensive rather than observed: the logstore populates the two counters as a pair. But the resolver deliberately guards against a partial payload, and the gauge should not contradict it.

**Minor — the tooltip's first sentence.** Confirmed, and it is the same claim round 2's finding 3 removed from the caption. Nil counters also cover a failed cache-hit aggregation query (`rdb.go:1544-1546`), which the payload cannot distinguish from a window nothing engaged. Round 2 made the caption observational and left the tooltip asserting the stronger claim. Fixed.

I did not take CodeRabbit's suggested second half. The tooltip's mechanism sentence stays as @akshaydeo worded it in round 1 — plugin enablement first, then the cache key — because "configure an `x-bf-cache-key` header or `default_cache_key`" is inaccurate for the header half: `x-bf-cache-key` is sent per request by the caller (`resolveCacheKey` checks the context key before the config default), while `default_cache_key` is gateway configuration. That distinction is what round 1's finding 1 was about, so collapsing both into "configure" would undo it.

No resolver or test change, and no new render probe: a closed Radix tooltip renders no content and the pie segments do not render at zero container size, so neither change is observable in `renderToStaticMarkup`. Re-verified on `6c21d60` — `vitest` 299 passed (unchanged), `tsc --noEmit` clean, `oxlint` and `oxfmt --check` clean on the changed file, `vite build` succeeds. The PR description is updated; its round-2 section quoted the old tooltip string.

Finding 4 from round 2 is still the post-merge follow-up: moving `clampPercentage` to `lib/utils/numbers.ts` and adopting it at `externalCacheTokenMeterChart.tsx:29` and `providerGovernanceTable.tsx:99`. Worth noting that the second site rounds before clamping and has no `Number.isFinite` guard, so adopting the shared helper there turns a rendered `NaN` into `0` — a real behaviour change, and the reason it wants its own PR rather than being folded in here.

wangkanai:
Merged current `dev` (`ce56d2d`) in **92486f7** — the PR was showing as conflicting.

The conflict was entirely `transports/changelog.md`, and it is not a code drift: the **v2.0.0 release rotation emptied that file to 0 bytes** on `dev`, so the one-line entry this PR added under `## 🐞 Fixed` no longer had an anchor. Nothing else on `dev` overlaps — comparing this PR's base to `ce56d2d` (56 commits, 124 files), the only file in common was the changelog. No `ui/**` chart, `ui/lib/utils/`, or `framework/logstore` change on `dev` touches anything this PR depends on.

**The changelog entry is dropped rather than re-seeded.** Writing a fresh `## 🐞 Fixed` header into the now-empty file would simply re-conflict with the next PR that does the same. I'd rather leave it to the release process than churn it — happy to add it back in whatever form you prefer. The PR body is updated to say so.

No review feedback is affected: the merge changes no line of the three remaining files, so every thread stays anchored and nothing needed re-resolving.

The PR is now **3 files**:

- `ui/lib/utils/cacheGauge.ts` (new)
- `ui/lib/utils/cacheGauge.test.ts` (new)
- `ui/app/workspace/dashboard/components/charts/localCacheTokenMeterChart.tsx`

### Re-verified on `92486f7`

| check | result |
|---|---|
| `vitest run` | **323 passed** |
| `tsc --noEmit` | clean |
| `oxlint` (3 changed files) | 0 warnings, 0 errors |
| `oxfmt --check` (3 changed files) | clean |
| `vite build` | succeeds |

Baselined against a clean `ce56d2d` worktree rather than quoted from the earlier run, since `dev` moved 56 commits:

- **323 = 310 on clean `dev` + the 13 tests this PR adds.** Exact delta, no test lost or gained elsewhere.
- `app/workspace/logs/views/columns.test.ts` still fails to *load* (`Failed to parse source for import analysis` on `columns.tsx`, a file this PR does not touch) — reproduced identically on unmodified `ce56d2d`, so it is unchanged pre-existing behaviour, not a regression. Re-flagging only in case it is still news.
- `tsc` is fully clean here. The 93 errors I reported earlier were all knock-on from a missing generated `app/routeTree.gen.ts`; with it generated, there are none, and none in this PR's files either way.

PR #6407, opened 2026-08-21
[fix]: retry KNN search without SORTBY when the backend rejects the result alias (valkey-search)

### LINKED ISSUE #4399: [Bug]: Semantic cache never hits on valkey-search: "Index field `score` does not exist
### Prerequisites

### What happened
With the semantic-cache plugin and a Redis vector store backed by **valkey-search** (valkey-bundle), semantic lookups never hit. Direct (exact) caching works; index creation and vector writes work. Debug logs show:
semantic search skipped: failed to search semantic cache: native vector search failed: Index field score does not exist

### Why
`framework/vectorstore/redis.go` `GetNearest` issues `...=>[KNN k @embedding $vec AS score] ... SORTBY score`. valkey-search doesn't expose the KNN `AS score` alias as an index field and rejects `SORTBY score`. The existing without-SORTBY retry only triggers on RediSearch's `unexpected argument SORTBY` wording, not valkey-search's, so it never fires.

### Affected
- Bifrost: v1.5.13 through current `main` (HEAD has the same narrow match)
- valkey-search: 1.0.2, 1.2.0, `main` — all reproduce

### Impact
Semantic similarity caching is effectively non-functional on valkey-search despite docs listing Valkey as supported. Direct caching unaffected.

### Fix
PR: broaden the retry detection to also match valkey-search's error. (link the MR)

### Steps to reproduce

1. run valkey-bundle
2. configure it in bifrost as semantic cache
3. log will show `{"level":"warn","time":"2026-06-15T10:43:00Z","message":"semantic search skipped: failed to search semantic cache: native vector search failed: Index field `score` does not exist"}`

### Expected behavior

semantic cache to work

### Actual behavior

it fails when backend is valkey

### Affected area(s)

Plugins

### Version

valkey version 9.1, valkey search version 1.2.0

### Environment

```text
K8s
```

### Relevant logs/output

```shell
{"level":"warn","time":"2026-06-15T10:43:00Z","message":"semantic search skipped: failed to search semantic cache: native vector search failed: Index field `score` does not exist"}
```

### Regression?

_No response_

### Severity

wangkanai:
## Summary

Fixes #4399 — semantic cache lookups fail permanently on valkey-search.

`GetNearest` sorts on the KNN result alias:

```go
fmt.Sprintf("%s=>[KNN %d @embedding $vec AS score]", hybridQuery, knnLimit),
"PARAMS", "2", "vec", queryBytes,
"SORTBY", "score",
```

`score` is the alias FT.SEARCH assigns to the KNN distance, not an index attribute. RediSearch accepts `SORTBY` on it. **valkey-search resolves `SORTBY` against the index schema only**, finds no `score` field, and rejects the entire command:

```
semantic search skipped: failed to search semantic cache:
native vector search failed: Index field `score` does not exist
```

A retry-without-`SORTBY` for exactly this case already exists at `redis.go:1198` — the comment even says *"Some Valkey implementations reject SORTBY in KNN search (already distance-ordered)"* — but it only matches RediSearch's wording:

```go
if strings.Contains(errMsg, "unexpected argument `sortby`") || strings.Contains(errMsg, "unexpected argument sortby") {
```

valkey-search's message contains neither string, **so the fallback never fires** and the error propagates. Direct (exact) caching is unaffected, which is why the failure is easy to miss: the plugin keeps writing entries and simply never returns a hit.

This PR makes that existing fallback reachable. It does not change what the fallback does.

## Changes

- **`framework/vectorstore/redis.go`**
  - `isKNNSortByUnsupported(errMsg string) bool` — extracted from the inline condition and widened to also match valkey-search's `Index field \`score\` does not exist`. Signature mirrors the neighbouring `isQuerySyntaxError(errMsg string)`.
  - `const knnScoreAlias = "score"` — the alias is emitted in the KNN clause and then reused in `SORTBY`, in `RETURN`, in the error check, and when reading `Properties["score"]` back. Naming it keeps those five sites from drifting apart, which is the underlying cause of this bug class.
- **`framework/vectorstore/redis_test.go`** — table test for the detector, plus a regression test that drives `GetNearest` against a fake server which rejects `SORTBY` the way valkey-search does.
- **`framework/changelog.md`** — entry under the existing `fix:` lines.

The widened match is anchored on the whole backticked phrase — ``index field `score` `` **and** `does not exist` — so neither an unrelated missing field (``Index field `provider` does not exist``) nor one whose name merely contains the alias (``Index field `llm_scores` does not exist``) is swallowed; both cases are covered by tests. A false positive would cost one extra round trip and nothing else, since KNN results are already distance-ordered.

### Review round 2

Three of the maintainer's nits are folded in, plus one test-coverage item that was scoped as a follow-up:

- **`isKNNSortByUnsupported` is anchored** on ``index field `score` `` rather than on the alias appearing anywhere in the message. This is the second of the two options in the review; the three-substring form (`index field` + alias + `does not exist`) still matches ``index field `llm_scores` does not exist``, because `score` is a substring of `llm_scores`. Two table cases pin the narrowing, and both fail against the previous loose match.
- **The fallback now logs** `s.logger.Debug` before retrying, in the same format as the sibling `DIALECT` fallback, so an operator can see it fire.
- **`parseSearchResultDocument` uses `knnScoreAlias`** instead of a `"score"` literal — the last alias site, and the one where a rename would silently stop populating `SearchResult.Score` rather than erroring. The remaining `score` mentions in the file are prose or local variable names, including the FT.SEARCH syntax example at `redis.go:1183`, which is left literal because it documents wire syntax rather than referencing the alias.
- **The fake server returns one scored document**, so the regression test asserts a parsed `SearchResult` — `Properties[knnScoreAlias]` resolved into `Score` — rather than only that a retry was issued. The fixture uses a KNN distance of `0.2`, which `GetNearest` reports as a similarity of `0.8`.

### Review round 3

The compat retry rebuilt its argument list by searching for the `"SORTBY"` literal:

```go
for i := 0; i < len(args); i++ {
    if i+1 < len(args) && args[i] == "SORTBY" {
        i++ // skip sort field value too
        continue
    }
    compatArgs = append(compatArgs, args[i])
}
```

`namespace` and the `RETURN` field names go into that same slice, so a caller-supplied value equal to `"SORTBY"` matched and took the argument after it along. With `selectFields: []string{"SORTBY"}` the retry drops the projected field *and* `LIMIT`, sending a malformed `FT.SEARCH` — so on valkey-search the fallback fails for an input the fallback exists to rescue.

The clause's offset is now recorded when it is appended, and the retry splices exactly those two arguments out:

```go
sortByAt := len(args)
args = append(args, "SORTBY", knnScoreAlias)
...
compatArgs = append(compatArgs, args[:sortByAt]...)
compatArgs = append(compatArgs, args[sortByAt+2:]...)
```

That covers the namespace case structurally as well, so only the field case needs a test. `TestRedisStore_GetNearest_RetryKeepsSelectFieldNamedSortBy` asserts the retry keeps `LIMIT`, keeps exactly one `SORTBY` (the projected field, not the clause), and keeps `RETURN 2 score SORTBY` intact. All three assertions fail against the scanning loop.

The fake server also had to become faithful: it rejected any command containing the literal, which would fail a retry that merely *projects* a field named `SORTBY`. It now rejects `SORTBY` only in clause position, matching probes B and E against the live server — valkey-search rejects sorting on the alias, not returning it.

To be clear about severity: no caller in this repo passes such a field. `plugins/semanticcache/search.go` passes a fixed projection list, and a namespace named `SORTBY` is operator configuration nobody has. This hardens an exported method rather than fixing an observed failure, and the maintainer's assessment of the deployed code was right. The loop also predates this PR — it becomes reachable on valkey-search only because this PR makes the retry fire there, which is why it is fixed here rather than left as someone else's cleanup.

Not folded in: memoizing the outcome so `SORTBY` is dropped up front. It adds cross-request mutable state to `RedisStore` and warrants its own regression tests, and it can only be written once this merges, since the latch belongs inside the branch introduced here. I will open it as a follow-up after merge.

The forward-looking ordering point (finding 5) needs no change: dropping `SORTBY` relies on KNN results already being distance-ordered, which the doc comment on `isKNNSortByUnsupported` states, and today's only caller passes `limit=1`.

### Review round 4

Test-only. The fake server's clause detector still read *any* pre-`RETURN` occurrence of the literal, so a namespace named `SORTBY` would have been taken for a sort clause and the retry rejected a second time. It now matches `SORTBY` with the field that follows it, the way the real parser does: valkey-search rejects the clause because the KNN alias is not sortable, and the namespace is positional.

`TestCommandHasSortByClause` pins the three shapes that separate a clause from a value spelled the same way — the real clause, a namespace named `SORTBY` after the clause was dropped, and a projected field named `SORTBY`. Against the previous helper only the namespace case fails; the projected-field case was already fixed in round 3.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [ ] Core (Go)
- [ ] Transports (HTTP)
- [ ] Providers/Integrations
- [x] Plugins  *(semantic cache, via `framework/vectorstore`)*
- [ ] UI (React)
- [ ] Docs

## How to test

```sh
cd framework
gofmt -l vectorstore/
go vet ./vectorstore/
go test ./vectorstore/ -run 'TestIsKNNSortByUnsupported|TestRedisStore_GetNearest_RetriesWithoutSortBy' -v
```

Results on this branch:

- Both new tests pass; `go vet` clean; `gofmt` clean on both changed files.
- **The tests fail without the production change.** Reverting just the widened branch to `return false` reproduces the reported error verbatim:
  ```
  --- FAIL: TestRedisStore_GetNearest_RetriesWithoutSortByWhenAliasRejected
      Received unexpected error: native vector search failed: Index field `score` does not exist
      Messages: the rejected SORTBY should be retried, not surfaced
  ```
- Full package: **28 failures before my change, the same 28 after** — all integration tests needing a live Pinecone/Qdrant/Weaviate or an unauthenticated Redis on `localhost:6379`. Identical sets, verified by diffing the failure lists with the change stashed and unstashed.
- Re-verified at the round-3 head: still **28 before / 28 after**, identical sets by diff, `go vet ./vectorstore/` clean, `gofmt` clean on both changed files, `go build ./vectorstore/` succeeds.

  Correcting a claim from the round-2 note: it said `go build ./...` succeeds. It does not — `framework/queryscope` fails with `undefined: schemas.BifrostContextKeyDimensionScope`. That is pre-existing and unrelated: it reproduces on this branch with both changed files stashed, and comes from the `dev` merge, not from anything here. `framework` resolves `core` from the module proxy, so a `framework` file can reference a `core` symbol that the pinned `core` version does not yet export.

### Verified against a live valkey-search

Reproduced on Valkey 9.0.3 + `search` module `v66048` by replaying `GetNearest`'s exact argument list against a real 1024-dim index, borrowing a stored embedding so the vector is valid:

| probe | arguments | result |
|---|---|---|
| A | `SORTBY score` + `RETURN 1 score` — what Bifrost sends today | ❌ `Index field \`score\` does not exist` |
| B | drop `SORTBY` only, keep `RETURN 1 score` | ✅ 5 hits |
| C | keep `SORTBY`, drop `RETURN` | ❌ same error |
| D | drop both | ✅ 5 hits |
| E | drop `SORTBY`, `RETURN 3 score response cache_key` | ✅ 5 hits |

`SORTBY` is the sole cause — **returning** the alias is fine on valkey-search, only **sorting** on it is rejected. That is what makes the existing drop-SORTBY fallback the correct fix rather than a workaround.

## Breaking changes

- [ ] Yes
- [x] No

RediSearch behaviour is unchanged: its own error wording still matches first, and the alias is now a constant with the same value. The only behavioural difference is on backends that previously failed outright.

## Related issues

Closes #4399

## Security considerations

None. No auth, secrets, PII or network destination is touched; the change affects only how one FT.SEARCH error is classified before an already-present retry.

## Checklist

- [x] I updated documentation where needed — `framework/changelog.md`
- [x] I verified builds succeed (Go and UI) — Go only; no UI files are touched

---

Note for maintainers: I put the changelog entry in `framework/changelog.md`, which uses the flat `- fix: … (#issue)` format documented in `raising-a-pr.mdx`, rather than `transports/changelog.md`, whose release-scoped sections I used for an earlier UI-only PR. Happy to move or duplicate it if that is the wrong file for a `framework/` change.


framework/vectorstore/redis.go
```diff
@@ -1196,7 +1223,7 @@ func (s *RedisStore) GetNearest(ctx context.Context, namespace string, vector []
 	if result.Err() != nil {
 		errMsg := strings.ToLower(result.Err().Error())
 		// Some Valkey implementations reject SORTBY in KNN search (already distance-ordered).
-		if strings.Contains(errMsg, "unexpected argument `sortby`") || strings.Contains(errMsg, "unexpected argument sortby") {
+		if isKNNSortByUnsupported(errMsg) {
```
akshaydeo, line 1237: Finding 1 (medium, efficiency) and finding 3 (low, observability).

Failure scenario: on a valkey-search deployment this branch is now reached on **every single** `GetNearest` call, forever. The first `FT.SEARCH` is known-bad from the second request onward, yet it is still sent, awaited, and thrown away. Semantic cache lookup sits in the request pre-hook, so every LLM request through the gateway pays two vector-store round trips instead of one. Before this PR the cost was one round trip and a permanent miss; after it, correctness is fixed but the extra trip becomes the steady state.

Suggested direction (follow-up PR is fine, this does not need to block):

```go
// on RedisStore
knnSortByUnsupported atomic.Bool

// when building args
if !s.knnSortByUnsupported.Load() {
    args = append(args, "SORTBY", knnScoreAlias)
}

// in this branch, before retrying
s.knnSortByUnsupported.Store(true)
s.logger.Debug(fmt.Sprintf("FT.SEARCH SORTBY fallback triggered for namespace %s: %s", namespace, result.Err()))
```

The `Debug` line also brings this in line with the two sibling fallbacks at redis.go:349 and redis.go:1431, which both log before retrying. Right now this one is silent, so an operator has no way to see that every lookup is burning a failed round trip.

framework/vectorstore/redis.go
```diff
@@ -1137,6 +1137,33 @@ func buildRedisQueryCondition(query Query, fieldTypes map[string]VectorStoreProp
...
+//	RediSearch:    "Unexpected argument `SORTBY`"
+//	valkey-search: "Index field `score` does not exist"
+//
+// Retrying without SORTBY is safe either way: KNN results are already ordered
+// by distance, so a false positive costs one extra round trip, never accuracy.
+func isKNNSortByUnsupported(errMsg string) bool {
+	if strings.Contains(errMsg, "unexpected argument `sortby`") || strings.Contains(errMsg, "unexpected argument sortby") {
+		return true
+	}
+	return strings.Contains(errMsg, knnScoreAlias) && strings.Contains(errMsg, "does not exist")
```
akshaydeo: Finding 2 (low-medium). The doc comment above names the wording precisely as ``Index field `score` does not exist``, but the code checks for `score` anywhere in the message and `does not exist` anywhere in the message, independently.

Failure scenario: `VectorStoreNamespace` is user-configured, so a namespace whose name contains `score` (say `llm_scores`) combined with any index-missing error produces a message containing both substrings. The guard fires, a pointless second round trip goes out, and the same error is returned. The blast radius is genuinely small, which is why this is a nit and not a blocker: the retry cannot turn a real failure into a wrong answer, only into one wasted round trip.

Tightening it costs nothing:

```go
return strings.Contains(errMsg, "index field") &&
    strings.Contains(errMsg, knnScoreAlias) &&
    strings.Contains(errMsg, "does not exist")
```

Your existing `index field \`provider\` does not exist` case still passes, and a new case like `index with name \`llm_scores\` does not exist` would document the narrowing.

framework/vectorstore/redis.go
```diff
@@ -1137,6 +1137,33 @@ func buildRedisQueryCondition(query Query, fieldTypes map[string]VectorStoreProp
 	}
 }
 
+// knnScoreAlias is the alias FT.SEARCH assigns to the KNN distance
+// ("... AS score"). It is also what GetNearest sorts and returns on, which is
+// why the two must stay in sync - see isKNNSortByUnsupported.
+const knnScoreAlias = "score"
```
akshaydeo, line 1143: Finding 4 (low). Introducing this constant is the right call and the reasoning in the PR description is sound. One site was missed though: `parseSearchResults` at redis.go:849 still reads the alias back via a literal.

```go
for fieldName, fieldValue := range attrs {
    if fieldName == "score" {   // redis.go:849
```

That is the very drift the constant exists to prevent, and it is the site where a rename would silently stop populating `SearchResult.Score` rather than erroring. Worth switching to `knnScoreAlias` (the comment at redis.go:1152 mentions the alias too, if you want it fully consistent). redis.go:849 is outside this diff, so it cannot be an inline comment there.

akshaydeo (commented):
Thanks for this one, @wangkanai. The write-up is genuinely excellent: the live probe matrix (A through E) isolating `SORTBY` from `RETURN` is exactly the evidence needed to justify reusing the existing fallback instead of inventing a new code path, and the "28 failures before, same 28 after" baseline diff is the right way to report a noisy package. The fix itself is minimal and correct, the detector is table-tested, and the regression test really does fail without the production change.

## Overview

The PR widens an already-present drop-`SORTBY` retry in `RedisStore.GetNearest` so it also fires on valkey-search's error wording, extracts the condition into `isKNNSortByUnsupported`, and names the KNN result alias as a constant. Nothing in the retry's behaviour changes, and RediSearch keeps matching on its own wording first. Nothing I found blocks the merge: the surviving findings are one meaningful efficiency point and a set of small tightening opportunities.

## Findings

|---|---|---|---|---|
| 1 | Medium | [redis.go:1226](https://github.com/maximhq/bifrost/blob/2cf3e0a309a2d9ba0195ca5bcb7d113514c2f491/framework/vectorstore/redis.go#L1226) | The retry is not remembered. On valkey-search every semantic cache lookup now sends a `SORTBY` query that is guaranteed to fail, then retries, so the steady state is two vector-store round trips per lookup on the request hot path. | CONFIRMED |
| 2 | Low-Med | [redis.go:1164](https://github.com/maximhq/bifrost/blob/2cf3e0a309a2d9ba0195ca5bcb7d113514c2f491/framework/vectorstore/redis.go#L1164) | The widened branch matches `score` anywhere plus `does not exist` anywhere, not the anchored `index field \`score\`` phrase the doc comment names. Any error text containing both substrings triggers a wasted retry. | CONFIRMED (impact bounded) |
| 3 | Low | [redis.go:1226](https://github.com/maximhq/bifrost/blob/2cf3e0a309a2d9ba0195ca5bcb7d113514c2f491/framework/vectorstore/redis.go#L1226) | This fallback logs nothing, unlike the sibling `isQuerySyntaxError` fallbacks at redis.go:349 and redis.go:1431 which both emit `s.logger.Debug`. Operators get no signal that every lookup is paying a failed round trip. | CONFIRMED |
| 4 | Low | [redis.go:849](https://github.com/maximhq/bifrost/blob/2cf3e0a309a2d9ba0195ca5bcb7d113514c2f491/framework/vectorstore/redis.go#L849) | `knnScoreAlias` misses one alias site: `parseSearchResults` still compares `fieldName == "score"` against a literal. That is exactly the drift the constant was introduced to prevent. | CONFIRMED |
| 5 | Low | [redis.go:1226](https://github.com/maximhq/bifrost/blob/2cf3e0a309a2d9ba0195ca5bcb7d113514c2f491/framework/vectorstore/redis.go#L1226) | Dropping `SORTBY` leans on the backend returning KNN hits in distance order. Harmless for the only caller today (`plugins/semanticcache/search.go:100` passes `limit=1`, so `knnLimit=1`), but silently order-sensitive if a future caller passes `limit>1`. | PLAUSIBLE |
| 6 | Low | framework/vectorstore/redis_test.go | The fake server answers the retry with `*1\r\n:0\r\n`, an empty result set, so the test proves the retry is issued but never exercises parsing a real hit back through `Properties[knnScoreAlias]`. | CONFIRMED (coverage) |

## Merge recommendation

**Merge after nits.** No finding is a correctness blocker: the change strictly turns a permanently failing path into a working one, and findings 2 through 6 are bounded in cost. Finding 1 is worth a follow-up because it makes a two-round-trip lookup the permanent steady state for every Valkey user, on a path that sits in front of every LLM request.

## Followups

1. **Follow-up PR** - `framework/vectorstore/redis.go:1226`: memoize the outcome. An `atomic.Bool` field on `RedisStore` (for example `knnSortByUnsupported`) set when the fallback first fires, checked when building `args` so `SORTBY` is omitted up front, removes the permanent extra round trip. It is per-store state, so a client repointed at a different backend is a non-issue in practice, and worst case it degrades to what the code does today.
2. **Follow-up PR** - `framework/vectorstore/redis.go:1164`: anchor the match, for example require `index field` alongside the alias and `does not exist`, or match the alias in backticks. Your existing `index field \`provider\` does not exist` test case already guards the direction; this just narrows the remaining surface.
3. **Follow-up PR** - `framework/vectorstore/redis.go:1226`: add an `s.logger.Debug` line mirroring redis.go:349, so the fallback is observable.
4. **Follow-up PR** - `framework/vectorstore/redis.go:849`: use `knnScoreAlias` in `parseSearchResults` instead of the `"score"` literal, closing the last drift site.
5. **Follow-up PR** - `redis_test.go`: have the fake server return one document with a `score` attribute on the retry, so the assertion covers the full path through to a populated `SearchResult.Score`.

None of these need to happen in this PR. If you would rather fold 2, 3 and 4 in here since they are three lines total, that is welcome too.

- **Retry loop / repeated retries**: refuted. The retry is issued exactly once and its result is never re-classified; `if result.Err() != nil` at redis.go:1235 returns unconditionally.
- **The fallback masking a genuine error**: refuted. Dropping `SORTBY` only changes ordering, so a query failing for any other reason fails identically on the retry and the original error still surfaces wrapped as `native vector search failed`.
- **Wrong top-1 cache hit on valkey**: refuted for today's code. `plugins/semanticcache/search.go:100` passes `limit=1`, so the KNN clause itself caps at one document and ordering cannot change the answer. Kept only as forward-looking finding 5.
- **The constant coupling error matching to the query alias**: refuted. valkey-search echoes back the field name we sent in `SORTBY`, so the alias and the error text genuinely do stay in sync when the constant changes.
- **`SORTBY`-stripping loop skipping the wrong argument pair**: refuted. The only `"SORTBY"` literal in `args` is the sort clause itself; the namespace and `RETURN` field names come from config and the projection list.
- **Changelog placement**: cleared. `framework/changelog.md` uses the flat `- fix: ...` form and already carries `thanks [@user]` credits, so your entry matches both the file and the format. Your instinct here was right.
- **AGENTS.md conventions**: cleared. No new Go filenames, gofmt-clean, and no error strings changed, so nothing in the Code Style section applies.

framework/vectorstore/redis.go
wangkanai: Finding 3 is fixed in 9063c03 — `s.logger.Debug` before the retry, same format as the `DIALECT` fallback:

```go
s.logger.Debug(fmt.Sprintf("FT.SEARCH SORTBY fallback triggered for namespace %s: %s", namespace, result.Err()))
```

Finding 1 I am taking as the follow-up you scoped it as, for a reason beyond the "does not need to block": the latch has to live inside the branch this PR introduces, so a follow-up cannot be branched from `dev` until this merges. I will open it once it does.

Two things I want to flag about that follow-up rather than smuggle into this one. It puts cross-request mutable state on `RedisStore`, which is a different risk class from the three-line nits and wants its own tests — at minimum that the second call omits `SORTBY` up front, and that the latch is per-store. And it makes finding 5's ordering assumption permanent rather than per-retry: once `SORTBY` is never sent, every valkey lookup relies on the backend returning KNN hits in distance order, where today that only holds on the retry path. Harmless at `limit=1`, but it is the same assumption with a longer lifetime, so it belongs in the PR that introduces the latch and not spread across two.

framework/vectorstore/redis.go
wangkanai: Fixed in 9063c03, but with the second of your two options rather than the snippet — the three-substring form does not actually close the case you raised:

```go
strings.Contains(errMsg, "index field") &&
    strings.Contains(errMsg, knnScoreAlias) &&
    strings.Contains(errMsg, "does not exist")
```

``index field `llm_scores` does not exist`` still matches all three, because `score` is a substring of `llm_scores`. Your `llm_scores` example was the one I needed to exclude, so I anchored on the whole backticked phrase instead:

```go
return strings.Contains(errMsg, "index field `"+knnScoreAlias+"`") && strings.Contains(errMsg, "does not exist")
```

Built from the constant, so it stays a `knnScoreAlias` site. Backticks survive the `ToLower` upstream, and this is the wording I reproduced against valkey-search v66048 in probes A and C.

Two table cases pin the narrowing, and both fail against the previous loose match — ``index field `llm_scores` does not exist``, and an alias mention outside an index-field phrase.

framework/vectorstore/redis.go
wangkanai: Fixed in 9063c03 — `parseSearchResultDocument` now compares against `knnScoreAlias`. You are right that it is the worst of the sites to have missed: a rename there stops populating `SearchResult.Score` silently, where every other site would fail loudly.

I swept the rest of the file. `grep -n score framework/vectorstore/redis.go` leaves only prose and local variable names now, with one deliberate exception — the FT.SEARCH syntax example at `redis.go:1183`:

```go
// The correct syntax is: (metadata_filter)=>[KNN k @embedding $vec AS score]
```

I left that literal because it documents the wire syntax rather than referencing the alias, and splicing a Go constant name into a query-syntax example makes it harder to match against what actually goes on the wire. It also cannot drift into a bug, unlike line 849. Happy to change it if you would rather have zero literals in the file.

wangkanai:
Review round 2 is in **9063c03**. All three threads are resolved; details are in each thread, summary here.

| # | Severity | Status |
|---|---|---|
| 1 | Medium | Deferred to a follow-up, per your scoping — see below |
| 2 | Low-Med | Fixed — anchored on the backticked phrase |
| 3 | Low | Fixed — `s.logger.Debug` before the retry |
| 4 | Low | Fixed — `knnScoreAlias` in `parseSearchResultDocument` |
| 5 | Low | No change needed |
| 6 | Low | Fixed — folded in, see divergence below |

**Finding 2** took the second of your two options, not the snippet. The three-substring form still matches ``index field `llm_scores` does not exist``, since `score` is a substring of `llm_scores` — the exact case you raised. Anchoring on ``"index field `" + knnScoreAlias + "`"`` closes it and stays a constant site. Two table cases pin the narrowing and both fail against the previous loose match.

**Finding 4** is fixed, with one literal deliberately left: the FT.SEARCH syntax example at `redis.go:1183` documents wire syntax rather than referencing the alias, and cannot drift into a bug the way line 849 could. Say the word if you would rather have zero literals in the file.

**Divergence — finding 6 is folded in**, though you scoped it as a follow-up. It touches only the fake server and the regression test this PR introduces, so a separate PR would land on the same new lines. The fake now serves one document at KNN distance `0.2`, and the test asserts a parsed `SearchResult` with `Score` of `0.8` — `GetNearest` overwrites the raw distance with `1 - distance` — instead of only asserting that a retry was issued. Happy to back it out if you would rather keep the diff to what you asked for.

**Finding 1 stays a follow-up**, and not only because it does not block. The latch lives inside the branch this PR introduces, so it cannot be branched from `dev` until this merges — I will open it then. It also deserves separation on its own merits: it puts cross-request mutable state on `RedisStore`, and it makes finding 5's ordering assumption permanent rather than per-retry, since `SORTBY` would then never be sent on a valkey deployment. Both want their own tests, in the PR that introduces them.

**Finding 5** needs no change here. The reliance on KNN results already being distance-ordered is stated in the `isKNNSortByUnsupported` doc comment, and today's only caller passes `limit=1`. I considered sorting client-side after the fallback and decided against it: it would drag in RESP3 result-map ordering and the missing-score branch at `redis.go:1272`, for an assumption that is currently unreachable.

Verification at the new head — `gofmt` clean on both changed files, `go vet ./vectorstore/` clean, `go build ./...` succeeds, both targeted tests pass, and the full-package baseline is unchanged at **28 failures before / 28 after**, identical sets by diff. The negative control still holds: reverting the detector reproduces the two new table failures.

The PR description is updated — its "deliberately conservative" sentence described the pre-tightening match.

framework/vectorstore/redis.go
```diff
@@ -1196,7 +1227,8 @@ func (s *RedisStore) GetNearest(ctx context.Context, namespace string, vector []
 	if result.Err() != nil {
 		errMsg := strings.ToLower(result.Err().Error())
 		// Some Valkey implementations reject SORTBY in KNN search (already distance-ordered).
-		if strings.Contains(errMsg, "unexpected argument `sortby`") || strings.Contains(errMsg, "unexpected argument sortby") {
+		if isKNNSortByUnsupported(errMsg) {
+			s.logger.Debug(fmt.Sprintf("FT.SEARCH SORTBY fallback triggered for namespace %s: %s", namespace, result.Err()))
```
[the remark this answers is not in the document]
wangkanai: Fixed in 2cc49c8, with one correction to the severity.

The failure is real. With `selectFields: []string{"SORTBY"}` the loop matches the projected field, skips it *and* the argument after it, and the retry goes out as `... RETURN 2 score 0 <n> DIALECT 2` — `LIMIT` eaten, the projection short by one. On valkey-search that means the fallback fails for exactly the input it exists to rescue.

The retry no longer scans. The clause's offset is recorded when it is appended, and the two arguments are spliced out by position:

```go
sortByAt := len(args)
args = append(args, "SORTBY", knnScoreAlias)
...
compatArgs = append(compatArgs, args[:sortByAt]...)
compatArgs = append(compatArgs, args[sortByAt+2:]...)
```

That covers the namespace case structurally too, so only the field case needed a test. `TestRedisStore_GetNearest_RetryKeepsSelectFieldNamedSortBy` asserts the retry keeps `LIMIT`, keeps exactly one `SORTBY` — the projected field, not the clause — and keeps `RETURN 2 score SORTBY` intact. All three fail against the scanning loop.

Writing that test also exposed a flaw in the fake server: it rejected any command containing the literal anywhere, so it failed a retry that merely *projects* a field named `SORTBY`. Real valkey-search rejects sorting on the alias, not returning it — probes B and E in the description. The fake now rejects `SORTBY` only in clause position.

On severity: this is not Major, and @akshaydeo's "refuted" on the same point was right about the code as deployed. No caller in this repo can reach it — `plugins/semanticcache/search.go` passes a fixed projection list, and a namespace named `SORTBY` is operator configuration nobody has. So this hardens an exported method rather than fixing an observed failure. It is worth doing anyway because the positional form is shorter than the argument about reachability, and because the loop predates this PR — it only becomes reachable on valkey-search because this PR makes the retry fire there.

wangkanai:
Review round 3 is in **2cc49c8**. The one CodeRabbit thread is resolved.

@akshaydeo — this one lands on a point you explicitly cleared, so it is worth being direct about it. Your "checked and cleared" entry read:

> **`SORTBY`-stripping loop skipping the wrong argument pair**: refuted. The only `"SORTBY"` literal in `args` is the sort clause itself; the namespace and `RETURN` field names come from config and the projection list.

That is correct for the code as deployed, and I am not treating it as an error. It is a reachability argument, and reachability holds: `plugins/semanticcache/search.go` passes a fixed projection list, and a namespace named `SORTBY` is operator configuration nobody has. CodeRabbit's claim is structural — `GetNearest` is exported, and a caller-supplied `selectFields: []string{"SORTBY"}` does mangle the retry: the loop takes the projected field *and* `LIMIT`, so the command goes out as `... RETURN 2 score 0 <n> DIALECT 2`.

Both readings are right about different things. I fixed it because the positional form is four lines and shorter than maintaining the argument about whether a namespace can be called `SORTBY` — not because the severity is Major, which it is not. The clause offset is recorded when it is appended and spliced out by index, so the namespace case is covered structurally and only the field case needed a test. All three of its assertions fail against the scanning loop.

The loop predates this PR. It becomes reachable on valkey-search only because this PR makes the retry fire there, which is why it is fixed here rather than left as separate cleanup.

Writing the test also caught a flaw in my own fake server: it rejected any command containing `SORTBY` anywhere, so it failed a retry that merely *projects* a field of that name. Probes B and E showed the live server rejects sorting on the alias, not returning it, so the fake now rejects `SORTBY` only in clause position. That made the previous round's fake less faithful than the probe table it was built from.

**A correction to my round-2 verification note.** It claimed `go build ./...` succeeds. It does not — `framework/queryscope` fails with `undefined: schemas.BifrostContextKeyDimensionScope`. That is pre-existing and unrelated: it reproduces on this branch with both changed files stashed, and comes from the `dev` merge rather than from anything here (`framework` resolves `core` from the module proxy, so a `framework` file can reference a `core` symbol the pinned version does not export yet). My round-2 check piped the build through `tail`, which masked it. `go build ./vectorstore/` succeeds.

Otherwise unchanged at the round-3 head: `gofmt` clean on both changed files, `go vet ./vectorstore/` clean, and the full-package baseline still **28 failures before / 28 after**, identical sets by diff.

Finding 1 from your review — memoizing the outcome so `SORTBY` is dropped up front — is still the post-merge follow-up, unchanged.

framework/vectorstore/redis_test.go
```diff
@@ -463,6 +505,47 @@ func (s *fakeRedisSearchServer) stats() (ftSearchCalls int, sawScan bool) {
...
+// precedes RETURN in the command GetNearest builds; projected names always follow it.
+func commandHasSortByClause(command []string) bool {
+	limit := len(command)
+	for i, part := range command {
+		if strings.EqualFold(part, "RETURN") {
+			limit = i
+			break
+		}
+	}
+	return commandContainsArg(command[:limit], "SORTBY")
```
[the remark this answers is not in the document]
wangkanai: Fixed in `2396b88`. The fake now matches `SORTBY` together with the field that follows it, which is what the real parser does — valkey-search rejects the clause because the KNN alias is not sortable, and the namespace is positional, so it can be spelled `SORTBY` without being one.

Kept the `RETURN` boundary as well. It is redundant against the pair match for the case at hand, but it keeps the helper's intent readable and costs nothing.

Pinned with a table test (`TestCommandHasSortByClause`) over the three shapes that separate a clause from a value spelled the same way: the real clause, a namespace named `SORTBY` after the clause was dropped, and a projected field named `SORTBY`. Negative control against the previous helper: only the namespace case fails, which is exactly the gap you described — the projected-field case was already covered by `2cc49c8`.

`gofmt`/`go vet ./vectorstore/` clean, `go build ./vectorstore/` succeeds, and the full-package failure set is unchanged at 28 before / 28 after (those 28 need a live Redis and fail on `dev` too).

wangkanai:
Closing this in favour of **#6597**, which carries the same fix rebased onto current `main`.

This branch was opened against an older base and had drifted into conflict. Rather than force-push over the review history here, I opened a clean branch — the feedback from this PR is incorporated in it:

- the retry splices `SORTBY` out **positionally**, using an index recorded when the clause is appended, rather than scanning `args` for the literal
- the matcher requires the ``index field `<alias>` `` phrase *together with* `does not exist`, so an unrelated missing field or a field merely containing the alias does not match
- the fake backend in the tests matches the `SORTBY` + alias pair the way the real parser does

Since this PR was last updated, the fix has also been **verified in production against a live valkey-search backend**. Before it, every `FT.SEARCH` failed and the semantic cache had never served a hit; after it, a paraphrase pair returns:

```json
{"cache_hit": true, "hit_type": "semantic", "similarity": 0.9309202432632,
 "threshold": 0.8, "cache_hit_latency": 381}
```

Thanks to the reviewers here — the positional-splice and parser-fidelity points both made the fix correct rather than merely working. Continuing in #6597.

PR #6408, opened 2026-08-21
tests: harness tests fixes

TejasGhatte:
## Summary

This PR fixes several correctness issues in the E2E test harness: the wrong AWS endpoint form was being used for OpenAI-family models on Bedrock (causing 404s on the direct leg of the token-parity matrix), embedding extra params were being sent inside an `extra_params` wrapper that is silently ignored on the native `/v1/embeddings` route, prompt-caching round-trip tests could fail non-deterministically when sliced or rerun in isolation, and a Vertex (Claude) caching backend was incorrectly marked as read-guaranteed despite routing through a global multi-region deployment.

## Changes

- **Bedrock OpenAI-family model id split**: Introduced a separate `bedrockOpenaiDirectModel` collection variable holding the cross-Region inference profile form (`global.openai.gpt-5.6-sol`), while `bedrockOpenaiModel` now holds the bare id (`openai.gpt-5.6-sol`). The direct leg of the token-parity matrix calls `bedrock-runtime` Converse, which requires the profile form; the Bifrost leg routes through Bedrock Mantle, which requires the bare id and 404s on a profile-prefixed id. The two variables are now used by the correct legs respectively.

- **Embedding extra params routing fix**: Removed `extra_params` wrappers from all `/v1/embeddings` test bodies for Titan and Cohere on Bedrock. The `extra_params` unwrapping only applies to integration/drop-in routes served by `GenericRouter`; on the native `/v1/embeddings` route, `extractExtraParams` collects unknown top-level keys directly. Sending them nested under `extra_params` caused the wrapper itself to be forwarded to Bedrock, which rejected it. Affected fields (`normalize`, `embeddingTypes`, `input_type`) are now sent as plain top-level keys. The `normalize` test rows also now include `dimensions: 256` so the effect of `normalize: false` is actually observable (at 1024 dimensions Titan V2 already returns a unit-length vector).

- **Prompt-caching chain variable ordering**: The three caching rounds (write → read → read) are now linked via chained collection variables. Each round publishes a variable on any non-4xx response, and the next round consumes it via a `?_chain=` query parameter on the URL. This integrates with the existing `filter-collection.mjs` and `augment-provider-harness.mjs` machinery so that sliced or rerun selections automatically pull in prerequisite rounds, and a missing prerequisite reports a clear error instead of a misleading cache miss.

- **Explicit-cache breakpoint assertion**: Read rounds for explicit-cache backends (Anthropic, Bedrock Claude) now assert that at least one of `cached_tokens` or `cache_write_tokens` is non-zero, catching the case where Bifrost silently drops a `cache_control` breakpoint before it reaches the provider.

- **Vertex (Claude) caching backend marked non-guaranteed**: The `vertex/claude-sonnet-4-6` backend is now `cacheReadGuaranteed: false`. The deployment is configured at `location=global`, which distributes requests across regions; an Anthropic cache entry lives only in the region that wrote it, so a repeat read is not guaranteed to land on the same region.

- **int8 vector assertion fix**: The assertion checking that binary embedding values are integers now uses `Number.isInteger(v[i])` instead of `v[i] % 1 === 0`. Chai's `eql` uses `SameValue` semantics, so `-0 % 1` (which is `-0`) never equalled `+0`, causing the assertion to fail on the first negative value in every int8 vector.

- **Image generation model swap**: Test cases using `vertex/imagen-4.0-generate-001` are replaced with `vertex/gemini-2.5-flash-image`.

- **Gemini Converse `maxTokens` increase**: Bedrock Converse requests targeting `gemini/gemini-2.5-pro` and `vertex/gemini-2.5-pro` now use `maxTokens: 4096` instead of `1024`.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [ ] Core (Go)
- [ ] Transports (HTTP)
- [ ] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

Run the E2E provider harness against a live environment:

```sh
# Run the full collection
cd tests/e2e/api
node runners/run-collection.mjs --provider bedrock
node runners/run-collection.mjs --provider vertex

# Validate prompt-caching chain ordering with a sliced run
node runners/run-collection.mjs --provider anthropic --folder "Prompt caching"

# Validate embedding extra-param routing
node runners/run-collection.mjs --provider bedrock --folder "53. Embeddings"
```

Expected: all token-parity matrix rows for `bedrock_openai` pass on both direct and Bifrost legs; embedding rows for Titan and Cohere pass without Bedrock rejecting an `extra_params` key; prompt-caching rounds 2 and 3 pass when run in isolation via `--rerun-failed`.

## Breaking changes

- [ ] Yes
- [x] No

## Security considerations

None.

## Checklist

### DIFF
--- a/tests/e2e/api/collections/provider-harness.json
+++ b/tests/e2e/api/collections/provider-harness.json
@@ -290,6 +290,11 @@
     },
     {
       "key": "bedrockOpenaiModel",
+      "value": "openai.gpt-5.6-sol",
+      "type": "string"
+    },
+    {
+      "key": "bedrockOpenaiDirectModel",
       "value": "global.openai.gpt-5.6-sol",
       "type": "string"
     },
@@ -4554,7 +4559,7 @@
                     ],
                     "body": {
                       "mode": "raw",
-                      "raw": "{\n  \"messages\": [{ \"role\": \"user\", \"content\": [{ \"text\": \"Hello from Bedrock Converse.\" }] }],\n  \"inferenceConfig\": { \"maxTokens\": 1024 }\n}"
+                      "raw": "{\n  \"messages\": [{ \"role\": \"user\", \"content\": [{ \"text\": \"Hello from Bedrock Converse.\" }] }],\n  \"inferenceConfig\": { \"maxTokens\": 4096 }\n}"
                     },
                     "url": {
                       "raw": "{{baseUrl}}/bedrock/model/gemini%2Fgemini-2.5-pro/converse",
@@ -4610,7 +4615,7 @@
                     ],
                     "body": {
                       "mode": "raw",
-                      "raw": "{\n  \"messages\": [{ \"role\": \"user\", \"content\": [{ \"text\": \"Hello from Bedrock Converse.\" }] }],\n  \"inferenceConfig\": { \"maxTokens\": 1024 }\n}"
+                      "raw": "{\n  \"messages\": [{ \"role\": \"user\", \"content\": [{ \"text\": \"Hello from Bedrock Converse.\" }] }],\n  \"inferenceConfig\": { \"maxTokens\": 4096 }\n}"
                     },
                     "url": {
                       "raw": "{{baseUrl}}/bedrock/model/vertex%2Fgemini-2.5-pro/converse",
@@ -7365,7 +7370,7 @@
                   }
                 },
                 {
-                  "name": "vertex/imagen-4.0-generate-001",
+                  "name": "vertex/gemini-2.5-flash-image",
                   "request": {
                     "method": "POST",
                     "header": [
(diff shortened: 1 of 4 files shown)


TejasGhatte:
* **#6408** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/6408?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/maximhq/bifrost/6408?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `dev`


<h2></h2>

Pratham-Mishra04 (approved):

Pratham-Mishra04:
### Merge activity

* **Aug 21, 10:46 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/6408).
* **Aug 21, 10:47 AM UTC**: @Pratham-Mishra04 merged this pull request with [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/6408).

PR #6414, opened 2026-08-21
fix: handle redaction-shaped admin passwords

### LINKED ISSUE #6413: [Bug]: Valid 32-character admin passwords matching the redaction pattern are silently ignored
### Prerequisites

### Description

A valid admin password is incorrectly treated as an unchanged credential when it matches generic 32-character redaction pattern:
```text
4 characters + 24 asterisks + 4 characters
```
For example:
```text
Aa1b************************cdef
```
This password satisfies the current password requirements. However, `SecretVar.IsRedacted()` classifies it as redacted, causing `ShouldPreserveStored()` to return true.

The admin configuration handler then preserves the existing database password instead of validating, hashing, and storing the newly submitted password. The configuration request succeeds, so the password change is silently discarded.

### Steps to reproduce

  1. Enable dashboard password authentication and configure an existing admin account.
  2. Open `/workspace/config/security`.
  3. Enter the following as the new admin password:
```text
     Aa1b************************cdef
```
  4. Save the configuration.
  5. Log out.
  6. Attempt to sign in using the new password.

### Expected behavior

The submitted password should be treated as a new credential, validated against the password policy, hashed, and stored.

### Actual behavior

The password is treated as an unchanged redacted credential. The save request succeeds, but the existing password remains stored and the newly submitted password cannot be used to sign in.

### Affected area(s)

Transports (HTTP)

### Version

latest

### Environment

```text

```

### Relevant logs/output

```shell

```

### Regression?

_No response_

### Severity

G-XD:
## Summary

Fixes an admin-password update issue where a valid password matching generic 32-character secret mask was treated as an unchanged credential.

## Changes

- Treat mask-shaped admin passwords as new passwords.
- Validate passwords before updating any other config.
- Hash new passwords and preserve env/vault metadata.
- Keep the generic SecretVar.IsRedacted() behavior unchanged.
- Add regression tests and update the transport changelog.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [ ] Core (Go)
- [x] Transports (HTTP)
- [ ] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

```sh
cd transports
go test ./bifrost-http/handlers -count=1
```

## Screenshots/Recordings

No

## Breaking changes

- [ ] Yes
- [x] No

## Related issues

Closes #6413

## Security considerations

New passwords still have to pass the existing password policy. Invalid passwords are rejected before any other config is changed.

## Checklist


transports/bifrost-http/handlers/config_test.go
```diff
@@ -1,10 +1,80 @@
...
+	ConfigManager
+	store configstore.ConfigStore
+}
+
+func (m *configHandlerTestManager) UpdateAuthConfig(ctx context.Context, config *configstore.AuthConfig) error {
+	if config.IsEnabled && (config.AdminUserName == nil || config.AdminUserName.GetValue() == "" ||
+		config.AdminPassword == nil || config.AdminPassword.GetValue() == "") {
+		return errors.New("username and password are required when auth is enabled")
+	}
+	return m.store.UpdateAuthConfig(ctx, config)
```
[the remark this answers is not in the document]
G-XD: Fixed in da0b056ec. The successful update payloads now include `compat: {}`, keeping them aligned with the fixture and avoiding unrelated plugin removal. `ReloadClientConfigFromConfigStore` already has a no-op test implementation.

transports/bifrost-http/handlers/config.go
```diff
@@ -312,6 +353,12 @@ func (h *ConfigHandler) updateConfig(ctx *fasthttp.RequestCtx) {
 		SendError(ctx, fasthttp.StatusBadRequest, "Invalid request payload")
 		return
 	}
+	if payload.AuthConfig != nil {
+		if err := validateSubmittedAdminPassword(payload.AuthConfig.AdminPassword); err != nil {
```
akshaydeo: **Blocking (medium).** This now runs for every payload that carries `auth_config`, including `is_enabled: false`. Before this PR the equivalent checks lived only inside the `if payload.AuthConfig.IsEnabled` branch, so the auth-disable path always succeeded.

Failure scenario: the admin password is stored as `env.ADMIN_PW`. `GET /api/config` returns `{value: "<REDACTED>", ref: "env.ADMIN_PW", type: "env"}`, and the UI resubmits that whole block on every save. `SecretVar.UnmarshalJSON` re-resolves env refs, so if the variable is no longer set, `GetValue()` is `""` and `IsFromSecret()` is true, `shouldPreserveAdminPassword` returns false, and this returns 400 `external reference env.ADMIN_PW for admin_password resolved to an empty value`. The operator can no longer disable auth to recover. The same happens for a resolvable but below-policy secret-backed password, which the config.json bootstrap path (`lib/config.go:4261`) never policy-checks.

Moving validation ahead of the client-config writes is the right idea; it just needs to stay scoped to the path that actually consumes the password:

```go
if payload.AuthConfig != nil && payload.AuthConfig.IsEnabled {
	if err := validateSubmittedAdminPassword(payload.AuthConfig.AdminPassword); err != nil {
		SendError(ctx, fasthttp.StatusBadRequest, err.Error())
		return
	}
}
```

and the disable branch below should treat an unresolved secret reference as preserve-stored before it reaches `hashAdminPassword`.

transports/bifrost-http/handlers/config.go
```diff
@@ -83,6 +83,47 @@ func getPasswordPolicyFailures(password string) []string {
...
+func shouldPreserveAdminPassword(password *schemas.SecretVar) bool {
+	if password == nil {
+		return true
+	}
+	if password.IsFromSecret() {
+		return false
+	}
+
+	value := password.GetValue()
+	return value == "" || strings.EqualFold(value, "<redacted>") || strings.EqualFold(value, "[redacted]")
```
akshaydeo: **Follow-up (medium).** This list is a copy of the sentinel arm of `SecretVar.IsRedacted()` (`core/schemas/secretvar.go:236-243`), with no shared source of truth. The failure mode if they drift is severe and silent: if core adds a third sentinel, or `FullyRedacted()` stops emitting `<REDACTED>`, this function returns false for the placeholder, `validateSubmittedAdminPassword` policy-checks it, and `hashAdminPassword` stores it - locking the operator out of the dashboard with no recovery short of editing the config store.

Suggest exporting one helper in core and calling it from both sides:

```go
// core/schemas/secretvar.go
func IsRedactionSentinel(val string) bool {
	return strings.EqualFold(val, "<redacted>") || strings.EqualFold(val, "[redacted]")
}
```

then `return value == "" || schemas.IsRedactionSentinel(value)` here, and reuse it inside `IsRedacted()`.

transports/bifrost-http/handlers/config.go
```diff
@@ -859,9 +906,9 @@ func (h *ConfigHandler) updateConfig(ctx *fasthttp.RequestCtx) {
 		} else {
 			// Compare with existing config using value comparison (not pointer comparison)
 			// Password is considered changed when it was intentionally submitted —
-			// ShouldPreserveStored() returns false for both plain values and secret refs.
+			// shouldPreserveAdminPassword() only recognizes admin-password sentinels.
 			passwordChanged := payload.AuthConfig.AdminPassword != nil &&
-				!payload.AuthConfig.AdminPassword.ShouldPreserveStored()
+				!shouldPreserveAdminPassword(payload.AuthConfig.AdminPassword)
 			usernameChanged := payload.AuthConfig.AdminUserName != nil &&
```
akshaydeo, line 1012: **Nit.** `usernameChanged` is computed here, but the new backfill at line 935 (`if authConfig != nil && payload.AuthConfig.AdminUserName.GetValue() == ""`) runs afterwards. A client that sends `"admin_username": {"value": ""}` explicitly (rather than omitting the field, which is what the new test covers) gets `usernameChanged == true`, because `Equals` compares `Val`, `ref` and `SecretType` against the stored `admin`. That sets `authChanged`, which flushes every session and signs all admins out even though the backfill restores the identical username.

Easiest fix is to compute `usernameChanged` after the backfill, or to treat an empty non-secret username the same way `AdminPassword == nil` is treated: not submitted.

transports/bifrost-http/handlers/config.go
```diff
@@ -947,8 +976,16 @@ func (h *ConfigHandler) updateConfig(ctx *fasthttp.RequestCtx) {
 			}
 		} else if authConfig != nil {
 			// Auth is being disabled but there's an existing config - preserve credentials and update disabled state
-			if payload.AuthConfig.AdminPassword.ShouldPreserveStored() {
+			if shouldPreserveAdminPassword(payload.AuthConfig.AdminPassword) {
```
akshaydeo: Worth calling out that this branch is a second, separate security fix and not just a refactor: on `dev` the disable-auth path had no `else`, so a submitted plaintext password went straight to `UpdateAuthConfig` unhashed. Any later login then ran `encrypt.CompareHash(<plaintext>, password)` in `session.go:128` and could never succeed. Nice catch.

Two small asks: none of the new tests exercise this branch (they all use `is_enabled: true`), and the changelog entry only mentions the mask collision. A test that disables auth while submitting a new password plus its own changelog line would make this fix visible to operators.

akshaydeo (changes requested):
Thanks for this, @G-XD - it is a well-scoped fix with a clear root-cause writeup in #6413, real regression tests, and a good judgement call in keeping the generic `SecretVar.IsRedacted()` semantics untouched instead of weakening redaction for every credential in the system. The narrowing to `<redacted>` / `[redacted]` is sound: `getConfig` (config.go:186-192) only ever emits `<redacted>` for plaintext admin passwords and `FullyRedacted()`'s `<REDACTED>` for secret-backed ones, so the 4+24*+4 mask is never echoed on this field and can safely be treated as a real password. Both sentinels are also 10 characters, below the 12-character minimum, so neither can ever be a legitimate password that gets silently swallowed.

The one thing holding this back is the relocation of the credential checks to the top of `updateConfig`. Moving them ahead of the client-config writes is the right instinct, but the checks now also run on the auth-disable path, which closes an operator recovery route that worked before.

## Findings

| 1 | Medium | [config.go:357](https://github.com/maximhq/bifrost/blob/da0b056ec79a99d93ec16ff21d0db10f9e575de3/transports/bifrost-http/handlers/config.go#L357) | Early `validateSubmittedAdminPassword` now runs for every payload carrying `auth_config`, including `is_enabled: false`. A secret-backed admin password whose reference no longer resolves - or one that predates the password policy - makes disabling auth fail with 400, where it previously succeeded | CONFIRMED |
| 2 | Medium | [config.go:99](https://github.com/maximhq/bifrost/blob/da0b056ec79a99d93ec16ff21d0db10f9e575de3/transports/bifrost-http/handlers/config.go#L99) | The sentinel list is duplicated from `SecretVar.IsRedacted()` with no shared source. If core ever adds a sentinel or changes `FullyRedacted()`'s placeholder, this path hashes the placeholder as the new admin password and locks the operator out | PLAUSIBLE |
| 3 | Low | [config.go:912](https://github.com/maximhq/bifrost/blob/da0b056ec79a99d93ec16ff21d0db10f9e575de3/transports/bifrost-http/handlers/config.go#L912) | `usernameChanged` is computed before the new username backfill at line 935, so an explicitly empty `admin_username` still counts as a change and flushes every session | CONFIRMED |
| 4 | Low | [config.go:979](https://github.com/maximhq/bifrost/blob/da0b056ec79a99d93ec16ff21d0db10f9e575de3/transports/bifrost-http/handlers/config.go#L979) | The disable-auth branch now hashes a submitted password instead of storing it verbatim. This is a real second security fix (the old code persisted it in plaintext), but it has no test and no changelog line | CONFIRMED |
| 5 | Nit | transports/changelog.md:24 | The new entry omits the `(thanks [@G-XD](https://github.com/G-XD)!) (#6414)` attribution and PR number that every neighbouring entry carries | CONFIRMED |
| 6 | Nit | config_test.go | `configHandlerTestManager` embeds a nil `ConfigManager`, so any future `updateConfig` call to an un-overridden method (`ValidateSetupToken`, `UpdateSyncConfig`) panics on a nil interface instead of failing readably | PLAUSIBLE |

## Detail on finding 1

`SecretVar.UnmarshalJSON` re-resolves `type: "env"` / `type: "vault"` payloads from the environment on every request (secretvar.go:376-383), and the UI resubmits the whole `auth_config` block it received from `GET /api/config` on every save (`securityView.tsx` spreads `...bifrostConfig!`). So for an env-backed admin password:

```mermaid
flowchart TD
    A["GET /api/config"] --> B["admin_password:<br/>value &lt;REDACTED&gt;, ref env.ADMIN_PW, type env"]
    B --> C["UI resubmits with is_enabled: false"]
    C --> D["UnmarshalJSON re-resolves env.ADMIN_PW"]
    D --> E{"env var still set?"}
    E -->|no| F["Val = &quot;&quot;, IsFromSecret = true"]
    F --> G["shouldPreserveAdminPassword -> false"]
    G --> H["validateSubmittedAdminPassword -> 400"]
    H --> I["auth cannot be disabled"]
```

Before this PR the same check lived inside the `if payload.AuthConfig.IsEnabled` branch only (dev config.go:888-891), so the `else if authConfig != nil` disable path preserved the stored credential and returned 200. The same mechanism applies to a resolvable but below-policy secret-backed password, which the config.json bootstrap path (`lib/config.go:4261`) never policy-checks: after this change such an instance can no longer disable auth, or save any config edit while `auth_config` is echoed.

Suggested shape - keep the early call for the enabled path, and let the disable path fall back to the stored credential rather than erroring:

```go
if payload.AuthConfig != nil && payload.AuthConfig.IsEnabled {
    if err := validateSubmittedAdminPassword(payload.AuthConfig.AdminPassword); err != nil {
        SendError(ctx, fasthttp.StatusBadRequest, err.Error())
        return
    }
}
```

and in the disable branch at line 979, treat an unresolved secret reference as preserve-stored before reaching `hashAdminPassword`.

## Merge recommendation

**Not yet - request changes**, on finding 1 alone. The core fix and its tests are right and I would happily approve once the early validation stops applying to `is_enabled: false` payloads; findings 2-6 are all follow-ups or nits and none of them need to block.

## Followups

1. **In this PR (blocking)** - `transports/bifrost-http/handlers/config.go:357`: gate `validateSubmittedAdminPassword` on `payload.AuthConfig.IsEnabled`, and in the disable branch at line 979 preserve the stored credential when the submitted password is an unresolved secret reference. A test that disables auth with an env-backed password pointing at an unset variable would pin it.
2. **Follow-up PR** - `core/schemas/secretvar.go`: export a single `IsRedactionSentinel(string) bool` and have both `SecretVar.IsRedacted()` and `shouldPreserveAdminPassword` call it, so the two lists cannot drift.
3. **In this PR (nit)** - `transports/bifrost-http/handlers/config.go:912`: move the `usernameChanged` computation below the backfill at line 935, or treat an empty non-secret username as "not submitted", so a no-op save does not sign everyone out.
4. **In this PR (nit)** - add a test for the disable-auth hashing branch (line 979) and give it its own changelog line: silently persisting a plaintext admin password on that path was a separate bug and is worth calling out.
5. **In this PR (nit)** - `transports/changelog.md`: add `(thanks [@G-XD](https://github.com/G-XD)!) (#6414)` to match the surrounding entries.
6. **Follow-up PR** - `config_test.go`: replace the nil embedded `ConfigManager` in `configHandlerTestManager` with an explicit fake so future handler changes fail loudly rather than panicking.

## Candidates checked and cleared

- *The 32-character generic mask could be echoed by the API and would now be hashed as the admin password.* Refuted: `getConfig` (config.go:186-192) emits only `<redacted>` or `FullyRedacted()`'s `<REDACTED>` for `admin_password`; `Redacted()`'s 4+24*+4 mask is never applied to this field on any handler path.
- *A user could set the literal `<redacted>` sentinel as their password and have it silently ignored.* Refuted: both sentinels are 10 characters and fail the 12-character minimum in `getPasswordPolicyFailures`, so they can never be valid passwords.
- *`strings.EqualFold` introduces a timing side channel.* Refuted: the comparison is against fixed public sentinels, and the actual credential check remains `encrypt.CompareHash` (bcrypt) in `session.go:128`. No secret-dependent branch is added.
- *The UI still classifies the 32-character mask as redacted and would undo the fix client-side.* Refuted: the admin password field uses `passwordUnchangedRef` and `getPasswordPolicyFailures` (`securityView.tsx:168-191`); `isRedacted` is only used for provider keys and MCP URLs.
- *Removing the inline policy check leaves `hashAdminPassword` reachable without validation.* Refuted: both call sites sit inside the same `if payload.AuthConfig != nil` block the early validation guards, and `AdminPassword` is only reassigned in between to an empty (preserve-shaped) `SecretVar`.
- *The config.json bootstrap path has the same mask collision.* Refuted: `lib/config.go:4261` hashes any non-bcrypt file password unconditionally and never consults `IsRedacted()`.

Issue #3337, opened 2026-05-08
[Feature]: env support for network_config.base_url


ilkersigirci:
### Prerequisites

- [x] I have searched existing issues and discussions to avoid duplicates

### Problem to solve

`providers.*.network_config.base_url` currently behaves like a plain string and does not reliably support `env.VAR_NAME` references, unlike many other config fields.  
This forces users to maintain custom pre-render scripts to inject environment values before Bifrost starts. That adds operational complexity, creates drift between environments.

### Proposed solution

Add first-class `env.VAR_NAME` resolution for `providers.*.network_config.base_url` during config load, consistent with other env-backed fields.

Expected behavior:
- Allow both literal URLs and `env.VAR_NAME`.
- Resolve env values at startup/load time.
- If the env var is missing, return a clear validation/startup error naming the missing variable.

### Area(s)

Framework, Core (Go)

ilkersigirci:
Closed as duplicate of #3244

Issue #3244, opened 2026-05-06
[Feature]: Better Support for Environment Variables in config


ccreighton-apptio:
### Prerequisites

- [X] I have searched existing issues and discussions to avoid duplicates

### Problem to solve

There is a request [specifically for honoring the standard OTel Environment Variables](<https://github.com/maximhq/bifrost/issues/2985>), which is one use case, but every config value should support using environment variables.

For example:

* We should be able to specify model identifiers via environment variables in the `models` array, so that region-specific identifiers are supplied in deployment-specific variables with a single unified config.json across all deployments
* We should be able to also use these in model aliases
* We should be able to toggle things like request/response payload logging in different deployment tiers
* By changing an environment variable we should be able to adjust the effective configuration without modifications
* (Per maximhq/bifrost#3337) We should be able to specify providers.\*.network_config.base_url via environment variables (not my requirement, but I understand why not being able to do this is similarly problematic - this is why ***every*** value should support them!)

### Proposed solution

Every config option should support the ability to use an environment variable to supply the value

### Alternatives considered

* Use LiteLLM
* Maintain per-tier per-region configuration files and copy/paste changes between them

### Area(s)

Core (Go)

### Additional context

*No response*

akshaydeo:
This is a fair ask - we will start increasing the coverage in upcoming releases

PR #6435, opened 2026-08-22
fix: resolve env./vault. references in network_config.base_url

### LINKED ISSUE #6119: network_config.base_url doesn't support env.VARNAME substitution like key values do
## Description

`Key.Value` and some provider-specific fields (e.g. Azure's endpoint, Vertex's project ID) support `env.VARNAME`-style resolution via `SecretVar`. `NetworkConfig.BaseURL` is a plain `string` and doesn't get the same treatment — there's no way to point a provider's base_url at an environment variable natively.

## Why this matters

For anyone who wants to keep `config.json` committed to version control (so provider topology is reviewable, diffable, reproducible across machines) but still have a base_url that varies by environment (dev vs. prod endpoint) or just isn't something they want hardcoded in a git-tracked file, the only option today is an external wrapper that renders the JSON from a template before Bifrost starts. That's exactly what we ended up building — a small script that does placeholder substitution on `config.json` before every `start`, purely because `base_url` can't read from the environment the way `value` already can.

## Suggestion

Extend `SecretVar`-style env resolution (the same `env.VARNAME` convention already used for `Key.Value` and the Azure/Vertex-specific config fields) to `NetworkConfig.BaseURL`, and ideally to other plain-string `NetworkConfig` fields where the same need could come up. This would let a fully environment-parameterized `config.json` be committed to source control without any external templating step.

pryzmatical:
## Summary

`network_config.base_url` was a plain string with no support for `env.VAR_NAME` (or, on Enterprise, `vault.path`) references, unlike `ca_cert_pem` and every key/credential field on the same `NetworkConfig` struct. This meant users who wanted `config.json` to stay committed and diffable while `base_url` varied by environment had to template the file before starting Bifrost. Reported as #6119; the underlying gap was already reported twice before as #3244 and #3337, both closed as duplicates without a fix landing.

## Changes

- **`core/schemas/provider.go`**: `NetworkConfig.UnmarshalJSON` now resolves `BaseURL` through `NewSecretVar(...).GetValue()` — the same `SecretVar` machinery `CACertPEM` already uses two lines below it. A plain literal URL passes through unchanged; an `env.*`/`vault.*` reference resolves the same way it already does for every other secret-capable field on this struct.
- **`core/schemas/serialization_test.go`**: three new tests — an `env.*` reference resolves correctly, a plain URL is unaffected, and an unset env var reference resolves to empty (matching every other `SecretVar`-backed field's behavior on a missing reference, rather than leaving a broken literal string in `BaseURL`).
- **`docs/deployment-guides/config-json.mdx`**: documents that `network_config.base_url` now accepts the same `env.`/`vault.` reference form as the rest of the "Environment Variable References" section, with an example.
- **`core/changelog.md`**: entry added.

I deliberately did not change `BaseURL`'s type to `SecretVar` — 36 files across `core/providers/*` read `.BaseURL` directly, and a type change would need to touch all of them for a fix this scoped. Resolving in `UnmarshalJSON` gets the same practical outcome (config.json portable across environments) without that blast radius.

One tradeoff worth flagging: this resolves the reference once, at parse time. Unlike `CACertPEM`, `BaseURL` doesn't retain its original `env.*` ref for re-serialization, so an API response or a re-saved config would show the resolved literal rather than the reference. Happy to follow up with a `SecretVar`-backed version if that round-trip fidelity turns out to matter for how the config is actually consumed downstream — wanted to get the reported gap closed first without assuming that scope.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [x] Core (Go)
- [ ] Transports (HTTP)
- [ ] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [x] Docs

## How to test

```sh
cd core
go test ./schemas/... -run TestNetworkConfig -v
```

Or manually:

```sh
export MY_BASE_URL="https://your-endpoint.example.com/v1"
# in config.json: "network_config": { "base_url": "env.MY_BASE_URL" }
```

## Breaking changes

- [ ] Yes
- [x] No

Existing plain-string `base_url` values are unaffected — only values with an `env.`/`vault.` prefix change behavior, and previously that prefix would have been sent to the provider as a literal (broken) URL, so there's no working configuration this could regress.

## Related issues

Closes #6119

## Security considerations

None beyond what `SecretVar` already handles for its other fields (env var lookup only, no new external calls in the OSS path).

## Checklist


pryzmatical:
Pushed a follow-up commit addressing the merge-risk note above: "the current configuration schema may still reject those documented reference forms."

Confirmed and fixed in two places:

1. **`transports/config.schema.json`**: `base_url` had `"format": "uri"`, which schema-aware editors (VS Code, JetBrains) would flag as invalid for an `"env.VAR_NAME"` value. Dropped it, matching how `ca_cert_pem` is already declared on the same struct. Verified against `.github/workflows/scripts/validate-schema-sync.sh` (the repo's own Go↔schema sync check) — identical 16 pre-existing warnings and the same `OK` result before and after my change.

2. **`ui/lib/schemas/providerForm.ts`**: the custom-provider form had a hard `/^https?:\/\/.+/` check on `base_url` with no escape hatch, so submitting an `env.X` value through the web UI would block the form with "Base URL must start with http:// or https://". Added the same `isSecretRef` short-circuit already used for this in `ui/lib/utils/validation.ts`'s `isValidVertexAuthCredentials`.

Verified `tsc --noEmit` and `oxlint` against a clean-checkout baseline diff — identical pre-existing error sets before/after, zero new issues.

The other risk note (env state in tests) — happy to switch `os.Setenv`/`os.Unsetenv` to `t.Setenv` in the Go tests if preferred; I matched the existing convention already used in `secretvar_test.go` for consistency, but can change it either way.

pryzmatical:
Pushed a fix for all three review comments. Verified each independently rather than applying the suggested diffs as-is:

**1. Test env cleanup** — switched to `t.Setenv` (auto-restores prior state) for the set case, and added `os.LookupEnv` + `t.Cleanup` restoration for the unset case, since `t.Setenv` has no unset counterpart.

**2 & 3. Schema and UI validation were both too loose, not just too strict** — my previous fix solved the false negative (rejecting `env.X`) but introduced a false positive: any garbage string, or a malformed reference like `env.` or `vault.` with nothing after it, would now silently pass and resolve to an empty `base_url` at runtime instead of being caught early.

Fixed both with the same precise pattern:
- **Schema**: found the repo already has an established convention for exactly this (`anyOf: [{format: uri}, {pattern: env.VAR_NAME}]`, used by `webhook_url` for Slack/Teams) — matched it, and extended with a `vault.path` pattern since `base_url`'s Go-side resolution supports vault too (webhook_url doesn't). Verified against a real format-enforcing validator (`ajv` + `ajv-formats` — my first verification pass used a Python jsonschema setup that turned out to have a no-op format checker, which I caught by testing adversarial cases like `"not a url at all"` and seeing it incorrectly pass; ajv with formats loaded correctly rejects it).
- **UI**: same full-pattern regex, verified with a standalone Node script against the same 7 cases (literal URL, valid env ref, valid vault ref, `env.`, `env.$BAD`, `vault.`, garbage string) -- all pass/fail exactly as intended.

`validate-schema-sync.sh` still reports the same 16 pre-existing warnings and `OK`.

core/schemas/provider.go
```diff
@@ -108,7 +110,10 @@ func (nc *NetworkConfig) UnmarshalJSON(data []byte) error {
 	}
 
 	// Copy all non-duration fields
-	nc.BaseURL = alias.BaseURL
+	// Resolve BaseURL through SecretVar so "env.VAR_NAME" (and, for Enterprise,
+	// "vault.path") references work the same way they already do for CACertPEM
+	// just below. A plain literal URL passes through unchanged.
+	nc.BaseURL = NewSecretVar(alias.BaseURL).GetValue()
```
akshaydeo: We can keep secret instead of resolving and storing as string - if we store and save as string - next update will not be picked up directly

akshaydeo (changes requested):

core/schemas/provider.go
pryzmatical: Good catch, that's the exact tradeoff I flagged in the PR description as something I deliberately punted on. Before I write the follow-up, I want to confirm scope though.

BaseURL is read directly as a plain string in 31 files (around 250 call sites) across core/providers/*. I see two ways to fix the round-trip issue:

1. Full parity with CACertPEM: change NetworkConfig.BaseURL to *SecretVar and update all 31 provider files to use .BaseURL.GetValue(). It's a bigger diff, but it's fully consistent with the existing pattern and closes the gap for good.
2. Scoped fix: keep BaseURL as a resolved string for the provider read sites, but hang onto the original env./vault. reference in a second field that's only used when the config gets re-serialized (API response, config saved from the UI). That way a round-trip doesn't bake in the literal, and future secret rotations still get picked up.

Happy to do either one. Which do you prefer, or did you have a third approach in mind?

akshaydeo: Two blocking problems land on this one line, both because `BaseURL` is a plain `string` rather than a `SecretVar`.

**1. Unredacted env-var read primitive.** `ProviderConfig.Redacted` (framework/configstore/clientconfig.go) masks every `SecretVar`-typed field before a provider is returned by the API, but it does not touch `NetworkConfig` at all. So `{"network_config": {"base_url": "env.OPENAI_API_KEY"}}` now resolves at parse time and the raw value is handed back on the next `GET` of that provider. Any env var in the process is readable this way. Before this change the literal `env.OPENAI_API_KEY` round-tripped harmlessly.

**2. Unset reference silently reroutes traffic.** `NewSecretVar` yields `Val == ""` for a missing var, and downstream empty means "use the default":

```go
// core/providers/openai/openai.go:64
if config.NetworkConfig.BaseURL == "" {
	config.NetworkConfig.BaseURL = "https://api.openai.com"
}
```

`core/providers/cerebras/cerebras.go:48` does the same with `https://api.cerebras.ai`, the provider used in the new docs example. A typo'd or unplumbed variable therefore sends requests plus the configured key to the public endpoint instead of the intended private one, with no error. Previously the same typo failed loudly on the first request.

A fix that closes both:

```go
if schemas.IsSecretRef(alias.BaseURL) {
	resolved := NewSecretVar(alias.BaseURL).GetValue()
	// Do not echo the resolved value in the error: it may be an unrelated secret.
	if !strings.HasPrefix(resolved, "http://") && !strings.HasPrefix(resolved, "https://") {
		return fmt.Errorf("base_url reference %q did not resolve to an http(s) URL", alias.BaseURL)
	}
	nc.BaseURL = resolved
} else {
	nc.BaseURL = alias.BaseURL
}
```

Separately, consider keeping an unexported `baseURLRef` here and re-emitting it from the existing `NetworkConfig.MarshalJSON`, so the reference survives `TableProvider.BeforeSave` / `AfterFind` instead of being replaced by the resolved literal on the first DB write.

ui/lib/schemas/providerForm.ts
```diff
@@ -283,10 +283,17 @@ export const ProviderFormSchema = z
...
 
-			if (data.networkConfig?.base_url && !/^https?:\/\/.+/.test(data.networkConfig.base_url)) {
+			const baseURL = data.networkConfig?.base_url;
+			// env./vault. references resolve on the backend (see NetworkConfig.UnmarshalJSON) --
+			// they won't match the http(s) pattern and shouldn't be forced to. Matched against
+			// the full reference syntax (not just the prefix) so a typo like "env." or "env.$BAD"
+			// still gets caught here instead of silently reaching the backend as an unresolvable,
+			// empty base_url -- same patterns as transports/config.schema.json's base_url entry.
+			const isSecretRef = !!baseURL && (/^env\.[A-Za-z_][A-Za-z0-9_]*$/.test(baseURL) || /^vault\..+$/.test(baseURL));
+			if (baseURL && !isSecretRef && !/^https?:\/\/.+/.test(baseURL)) {
```
akshaydeo, line 296: The regexes are careful and they do match the schema entry, so the intent here is right. The problem is what happens after the user takes the offer: `NetworkConfig.UnmarshalJSON` resolves the reference and `TableProvider.BeforeSave` persists the resolved literal, so the reference the user typed is gone. Concretely, on a custom provider (the only branch this runs for, since `baseURLRequired` is `isCustomProvider`): type `env.MY_URL`, save, reload the form, and the field shows `https://...`. If `MY_URL` is unset it persists `""` and this required field comes back empty with no explanation.

Suggest holding this UI relaxation until the backend preserves the reference across the DB round-trip, so the form can show back what was typed.

transports/config.schema.json
```diff
@@ -3918,8 +3918,8 @@
       "properties": {
         "base_url": {
           "type": "string",
```
akshaydeo, line 4214: This `anyOf` does not constrain anything. `format` is annotation-only unless the validator opts in, and `transports/bifrost-http/lib/validator.go` never calls `AssertFormat` on the `santhosh-tekuri/jsonschema/v6` compiler, so the first branch matches every string and the two patterns never get a say.

The same reasoning applies to the pre-change `"format": "uri"`: it never rejected `env.FOO` either, so schema validation was not what blocked this feature.

Either drop the `anyOf` and keep just the improved `description`, or make the first branch a real assertion:

```json
"anyOf": [{ "pattern": "^https?://" }, { "pattern": "^env\\.[A-Za-z_][A-Za-z0-9_]*$" }, { "pattern": "^vault\\..+$" }]
```

akshaydeo (changes requested):
Thanks for this - the problem statement is real, the write-up is unusually clear about what you deliberately did not do, and you already flagged the round-trip tradeoff yourself. The tests, the schema entry, the UI validator and the docs are all touched, which is the right blast radius for a config change. The reason I am asking for changes is that `BaseURL` differs from `CACertPEM` in two ways that the one-line change inherits: `SecretVar` keeps its `ref` through marshal, and every `SecretVar`-typed field is redacted before it reaches an API response. A plain `string` `BaseURL` has neither property, so "resolve it like `CACertPEM`" gives resolution without the two guarantees that make resolution safe.

### Findings

|---|---|---|---|---|
| 1 | High (security) | [core/schemas/provider.go:116](https://github.com/maximhq/bifrost/blob/fix/base-url-env-reference/core/schemas/provider.go#L116) | `base_url` becomes an arbitrary env-var read primitive whose result is returned unredacted by the provider API | CONFIRMED |
| 2 | High | [core/schemas/provider.go:116](https://github.com/maximhq/bifrost/blob/fix/base-url-env-reference/core/schemas/provider.go#L116) | An unset or misspelled reference resolves to `""`, and providers silently fall back to their public default endpoint instead of failing | CONFIRMED |
| 3 | Medium | [framework/configstore/tables/provider.go:77](https://github.com/maximhq/bifrost/blob/dev/framework/configstore/tables/provider.go#L77) | The reference is destroyed on the first DB write-back, because `NetworkConfig` marshals the resolved literal | CONFIRMED |
| 4 | Medium | [ui/lib/schemas/providerForm.ts:293](https://github.com/maximhq/bifrost/blob/fix/base-url-env-reference/ui/lib/schemas/providerForm.ts#L293) | The UI now accepts a reference the backend cannot store, so the field silently changes value (or empties) after save | CONFIRMED |
| 5 | Low | transports/config.schema.json:3920 | The new `anyOf` is vacuous: the `{"format":"uri"}` branch always passes, so the schema constrains nothing | CONFIRMED |
| 6 | Low | [core/schemas/provider.go:61](https://github.com/maximhq/bifrost/blob/fix/base-url-env-reference/core/schemas/provider.go#L61) | The doc example uses `cerebras`, but the comment being edited still lists only OpenAI, Anthropic, Cohere, Mistral and Ollama | CONFIRMED |

### Finding 1 - env-var disclosure through the provider API (high)

`SecretVar` resolution is safe for `CACertPEM` because the value never leaves the process in cleartext: `ProviderConfig.Redacted` masks every `SecretVar`-typed field before the config is returned (see `framework/configstore/clientconfig.go`, for example `ollamaConfig.URL = *key.OllamaKeyConfig.URL.Redacted()`). `NetworkConfig.BaseURL` is a plain `string` and is not in that redaction pass at all.

So after this change, anyone who can create or update a provider and then read it back can do:

```json
{ "network_config": { "base_url": "env.OPENAI_API_KEY" } }
```

and `GET` the provider to receive the resolved value verbatim. Any env var in the gateway process is readable this way - API keys, `BIFROST_ENCRYPTION_KEY`, DB DSNs. Before this PR the same input round-tripped as the literal string `env.OPENAI_API_KEY`, so nothing leaked.

This is the one that makes me request changes rather than nit. Two ways out, either is fine:

- Only accept a resolved value that parses as an `http(s)` URL, and reject otherwise without echoing the value. This kills almost the entire disclosure surface, since few secrets are valid HTTP URLs, and it also fixes finding 2.
- Or resolve only in the config-file load path (`transports/bifrost-http/lib/config.go`), never for `NetworkConfig` values that arrive from an API request body. A reference in a file the operator controls is a very different trust level from one in a request body.

### Finding 2 - unset reference silently reroutes traffic (high)

`NewSecretVar` sets `Val = ""` when the env var is missing, and `TestNetworkConfig_BaseURLUnsetEnvReference` locks that in. Downstream, an empty `BaseURL` is not an error, it is the "use the default" signal:

```go
// core/providers/openai/openai.go:64
// Set default BaseURL if not provided
if config.NetworkConfig.BaseURL == "" {
	config.NetworkConfig.BaseURL = "https://api.openai.com"
}
```

`core/providers/cerebras/cerebras.go:48` does the same with `https://api.cerebras.ai`, which is the exact provider in the new docs example. So an operator who typos `env.CEREBRAS_BASE_URL` as `env.CEREBRAS_BASEURL`, or who deploys to an environment where the var was not plumbed through, gets a gateway that boots clean and quietly sends every request plus the configured API key to the public endpoint instead of the intended private one. Before this PR the same typo produced an immediate hard failure on the first request, because `env.CEREBRAS_BASEURL/v1/chat/completions` is not a dialable URI.

Empty is a defensible default for `CACertPEM` (fall back to system roots). It is not a defensible default for a field whose empty value changes where credentials are sent. Suggestion: when `IsSecretRef(alias.BaseURL)` is true and the resolved value is empty, return an error from `UnmarshalJSON` naming the reference (not the value). That turns a silent misroute into a startup failure, which is what an operator wants here.

### Finding 3 - the reference does not survive a DB round-trip (medium)

You flagged this in the description as "an API response or a re-saved config would show the resolved literal". It is a little sharper than that in practice, because the resolved literal is what gets persisted:

```mermaid
flowchart LR
  A["config.json / API body<br/>base_url: env.FOO"] --> B["NetworkConfig.UnmarshalJSON<br/>resolves to literal"]
  B --> C["TableProvider.BeforeSave<br/>json.Marshal(NetworkConfig)"]
  C --> D["config_providers.network_config_json<br/>holds the literal, ref is gone"]
  D --> E["AfterFind: literal<br/>nothing left to re-resolve"]
```

`CACertPEM` does not have this problem: `SecretVar.MarshalJSON` emits `ref` and `type`, and `SecretVar.UnmarshalJSON` re-resolves from `ref` on the way back in, so the indirection is preserved across the DB hop and a later env-var change still takes effect. `BaseURL` loses the indirection on the first write.

For file-managed providers this partly self-heals, since `GenerateConfigHash` hashes the resolved `NetworkConfig` and a changed env var re-syncs from `config.json`. For a provider created through the UI or API there is no file to re-sync from, so the reference is simply gone.

A targeted fix that avoids the `SecretVar` type change you (reasonably) ruled out: keep an unexported `baseURLRef string` on `NetworkConfig`, set it in `UnmarshalJSON` when the input was a reference, and re-emit it instead of the resolved value in the existing `NetworkConfig.MarshalJSON`. That is a handful of lines in one file, leaves all 36 `.BaseURL` readers untouched, and fixes findings 3 and 4 together (and most of 1, since the API response would then carry the ref rather than the value).

### Finding 4 - the UI validator promises more than the backend delivers (medium)

The zod change is careful and the regexes match the schema, so the intent is right. But this branch only runs for custom providers, where `base_url` is required, and it lets the user type `env.MY_URL` into a field whose value cannot survive the save. Concretely: user types `env.MY_URL`, saves, the backend resolves and persists `https://...`, the form reloads showing the literal. If the var is unset the backend persists `""` and the required field comes back empty with no explanation. Worth holding this piece until finding 3 is fixed, so the field shows back what the user typed.

### Finding 5 - the schema `anyOf` does not constrain anything (low)

```json
"anyOf": [{ "format": "uri" }, { "pattern": "^env\\.[A-Za-z_][A-Za-z0-9_]*$" }, { "pattern": "^vault\\..+$" }]
```

`format` is annotation-only unless the validator opts in, and `transports/bifrost-http/lib/validator.go` never calls `AssertFormat` on the `santhosh-tekuri/jsonschema/v6` compiler. So branch one matches every string, the `anyOf` is always satisfied, and the two patterns are decorative. The same reasoning means the pre-change `"format": "uri"` never rejected `env.FOO` either, so schema validation was not the thing blocking this feature. Either drop the `anyOf` and keep the improved `description`, or make branch one a real constraint such as `{ "pattern": "^https?://" }` so the three-way choice actually validates.

### Finding 6 - docs example contradicts the comment being edited (low)

The mdx example uses `cerebras`, while the comment two lines above your change still reads "BaseURL is supported for OpenAI, Anthropic, Cohere, Mistral, and Ollama providers". `cerebras.go:48` does honor `BaseURL`, so the comment is the stale half. Since this PR is already editing that exact comment, it is a good moment to widen it or drop the provider list.

### Merge recommendation

**Not yet - request changes.** Finding 1 turns a config field into an unredacted env-var read primitive, and finding 2 converts a loud misconfiguration into a silent credential misroute to a public endpoint. Findings 3 to 6 are cleanups that mostly fall out of fixing 1 and 2.

### Followups

1. **In this PR (blocking)** - `core/schemas/provider.go:116`: constrain what a reference is allowed to resolve to (require an `http(s)` URL), or resolve only on the config-file path so a request body cannot read env vars. Fixes finding 1.
2. **In this PR (blocking)** - `core/schemas/provider.go:116`: when the input is a secret reference and the resolved value is empty, return an error naming the reference instead of leaving `BaseURL` empty. Fixes finding 2. `TestNetworkConfig_BaseURLUnsetEnvReference` should then assert the error.
3. **In this PR (blocking)** - `core/schemas/provider.go`: add an unexported `baseURLRef` and re-emit it from `NetworkConfig.MarshalJSON`, so the reference survives `TableProvider.BeforeSave` / `AfterFind`. Fixes findings 3 and 4.
4. **In this PR** - `transports/config.schema.json:3920`: replace the `{"format":"uri"}` branch with a real constraint, or drop the `anyOf`. Fixes finding 5.
5. **Follow-up PR** - `core/schemas/provider.go:61`: refresh the stale provider list in the `BaseURL` comment. Fixes finding 6.

Happy to look again as soon as 1 and 2 are addressed - the rest is small.

- *A plain URL could be misparsed by `parseSecretRef`'s JSON branch.* Refuted: that branch is gated on `sonic.Valid([]byte(value))` plus a `value`/`ref`/`env_var` node, and a bare URL string is not valid JSON.
- *The ollama/sgl backfill migration at `framework/configstore/migrations.go:7356` would bake a resolved value into a key.* Refuted: the DB never holds an env reference for `base_url` in the first place (see finding 3), so that migration always reads a literal.
- *Vault resolution now runs on every `NetworkConfig` unmarshal, including per-read `AfterFind`.* Refuted: `LookupVault` returns immediately unless the string carries a `vault.` prefix.
- *Existing configs whose `base_url` legitimately starts with `env.` or `vault.` would break.* Refuted: a usable base URL needs a scheme, so no working configuration can start with those prefixes. Your reasoning in the description holds.
- *The new tests leak env state across the suite.* Refuted: `t.Setenv` restores, and the unset test's manual `t.Cleanup` restore is correct.

pryzmatical:
Pushed a commit addressing all six findings from the review above (9530b23). Walking through each below, plus one spot where I deliberately didn't do what was suggested, and I want to explain why before you see the diff.

Findings 1 and 2 (env-var disclosure, silent misroute): UnmarshalJSON now only accepts a resolved reference if it starts with http:// or https://. Anything else, an unset var, a typo, a resolved value that isn't a URL, never gets stored in BaseURL and never leaves the function. So there's nothing left to leak.

Now here's the part where I went off-script. The suggested fix returns an error in that case. I built that first, then went looking for where else UnmarshalJSON runs before I trusted it, and found a landmine: it also fires inside TableProvider's GORM AfterFind hook. I spun up an actual gorm.io/driver/sqlite instance to check, and confirmed an AfterFind error on one row takes the whole batch a Find() call loaded down with it, not just that row. GetProvidersConfig treats that as total failure, loadProviders reads that as "no providers in the store," and listProviders just 500s. So one provider with a stale env var would quietly wipe out every other DB-configured provider on the next restart and take the whole API down with it. That's a worse bug than the one we're trying to close.

So instead of erroring, an unresolved or non-URL reference just leaves BaseURL as the raw reference string, e.g. env.CEREBRAS_BASE_URL. It's never empty, so it doesn't trip the "" = use the public default fallback sitting in openai.go and cerebras.go, and it still fails the first real request to that provider immediately and loudly. Checked that against a real fasthttp client too: clean "missing port in address" error, no panic, no surprise connection anywhere. Scoped to the one broken provider, and it self-heals the moment the reference starts resolving.

Added TestNetworkConfig_BaseURLNeverErrorsOnUnmarshal to nail down that base_url can never fail UnmarshalJSON, and TestUpdateProvidersConfig_BrokenBaseURLReferenceDoesNotDropOtherProviders in framework/configstore, which runs the real GORM round trip (BeforeSave -> NetworkConfigJSON column -> AfterFind) instead of just calling UnmarshalJSON directly. Put the error-returning version back in temporarily to watch that second test fail before locking in the fix, so it's a real regression guard, not just a test that happens to pass.

Finding 3 (reference lost on DB round-trip): added an unexported baseURLRef on NetworkConfig, set in UnmarshalJSON, re-emitted by MarshalJSON instead of the resolved value. Didn't want to just trust encoding/json in a test either, so checked it against what the API handlers actually use, sonic.ConfigStd.Marshal going out, sonic.Unmarshal coming in. Both invoke the custom methods correctly, round trip holds end to end.

Finding 4 (UI validator ahead of backend): no code change needed here. Once the reference survives the round trip, the form gets back what the user actually typed instead of the resolved literal or a mystery empty field, so the existing validation was already fine.

Finding 5 (vacuous schema anyOf): swapped {"format":"uri"} for {"pattern":"^https?://"}, matching what the other two branches already do. Ran it through ajv+ajv-formats against nine cases, including the adversarial ones you'd want to try and break it with.

Finding 6: fixed the stale comment.

validate-schema-sync.sh still comes back with the same 16 pre-existing warnings and OK. Full core/schemas and framework/configstore suites pass.

(Posting as a top-level comment rather than an inline reply -- GitHub's API is 404ing on replies to the review threads right now, likely still reprocessing positions after the push.)

pryzmatical:
Nice catches, CodeRabbit. Pushed a follow-up (40b3631) for all three.

Changelog wording: fair, "rejected outright" was just wrong. It's not rejected, it's substituted. Fixed the wording to actually match what the code does.

The real one is the URL check. `strings.HasPrefix(resolved, "https://")` only checks the scheme, so a resolved value of exactly `"https://"` with nothing after it would've sailed right through. That's on me, I copied the maintainer's suggested check verbatim and didn't stress test it. Swapped it for `net/url.Parse` plus a check that `Host` is actually non-empty, same idea applied to the schema pattern (`^https?://[^/\s]`). Verified both against 13 cases, plain URLs, the reference forms, bare scheme, scheme with an empty path, scheme with trailing whitespace.

Since that check is load-bearing for the security fix, I didn't stop at just fixing the reported gap, I went and tried to break it myself. Threw CRLF, embedded NUL, percent-encoded CRLF, tabs, and a few header-injection-shaped payloads at `url.Parse` to see if anything could sneak through while still parsing as a valid scheme+host. All of it gets rejected outright with a parse error, so nothing gets past the check that shouldn't. Also traced where BaseURL actually ends up: `ConfigureDialer` resolves DNS itself and dials the IP literal directly, blocking link-local and private ranges unless the operator opts in, completely independent of whether BaseURL came from a literal or a resolved reference. So there's no SSRF surface here that didn't already exist for a literal base_url.

Test env leak: yeah, fair, my new test assumed the var was unset without guarding it. Brought it in line with the pattern the schemas tests already use.

All three should be closed now. Let me know if anything else jumps out.

pryzmatical:
Rebased onto current `dev` to clear the merge conflict, still green on CodeRabbit after the rebase. Happy to address anything else if there's more to look at.

PR #6436, opened 2026-08-22
Fix/bifrost 6102 genai resumable file upload issue

### LINKED ISSUE #6102: [Bug]: `/genai` Files API: resumable upload finalize step returns 500 "failed to convert file request"
### Prerequisites

### Description

**Summary:**
The Google GenAI-compatible `/genai` passthrough correctly handles step 1 of the resumable file-upload protocol (`X-Goog-Upload-Command: start`), returning a valid `x-goog-upload-url` session. However, step 2 — sending the actual file bytes with `X-Goog-Upload-Command: upload, finalize` — always fails with:

```json
{"error":{"code":500,"message":"failed to convert file request","status":"","details":null}}
```

##  Related issues

What I found on maximhq/bifrost's GitHub issues:

No open or closed issue matches the exact error text "failed to convert file request" — that specific 500 on the resumable-upload finalize step doesn't appear to be filed yet. However
- #2482("/genai 502 on generateContent with inline PDF + response_schema", closed Apr 9) is closely related — its notes explicitly mention: "Also: Files API upload via /genai loses MIME type (returns application/octet-stream)". That's a different symptom (MIME-type mangling vs. our 500 on finalize) but points at the same subsystem (Bifrost's /genai Files API handling) having known issues, and it was closed without that specific MIME sub-note apparently being fully addressed.
- #123 "Files API Support" is the original feature-add (closed, shipped in v1.2.0) — worth checking if the finalize-step bug is a regression introduced after that.
- No dedicated open issue for our exact repro.

**Impact:** Blocks any use case relying on Gemini Files API (document/PDF uploads for multimodal grounding) through the `/genai` passthrough, while `generateContent`, `cachedContents` CRUD, and other native-Gemini-format endpoints work correctly.

### Steps to reproduce

**Steps to reproduce (raw HTTP, no SDK):**
```python
import httpx

BASE = "https://<gateway-host>/genai"
VK = "<virtual-key>"
pdf_bytes = b"%PDF-1.1\n...minimal valid PDF...\n%%EOF"

# Step 1 — succeeds, returns 200 + x-goog-upload-url
r1 = httpx.post(f"{BASE}/upload/v1beta/files",
    headers={
        "x-goog-api-key": VK,
        "X-Goog-Upload-Protocol": "resumable",
        "X-Goog-Upload-Command": "start",
        "X-Goog-Upload-Header-Content-Length": str(len(pdf_bytes)),
        "X-Goog-Upload-Header-Content-Type": "application/pdf",
        "Content-Type": "application/json",
    },
    json={"file": {"display_name": "test.pdf"}},
)
upload_url = r1.headers["x-goog-upload-url"]

# Step 2 — always returns 500 "failed to convert file request"
r2 = httpx.post(upload_url,
    headers={
        "x-goog-api-key": VK,
        "X-Goog-Upload-Command": "upload, finalize",
        "X-Goog-Upload-Offset": "0",
        "Content-Length": str(len(pdf_bytes)),
    },
    content=pdf_bytes,
)
print(r2.status_code, r2.text)  # 400 / {"error":{"code":500,"message":"failed to convert file request"...}}
```

### Expected behavior

 File finalizes and a files/... resource is returned (matching real Gemini Files API behavior), and works identically via google-genai SDK's client.aio.files.upload().

### Actual behavior

500 error, reproduced consistently across multiple header/Content-Type variations on step 2. Step 1 (session init) works correctly, so the bug appears isolated to how the gateway parses/converts the raw binary continuation request.

### Affected area(s)

Core (Go)

### Version

v1.6.7

### Environment

```text
- Bifrost v1.6.7 self-hosted, endpoint: `/genai`
- Auth: `x-goog-api-key` with a Bifrost virtual key
- Client: `google-genai` Python SDK (`client.aio.files.upload()`), and reproduced with raw `httpx` requests bypassing the SDK entirely.
```

### Relevant logs/output

```shell

```

### Regression?

_No response_

### Severity

priyankac-scalent:
## Summary

Fixes GenAI resumable file uploads where retrying an interrupted upload could fail with a `500 failed to convert file request`.

The issue was caused by the resumable upload session being deleted from the KV store when the file request was converted. The session is now retained while the upload is in progress, allowing the same upload session to be reused for retries.

## Changes

* Updated GenAI file request conversion to retrieve the resumable upload session without deleting it.
* Changed KV store access from `GetAndDelete` to `Get` for resumable upload sessions.
* Preserved the upload session across multiple requests so interrupted uploads can continue using the same upload URL/session.
* Updated upload callback handling to correctly support the resumable upload flow.
* Added test coverage for:

  * Normal file upload.
  * Resumable file upload.
  * Resumable upload retry using the same session.
  * Verifying that the resumable upload session is retained after request conversion.

### Design decisions

* The resumable upload session must remain available until the upload is completed.
* Session cleanup should happen only after the upload is successfully finalized, rather than when the session is first read.
* This avoids breaking retries while keeping the existing KV-store-based session management.

## Type of change

* [x] Bug fix
* [ ] Feature
* [ ] Refactor
* [ ] Documentation
* [ ] Chore/CI

## Affected areas

* [x] Core (Go)
* [x] Transports (HTTP)
* [x] Providers/Integrations
* [ ] Plugins
* [ ] UI (React)
* [ ] Docs

## How to test

Run the complete Go test suite:

```sh
go test ./...
```

Run the relevant integration tests:

```sh
go test ./core/providers/gemini/... -v
```

### Scenarios to verify

**Normal file upload**

* Upload a file and verify it completes successfully.

**Resumable file upload**

* Start and complete a resumable upload successfully.

**Interrupted resumable upload**

* Interrupt a partial upload, retry with the same session, and verify it continues successfully.

**Session retention**

* Verify the resumable session remains available after request conversion.

**Finalize upload**

* Complete and finalize the upload successfully without `failed to convert file request`.

**Expected result:** Resumable uploads can be interrupted and retried using the same upload session without losing the session or returning a `500 failed to convert file request` error.

## Screenshots/Recordings

Not applicable — no UI changes.

## Breaking changes

* [ ] Yes
* [x] No

## Related issues

Closes #6102

## Security considerations

No new security or authentication changes were introduced.

The existing upload session and KV-store mechanisms continue to be used. Resumable upload sessions remain available while the upload is in progress and should be cleaned up after successful completion.

## Checklist

* [x] I read `docs/contributing/README.md` and followed the guidelines
* [ ] I verified the CI pipeline passes locally if applicable


transports/bifrost-http/integrations/genai.go
```diff
@@ -33,6 +36,11 @@ const requestedGeminiModelMetadataContextKey schemas.BifrostContextKey = "bifros
 
 const genAIRawRequestBodyContextKey schemas.BifrostContextKey = "bifrost-genai-raw-request-body"
 
+const (
+	MaxResumableUploadSize = 100 * 1024 * 1024 // 100 MB, aligned with the maximum request body size.
```
akshaydeo: this is configurable in bifrost - so we should us that vs hardcoding it to 100 MB

transports/bifrost-http/integrations/genai.go
```diff
@@ -33,6 +36,11 @@ const requestedGeminiModelMetadataContextKey schemas.BifrostContextKey = "bifros
...
+	MaxResumableChunkSize  = 10 * 1024 * 1024  // 10 MB, limits the size of each individual resumable upload chunk.
```
akshaydeo: This is also not required

akshaydeo (changes requested):

transports/bifrost-http/integrations/genai.go
```diff
@@ -343,10 +351,28 @@ func CreateGenAIFileRouteConfigs(pathPrefix string, handlerStore lib.HandlerStor
 		RequestParser: func(ctx *fasthttp.RequestCtx, req interface{}) error {
 			r := req.(*gemini.GeminiFileUploadHandlerReq)
 			uploadID := string(ctx.QueryArgs().Peek("upload_id"))
+
 			if uploadID != "" {
 				// Step 2: body is raw binary — do not JSON-decode.
+				body := ctx.Request.Body()
+
+				if len(body) > MaxResumableChunkSize {
```
akshaydeo: Blocking: this 10 MB per-chunk cap rejects the upload flow Google documents, which sends the whole file in one request after `start`:

```bash
curl "${upload_url}" \
  -H "X-Goog-Upload-Offset: 0" \
  -H "X-Goog-Upload-Command: upload, finalize" \
  --data-binary "@${PDF_PATH}"
```

Source: https://ai.google.dev/gemini-api/docs/document-processing

Failure scenario: a client uploads a 30 MB PDF in one request today and it succeeds, since the server accepts up to `MaxRequestBodySizeMB` (default 100, `framework/configstore/tables/clientconfig.go:43`). After this PR the same request fails with `upload chunk exceeds maximum size of 10 MB`.

Suggested direction: derive the ceiling from the configured body limit instead of hardcoding it, so the cap is never below what the server already accepts.

```go
// maxBodyBytes threaded in from ClientConfig.MaxRequestBodySizeMB * 1024 * 1024
if int64(len(body)) > maxBodyBytes {
    return errors.New("upload chunk exceeds maximum request body size")
}
```

Minor while you are here: `fmt.Errorf` on this line has no format verbs, so `errors.New` is the right call.

transports/bifrost-http/integrations/genai.go
akshaydeo: Blocking: the comment says this is "aligned with the maximum request body size", but that limit is `MaxRequestBodySizeMB`, which is user-configurable (default 100, `framework/configstore/tables/clientconfig.go:43`). An operator who raises it to 500 to push large media through the gateway now hits a hardcoded 100 MB ceiling with no knob to change it.

Upstream, the Gemini Files API allows "2 GB" per file and 20 GB per project (https://ai.google.dev/gemini-api/docs/files), so this is a Bifrost-only restriction that did not exist before this PR.

Suggested direction: make both constants configuration-driven (thread `ClientConfig.MaxRequestBodySizeMB` into `CreateGenAIFileRouteConfigs`), and if a hard total ceiling is still wanted, surface it in `config.schema.json` rather than pinning it in code.

akshaydeo (changes requested):
Thanks for this, and thanks for the unusually thorough PR description and test suite - the root cause analysis (`GetAndDelete` destroying the session on first read) is exactly right, and the e2e test that fails the provider once and then asserts the session survived is a genuinely good test to have written.

## Overview

The PR replaces the single-shot `GetAndDelete` session read with a retained session plus a per-offset chunk buffer, a CAS loop in `kvstore` for atomic chunk accumulation, a `Handled` short-circuit in the shared file-request path so non-final chunks never reach the provider, and session deletion moved to `PostCallback` after a successful finalize. The core fix and its lifecycle are right, and the tests are broad. What holds it back is that the change also introduces two hardcoded size ceilings that are tighter than what the gateway accepts today, which turns a bug fix into a regression for existing large-file uploads, plus a concurrency hole on the finalize path.

## Findings

|---|---|---|---|---|
| 1 | High | genai.go:359 (`RequestParser`, `len(body) > MaxResumableChunkSize`) | The new 10 MB per-chunk cap rejects the upload flow Google itself documents, which sends the whole file in one request. Regression for anyone uploading >10 MB today. | CONFIRMED |
| 2 | High | genai.go:41 (`MaxResumableUploadSize`) | 100 MB total cap is hardcoded and does not track the configurable `MaxRequestBodySizeMB`; Gemini itself allows 2 GB per file. | CONFIRMED |
| 3 | High | genai.go, converter finalize branch | Two concurrent `upload, finalize` requests with identical bytes both pass the CAS and both call the provider, creating two files at Gemini. | CONFIRMED |
| 4 | Medium | genai.go, CAS loop (`newSession.Chunks[offset] = append(...)`) | Every chunk request deep-copies the entire accumulated session; a 100 MB upload in 10 MB chunks copies ~450 MB, plus a full `sonic.Marshal` per CAS when a sync delegate is installed. | CONFIRMED |
| 5 | Medium | genai.go `PostCallback` (`if !hasFinalize` branch) | Dead code: the converter returns `Handled: true` for every non-finalize command, and `PostCallback` only runs from `handleFileRequest` (router.go:2057), which the `Handled` path returns before. | CONFIRMED |
| 6 | Medium | genai_test.go `TestFileUploadPostCallback_PreservesSessionForUploadOnly` | The test sets the command as a fasthttp header, but `PostCallback` reads `r.UploadCommand`, which the test leaves empty. It passes through the empty-command path, not the upload-only path. | CONFIRMED |
| 7 | Medium | genai.go, `normalizedCmd` allow-list | `X-Goog-Upload-Command: query` is now rejected. That command is how a client that lost its offset resumes, which is the scenario this PR is about. | PLAUSIBLE |
| 8 | Low | genai.go:377-386 (session created with `SetWithTTL(..., 1*time.Minute)`) and the CAS TTL | A 1-minute session TTL is very tight for interrupted-then-retried uploads with client backoff. | PLAUSIBLE |

### 1. The 10 MB per-chunk cap breaks the documented single-request upload (blocking)

Google's own Files API flow sends the entire file body in one request after `start`:

```bash
curl "${upload_url}" \
  -H "Content-Length: ${NUM_BYTES}" \
  -H "X-Goog-Upload-Offset: 0" \
  -H "X-Goog-Upload-Command: upload, finalize" \
  --data-binary "@${PDF_PATH}"
```

Source: https://ai.google.dev/gemini-api/docs/document-processing

Before this PR that request succeeded for any body the server accepted (`MaxRequestBodySizeMB`, default 100, `framework/configstore/tables/clientconfig.go:43`). After it, a 30 MB PDF - well inside both Gemini's 50 MB PDF limit and the gateway's default body limit - fails with `upload chunk exceeds maximum size of 10 MB`. Every existing user uploading between 10 MB and their configured body limit in a single shot breaks.

Concrete fix: derive both caps from the configured body-size limit rather than hardcoding them, e.g. pass `ClientConfig.MaxRequestBodySizeMB` into `CreateGenAIFileRouteConfigs` and use it as the per-chunk ceiling. At minimum, raise the per-chunk cap so it is never below what the server already accepted.

### 2. The 100 MB total cap is hardcoded and lower than the provider's limit (blocking)

The comment reads `100 MB, aligned with the maximum request body size`, but that limit is `MaxRequestBodySizeMB`, which is user-configurable (default 100). An operator who raises it to 500 to push large media through the gateway now hits a silent 100 MB ceiling with no config knob. Upstream, the Gemini Files API allows "2 GB" per file and 20 GB per project (https://ai.google.dev/gemini-api/docs/files), so this ceiling is a Bifrost-only restriction. Same fix as #1: make it configuration-driven, and if a hard ceiling is still wanted, document it in the schema.

### 3. Concurrent finalize can upload the file twice (blocking)

The CAS makes chunk accumulation atomic, but nothing guards the transition from "session complete" to "call the provider". Two `upload, finalize` requests carrying the same final chunk (a client retry racing a slow first attempt) both take the `existingChunk` identical-data branch, both CAS successfully, and both fall through to return a `FileRequest` with the assembled bytes. Result: two file resources created at Gemini, double storage and double billing, and the second `PostCallback` delete is a no-op. The PR's own `TestFileRequestConverter_ConcurrentRequestsSameUploadID` covers only the non-final `upload` command, so this path is untested.

Suggested fix: carry the finalize decision through the same CAS, e.g.

```go
if normalizedCmd == "upload,finalize" {
    if currentSession.Finalizing {
        return nil, errors.New("upload already being finalized")
    }
    newSession.Finalizing = true
}
```

so exactly one request wins the CAS and proceeds to the provider.

### 4. Memory and serialization cost grows quadratically with chunk count

Each chunk request copies the whole accumulated `Chunks` map before mutating it, then finalize builds another full copy in `assembledFile`, then the multipart encoder holds another. A 100 MB upload in 10 MB chunks does roughly 450 MB of copying in the CAS loop alone, and the copy happens before the validation error paths, so even a rejected chunk pays for it. If a `SyncDelegate` is installed (`framework/kvstore/kvstore.go:74`), `CompareAndSwap` additionally `sonic.Marshal`s the entire session on every chunk, and `[]byte` marshals as base64 (~1.33x), so the same upload serializes and gossips several hundred MB.

Because offsets are already validated as strictly sequential (`r.UploadOffset != newSession.NextOffset` is an error), the map plus `sort` plus reassembly is not buying anything a single append-only buffer would not. A `Data []byte` field with `append(session.Data, chunk...)`, and a `bytes.Equal` against the tail for the retry case, removes the map, the sort, the assemble pass, and most of the copying.

### 5. Dead branch and two parsers for one header

`PostCallback`'s `if !hasFinalize { ... "active" ... }` branch cannot run: the converter returns `Handled: true` for every non-finalize command, and `createHandler` returns before `handleFileRequest`, which is the only caller of `PostCallback` (`router.go:2057`). Alongside it, the same header is parsed two different ways - `hasGeminiUploadCommand` splits on commas and compares case-insensitively, while the converter builds `normalizedCmd` by lowercasing and stripping spaces against a two-value allow-list. They will disagree on inputs like `finalize` alone. Please delete the unreachable branch and keep a single parser.

### 7. `query` is rejected

The `X-Goog-Upload-*` protocol defines `X-Goog-Upload-Command: query` so that a client whose upload was interrupted can ask how many bytes the server received and resume from that offset (https://developers.google.com/photos/library/guides/resumable-uploads). Since resuming an interrupted upload is precisely what this PR enables, it would be worth handling `query` by returning `X-Goog-Upload-Status: active` plus `X-Goog-Upload-Size-Received: <NextOffset>`. The `active` response for a normal chunk does not carry `X-Goog-Upload-Size-Received` either, so a client currently has no way to discover the server's offset.

## Cleanups and conventions

- `sort.Slice(offsets, ...)`: `slices` is already imported at `genai.go:9`; `slices.Sort(offsets)` is the drop-in, and the new `sort` import goes away. If #4 is adopted the sort disappears entirely.
- `github.com/bytedance/gopkg/util/logger`: this import appears nowhere else in the repo, and the PR promotes the module to a direct dependency in `transports/go.mod`. It writes to a package-global logger that ignores Bifrost's configured level and destination. `GenericRouter` already carries a `schemas.Logger`; passing one into `CreateGenAIFileRouteConfigs` keeps these lines inside the gateway's logging.
- `fmt.Errorf("upload chunk exceeds maximum size of 10 MB")` has no format verbs; `errors.New` is the right call. The message also hardcodes the constant, while the sibling error derives it with `MaxResumableUploadSize/(1024*1024)`.
- Header parsing now happens in three places: `RequestParser`, `extractGeminiFileUploadParams` (PreCallback, which runs after RequestParser at `router.go:779`), and the converter's validation block. `extractGeminiFileUploadParams` sets `UploadOffset` without setting `HasUploadOffset`, so the two are only consistent by accident of ordering. Worth collapsing into one parse.
- `CompareAndSwap` is the only exported method in `framework/kvstore/kvstore.go` without a doc comment, and its comparison semantics are surprising for a general-purpose store: `reflect.DeepEqual` means two structurally identical but distinct values compare equal (classic ABA), and comparing a 100 MB value walks all of it on every attempt. Please document the semantics, and consider a monotonic version counter on the entry instead, which would make the primitive cheap and generally useful.
- Distributed note, not a request to fix here: the KV store is explicitly built for cross-node replication with last-write-wins by wall clock (see the `SyncDelegate` doc comment). `CompareAndSwap` is atomic only within one process, so two chunks of the same upload landing on different nodes will silently diverge. Single-shot uploads had the same node affinity requirement, but N chunks multiply the exposure. A comment at the session site stating that resumable uploads require sticky routing would save someone a bad afternoon.
- Test hygiene: `TestFileRequestConverter_AllowsUploadAtMaxSize` and `TestFileRequestConverter_AllowsMultipleChunksUpToMaxSize` are the same scenario with the same session, request and assertions - one can go. Several `t.Parallel()` tests allocate 90-100 MB buffers and then deep-copy them, so peak CI memory for this file is comfortably over 1 GB. Consider making the caps injectable so the limit tests can run with kilobyte-scale values.
- `ctx := &schemas.BifrostContext{}` in the size-limit tests versus `schemas.NewBifrostContext(...)` elsewhere - please use the constructor consistently.
- AGENTS.md, Testing: "Any change under `core/` that a client can observe on the wire must land together with a case in `tests/e2e/api/collections/provider-harness.json`". This PR touches `core/providers/gemini/types.go` and adds new wire-visible behaviour (new response headers, new error shapes). The narrow exemption is allowed, but the same section says "If a change is exempt, say so explicitly in the PR rather than leaving the omission unexplained" - so either add the harness case or state the exemption in the description.

## Merge recommendation

**Not yet - request changes.** Findings 1 and 2 turn a bug fix into a regression for anyone uploading more than 10 MB, which is the common case for the documented single-request flow, and finding 3 can duplicate a file at the provider. Findings 4 through 8 and the cleanups are not blockers.

## Followups

1. **In this PR (blocking)** - `genai.go` `RequestParser` chunk-size check and `MaxResumableChunkSize`: derive the per-chunk ceiling from `ClientConfig.MaxRequestBodySizeMB` so the documented single-request `upload, finalize` keeps working.
2. **In this PR (blocking)** - `genai.go` `MaxResumableUploadSize`: make the total ceiling configuration-driven rather than a hardcoded 100 MB, given Gemini allows 2 GB per file.
3. **In this PR (blocking)** - `genai.go` converter finalize branch: claim finalize inside the CAS so only one concurrent request reaches the provider, and add a concurrency test using `upload, finalize`.
4. **In this PR (blocking)** - `genai_test.go` `TestFileUploadPostCallback_PreservesSessionForUploadOnly`: set `UploadCommand: "upload"` on the request so the test exercises the branch it claims to.
5. **In this PR (non-blocking, small)** - `genai.go` `PostCallback`: delete the unreachable `!hasFinalize` branch and collapse `hasGeminiUploadCommand` and `normalizedCmd` into one parser; swap `sort.Slice` for `slices.Sort`; replace the `bytedance/gopkg` logger with the injected `schemas.Logger` and revert the `transports/go.mod` promotion.
6. **Follow-up PR** - `genai.go` session storage: replace the `Chunks` map with an append-only `Data []byte` to remove the per-chunk deep copy, the sort, and the assemble pass.
7. **Follow-up PR** - `genai.go`: support `X-Goog-Upload-Command: query` and emit `X-Goog-Upload-Size-Received` on `active` responses so interrupted clients can discover the server offset.
8. **Follow-up PR** - `framework/kvstore/kvstore.go`: document `CompareAndSwap`'s deep-equality semantics, or move to a version counter; revisit the 1-minute session TTL for resumable uploads.

## Checked and cleared

- *CAS always succeeds because it compares the same pointer* - refuted: `Get` returns the stored `any`, and after another goroutine swaps, the store holds a different session whose contents differ, so `reflect.DeepEqual` detects it.
- *`PostCallback` deletes the session even when the provider call failed* - refuted: `PostCallback` runs only after a successful provider response inside `handleFileRequest` (`router.go:2057`), and your e2e test asserts the session survives a provider 500.
- *KV decoder prefix mismatch breaks session decoding* - refuted: `generateUploadID` returns `"genai_upload_session:" + encoded`, which matches the prefix registered in `RegisterKVDecoders`.
- *Retrying the same offset double-counts `NextOffset`* - refuted: the existing-chunk branch returns without touching `NextOffset`, and your retry tests pin it.
- *The `Handled` path returns no status code and 500s* - refuted: fasthttp defaults to 200, and the e2e test asserts 200 with `X-Goog-Upload-Status: active`.
- *`r.FileData` aliases a reused fasthttp buffer* - refuted, and worth calling out as an improvement: this PR adds the explicit `append([]byte(nil), body...)` copy, which the retained-session design needs.

PR #6439, opened 2026-08-23
fix(anthropic): latch input_tokens from the strongest cache frame, not an independent max

wangkanai:
Follow-up to #6378, requested in @akshaydeo's review (followup 5).

> **Stacks on #6378.** Until that merges, the diff here also shows its commits. Merge #6378 first; this branch is `fix/passthrough-stream-usage-cache-double-count` + one commit.

## Problem

#6378 stops `input_tokens` being max-merged across cache-less and cache-aware events. But `input_tokens`, `cache_read_input_tokens` and `cache_creation_input_tokens` are still max-merged **independently**, so the reported prompt total can be assembled from values that never coexisted in a single frame.

A `message_start` carrying a prompt-scale `input_tokens` **together with a partially known `cache_read`** is exactly that shape — the frame #5510's hierarchical-cache explanation predicts, where the final split is not settled when `message_start` is emitted. It clears any authority test, and the independent max then keeps the inflated `input_tokens`:

```jsonc
{"type":"message_start","message":{"usage":{"input_tokens":173306,"cache_creation_input_tokens":0,"cache_read_input_tokens":5000,"output_tokens":0}}}
{"type":"message_delta","usage":{"input_tokens":250,"cache_creation_input_tokens":0,"cache_read_input_tokens":173056,"output_tokens":166}}
```

```
max(173306, 250) + max(5000, 173056) + 0 = 346362      // want 173306
```

`dev` reports 346362 here too, so this is a pre-existing limitation that #6378 narrows but does not close.

## Fix

Latch `input_tokens` from the single authoritative event with the **largest cache total**, instead of taking a max across events. The prompt total is then always `input + cache` from frames that agree.

Every case already in the table is reproduced unchanged — Qwen 8849, Kimi 173306, Kimi cache-aware 8931, MiniMax 8991, stock Anthropic 8849, the doc's web-search stream 10682 — and the merge stays order-independent.

### Tie-break

On equal cache totals the **smaller** `input_tokens` wins.

The review proposed the larger. I measured that and it selects the prompt-scale frame, which is the value this whole split exists to distrust: a breakdown-only `message_start` at `input_tokens 50000 / ephemeral_5m 49000` against a `message_delta` at `input_tokens 1000 / cache_creation_input_tokens 49000` ties at 49000 and reports **99000**. For a compliant upstream `input_tokens` does not vary across frames, so the tie-break never fires there.

Also note `cache_creation` is documented as the 5m/1h split **of** `cache_creation_input_tokens`, so it is a breakdown rather than an addend — it only contributes to the cache total when the top-level counter is absent.

## Tests

Two cases added, both order-independent like the rest of the table. Both **fail on #6378's head** and pass here:

| case | on #6378 | here |
|---|---:|---:|
| partially known cache on `message_start` | 346362 | **173306** |
| tie-break, breakdown-only vs top-level | 99000 | **50000** |

Full `providers/anthropic`, `providers/azure`, `providers/utils` and `schemas` suites pass.

## Not fixed here

Finding 2 from the same review — a cache-less event with a **larger** `input_tokens` arriving after an authoritative one is discarded (8931, where an unconditional max reports 9181). The review expected the tuple-latch to fix this as a side effect; measured, it does not, because the loose bucket is still discarded once any authoritative event exists. The latch changes *which* authoritative event wins, not whether a cache-less event may contribute. Deciding that a cache-less `input_tokens` can be added to an authoritative cache total would reintroduce precisely the composite-of-values-that-never-coexisted problem this PR removes, so #6378 pins the current answer as a test instead.


wangkanai:
Folded back into #6378 as `0ce8682`, together with a follow-on fix (`8e7f51f`) for a bug CodeRabbit found in the breakdown handling.

This PR only existed because the review offered to split it; both reviewers then asked for the latch in #6378 itself, so keeping it stacked on an unmerged branch added review friction without adding anything. Closing in favour of #6378, which now carries the whole fix.

PR #6449, opened 2026-08-23
feat: support env.VAR_NAME substitution for network_config.base_url

### LINKED ISSUE #6119: network_config.base_url doesn't support env.VARNAME substitution like key values do

amtulifra:
## Summary

`network_config.base_url` was a plain `string` with no env-var resolution, unlike `Key.Value` and other provider fields (Azure endpoint, Vertex project ID) that already support `"env.VAR_NAME"` via `SecretVar`. This left no way to keep `config.json` committed to version control while pointing `base_url` at an environment-specific value, short of an external templating wrapper (see #6119).

## Changes

- `NetworkConfig.UnmarshalJSON` (`core/schemas/provider.go`) now resolves an `"env.VAR_NAME"`-prefixed `base_url` against the environment at config-load time, via a new `resolveEnvString` helper.
- The prefix check is explicit (`strings.CutPrefix(value, "env.")`), so a URL that merely contains `env.` elsewhere (e.g. `https://host/env.path`) is left untouched — mirrors the existing `pricing_url` precedent in `ResolveFrameworkPricingConfig`.
- A missing environment variable fails the config load loudly (returns an error from `UnmarshalJSON`) rather than silently falling back to the provider's default endpoint — a wrong-but-quiet `base_url` is a worse failure mode than a blocked startup here.
- `BaseURL` stays a plain `string` (not `*SecretVar`): it has ~250 read sites across every provider package that treat it as a plain string, so converting the type would force `.GetValue()` everywhere for no real benefit.
- Documented the new `env.VAR_NAME` support in `docs/quickstart/gateway/provider-configuration.mdx` and in `transports/config.schema.json`'s `base_url` description.

### Design note
`config.json` is read-only at startup (never rewritten by Bifrost), so a user's `"env.MY_VAR"` reference stays intact in their git-tracked file. If a provider is later saved via the management API/UI, the DB-backed config store persists the *resolved* literal URL rather than the reference — this matches the existing accepted behavior of `pricing_url` (verified in `ResolveFrameworkPricingConfig`), not a new trade-off introduced here.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [x] Core (Go)
- [ ] Transports (HTTP)
- [ ] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [x] Docs

## How to test

```sh
cd core
go test ./schemas/... -run TestNetworkConfig -v
go test ./schemas/... ./providers/...
```

Manual check:
```sh
export MY_PROVIDER_BASE_URL="https://custom.internal/v1"
# in config.json: "network_config": { "base_url": "env.MY_PROVIDER_BASE_URL" }
```
Bifrost should resolve `base_url` to `https://custom.internal/v1` at startup; an unset env var should fail config load with a clear error naming the missing variable.

## Breaking changes

- [ ] Yes
- [x] No

Existing plain-string `base_url` values are unaffected — only values with the literal `"env."` prefix are resolved.

## Related issues

Closes #6119

## Security considerations

None beyond existing `env.*` resolution behavior already used for `Key.Value`, Azure endpoint, Vertex project ID, and `pricing_url` — no new secret-handling surface is introduced. Values are only read from the process's own environment via `os.LookupEnv`, exactly as those existing code paths do.

## Checklist


amtulifra:
Thanks for the review! Addressed in f796bf1a2:

1. **Malformed env var names** — fixed. `resolveEnvString` now validates the suffix against `^[A-Za-z_][A-Za-z0-9_]*$` instead of trimming whitespace and accepting anything non-empty. `env. FOO` and `env.FOO-BAR` now error instead of silently mangling the variable name. Added table-driven test cases covering whitespace, hyphens, dots, and leading digits.

2. **"at startup" wording** — fixed. `NetworkConfig.UnmarshalJSON` also runs for API-submitted provider payloads (`createProvider`/`updateProvider`), not just config.json parsing at process start, so the docs now say "at configuration load time."

3. **Schema `format: "uri"` rejecting `env.VAR_NAME`** — checked this empirically rather than assuming either way. `env.VLLM_BASE_URL` already passes schema validation against the current `format: "uri"` constraint, because this repo's `jsonschema` compiler setup (`transports/bifrost-http/lib/validator.go`) never calls `AssertFormat()`, so format is documentation-only here, not an enforced assertion — same reason the existing `pricing_url` field works today with the same `format: "uri"` + `env.*` combination. I added `TestValidateConfigSchema_NetworkConfig_BaseURL_EnvReference` as a permanent regression test proving this holds, rather than adding an `anyOf`/regex escape hatch to the schema for something that isn't actually broken.

amtulifra:
@coderabbitai review

amtulifra:
@coderabbitai full review

PR #6467, opened 2026-08-24
fix(responses): default streamed message content to an empty array

### LINKED ISSUE #6463: fix(responses): ensure streamed message items always include content array
## Problem

Some OpenAI-compatible Responses upstreams emit `response.output_item.added` for a message item without a `content` field. The OpenAI Responses wire shape requires a message `content` array; strict streaming clients reject the missing field before later deltas arrive.

## Expected behavior

For a streamed message item, default a missing `content` field to `[]` without mutating the source event.

## Proposed fix

Normalize only message items in `BifrostResponsesStreamResponse.WithDefaults`, alongside the existing default `in_progress` status, and add a schema-level regression test.

I have a small tested PR ready for this behavior.

Alex-wangyang:
## Summary

Closes #6463.

Normalizes a missing `content` field on streamed message `response.output_item.added` events to an empty array. The normalization copies the item first, so the source event is unchanged.

## Test plan

- `cd core && go test ./schemas -run TestBifrostResponsesOutputItemAddedDefaultsMessageContentToArray -count=1`

### DIFF
--- a/core/schemas/responses.go
+++ b/core/schemas/responses.go
@@ -3830,9 +3830,20 @@ func (resp *BifrostResponsesStreamResponse) WithDefaults() *BifrostResponsesStre
 	// Apply event-specific defaults
 	switch resp.Type {
 	case ResponsesStreamResponseTypeOutputItemAdded:
-		// Default item status to "in_progress"
-		if result.Item != nil && result.Item.Status == nil {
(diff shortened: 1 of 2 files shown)


Alex-wangyang:
Closing as a duplicate of #6456, which contains the same fix plus the required core changelog entry.

PR #6469, opened 2026-08-24
fix(responses): retry without provider-scoped reasoning item IDs

### LINKED ISSUE #6465: fix(responses): retry once without provider-scoped reasoning item IDs
## Problem

Some OpenAI-compatible Responses upstreams return reasoning items with `encrypted_content`, but issue server-side reasoning item IDs that cannot be replayed in a later request. The subsequent request fails with HTTP 400 such as `Referenced reasoning item ... was not found or has expired`, even though stateless encrypted reasoning remains usable.

## Expected behavior

On that narrow 400 error, Bifrost should retry exactly once after removing only the non-replayable reasoning-item IDs, preserving the reasoning summary and `encrypted_content`. The retry should not consume the configured retry budget or introduce a backoff because its payload changed.

## Proposed fix

Add a narrow detector, rewrite both typed and raw Responses input safely, and add regression tests for detection, raw passthrough, and end-to-end retry behavior. This follows the existing encrypted-reasoning fail-soft pattern.

I have a tested PR ready for this behavior.

Alex-wangyang:
## Summary

Closes #6465.

Some Responses-compatible providers issue reasoning IDs that are scoped to a prior response, while still returning stateless `encrypted_content`. If replaying the ID receives the narrow 400 `Referenced reasoning item ... was not found or has expired` response, retry exactly once after removing only that ID. The encrypted reasoning payload and visible summary are preserved.

The retry follows the existing encrypted-reasoning fail-soft behavior: it does not consume the configured retry budget and skips backoff because the payload changes. Both typed and raw-passthrough Responses input are handled.

## Test plan

- `cd core && go test . -run 'TestIsReasoningItemIDRejection|TestResponsesRetryDropsOnlyNonReplayableReasoningItemID|TestStripResponsesReasoningItemIDsRewritesRawPassthrough' -count=1`

### DIFF
--- a/core/bifrost.go
+++ b/core/bifrost.go
@@ -6081,11 +6081,17 @@ func executeRequestWithRetries[T any](
 	// Index in BifrostContextKeyAttemptTrail of an attempt that hit a rate limit and is waiting
 	// to learn whether the *next* key selection actually picks a different key. -1 = no pending.
 	pendingRotationAttemptIdx := -1
-	// Attempts granted outside the configured retry budget. Only the encrypted-reasoning
-	// fail-soft below adds one: MaxRetries defaults to 0, and a request that would
-	// otherwise die on a rejected replay deserves its one stripped attempt regardless of
-	// how the retry budget is tuned.
+	// Attempts granted outside the configured retry budget. The reasoning-replay
+	// fail-softs below add them: MaxRetries defaults to 0, and a request that would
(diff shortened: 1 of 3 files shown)


Alex-wangyang:
Closing as a duplicate of #6460, which contains the same fix plus the required core changelog entry and already has maintainer review.

PR #5892, opened 2026-08-06
fail soft on invalid_encrypted_content by stripping replayed reasoning and retrying once

akshaydeo:
## Summary

When a gateway rotates API keys across a multi-key pool, falls back to a different provider, or starts/stops proxying mid-session, replayed `encrypted_content` reasoning items become undecryptable at the new upstream identity. The upstream returns a 400 `invalid_encrypted_content` error, and no amount of retrying the same payload will fix it. This PR adds a fail-soft path that detects this specific rejection, strips `encrypted_content` from the offending reasoning items, and grants one extra retry attempt outside the normal retry budget so the conversation continues with summaries only rather than failing outright.

## Changes

- Added `isEncryptedReasoningRejection` to detect upstream 400s caused by unverifiable encrypted reasoning content, matching on the `invalid_encrypted_content` error code or message text as a fallback for older deployments that omit the code field.
- Added `stripResponsesEncryptedContent` to remove `encrypted_content` from all reasoning items in a Responses request. Reasoning items left with no summary and no content blocks are dropped entirely. Both typed input and raw body passthrough are handled; large payload mode (streamed body) is explicitly declined since the body cannot be rewritten.
- In `executeRequestWithRetries`, when a non-retryable error is identified as an encrypted reasoning rejection and the strip succeeds, an `extraAttempts` counter is incremented to grant one additional attempt beyond the configured `MaxRetries`. A `strippedEncryptedContent` flag ensures this fires at most once per request. A `lastWasEncryptedContentStrip` flag skips backoff for this retry since the payload changed and there is nothing to wait out.
- Retry log messages now report `MaxRetries+extraAttempts` as the budget ceiling so the counter never displays as `1/0`.
- Reasoning structs are cloned rather than mutated in place so plugins and the transport layer holding the same pointers are not affected.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [x] Core (Go)
- [ ] Transports (HTTP)
- [ ] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

```sh
go version
go test ./...
```

The new test file `core/encryptedreasoning_test.go` covers:
- The stripped retry succeeds and `encrypted_content` is absent on the second attempt while the summary and item ID are preserved.
- An upstream that keeps rejecting after the strip receives exactly 2 total attempts and then returns the error — no unbounded loop.
[...]

PR #6491, opened 2026-08-24
[fix]: retry without non-replayable reasoning item IDs

### LINKED ISSUE #6488: [Bug]: Replayed Responses reasoning item IDs can be non-replayable across turns
### Prerequisites

- [x] I searched existing issues and pull requests for duplicates.
- [x] I reproduced the behavior on the current `dev` branch.

### Description

Some OpenAI-compatible Responses providers return reasoning items with both `encrypted_content` and a server-side item `id`, but the `id` cannot be resolved when the client replays the item on a later turn. The upstream rejects the second request even though stateless replay through `encrypted_content` remains valid.

This was reproduced with Prime using a Responses-compatible OpenCode Go endpoint: the first turn returned a reasoning item and a tool call; the second turn replayed the reasoning item and failed with a 400.

### Steps to reproduce

1. Send a Responses request with `store:false` to a provider that returns a reasoning item containing `id` and `encrypted_content`.
2. Execute the returned tool call.
3. Replay the complete reasoning item alongside the tool result in the next request.
4. Observe the provider reject the reasoning item ID.

Protocol probes showed:

- replaying `id` plus `encrypted_content`: 400
- replaying the same item after removing only `id`: 200
- changing to `store:true`: still 400

### Expected behavior

When the upstream specifically reports that a referenced reasoning item was not found or expired, Bifrost should retry once on the same key after removing only the reasoning item IDs. It should preserve `encrypted_content`, summaries, and unrelated input fields.

### Actual behavior

Bifrost returns the upstream 400 directly, terminating otherwise valid multi-turn tool use.

### Affected area(s)

Core (Go)

### Version

Reproduced on `transports/v1.6.11`; the current `dev` branch (`0ac7f7142`) does not handle this error variant.

### Environment

```text
Client: Prime 0.7.1
API: Responses, multi-turn tool call
Provider shape: OpenAI-compatible Responses
Observed error: Referenced reasoning item ... was not found or has expired
```

### Severity

Alex-wangyang:
## Summary

Fail soft when a Responses upstream cannot resolve a replayed reasoning item ID but still accepts the item's stateless `encrypted_content`. Bifrost retries once on the same key after removing only the server-side handles, preserving encrypted reasoning, summaries, and unrelated input fields.

## Changes

- Detect the specific 400 error where a referenced reasoning item was not found or expired.
- Remove reasoning item IDs from typed Responses input without mutating caller-owned slices.
- Rewrite raw passthrough input while preserving unknown fields.
- Grant one immediate extra attempt outside the normal retry budget, guarded to run at most once.
- Add a routing-engine warning so the compatibility retry is auditable.
- Add predicate, typed retry, and raw passthrough regression tests.
- Add a core changelog entry.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [x] Core (Go)
- [ ] Transports (HTTP)
- [ ] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

```sh
cd core
go test . -run 'Test(IsReasoningItemIDRejection|ResponsesRetryDropsOnlyNonReplayableReasoningItemID|StripResponsesReasoningItemIDsRewritesRawPassthrough)$' -count=1
```

Expected: all three regression tests pass.

The full local `go test .` also reaches the unrelated `TestValidateExternalURLBoundsDNSLookup`, which fails in this macOS environment because DNS interception resolves the reserved `.invalid` hostname instead of returning an error. The focused tests and compilation pass.

## Screenshots/Recordings

Not applicable.

## Breaking changes

- [ ] Yes
- [x] No

## Related issues

Closes #6488

Related design precedent: #5892 strips rejected encrypted reasoning content and retries once. This change handles the neighboring case where only the item ID is non-replayable and the encrypted content remains valid.

## Security considerations

The retry preserves opaque encrypted reasoning content and never logs it. Only upstream-issued item IDs are removed after a specific rejection. The action is recorded without request content.

## Checklist

- [x] I read the current contribution guidance and followed it.
- [x] I added/updated tests where appropriate.
- [x] I updated the core changelog.
- [x] I verified the focused Go tests and compilation locally.
- [ ] I verified the complete repository CI pipeline locally.


Alex-wangyang:
Closing this duplicate in favor of canonical PR #6460, which is updated to the latest dev and is now mergeable.

PR #6500, opened 2026-08-25
fix: Anthropic Compaction Token Undercounting, Resulting In Big Output Token Count Miss

### LINKED ISSUE #6501: [Bug]: Anthropic Compaction Billing Gap
### Prerequisites

### Description

When Claude summarizes a long conversation (compaction), Anthropic charges for **the summary and the reply**. Bifrost only recorded the reply, so gateway cost and token logs were low compared with the Anthropic Console. We found a consistent gap between what Bifrost showed and what we were being charged. This appears to be the root cause of that gap.

### Steps to reproduce

It needs ~50k input because that is Anthropic’s lowest compaction trigger.

```python
import os
from anthropic import Anthropic

client = Anthropic(
    base_url=os.environ["BIFROST_BASE_URL"].rstrip("/") + "/anthropic",
    api_key=os.environ["BIFROST_API_KEY"],
)

blob = "Warehouse aisle snapshot with SKU, units, and location. " * 80
messages = []
for i in range(30):
    messages.append({"role": "user", "content": f"Turn {i}. {blob}"})
    messages.append({"role": "assistant", "content": f"Ack {i}. {blob}"})
messages.append({"role": "user", "content": "Reply with just the word ok."})

with client.beta.messages.stream(
    model="anthropic/claude-sonnet-5",
    max_tokens=256,
    betas=["compact-2026-01-12"],
    context_management={
        "edits": [{
            "type": "compact_20260112",
            "trigger": {"type": "input_tokens", "value": 50_000},
        }]
    },
    messages=messages,
) as stream:
    msg = stream.get_final_message()

u = msg.usage
iters = list(u.iterations or [])
print("top-level output_tokens:", u.output_tokens)
print("iterations:", [(i.type, i.output_tokens) for i in iters])
print("sum of iterations output:", sum(i.output_tokens for i in iters))
```

**Bug:** SSE / Bifrost log `output_tokens` equals the last pass only (the short `ok`). Anthropic Console equals the **sum**. You should see a `compaction` pass with extra output that never lands in the log.

**Fixed:** log `completion_tokens` matches `sum of iterations output` and the Console minute for that request.

### Expected behavior

Bifrost logs should match Anthropic logs, billing

### Actual behavior

They do not when compaction is used. 

### Affected area(s)

Core (Go)

### Version

v1.6.11

### Environment

```text

```

### Relevant logs/output

```shell

```

### Regression?

_No response_

### Severity

17jmumford:
## Summary
Resolve https://github.com/maximhq/bifrost/issues/6501. See Problem description below for details. 
- Add `billableAnthropicUsage()` to fold `usage.iterations` compaction-pass tokens into Anthropic billing totals when top-level usage only reflects the reply pass.
- Wire the normalizer into chat, responses, streaming accumulation, and passthrough usage paths while preserving raw `iterations` on the wire for fidelity.
- Add unit tests with replica token fixtures and a provider-harness case (`compaction-billing`) for end-to-end billing validation.

## Problem

Anthropic compaction responses can include `usage.iterations` with a separate `type: "compaction"` entry. Bifrost was billing from top-level usage only, undercounting input/output/cache tokens when a compaction pass ran before the message pass.

## Approach

Start from top-level usage (reply pass) and add token fields from compaction iterations only. Do not sum all iterations, to avoid double-counting server-side fallback declined attempts.

## Test plan

- [x] `cd core/providers/anthropic && GOWORK=off GOTOOLCHAIN=auto go test . -run 'BillableAnthropic|Compaction|StreamReplica|PassthroughStream_Compaction' -count=1`
- [x] `USE_INFISICAL=0 make run-provider-harness-test PROVIDER=anthropic FEATURE="compaction-billing"` (live, paid; requires `ANTHROPIC_API_KEY` and running Bifrost with `tests/integrations/python` config DB seeded)
- [x] `USE_INFISICAL=0 make test-core PROVIDER=anthropic` (optional broader live sweep)


Pratham-Mishra04:
Hey @TejasGhatte can you check this

core/providers/anthropic/types.go
```diff
@@ -1915,6 +1915,74 @@ type AnthropicOutputTokensDetails struct {
...
+const AnthropicUsageIterationTypeCompaction = "compaction"
+
+// billableAnthropicUsage returns the usage Anthropic actually charges for billing.
+// Top-level input/output/cache fields mirror the reply (message) pass only; the
+// compaction pass is listed separately under usage.iterations. When iterations
+// are present, fold compaction iteration token counts into the top-level fields.
+// Server-side fallback iterations are not added — top-level already mirrors the
+// billed serving attempt and declining attempts must not be summed. When
+// iterations are absent, top-level fields are already complete.
+func billableAnthropicUsage(u *AnthropicUsage) *AnthropicUsage {
```
17jmumford, line 1929: This is the bulk of the change. Testing seems to show some slight discrepancies still in how 5min vs 1hr cache tokens are counted, but overall billed tokens appears to be correct now, with compaction tokens being included. Huzzah!

akshaydeo (dismissed):

PR #6522, opened 2026-08-25
fix: exclude batch echo rows from missing-cost filter, prevent recalculation billing them, and enforce model allowlist for inline batch requests

sammaji:
## Summary

Batch `/results` calls that did not settle a batch were getting their own log rows treated as billable aggregate rows during cost recalculation, causing the batch to be billed once per `/results` fetch. This PR introduces an `Echo` marker on batch accounting rows to distinguish read-only display copies from the single settlement row that owns the bill, and extends governance model-allowlist enforcement to cover inline batch create requests.

## Changes

- **Echo marker on batch accounting rows**: A new `Echo bool` field on `BatchAccountingDebug` marks log rows written by `/results` calls that did not settle the batch. These rows carry a snapshot of the settled price for display but must never be billed.
- **`MissingCostOnly` filter exclusion**: The logstore query for missing-cost rows now excludes echo rows (`batch_debug NOT LIKE "%\"echo\":true%"`), since their NULL cost is final and no recalculation will ever fill it.
- **`batchRowRoleOf` replaces `isBatchAggregateRow`**: The classification function now returns one of three roles — `None`, `Aggregate`, or `Echo`. Rows without the echo marker are classified by whether their ID matches the deterministic aggregate ID derived from `(provider, batch_id)`, providing backward compatibility for rows written before the marker existed.
- **Echo rows reprice display-only**: `calculateBatchAggregateCost` accepts a `refreshSnapshotCost bool` parameter. When true, `Accounting.Cost` is updated so the displayed price stays current, but the row's `cost` column is left NULL via the new `batchDebugOnly` path in `persistRecalcOutcomes`.
- **Governance allowlist applied to inline batch create**: `PreLLMHook` now iterates every distinct model named across batch item bodies/params via `BatchCreateModels`, evaluating governance for each. `IsModelCheckedWhenPresent` is extracted into a shared utility and extended to include `BatchCreateRequest`, so the model allowlist applies whenever a model is present even if it is not required.
- **Anthropic batch route: mixed-model handling**: The Anthropic integration now tracks when items carry different models and sets the top-level model to nil rather than erroring, allowing mixed-model Anthropic batches while still rejecting mixed models for non-Anthropic providers.
- **`batchCreate` handler model extraction**: The handler now checks both `Body` and `Params` when inferring the model from the first batch item, matching the Anthropic integration's dual-shape awareness.
- **Test IDs use `AccountingLogID`**: Existing cost-fidelity tests now derive their log IDs from `batchaccounting.AccountingLogID` so the `batchRowRoleOf` classification correctly identifies them as aggregate rows.

## Type of change

- [x] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [x] Core (Go)
- [x] Transports (HTTP)
- [x] Providers/Integrations
- [x] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

```sh
go test ./core/schemas/... ./framework/logstore/... ./plugins/governance/... ./plugins/logging/... ./transports/bifrost-http/...
```

Key scenarios to validate:
- A batch that is fetched via `/results` multiple times produces exactly one billed row; subsequent fetch rows have `NULL` cost and `echo: true` in `batch_debug`.
- `MissingCostOnly` search does not return echo rows.
- After `RecalculateCosts`, echo rows have an updated `Accounting.Cost` snapshot but their `cost` column remains `NULL`.
- A `BatchCreateRequest` with a model on the virtual key's disallowed list is rejected by governance.
- A `BatchCreateRequest` with no model passes governance without restriction.

## Breaking changes

- [ ] Yes
- [x] No

## Security considerations

None beyond the governance enforcement fix, which tightens model allowlist checks to cover inline batch requests that were previously bypassing them.

## Checklist

### DIFF
--- a/plugins/governance/main.go
+++ b/plugins/governance/main.go
@@ -1052,6 +1052,21 @@ func (p *GovernancePlugin) PreLLMHook(ctx *schemas.BifrostContext, req *schemas.
 		Model:      model,
 		UserID:     userID,
 	}
+	// A batch create fans out to many completions, each naming its own model, so
+	// every model it will run is evaluated before the request itself. Every check
+	// reached from here is read-only, so the extra passes cannot double-count usage.
+	if req.RequestType == schemas.BatchCreateRequest && req.BatchCreateRequest != nil && len(req.BatchCreateRequest.Requests) > 0 {
+		for _, batchModel := range BatchCreateModels(req, model) {
+			batchEvaluationRequest := *evaluationRequest
+			batchEvaluationRequest.Model = batchModel
+			_, bifrostError := p.EvaluateGovernanceRequest(ctx, &batchEvaluationRequest, req.RequestType)
+			if bifrostError != nil {
+				return req, &schemas.LLMPluginShortCircuit{
+					Error: bifrostError,
+				}, nil
+			}
+		}
+	}
 	// Evaluate governance using common function
 	_, bifrostError := p.EvaluateGovernanceRequest(ctx, evaluationRequest, req.RequestType)
 	// Convert BifrostError to LLMPluginShortCircuit if needed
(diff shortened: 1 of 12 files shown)


sammaji:
* **#6522** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/6522?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/maximhq/bifrost/6522?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `dev`


<h2></h2>

plugins/governance/main.go
```diff
@@ -1053,17 +1053,60 @@ func (p *GovernancePlugin) PreLLMHook(ctx *schemas.BifrostContext, req *schemas.
...
 	}
 	// Evaluate governance using common function
-	_, bifrostError := p.EvaluateGovernanceRequest(ctx, evaluationRequest, req.RequestType)
-	// Convert BifrostError to LLMPluginShortCircuit if needed
-	if bifrostError != nil {
-		return req, &schemas.LLMPluginShortCircuit{
-			Error: bifrostError,
-		}, nil
+	models := []string{model}
+	if req.RequestType == schemas.BatchCreateRequest && req.BatchCreateRequest != nil && len(req.BatchCreateRequest.Requests) > 0 {
```
Pratham-Mishra04, line 1058: have a separate branch for this - pure additive change

Pratham-Mishra04 (approved):

Pratham-Mishra04:
### Merge activity

* **Aug 25, 4:35 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/6522).
* **Aug 25, 4:39 PM UTC**: @Pratham-Mishra04 merged this pull request with [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/6522).

PR #6528, opened 2026-08-25
fix: harden tool results and weighted key selection

hbmartin:
## Summary

- validate Bedrock tool-message metadata before reading optional content
- preserve content-less tool results instead of panicking or dropping them
- make weighted key selection safe for empty, zero, tiny, and negative weights
- add focused regression coverage for malformed tool messages and key-selection edge cases

## Testing

- `go test ./keyselectors ./providers/bedrock`
- `git diff --check`

### DIFF
--- a/core/keyselectors/weightedrandom.go
+++ b/core/keyselectors/weightedrandom.go
@@ -1,20 +1,28 @@
 package keyselectors
 
 import (
+	"fmt"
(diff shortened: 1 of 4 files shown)


hbmartin:
Closing and reopening against the fork repository as requested.

PR #6533, opened 2026-08-26
feat: apply pricing schedules in cost engine

qixiangyang:
## Summary
- persist provider-generic `pricing_schedule` JSON with each model pricing row and load it into the runtime pricing store
- apply time-based pricing multipliers in the cost engine using the authoritative attempt-start timestamp
- compose schedules with the final resolved service/context tier rate rather than replacing tier pricing
- expose explicit-billing-time cost APIs for bare usage paths, including guardrail judge calls and failed/cancelled requests
- report schedule evaluation diagnostics and prevent mutation of provider-supplied costs

## Semantics
- Pricing time is the start of the provider attempt, not completion time or log time.
- Schedule multipliers apply after base, service-tier, context-tier, and flat per-request pricing resolution.
- A missing attempt timestamp intentionally falls back to base pricing and is reported through `pricing_schedule.timestamp_available`.
- Provider-reported costs are trusted and are not rescaled.

## Tests
- `cd framework && GOTOOLCHAIN=go1.26.6 go test ./modelcatalog/... ./configstore ./configstore/tables -count=1`
- `cd plugins/governance && GOTOOLCHAIN=go1.26.6 go test ./...`
- `cd plugins/logging && GOTOOLCHAIN=go1.26.6 go test ./...`

Depends on #6514 and #6516.


qixiangyang:
Thanks @coderabbitai — reviewed each point and addressed the valid functional findings in `e30a2d829`:

- **Schedule lookup candidate precedence**: agreed. `applyPricingSchedule` now uses the same precedence as `resolvePricing` (`ServerSideFallbackModel -> alias canonical model name -> alias model ID -> routing model`) and stops at the first model with a schedule. Added an alias/canonical-model regression test.
- **Cross-midnight weekday semantics**: agreed. Synced the hardened schedule implementation and tests from #6516, including wrapped-window weekday handling, validation, and clock parsing.
- **Streaming attempt timestamp ownership**: agreed. Chat/Responses accumulation now retains the `BillingAttemptStartedAt` value from the highest chunk index carrying it, rather than the last arrival. `ImageStreamChunk` now carries the timestamp, the image final chunk propagates it, image accumulation uses the same highest-index rule, and pooled chunks reset it. Added regression tests.
- **Governance `go.mod` dependency**: not changing it here. This monorepo builds modules through the root `go.work`, so governance resolves the local framework source in CI. The new framework API has no released tag yet; bumping `plugins/governance/go.mod` to an unavailable version would be less safe. Version bumps are handled by the repository's release/module-sync process after the framework API ships. Happy to update if maintainers prefer an explicit pseudo-version before merge.

qixiangyang:
Synced the non-finite multiplier validation fix from #6516 so the stacked branch remains consistent. Datasheet tests pass.

qixiangyang:
Valid catch — fixed. The image stream error path now copies `BifrostError.ExtraFields.BillingAttemptStartedAt` into the accumulated chunk, so failed/cancelled image attempts retain the same billing timestamp. Added an error-path regression test; `framework/streaming` passes.

qixiangyang:
Addressed in 927010209: when post-hooks recover a stream error into a response, the response now receives the provider attempt start timestamp. The branch was also rebased on the updated #6514 base.

qixiangyang:
Rebased on a55127b16 from #6514 so this PR inherits the nil-context streaming cancellation fix.

qixiangyang:
Addressed in 57bdd7568: moved the stream lifecycle helper out of  into a dedicated , renamed it to , and kept its semantics unchanged: cancel after producer close, close the forwarded channel, drain the producer when the owned context is already done, and leave caller-owned contexts untouched. Core tests pass.

qixiangyang:
Addressed in 57bdd7568: moved the stream lifecycle helper out of `core/bifrost.go` into dedicated `core/stream_context.go`, renamed it to `cancelOwnedContextAfterStream`, and kept its semantics unchanged: cancel after producer close, close the forwarded channel, drain the producer when the owned context is already done, and leave caller-owned contexts untouched. Core tests pass.

qixiangyang:
Addressed in 0697a8c57: the stream loop now watches `ctx.Done()` even while the producer is idle, then drains until producer close before running deferred cleanup. The cancellation regression test also keeps the producer blocked until cancellation so the cancellation path is exercised deterministically. Core tests pass.

qixiangyang:
Addressed in `0a3bcb1ef`: the producer now waits for `ctx.Done()` while the input stream is idle before sending the second chunk, so the test deterministically exercises the outer cancellation branch in `cancelOwnedContextAfterStream`. Full `core` tests and `git diff --check` pass.

qixiangyang:
Rebased on the updated schedule-config base so this PR inherits the streaming context lifecycle fix from #6514. Core tests and `git diff --check` pass.

qixiangyang:
Rebased on the updated billing-attempt base to inherit fallback timestamp resets, replacement-error restamping, nil-sentinel clearing, and public unary context isolation fixes. Core tests and `git diff --check` pass.

PR #6534, opened 2026-08-26
feat: configure model pricing schedules

qixiangyang:
## Summary

Adds provider-generic configuration support for time-based pricing schedules and wires it into the model catalog management API/UI.

- Adds `pricing_schedule` to model pricing persistence and the model catalog management API.
- Supports `pricing_schedule` and `clear_pricing_schedule` in `PUT /api/models/catalog`.
- Preserves omitted fields: schedule-only updates no longer rewrite `additional_attributes`, and attribute-only updates no longer clear schedules.
- Adds a generic schedule editor to Model Catalog for timezone, calendar, weekday rules, time windows, and multipliers.
- Documents DeepSeek V4's weekday peak windows: Monday–Friday 09:00–12:00 and 14:00–18:00 Asia/Shanghai; weekends are off-peak all day.
- Updates provider docs, architecture docs, and OpenAPI.

Stacked on #6533.

## Testing

```sh
cd framework
GOTOOLCHAIN=go1.26.6 go test ./modelcatalog/... ./configstore ./configstore/tables -count=1

cd ../transports
GOTOOLCHAIN=go1.26.6 go test ./bifrost-http/server ./bifrost-http/handlers -count=1

cd ../ui
./node_modules/.bin/tsc --noEmit -p tsconfig.json
```

Go tests pass. TypeScript has only the existing baseline errors listed in the PR discussion (unrelated layout route typing errors).


qixiangyang:
Synced the latest #6533 fix (`e30a2d829`) into this stacked PR. This keeps schedule model-candidate resolution and billing-attempt streaming propagation consistent across the stack. Framework, core, logging/governance plugin, and HTTP handler tests pass.

qixiangyang:
Synced the non-finite multiplier validation fix from #6516 through #6533. Datasheet tests pass.

qixiangyang:
Synced the image-stream error timestamp fix from #6533. Streaming tests pass.

qixiangyang:
Thanks @coderabbitai — all six points were valid. Pushed `dc0283d81`:

- `toggleScheduleDay` now updates only the matching schedule row.
- DeepSeek docs now describe the announced schedule as forthcoming and use the published daily windows rather than weekday-only behavior.
- All three transport Dockerfiles build with the `time/tzdata` tag.
- `parsePricingClockMinutes` now requires every `HH`/`MM` character to be an ASCII digit.
- Pricing schedules are indexed and resolved by provider + model, with regression coverage for two providers sharing a model name.
- Both OpenAPI representations retain the two properties and enforce mutual exclusion via `oneOf`.

Validated with framework modelcatalog/streaming/logstore tests, core and core/schemas tests, logging and governance plugin tests, Prettier, and `git diff --check`.

qixiangyang:
Good catch — pushed `f95a1cc25`. The three image builds now use the correct `timetzdata` build tag rather than `time/tzdata`. This keeps timezone data embedded without relying on the runtime image providing zoneinfo.

qixiangyang:
Addressed in 33cdcf1ec: errors returned or replaced by post-hooks are restamped with the provider attempt start time on both non-streaming and streaming paths. The branch also inherits the nil-context isolation fix from #6514 and is rebased on the updated #6533.

qixiangyang:
Rebased on 60b8c469d from #6533 so this PR inherits the nil-context streaming cancellation fix from #6514.

qixiangyang:
Rebased on 57bdd7568 from #6533 to inherit the stream-context helper relocation and current billing-attempt base.

qixiangyang:
Addressed in `879b9a45f`: successful non-streaming responses returned by post-hooks are restamped with the provider attempt start time after `RunPostLLMHooks`. Added an integration-style regression test that drives a real chat request through `tryRequest`, has the post-hook replace the provider response, and verifies `BillingAttemptStartedAt` on the final response. Full `core` tests and `git diff --check` pass.

qixiangyang:
Rebased on updated `origin/dev` and the deterministic cancellation-test fix from #6533. Core tests and `git diff --check` still pass.

qixiangyang:
Also addressed the follow-up in `b65c3d9b4`: recovered non-streaming responses are stamped with `BillingAttemptStartedAt` before metadata population. Full `core` tests and `git diff --check` pass.

qixiangyang:
The latest streaming context concern is valid: a returned stream can outlive `handleStreamRequest`, and an internally owned context has no caller-accessible cancel handle. This is now fixed at the base PR (#6514) by requiring callers to provide a non-nil, cancelable context for streaming requests; abandoned consumers can therefore cancel explicitly. The now-unused owned-context wrapper was removed. Rebased this PR on the updated stack. Core tests and `git diff --check` pass.

qixiangyang:
Rebased on the updated stack to inherit fallback timestamp resets, replacement-error restamping, nil-sentinel clearing, public unary context isolation, and the streaming context contract fixes. Core tests and `git diff --check` pass.

impoiler:
Hi @qixiangyang, same as the other PR:

We really appreciate the effort you put into this task. I completed it internally to avoid back-and-forth on the fields and dependencies in the datasheet. The feature is now available in v2.2.1, so I’m closing this PR.

Again, thank you very much for your contribution and effort.

PR #6536, opened 2026-08-26
fix: skip key selection on anthropic provider

TejasGhatte:
## Summary

`SkipKeySelection` was previously allowed for any provider that wasn't Azure, Bedrock, BedrockMantle, or Vertex. This meant that if a governance routing rule rewrote the provider/model after the Claude Code OAuth transport set the flag, a non-Anthropic provider (e.g. Fireworks) would skip key selection entirely — leaving it without a configured key and causing request construction to fall back to the OpenAI schema, which breaks providers like Fireworks that require `max_tokens`.

The fix tightens `isKeySkippingAllowed` to an allowlist of exactly one provider (`Anthropic`), since `SkipKeySelection` exists solely for Claude Code OAuth passthrough where the caller's token is the upstream credential and only the Anthropic provider forwards it.

## Changes

- `isKeySkippingAllowed` now returns `true` only for `schemas.Anthropic`, replacing the previous denylist approach. This ensures non-Anthropic providers always receive a configured key from the pool.
- The `selectKeyFromProviderForModelWithPool` call site now passes `baseProviderType` instead of `providerKey` to `isKeySkippingAllowed`, so the gate is evaluated against the resolved base provider.
- `SkipKeySelection` is intentionally **not** cleared in `clearAnthropicPassthroughForNonNativeProvider` — it also drives `IsClaudeCodeMaxMode`, which suppresses `x-api-key` on the Anthropic provider. Clearing it during a non-native attempt would cause an Anthropic fallback to send the account key alongside the caller's OAuth token. The flag is gated at the read site instead.
- A new test `TestSelectKeyFromProviderForModelWithPool_SkipKeySelectionGatedOnBaseProvider` verifies that Anthropic skips key selection while Fireworks (with `UseAnthropicEndpoints`) still selects its own key.
- `TestClearAnthropicPassthroughForNonNativeProvider` is updated to assert that `SkipKeySelection` survives the clear operation and documents why.

## Type of change

- [x] Bug fix

## Affected areas

- [x] Core (Go)
- [x] Providers/Integrations

## How to test

```sh
go test ./core/... -run TestSelectKeyFromProviderForModelWithPool_SkipKeySelectionGatedOnBaseProvider
go test ./core/... -run TestClearAnthropicPassthroughForNonNativeProvider
go test ./...
```

The new test covers the regression directly: with `SkipKeySelection` set and a Fireworks provider, the selected key must be present and must have `UseAnthropicEndpoints = true`. Without this fix, the key pool would be empty and that assertion would fail.

## Breaking changes

- [ ] Yes
- [x] No

## Security considerations

`SkipKeySelection` bypasses key injection entirely, relying on the caller's OAuth token as the upstream credential. Tightening the allowlist to Anthropic-only reduces the surface where a misconfigured or rewritten routing rule could cause a request to be sent without any credential, or with the wrong credential type for the target provider.

## Checklist

### DIFF
--- a/core/bifrost.go
+++ b/core/bifrost.go
@@ -8771,7 +8771,7 @@ func (bifrost *Bifrost) getKeysForBatchAndFileOps(ctx *schemas.BifrostContext, p
 // via the keyProvider closure built by the caller.
 //
 // canRotate=false is returned for cases where the caller must always use the same key:
-//   - SkipKeySelection (provider allows keyless requests; empty slice returned)
+//   - SkipKeySelection (Claude Code OAuth passthrough to Anthropic; empty slice returned)
 //   - Explicit BifrostContextKeyAPIKeyID / APIKeyName (user pinned a specific key)
 //   - Session stickiness (key persisted in KV store for the session lifetime)
 //   - Single-key pool (only one eligible key — rotation is a no-op, KV write skipped)
@@ -8786,8 +8786,8 @@ func (bifrost *Bifrost) selectKeyFromProviderForModelWithPool(ctx *schemas.Bifro
 		}
 	}
 
-	// SkipKeySelection: provider allows keyless requests — return empty pool, no rotation.
-	if skipKeySelection, ok := ctx.Value(schemas.BifrostContextKeySkipKeySelection).(bool); ok && skipKeySelection && isKeySkippingAllowed(providerKey) {
+	// SkipKeySelection: the caller's OAuth token is the credential — return empty pool, no rotation.
+	if skipKeySelection, ok := ctx.Value(schemas.BifrostContextKeySkipKeySelection).(bool); ok && skipKeySelection && isKeySkippingAllowed(baseProviderType) {
 		return []schemas.Key{}, false, nil
 	}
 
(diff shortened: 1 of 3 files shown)


TejasGhatte:
* **#6536** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/6536?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/maximhq/bifrost/6536?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `dev`


<h2></h2>

Pratham-Mishra04 (approved):

Pratham-Mishra04:
### Merge activity

* **Aug 26, 7:19 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/6536).
* **Aug 26, 7:20 AM UTC**: @Pratham-Mishra04 merged this pull request with [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/6536).

PR #6735, opened 2026-09-02
fix(core): resolve env.-indirected network_config.base_url

valentinyanakiev:
## Summary

`network_config.base_url` sits alongside config fields that already accept a `SecretVar` reference (`ca_cert_pem`, `proxy.url`, the Ollama/SGL key URLs), but a `"base_url": "env.SOME_VAR"` was never resolved: it was passed through and dialed as a literal hostname. Requests to that provider fail to connect with no diagnostic pointing at the real problem.

This makes `base_url` a `SecretVar` like the other reference-capable fields, so an upstream URL can be kept out of committed config the same way credentials already can.

## Changes

- `NetworkConfig.BaseURL` is now `*schemas.SecretVar` (`json:"base_url"`), the same shape as `CACertPEM` and `ProxyConfig.URL`. `SecretVar.UnmarshalJSON` handles plain URLs, `env.`, and `vault.` references; `MarshalJSON` emits `SecretVarAsString(nc.BaseURL)`, so config-store persistence, `ProviderConfig` hashing, and API responses stay reference-shaped. No shadow field.
- Every read site goes through `BaseURL.GetValue()`: all providers (including `databricks` and `github-copilot`, which landed on `dev` after this PR was opened), the OpenAI/ElevenLabs realtime paths, and the provider HTTP handler's `ValidateExternalURL` check.
- The per-provider default-URL / trailing-slash block in every constructor is folded into one `providerUtils.NormalizeBaseURL(&config.NetworkConfig, defaultURL)` helper (`""` for ollama, sgl, vllm, databricks, and github-copilot, whose base URL is optional). It clones the `SecretVar` before trimming, so a pointer shared with a config-store copy is never mutated in place, and the `env.`/`vault.` reference survives normalization.
- Fail-loud at decode on a reference that resolves to an empty value: `network_config.base_url references "env.X" but it resolved to an empty value`. Same fail-closed stance `ConfigureTLS` takes for `ca_cert_pem`; the alternative is dialing an empty host. Surrounding whitespace in a resolved value is trimmed.
- The ollama/sgl key-backfill migration copies the `SecretVar` into the key URL (reference preserved) instead of wrapping the resolved string.
- `config.schema.json`: `base_url` documents the reference form and drops `"format": "uri"`, which an `env.` reference would not satisfy. The wire form is still a string, so the UI types are unaffected.
- One line in `docs/quickstart/gateway/provider-configuration.mdx`.
- Review follow-ups (third commit): `NetworkConfig.Redacted` clones `BaseURL` and masks the resolved value when it came from a reference (literal URLs stay readable for the UI); `providerUtils.LoggableURL` keeps a reference-resolved base URL out of the four Gemini batch debug logs.
- UI (second commit): the provider network form, the provider-form schema, and the custom-provider sheet previously rejected anything that was not a URL. One shared `isValidBaseURL` predicate (http(s) URL or a non-empty `env.`/`vault.` reference) now backs all four validation sites via `baseURLSchema`, and the base URL inputs describe the reference form. Wire shape unchanged.

Revision history: v1 hand-rolled `os.LookupEnv` prefix handling; v2 resolved through `SecretVar` but kept `BaseURL` a `string` with a hidden retained reference (reviewer: a workaround); v3 (this) converts the field itself and updates the callers, per review.

## Type of change

- [x] Bug fix

## Affected areas

- [x] Core (Go)
- [x] Framework (configstore migration)
- [x] Transports (provider handler, `config.schema.json`)
- [x] Docs
- [x] UI

## How to test

```sh
# in core/, framework/, transports/ (go.work via `make setup-workspace`)
go build ./... && go vet ./...

cd core && go test -count=1 ./schemas/ ./providers/...
cd transports && go test -count=1 ./bifrost-http/lib/
cd framework && go test -count=1 ./configstore/

# UI
cd ui && npm ci --ignore-scripts && npx vitest run lib/utils/validation.test.ts
```

- `TestNetworkConfig_BaseURLSecretRef` (core/schemas): literal passthrough and round-trip; absent stays nil and is omitted; `env.` resolves, keeps its reference, and round-trips as the reference (never the resolved URL); whitespace trimmed; unset, empty, whitespace-only `env.` and unresolvable `vault.` fail loud.
- `isSecretReference` / `isValidBaseURL` (ui/lib/utils/validation.test.ts): accepts http(s) URLs and `env.`/`vault.` references; rejects bare hosts, non-http schemes, empty references, and `https://env.example.com`-style lookalikes.
- `TestNetworkConfig_Redacted_BaseURL` (core/schemas) and `TestLoggableURL` (core/providers/utils): redacted copies clone and mask reference-resolved base URLs, and debug-log rendering hides the resolved host.
- `TestNormalizeBaseURL` (core/providers/utils): default only when unset; trailing slashes trimmed; `env.` reference retained; caller's `SecretVar` cloned, not mutated; no-op with no default.

Verified locally on Go 1.27 (toolchain auto) against current `dev` @ `03ab39186`: `core`, `framework`, `transports`, and all `plugins/*` build and vet clean; `core` schemas/providers tests, `framework/configstore` tests, and `transports/bifrost-http/lib` tests pass, with one exception: `TestGenerateMCPClientHash_RuntimeVsMigrationParity` fails identically on a clean `dev` checkout and is unrelated to this change. `validate-schema-sync.sh` reports the same three errors on `dev` and on this branch. UI: `vitest` passes for the touched file (the one failing test file in the suite, `logs/views/columns.test.ts`, fails identically on `dev`); `tsc --noEmit` reports no errors in the touched files (the 97 route-typing errors it reports are identical on `dev`); `oxlint` and `oxfmt --check` are clean for the touched code.

## Breaking changes

- [x] Yes, for Go consumers of the `core` module only

`NetworkConfig.BaseURL` changes type from `string` to `*schemas.SecretVar`; downstream Go code reads it with `GetValue()` and sets it with `schemas.NewSecretVar(...)`. The JSON wire form (`config.json`, config store, HTTP API) is unchanged: a plain URL string still works, and `env.`/`vault.` strings that were previously dialed literally now resolve.

## Related issues

Found while migrating a downstream deployment ([LaneTally](https://github.com/feedback-loop-ai/lanetally)) from `transports/v1.5.16` to `transports/v2.0.0`.

## Security considerations

Lets operators keep upstream endpoints out of committed config and out of the config store, using the same reference mechanism already trusted for credentials. The resolved URL never appears in marshaled output while the reference still resolves to it.

## Checklist

### DIFF
--- a/core/schemas/provider.go
+++ b/core/schemas/provider.go
@@ -242,7 +242,11 @@ func (nc NetworkConfig) MarshalJSON() ([]byte, error) {
 	return json.Marshal(alias)
 }
 
-// Redacted returns a redacted copy of the network configuration with CACertPEM masked.
+// Redacted returns a redacted copy of the network configuration: CACertPEM is masked,
+// and a BaseURL that resolved from an env./vault. reference has its resolved value
+// masked too, so the copy exposes only the reference. A literal base_url is left
+// readable, since it is not a secret and the UI displays it. Both SecretVars are
+// cloned so the copy never shares pointers with the original.
 func (nc *NetworkConfig) Redacted() *NetworkConfig {
 	if nc == nil {
 		return nil
@@ -250,6 +254,13 @@ func (nc *NetworkConfig) Redacted() *NetworkConfig {
 	redacted := *nc
 	if nc.CACertPEM != nil && nc.CACertPEM.IsSet() {
 		redacted.CACertPEM = nc.CACertPEM.Redacted()
+	}
+	if nc.BaseURL != nil {
+		if nc.BaseURL.IsFromSecret() {
+			redacted.BaseURL = nc.BaseURL.Redacted()
+		} else {
+			redacted.BaseURL = nc.BaseURL.Clone()
+		}
 	}
 	return &redacted
 }
--- a/core/schemas/provider.go
+++ b/core/schemas/provider.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"maps"
+	"strings"
 	"time"
 )
 
@@ -58,8 +59,10 @@ const (
 //   - Duration string: "500ms", "5s", "1m" — parsed via time.ParseDuration (preferred)
 //   - Integer: treated as milliseconds (legacy format, e.g. 500 means 500ms)
 type NetworkConfig struct {
-	// BaseURL is supported for OpenAI, Anthropic, Cohere, Mistral, and Ollama providers (required for Ollama)
-	BaseURL                        string            `json:"base_url,omitempty"`                       // Base URL for the provider (optional)
+	// BaseURL is supported for OpenAI, Anthropic, Cohere, Mistral, and Ollama providers (required for Ollama).
+	// Accepts a plain URL or a SecretVar reference ("env.VAR" / "vault.path"), like CACertPEM and
+	// ProxyConfig.URL; read the dialable URL with BaseURL.GetValue().
+	BaseURL                        *SecretVar        `json:"base_url,omitempty"`                       // Base URL for the provider (optional, supports env.*)
 	ExtraHeaders                   map[string]string `json:"extra_headers,omitempty"`                  // Additional headers to include in requests (optional)
 	DefaultRequestTimeoutInSeconds int               `json:"default_request_timeout_in_seconds"`       // Default timeout for requests
 	MaxRetries                     int               `json:"max_retries"`                              // Maximum number of retries
@@ -76,7 +79,9 @@ type NetworkConfig struct {
 	AllowPrivateNetwork            bool              `json:"allow_private_network,omitempty"`          // Allow connections to RFC 1918 private IPs (for k8s pods, LAN deployments). Link-local (169.254.x.x) is always blocked.
 }
 
-// UnmarshalJSON customizes JSON unmarshaling for NetworkConfig.
+// UnmarshalJSON customizes JSON unmarshaling for NetworkConfig. base_url accepts
+// a plain URL or a SecretVar reference ("env.VAR" / "vault.path"); a reference that
+// resolves to an empty value is an error.
 //
 // RetryBackoffInitial and RetryBackoffMax accept two formats:
 //   - Duration string (preferred): "500ms", "5s", "1m" — parsed via time.ParseDuration
@@ -85,7 +90,7 @@ type NetworkConfig struct {
 func (nc *NetworkConfig) UnmarshalJSON(data []byte) error {
 	// Use an alias type to avoid infinite recursion
 	type NetworkConfigAlias struct {
-		BaseURL                        string            `json:"base_url,omitempty"`
+		BaseURL                        *SecretVar        `json:"base_url,omitempty"`
 		ExtraHeaders                   map[string]string `json:"extra_headers,omitempty"`
 		DefaultRequestTimeoutInSeconds int               `json:"default_request_timeout_in_seconds"`
 		MaxRetries                     int               `json:"max_retries"`
@@ -108,6 +113,21 @@ func (nc *NetworkConfig) UnmarshalJSON(data []byte) error {
 	}
 
 	// Copy all non-duration fields
+	// base_url accepts the same SecretVar reference forms ("env.VAR", "vault.path") already
+	// supported for ca_cert_pem, proxy.url, and the Ollama/SGL key URLs, so an upstream URL
+	// never has to be committed in plaintext. SecretVar.UnmarshalJSON performs the lookup and
+	// retains the reference, so MarshalJSON round-trips "env.VAR" rather than the resolved URL.
+	//
+	// A reference that resolves to an empty value is a fail-loud config error rather than a
+	// silent empty string (the fail-closed stance ConfigureTLS takes for ca_cert_pem): a
+	// misconfigured Key fails closed upstream with a 401, but a misconfigured base_url would
+	// otherwise be dialed as an empty host.
+	if alias.BaseURL != nil && alias.BaseURL.IsFromSecret() {
+		alias.BaseURL.Val = strings.TrimSpace(alias.BaseURL.Val)
+		if alias.BaseURL.Val == "" {
+			return fmt.Errorf("network_config.base_url references %q but it resolved to an empty value", alias.BaseURL.GetRawRef())
+		}
+	}
 	nc.BaseURL = alias.BaseURL
 	nc.ExtraHeaders = alias.ExtraHeaders
 	nc.DefaultRequestTimeoutInSeconds = alias.DefaultRequestTimeoutInSeconds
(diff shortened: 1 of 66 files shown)


core/schemas/provider.go
```diff
@@ -108,7 +110,25 @@ func (nc *NetworkConfig) UnmarshalJSON(data []byte) error {
...
+	// operator never intended.
+	if envKey, ok := strings.CutPrefix(alias.BaseURL, "env."); ok {
+		resolved, present := os.LookupEnv(envKey)
+		if !present || resolved == "" {
+			return fmt.Errorf("network_config.base_url references environment variable %q, which is unset or empty", envKey)
+		}
+		nc.BaseURL = resolved
+	} else {
+		nc.BaseURL = alias.BaseURL
+	}
```
BearTS: Hey this implementation is incorrect, I would suggest making use of schemas.SecretVar and then using the GetValue function 
https://github.com/maximhq/bifrost/blob/dev/core/schemas/secretvar.go

BearTS (changes requested):

core/schemas/provider.go
valentinyanakiev: Reworked onto `schemas.SecretVar` as suggested: `UnmarshalJSON` now goes through `IsSecretRef` → `NewSecretVar` → `GetValue()`, so `vault.` references work too. `BaseURL` stays a `string` at runtime (~250 read sites unchanged); the originating `SecretVar` is kept on an unexported field so `MarshalJSON` round-trips the reference instead of the resolved URL, matching `CACertPEM`. The unset-variable test now sets and explicitly unsets the variable so it no longer depends on the host environment. Pushed as a single commit rebased on current `dev`.

valentinyanakiev:
@coderabbitai The pre-merge Docstring Coverage check reports 33.33% over 3 functions in 2 files, but every declaration this diff touches carries an identifier-prefixed Go doc comment: `NetworkConfig.UnmarshalJSON` and `NetworkConfig.MarshalJSON` in `core/schemas/provider.go`, and `TestNetworkConfig_BaseURLSecretRef` in `core/schemas/serialization_test.go`. Please list the exact function declarations the check considers undocumented so I can fix the right ones. Do not trigger another full review.

valentinyanakiev:
@coderabbitai full review

valentinyanakiev:
@coderabbitai full review

valentinyanakiev:
@coderabbitai ignore pre-merge checks

Reason: the only failing pre-merge check is the Docstring Coverage warning (66.67%, 3 functions). Every declaration this diff touches carries an identifier-prefixed Go doc comment: `NetworkConfig.UnmarshalJSON` and `NetworkConfig.MarshalJSON` (both pre-existing comments, extended in this PR to describe the base_url reference handling) and `TestNetworkConfig_BaseURLSecretRef`. Two full reviews after the docstring additions moved the metric from 33.33% to 66.67% and no further, and the earlier request to name the undocumented declaration was not answered, so the residual is a detector false negative on a documented function, the same shape as the one recorded on PR 5985.

valentinyanakiev:
Rebased onto current `dev` @ 3585a9c82 (no merge commits, single commit, DCO signed-off). Head is now 7719fb5d2. Re-verified on the rebased tree with the module-directed toolchain (go 1.27.0 via GOTOOLCHAIN): `go build ./...` (core), `go vet ./schemas/...`, targeted `go test ./schemas/... -run 'NetworkConfig|Serializ|EnvVar|SecretVar'`, and full `go test ./providers/...` — all pass. The earlier maintainer request (use the SecretVar/GetValue machinery) is implemented in this head: `NetworkConfig.UnmarshalJSON` resolves via `NewSecretVar`/`GetValue` and round-trips the reference through `baseURLRef`/`MarshalJSON`. Maintainer re-review is pending.

valentinyanakiev:
@BearTS ready for another look — the implementation is reworked onto `schemas.SecretVar` as you suggested.

`NetworkConfig.UnmarshalJSON` now goes `IsSecretRef` → `NewSecretVar` → `GetValue()`, so `vault.` references work alongside `env.` for free instead of the hand-rolled prefix check.

Two decisions worth flagging, since they go slightly beyond the suggestion:

- **`BaseURL` stays a `string` on the struct.** There are ~250 read sites across providers, transports and the config store, and the runtime contract ("BaseURL holds the dialable URL") is unchanged, so none of them have to move.
- **The originating `SecretVar` is retained on an unexported `baseURLRef` field** so `MarshalJSON` emits the reference (`env.VAR`) rather than the resolved URL. Without it, config-store persistence and API responses would write back the resolved endpoint and defeat the point of the indirection. This mirrors how `CACertPEM` round-trips through `SecretVarAsString`.

An unresolved or empty reference is still a fail-loud config error rather than `SecretVar`'s usual silent-empty, matching the fail-closed stance `core/providers/utils` already takes for `proxy.url`: a bad provider key surfaces as a 401, but a bad `base_url` would otherwise be dialed as the literal reference string.

Tests cover literal passthrough, `env.` resolution, the reference round-trip (and that the resolved URL never appears in marshaled output), programmatic reassignment, unset/empty/whitespace-only, and an unresolvable `vault.` — plus CodeRabbit's host-environment isolation nit on the unset case.

valentinyanakiev:
@BearTS @akshaydeo @TejasGhatte gentle nudge on this one.

The change you asked for on 2 Sep is in: `base_url` now resolves through `schemas.SecretVar` → `GetValue()` exactly like `ca_cert_pem` and `proxy.url`, so `env.` and `vault.` references both work. The head (`7719fb5d`) is rebased on current `dev`, single commit, DCO signed, all three checks green, and CodeRabbit has re-approved.

The standing "changes requested" review is now the only thing blocking the merge button. @BearTS, could you re-review or dismiss it when you have a moment? For context, this is one of the focused PRs carved out of #5985 (opened 9 Aug), so the underlying fix has been waiting a while. Happy to adjust anything else you'd like to see.

BearTS:
Hey @valentinyanakiev , I would not recommend using this workaround, but instead convert the original baseUrl in secretVar and updating the callers

valentinyanakiev:
@BearTS done, thanks for the steer. `NetworkConfig.BaseURL` is now `*schemas.SecretVar` and every caller reads it through `GetValue()`; the hidden retained-reference field is gone and `MarshalJSON` just uses `SecretVarAsString`, exactly like `CACertPEM`.

While touching every constructor I folded the 25 copies of the default-URL / trailing-slash block into one `providerUtils.NormalizeBaseURL(&config.NetworkConfig, defaultURL)` helper. It clones the `SecretVar` before trimming so a pointer shared with a config-store copy is never mutated in place, and the `env.`/`vault.` reference survives. The ollama/sgl key-backfill migration now carries the `SecretVar` into the key URL instead of the resolved string. Also picked up `databricks` and `github-copilot`, which landed on `dev` after this PR was opened.

Head `a92aa655`, rebased on current `dev` @ `03ab39186`, single commit, DCO signed. `core`, `framework`, `transports`, and all plugins build and vet clean; `core` schemas/providers tests, `configstore` tests, and `transports/bifrost-http/lib` tests pass (the one failure there, `TestGenerateMCPClientHash_RuntimeVsMigrationParity`, fails identically on a clean `dev` checkout and is unrelated). The PR description is updated to match.

Note for reviewers: this is a Go-API change for `core` consumers (field type), while the JSON wire form is unchanged. When you have a moment, could you re-review or dismiss the earlier changes-requested so the merge button unblocks?

valentinyanakiev:
CodeRabbit's three findings on the refactored head, dispositioned in `b026f7742`:

1. **Redact `BaseURL` in `NetworkConfig.Redacted`** — done, with one deliberate nuance. The copy now clones the `SecretVar` (no shared pointer), and masks the resolved value only when the base_url came from an `env.`/`vault.` reference. A literal base_url stays readable: it is not a secret, the UI displays it, and fully masking it would blank every plain URL in the provider list. The JSON wire form already emitted the reference rather than the resolved URL, so this closes the in-process exposure path.
2. **Gemini debug logs print the resolved base URL** — done. New `providerUtils.LoggableURL` replaces the resolved scheme/host with the reference when base_url is secret-backed (e.g. `env.UPSTREAM_URL/v1beta/batches/123:cancel`) and returns the URL unchanged otherwise. Applied at the batch cancel/delete/results sites and the result-file download.
3. **Ollama/SGL migration rollback leaves created rows** — not addressed here on purpose. That migration predates this PR; the only change in it is carrying the `SecretVar` into the key URL instead of the resolved string. Happy to open a separate PR if maintainers want the rollback tightened.

Tests: `TestNetworkConfig_Redacted_BaseURL` and `TestLoggableURL` added; `core` builds and vets, schema/utils/gemini tests green.

BearTS:
@valentinyanakiev sure let me take a look at this

PR #6736, opened 2026-09-02
fix(transports): loadCustomPlugins fails closed on a required custom plugin load failure

valentinyanakiev:
## Summary

`loadCustomPlugins` logs a per-plugin instantiate failure, marks the plugin's status `error` via `UpdatePluginOverallStatus`, and `continue`s to the next plugin, but the function unconditionally returns `nil` regardless of how many enabled plugins failed. `Bootstrap` therefore never sees an error, never exits, and `server.Start()` proceeds: every surface (inference, admin, `/health`) keeps serving with an enabled, non-enterprise custom plugin silently absent, even though the deployment's own config declares it required.

Found via a downstream deployment's plugin-ABI conformance gate, whose negative leg (a plugin config pointing at a nonexistent `.so`) showed the process booting healthy and answering `GET /health` 200 with the configured plugin missing entirely. That is a fail-open defect for any deployment that treats an enabled custom plugin as required, such as a metadata-capture or policy-enforcement plugin with no safe degraded mode.

## Changes

- `loadCustomPlugins` counts every enabled, non-enterprise custom plugin that fails to instantiate or returns a nil plugin, and returns one aggregated error when the count is non-zero. It propagates through the existing, unmodified call chain (`loadCustomPlugins` → `LoadPlugins` → `Bootstrap` → `main.go`'s `os.Exit(1)`); no new fail-closed mechanism, the process simply never reaches `Start()`.
- The aggregate error is deliberately a count plus a generic message, never raw plugin name, path, or underlying error text. It reaches `main.go`'s top-level `"failed to bootstrap server: %v"` line, a different and less scoped logging surface than the per-plugin `UpdatePluginOverallStatus` / `logger.Error` calls, which already carry the full detail and are the right place for it (a plugin's path or URL can itself be sensitive).
- Enterprise-plugin skip behavior, per-plugin status bookkeeping, and every other loop body line are unchanged.

## Type of change

- [x] Bug fix

## Affected areas

- [x] Transports (HTTP)

## How to test

```sh
source .github/workflows/scripts/setup-go-workspace.sh   # multi-module workspace, as CI does
cd transports
go vet ./bifrost-http/server/
go test ./bifrost-http/server/ -run TestLoadCustomPlugins -count=1 -v
```

Four focused tests in `plugins_failclosed_test.go`, all green:

- `TestLoadCustomPluginsFailsClosedOnInstantiateError`: an enabled plugin whose shared object cannot be loaded makes the function return an error, and the error does not contain the plugin path (sentinel-checked).
- `TestLoadCustomPluginsFailsClosedOnNilPlugin`: a loader returning `(nil, nil)` reaches the defensive nil-plugin branch, which also counts as a failure.
- `TestLoadCustomPluginsSkipsEnterprisePluginFailure`: enterprise plugins keep today's skip-and-continue behavior.
- `TestLoadCustomPluginsReturnsNilWithNoEnabledCustomPlugins`: the success path is unchanged.

The two fail-closed tests fail against stock `dev` (the function returns `nil`).

## Breaking changes

- [x] No

A deployment whose config enables a custom plugin that cannot load previously booted with that plugin silently missing; it now exits at bootstrap with a count-only error and the per-plugin detail in the existing plugin status/log surface. A deployment with a broken but unwanted plugin should disable it in config rather than rely on the silent skip.

## Related issues

None filed. Same downstream migration as #6735, #6740, #6741.

## Security considerations

Closes a fail-open path: a policy-enforcement or capture plugin that failed to load no longer leaves the gateway serving without it. The new aggregate error is sanitized by construction (count and generic text only) so the less-scoped bootstrap log line never gains plugin paths or config text.

## Checklist

- [x] I verified builds succeed (Go, in the CI workspace)

### DIFF
--- a/transports/bifrost-http/server/plugins.go
+++ b/transports/bifrost-http/server/plugins.go
@@ -310,6 +310,13 @@ func (s *BifrostHTTPServer) loadBuiltinPlugins(ctx context.Context) error {
 
 // loadCustomPlugins loads plugins from PluginConfigs
 func (s *BifrostHTTPServer) loadCustomPlugins(ctx context.Context) error {
+	// Count enabled, non-enterprise custom plugins that fail to load so the
+	// function can fail closed below instead of letting Bootstrap proceed with
+	// a required plugin silently absent. The aggregate error stays a count plus
+	// a generic message: it reaches main.go's top-level bootstrap log line,
+	// while the per-plugin status/log calls in the loop already carry the full
+	// detail (a plugin path or URL can itself be sensitive).
+	failureCount := 0
 	for _, cfg := range s.Config.PluginConfigs {
 		// Skip built-ins (already loaded)
 		if lib.IsBuiltinPlugin(cfg.Name) {
@@ -347,6 +354,7 @@ func (s *BifrostHTTPServer) loadCustomPlugins(ctx context.Context) error {
 			// Use cfg.Name since plugin may be nil when InstantiatePlugin returns an error
 			s.Config.UpdatePluginOverallStatus(cfg.Name, cfg.Name, schemas.PluginStatusError,
 				[]string{fmt.Sprintf("error loading plugin %s: %v", cfg.Name, err)}, []schemas.PluginType{})
+			failureCount++
 			continue
 		}
 
@@ -355,6 +363,7 @@ func (s *BifrostHTTPServer) loadCustomPlugins(ctx context.Context) error {
 			logger.Error("plugin %s instantiated but returned nil", cfg.Name)
 			s.Config.UpdatePluginOverallStatus(cfg.Name, cfg.Name, schemas.PluginStatusError,
 				[]string{fmt.Sprintf("plugin %s instantiated but returned nil", cfg.Name)}, []schemas.PluginType{})
+			failureCount++
 			continue
 		}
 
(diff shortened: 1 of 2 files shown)


valentinyanakiev:
@coderabbitai full review

valentinyanakiev:
@coderabbitai full review

valentinyanakiev:
Rebased onto current `dev` @ 3585a9c82 (single commit, DCO signed-off). Head is now f282c8de0. Re-verified on the rebased tree using the CI workspace setup (`.github/workflows/scripts/setup-go-workspace.sh`): `go vet ./bifrost-http/server/...` and `go test ./bifrost-http/server/... -run TestLoadCustomPlugins -v` — all four focused tests pass.

valentinyanakiev:
Provenance correction: the commit message previously said this was verified only against the pinned `transports/v2.0.0` release because current `dev` had a `framework/grant` module-resolution gap. After the rebase onto `dev` @ 3585a9c82 that caveat no longer applies, so the message was amended (message-only — the tree is byte-identical to f282c8de0) to state the current truth: verified on current dev via the CI workspace setup, all four `TestLoadCustomPlugins` tests pass. Head is now 24bbe47d0, single DCO-signed commit.

valentinyanakiev:
@akshaydeo @Pratham-Mishra04 @roroghost17 gentle nudge: this has been ready for review since 2 Sep.

Head `24bbe47d` is rebased on current `dev`, single commit, DCO signed, both checks green, CodeRabbit approved. It is a small robustness fix: today a required custom plugin can fail to load and Bifrost still comes up healthy on `/health` without it, which is easy to miss in production. The PR makes that a startup failure instead. Happy to discuss the failure-mode choice if you would prefer a different one.

valentinyanakiev:
@akshaydeo @Pratham-Mishra04 @roroghost17 second nudge — this one has had no human review since it was opened on 2 Sep.

Head `24bbe47d` is a single DCO-signed commit, 2 files, +125/-0, mergeable, CodeRabbit approved, and all reporting checks green.

The behaviour: today a plugin listed as required in config can fail to load and Bifrost still comes up serving on a healthy `/health`, silently without it. That is easy to miss in production — you find out from the absence of whatever the plugin was supposed to enforce. The PR makes `loadCustomPlugins` fail closed on a required plugin's load failure, and leaves optional plugins behaving exactly as before.

One question while you're here: this PR shows no `license/cla` status at all, while my three other open PRs against this repo all carry `license/cla` green from the same account. Is that expected for a transports-only change, or does the check need a re-trigger here? Happy to push an empty commit if that is what re-fires it.

PR #6740, opened 2026-09-02
fix(anthropic): emit and preserve output_config.effort on DeepSeek's Anthropic endpoint

valentinyanakiev:
## Summary

DeepSeek's Anthropic-compatible API takes reasoning depth through `output_config.effort` (its docs: "output_config: only `effort` is supported") and documents `thinking.budget_tokens` as ignored. The DeepSeek provider builds its `/anthropic/v1/messages` requests through the shared Anthropic converters, whose effort handling is gated on the Claude model ladder (`DefaultSupportsNativeEffort`). `deepseek-v4-flash` and `deepseek-v4-pro` never match that ladder, so on `dev` today:

- a caller's `reasoning.effort` is converted into `thinking: {type: enabled, budget_tokens: N}`, which DeepSeek ignores, and no `output_config.effort` is emitted;
- a raw `output_config.effort` the caller sends is removed by `stripUnsupportedAnthropicFields` / `StripUnsupportedFieldsFromRawBody`.

Either way the wire carries no effort. Reproduced against a capturing `httptest` server standing in for DeepSeek: effort-only Chat and Responses requests arrive with `output_config` absent and `thinking.budget_tokens` set.

## Changes

- `ProviderFeatureSupport` gains `NativeEffort bool`: `output_config.effort` is accepted on every model the provider serves. Set for `schemas.DeepSeek`.
- New `defaultSupportsNativeEffort(caps)` in `anthropic/utils.go` is the fallback passed to `ModelCaps.SupportsNativeEffort` at the five existing call sites (typed strip, raw strip, the `SupportsNativeEffort` helper, and the two Responses-ladder branches). A datasheet row still overrides; a provider with the grant returns true; everyone else falls back to `DefaultSupportsNativeEffort(model)` exactly as before. This follows the pattern the `ModelCaps` comment describes ("the fallback every call site passes is the provider-level answer from the ProviderFeatures matrix").
- No model-name predicate, no change to the strip functions' signatures or the Claude ladder. With the grant, the existing "native effort + budget_tokens" branch of the Chat ladder emits `output_config.effort` alongside the `thinking` block DeepSeek already receives today; the Responses ladder's native-effort-only branch forwards the effort with thinking omitted.
- Changelog entry.

## Type of change

- [x] Bug fix

## Affected areas

- [x] Core (Go)
- [x] Providers/Integrations

## How to test

```sh
cd core
go vet ./providers/anthropic/ ./providers/deepseek/
[...]

PR #6741, opened 2026-09-02
fix(deepseek): reject document content before egress on the Anthropic endpoint

valentinyanakiev:
## Summary

DeepSeek's Anthropic-compatible API documents document content blocks as unsupported (its request-support table: "Documents: Not Supported"; images are supported) but does not reject them. A request carrying a document returns HTTP 200 with the document silently dropped from the model's view. Because there is no upstream error, no fallback can ever fire, and a caller who routed a document-bearing request through DeepSeek with an enumerated fallback gets a confident answer that never saw the document. I hit this live against the real endpoint before finding the documented contract.

## Changes

- New `core/providers/deepseek/content.go`: `rejectUnsupportedChatContent` / `rejectUnsupportedResponsesContent` scan the neutral request for file (document) blocks: user-turn content blocks, and file blocks inside function tool-call outputs (which reach the model as content the same way). Type discriminators only; no file data, URL, filename, or text is read.
- Called at the top of the Anthropic-endpoint branch in the four provider methods (`ChatCompletion`, `ChatCompletionStream`, `Responses`, `ResponsesStream`), before conversion or egress. The OpenAI-endpoint branch is untouched.
- The error is an Anthropic-shaped `invalid_request_error` (400, code `deepseek_unsupported_document_content`) with `AllowFallbacks: true`, so a routing rule with a next provider carries the untouched request onward, and a direct request gets a typed local error instead of a silent drop.
- Image blocks are deliberately not touched: DeepSeek documents them as supported (base64, url, file sources).
- Changelog entry.

## Type of change

- [x] Bug fix

## Affected areas

- [x] Core (Go)
- [x] Providers/Integrations

## How to test

```sh
cd core
go vet ./providers/deepseek/
go test ./providers/deepseek/ -run 'RejectsDocument|StillForwardsImages' -count=1 -v
go test ./providers/deepseek/ -count=1
```

`providers/deepseek/content_test.go` drives the real provider through a counting `httptest` server standing in for `/anthropic/v1/messages`:

- a document block on each of the four paths is rejected typed (400, the code above, `AllowFallbacks` true) with **zero** upstream arrivals; all four **fail on stock `dev`** with the guard removed (the request reaches the server and returns a success);
- an image block still reaches the upstream exactly once and succeeds.

## Breaking changes

- [x] No

A document-bearing request to DeepSeek's Anthropic endpoint previously "succeeded" with the document ignored; it now fails typed or falls through to the next configured provider.

## Related issues

Reference: https://api-docs.deepseek.com/guides/anthropic_api (content block support). Second focused replacement for the closed PR 5985 (maximhq/bifrost, "Preserve DeepSeek V4 Flash and Pro Anthropic fidelity"); PR 6740 is the first. Scope is deliberately narrower than that PR's content guard, which rejected images and documents for the two V4 models with a 415: DeepSeek's current docs list images as supported, so this PR rejects documents only, applies to every model the provider serves (the contract is the endpoint's, not a model's), and returns the Anthropic-shaped 400 invalid_request_error the rest of the provider surface uses.

## Security considerations

Reduces silent data-handling surprise: a document a caller believed the model read is no longer discarded upstream without signal. The guard inspects no payload bytes.

## Checklist

### DIFF
--- a/core/providers/deepseek/content.go
+++ b/core/providers/deepseek/content.go
@@ -1,0 +1,88 @@
+package deepseek
+
+import (
+	"github.com/maximhq/bifrost/core/schemas"
+)
+
+// DeepSeek's Anthropic-compatible API documents document content blocks as
+// unsupported ("Documents: Not Supported"; images are supported) but does not
+// reject them: a request carrying a document returns HTTP 200 with the
+// document silently dropped from the model's view. An upstream-error-driven
+// fallback can therefore never enforce the contract, so the provider checks
+// for document blocks itself before conversion or egress and returns a typed,
+// fallback-eligible error. The guard reads block type discriminators only; it
+// never reads file data, URLs, filenames, or text.
+//
+// Source: https://api-docs.deepseek.com/guides/anthropic_api
+
+const unsupportedDocumentCode = "deepseek_unsupported_document_content"
+
+// rejectUnsupportedChatContent fails a Chat request bound for the Anthropic
+// endpoint when any message carries a file (document) content block.
+func rejectUnsupportedChatContent(request *schemas.BifrostChatRequest) *schemas.BifrostError {
+	if request == nil {
+		return nil
+	}
+	for i := range request.Input {
+		content := request.Input[i].Content
+		if content == nil {
+			continue
+		}
+		for j := range content.ContentBlocks {
+			if content.ContentBlocks[j].Type == schemas.ChatContentBlockTypeFile {
+				return newUnsupportedDocumentError()
+			}
+		}
+	}
+	return nil
+}
+
+// rejectUnsupportedResponsesContent applies the same contract to Responses
+// requests, including file blocks inside function tool-call outputs, which
+// reach the model as content the same way a user turn does.
+func rejectUnsupportedResponsesContent(request *schemas.BifrostResponsesRequest) *schemas.BifrostError {
+	if request == nil {
+		return nil
+	}
+	for i := range request.Input {
+		message := &request.Input[i]
+		if message.Content != nil && hasResponsesFileBlock(message.Content.ContentBlocks) {
+			return newUnsupportedDocumentError()
+		}
+		if message.ResponsesToolMessage != nil && message.ResponsesToolMessage.Output != nil &&
+			hasResponsesFileBlock(message.ResponsesToolMessage.Output.ResponsesFunctionToolCallOutputBlocks) {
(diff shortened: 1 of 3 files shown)


valentinyanakiev:
@coderabbitai full review

valentinyanakiev:
Rebased onto current `dev` @ 3585a9c82 — this resolves the merge conflict GitHub was reporting (a `core/changelog.md` overlap; no code conflict). Head is now f19c933b7 (single commit, DCO signed-off). Re-verified on the rebased tree: `go vet ./providers/deepseek/...`, targeted `-run 'Document|Image|Content'` (5 passing tests), and full `go test ./providers/deepseek/...` — all pass.

valentinyanakiev:
@akshaydeo gentle nudge: ready for review since 2 Sep, and it is the content-guard half of #5985 (opened 9 Aug), cut down to a focused diff.

Head `f19c933b` is rebased on current `dev`, single commit, DCO signed, all checks green, CodeRabbit approved. The behaviour it fixes is a quiet one: DeepSeek's Anthropic endpoint accepts document blocks it cannot process, so today a request egresses with the payload and comes back degraded instead of failing fast before the bytes leave. Would appreciate a look when you can; happy to adjust.

valentinyanakiev:
@akshaydeo second nudge — no human review yet since 2 Sep.

Head `64ac0c69` is rebased on current `dev`, single DCO-signed commit, 3 files, +293/-0, mergeable, all checks green.

Same note as on #6740: I dropped the `core/changelog.md` entry, because the file is emptied at each release cut and the line was re-conflicting the PR every couple of days without helping review. Happy to re-add it when you look. The code diff is otherwise byte-identical to what it was.

Why it matters: DeepSeek documents document blocks as unsupported on its Anthropic-compatible endpoint, but accepts the request anyway and silently drops them. Because the upstream returns success, no fallback can ever trigger — the caller gets a confident answer computed without the document it supplied. This rejects document content before egress with a typed, fallback-eligible `invalid_request_error`, and leaves image blocks forwarding unchanged (covered by a test).

Worth a sanity check on your side: `dev` has since merged #5663 (preserve document blocks in Bedrock tool results) and the Cowork base64-document decode path. Neither conflicts with this textually and the full provider suite passes on the rebase, but you may want to confirm the interaction reads the way you expect.

akshaydeo:
@valentinyanakiev we have a bunch of PRs in the pipeline - so things gets slower 🙇  we will get to this asap

PR #6744, opened 2026-09-02
fix: seed user scope in model config allow-list and preserve runtime-owned governance rows from config.json prune

sammaji:
## Summary

Two related bugs caused data loss when `source_of_truth=config.json` was active alongside enterprise access profiles. First, the `user` model config scope was declared in the package but omitted from the seed map, so any write to a user-scoped row before the governance plugin's `Init` called `RegisterModelConfigScope` failed with `invalid scope "user" for model config`, aborting access profile assignment mid-way. Second, the config.json prune loop deleted every governance row not present in the file, including virtual keys, budgets, rate limits, and model configs minted at runtime by access profiles — rows that can never appear in a config file.

## Changes

- `ModelConfigScopeUser` is now included in the initial `validModelConfigScopes` seed so it is writable from process start without requiring a registration call. `RegisterModelConfigScope` is updated to document that it exists only for scopes this module does not declare.
- A `RuntimeOwnedGovernanceRows` type and a `RegisterRuntimeOwnedGovernanceResolver` hook allow downstream builds to identify governance rows that are runtime-owned. The config.json prune skips deletion of any row whose ID appears in the resolved set.
- `retainRuntimeOwned` merges file-declared rows with any surviving runtime-owned rows so the in-memory governance config stays consistent with what the database retained after pruning.
- If no resolver is registered (OSS default) or the resolver returns an error, the prune behaves exactly as before — nothing is protected and startup is not blocked.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [ ] Core (Go)
- [x] Transports (HTTP)
- [ ] Providers/Integrations
- [x] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

```sh
go test ./framework/configstore/tables/...
go test ./transports/bifrost-http/lib/...
```

`TestDeclaredModelConfigScopesAreValidByDefault` confirms all three declared scopes are valid without a registration call.

`TestSQLite_SourceOfTruthConfigJSON_RuntimeOwnedRowsSurvivePrune` seeds both protected and unprotected rows, registers a resolver that marks only the protected set, reloads with `source_of_truth=config.json`, and asserts that protected rows survive in both the database and the in-memory governance config while unprotected rows are pruned.

`TestSQLite_SourceOfTruthConfigJSON_NoResolverPrunesEverything` confirms OSS behavior is unchanged when no resolver is registered.

`TestSQLite_SourceOfTruthConfigJSON_ResolverErrorDoesNotBlockLoad` confirms a failing resolver degrades gracefully rather than aborting startup.

## Breaking changes

- [ ] Yes
- [x] No

## Security considerations

The resolver is called inside `LoadConfig` before any downstream store wrapper exists. A resolver error is logged and treated as "protect nothing" rather than failing startup, which is intentional — the prune it guards is itself skipped when the relevant config section is absent, so the failure mode is bounded.

## Checklist

### DIFF
--- a/framework/configstore/tables/modelconfig.go
+++ b/framework/configstore/tables/modelconfig.go
@@ -21,22 +21,29 @@ const (
 // with a nil provider it means all models on all providers.
 const ModelConfigAllModels = "*"
 
-// validModelConfigScopes is the runtime registry of accepted scope values.
-// OSS seeds it with global + virtual_key; downstream consumers (e.g. the
-// enterprise build registering "user") extend it at startup via
-// RegisterModelConfigScope. Guarded by validModelConfigScopesMu.
+// validModelConfigScopes is the runtime registry of accepted scope values,
+// seeded with every scope declared above. All three are known to this module —
+// the governance plugin reads and enforces user-scoped rows the same as the
+// other two — so none of them may depend on a registration call to become
+// writable. A scope that has to be registered before it can be persisted is one
+// this module does not declare. Guarded by validModelConfigScopesMu.
 var (
 	validModelConfigScopesMu sync.RWMutex
 	validModelConfigScopes   = map[string]bool{
 		ModelConfigScopeGlobal:     true,
 		ModelConfigScopeVirtualKey: true,
+		ModelConfigScopeUser:       true,
 	}
 )
 
-// RegisterModelConfigScope adds scope to the allow-list consulted by
-// IsValidModelConfigScope and TableModelConfig.BeforeSave. Intended to be
-// called once at process startup; safe to call concurrently. Whitespace-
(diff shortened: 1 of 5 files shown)


sammaji:
* **#6744** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/6744?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/maximhq/bifrost/6744?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `dev`


<h2></h2>

PR #6756, opened 2026-09-02
fix(governance): revoke in-flight VK rotation grace windows when cooldown is set to 0

impoiler:
## Problem

The security settings say:

> Leave empty (or 0) to have the old value stop working immediately.

It didn't. `vk_rotation_cooldown` is read **once, at rotation time**, and stamped onto `previous_value_expires_at`; authentication only ever reads that stamped timestamp, never the live setting. So saving `0` changed what *future* rotations would stamp and nothing else — a key rotated under a 1h cooldown kept accepting its old value for the rest of that hour. Nothing in the config path touched VK rows.

Reported from the UI: setting the cooldown to empty "doesn't expire all the grace period keys, it still works till the cooldown period ends".

## Fix

**`ConfigStore.RevokeVirtualKeyRotationGrace`** — one bulk `UPDATE` clearing `previous_value`, `previous_value_hash`, `previous_value_expires_at` on every key still carrying them, returning the affected IDs.

- Selects on `previous_value_expires_at IS NOT NULL` rather than "expiry in the future", so already-expired leftovers are swept in the same pass. Nothing purged these before, so retired credential material sat on the row indefinitely.
- Deliberately unscoped (`DB()`, not `ScopedDB`): the cooldown is a global setting, so running the revocation through the caller's DAC scope would revoke only the keys that caller can see and silently leave the rest authenticating.
- `SkipHooks` — targeted column write on rows that are not loaded; `BeforeSave` would otherwise run against a zero-valued model.
- `rotated_at` is audit history, not grace state, and survives.

**`LocalGovernanceStore.RevokeRotationGraceInMemory`** — pure in-memory sweep, no DB access, clearing retired values from the cached keys auth actually reads.

- Ranges the ID index, not the value map: the value map holds a second entry per grace value pointing at the same object, so ranging it would visit those keys twice and mutate an already-replaced object.
- Copy-on-write per key, matching every other in-memory VK update — readers hold these pointers without a lock.
- Drops each alias only where it still resolves to the rotated key. **The `previous_value_hash` index is deliberately non-unique, so one key's retired value can be another key's live current credential**; revoking must not deauthenticate that owner.

**`PUT /api/config`** fires the revocation when the incoming cooldown is zero — after the config persists, so a failed save cannot revoke live credentials, and failing the request on error rather than reporting success while retired values keep working. All three spellings of "no cooldown" hit it: `0`, `""`, `"0s"`.

Cost is one `UPDATE` plus one in-memory pass regardless of how many keys are affected. That is what lets the cluster propagate it as a single message rather than a reload per key — see the companion enterprise PR.

## Scope

A **non-zero** change still applies only to future rotations: shortening 1h → 5m leaves existing 1h windows running. That was a deliberate call; the UI copy now says so explicitly instead of leaving it to be discovered.

Not covered: setting the cooldown to 0 **via config.json** on a source-of-truth deployment does not revoke at boot — the load path never calls the revocation. Boot-time credential revocation is a different risk profile, so it is left as a separate decision.

## Drive-by

`plugins/governance/store_test.go` had a stale `NewLocalGovernanceStore` call (5 args, signature takes 6 since `modelCatalog` was added). Go stops at the first error, so the **entire package failed to build** and every governance test was silently skipped — including all eight existing rotation grace-period tests. One-token fix; they all pass.

## Tests

- `TestRevokeVirtualKeyRotationGrace` — live + expired + never-rotated keys; asserts the retired value stops resolving through `GetVirtualKeyByValue`, secret material is cleared, `rotated_at` survives, current values untouched, idempotent.
- `TestGovernanceStore_RevokeRotationGraceInMemory` and `..._KeepsOtherKeysCurrentValue` — the in-memory sweep, including the non-unique-hash edge above.
- `TestUpdateConfig_ZeroCooldownRevokesInFlightGracePeriods` (all four cooldown spellings) and `TestUpdateConfig_GraceRevocationFailureFailsRequest`.

`go vet` clean and full suites green across `core`, `framework`, `transports`, `plugins/governance`.

Two pre-existing failures on `dev`, unrelated and unchanged by this PR: `TestPreMCPConnectionHook_NoHealthCheckMarker_NoGrant_Warns` (was hidden behind the build break above) and, in enterprise, `TestRuntimeUserAccessProfileReload_NarrowsRelations`.


impoiler:
* **#6756** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/6756?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/maximhq/bifrost/6756?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `dev`


<h2></h2>

PR #6767, opened 2026-09-02
vk allowed models * handling for governance

### LINKED ISSUE #6657: [Bug]: Fireworks virtual key with allowed_models: ["*"] blocks every model (empty synced catalog; explicit list works)
### Prerequisites

### Description

A virtual key scoped to the `fireworks` provider with `allowed_models: ["*"]` rejects **every** Fireworks model with `403 model_blocked`, even though the wildcard should permit all of them.

The root cause appears to be that `["*"]` is resolved against Bifrost's **discovered model catalog** rather than being treated as an unconditional allow. For `fireworks`, that catalog is empty, so the wildcard matches nothing and every request is denied.

Evidence that the catalog is the deciding factor, not the governance config:

- `GET /v1/models` with the virtual key returns **0 models**.
- The `config_models` table is **empty** (0 rows) — for `fireworks` and for every other provider.
- The provider key's own sync status is `success` with an empty description, so nothing surfaces the failure.
- The same Fireworks key works fine when called directly against `https://api.fireworks.ai/inference/v1/chat/completions`, and `GET /v1/models` there returns 23 models, including `accounts/fireworks/models/glm-5p3`.

This looks like the same class of bug as #5224 (OpenRouter embedding models blocked under a wildcard because they are absent from the synced catalog), but here it affects **ordinary chat models** on `fireworks` via `/v1/chat/completions`.

**Workaround:** replacing `["*"]` with an explicit list containing the fully-qualified model id makes the exact same request succeed immediately. This is what isolates the bug to wildcard-vs-catalog resolution — nothing else about the request, key, or provider changes.

### Steps to reproduce

1. Configure the `fireworks` provider with a valid API key.
2. Create a virtual key whose only provider config is `fireworks`, with `allowed_models: ["*"]` and `blacklisted_models: []`.
3. Send a chat completion request:

```bash
curl -s http://127.0.0.1:8080/v1/chat/completions \
  -H "Authorization: Bearer <virtual-key>" \
  -d '{"model":"fireworks/accounts/fireworks/models/glm-5p3",
       "messages":[{"role":"user","content":"say OK"}],
       "max_tokens":50}'
```

4. Observe `403 model_blocked`.
5. Now change that virtual key's `allowed_models` to `["accounts/fireworks/models/glm-5p3"]`, restart, and re-run the identical request — it returns `200` with a valid completion.

Also fails identically with the short form `fireworks/glm-5p3`, and with `fireworks/glm-5p3-flash`.

### Expected behavior

`allowed_models: ["*"]` on a provider config permits every model for that provider.

### Actual behavior

Every Fireworks model returns 403 `model_blocked`; `/v1/models` reports 0 models for the key.

### Affected area(s)

Core (Go)

### Version

v2.0.0

### Environment

```text
Docker image: maximhq/bifrost@sha256:cf71be9fad4e0749b6e26cbb774c687413dad9a0970b83f4e1dadb6f503ea208
Host OS: Ubuntu 26.04 LTS (x86_64)
Deployment: single node, docker compose
config_store: sqlite (/app/data/config.db)
enforce_auth_on_inference: true
enforce_governance_header: true
```

### Relevant logs/output

```shell
# With allowed_models: ["*"]  -> blocked
{"type":"model_blocked","is_bifrost_error":false,"status_code":403,
 "error":{"message":"Model 'accounts/fireworks/models/glm-5p3' is not allowed for this virtual key"},
 "extra_fields":{"provider":"fireworks",
                 "original_model_requested":"accounts/fireworks/models/glm-5p3",
                 "resolved_model_used":"accounts/fireworks/models/glm-5p3",
                 "request_type":"chat_completion"}}

# Same request, allowed_models: ["accounts/fireworks/models/glm-5p3"]  -> 200 OK
{"id":"chatcmpl-...","choices":[{"index":0,"finish_reason":"length",
 "message":{"role":"assistant", ...}}],
 "model":"accounts/fireworks/models/glm-5p3"}
```

Ruled out during triage, each verified independently:

- Wrong model id — reproduced with the short id, the fully-qualified id, and a non-existent variant.
- Stale in-memory cache — full `docker compose down` / `up` between every attempt.
- Duplicate provider key rows — deleted the extra `config_keys` row for `fireworks`, no change.
- `blacklisted_models` misparsed — set to `NULL` directly in SQLite, no change.
- Virtual key staleness — reproduced on a freshly created, Fireworks-only virtual key.
- Governance team/customer inheritance — `team_id` and `customer_id` are both `NULL`.

For contrast, a wildcard on `openai` under the same setup resolves correctly and reaches the upstream provider (it returns the provider's own `insufficient_quota` error), so the wildcard path is not globally broken — it fails for `fireworks`.

### Regression?

Not known — first deployment of Bifrost on this host, so no earlier working version to compare against.

### Severity

akshaydeo:
## Summary

Fixes a regression where virtual keys scoped to a provider with `allowed_models: ["*"]` would return `403 model_blocked` for any model not yet present in the pricing datasheet. Because providers like Fireworks have no enumerable list-models endpoint, their catalog is built entirely from the pricing sheet, which can lag new model releases by weeks. This made the wildcard allowlist narrower than the provider itself, blocking legitimate requests.

## Changes

- In `IsModelAllowedForProvider`, when `allowedModels` is `["*"]` and `GetProvidersForModel` does not find the provider in the discovered catalog, the code now falls back to allowing the request if the live store returned no models for that provider but the datasheet has at least one row for it. This means the provider itself (which 404s unknown models) acts as the authority rather than the stale pricing sheet.
- Added `wildcardcatalog_test.go` with a regression test (`TestIsModelAllowedForProvider_WildcardAllowsModelMissingFromCatalog`) that reproduces the exact failure: a Fireworks key with `["*"]` must permit a model that postdates the pricing datasheet, both in full path form and short form.
- Updated a call site in `plugins/governance/store_test.go` to match a changed function signature.

## Type of change

- [x] Bug fix

## Affected areas

- [x] Core (Go)
- [x] Providers/Integrations
- [x] Plugins

## How to test

```sh
go test ./framework/modelcatalog/... -run TestIsModelAllowedForProvider_WildcardAllowsModelMissingFromCatalog -v
go test ./...
```

The regression test sets up a Fireworks catalog with a non-empty datasheet and an empty live store, then asserts that a model absent from the datasheet is still permitted under a `["*"]` allowlist.

## Breaking changes

- [x] No

## Related issues

Closes #6657

## Security considerations

The fallback only activates when the live list-models store is empty for a provider and the datasheet has at least one row, meaning providers with a fully enumerable live catalog are unaffected and still narrowed by that catalog. The provider itself remains the final authority and will 404 any model it does not recognise.

## Checklist

### DIFF
--- a/framework/modelcatalog/models.go
+++ b/framework/modelcatalog/models.go
@@ -297,7 +297,8 @@ func (mc *ModelCatalog) computeProvidersForModel(model string) []schemas.ModelPr
 // checks, not by the static keyconfig allow set).
 //
 //   - allowedModels=["*"]: defer to GetProvidersForModel (with custom-provider
-//     fast path when list-models is disabled).
+//     fast path when list-models is disabled), falling back to allow for a
+//     provider the datasheet describes but list-models cannot enumerate.
 //   - allowedModels=[]: deny-by-default.
 //   - explicit allowedModels: direct or provider-prefixed match against the
 //     provider's catalog.
@@ -313,7 +314,22 @@ func (mc *ModelCatalog) IsModelAllowedForProvider(provider schemas.ModelProvider
 		if isCustomProvider && hasListModelsEndpointDisabled {
 			return true
 		}
-		return slices.Contains(mc.GetProvidersForModel(model), provider)
+		if slices.Contains(mc.GetProvidersForModel(model), provider) {
+			return true
+		}
+		// A provider the datasheet describes but list-models cannot enumerate is
+		// known only through the pricing sheet, which lags new releases by weeks.
+		// Refusing on that list makes ["*"] narrower than the provider itself, so
+		// a wildcard denies every model released since the last sync (issue
+		// #6657). Defer to the provider instead: it 404s a model it does not
+		// have, and it is the authority on its own catalog.
+		//
+		// Both other cases are already right and stay untouched. A provider with
+		// no datasheet rows either is handled by computeProvidersForModel's
+		// keyconfig fallback, and a provider whose live list-models did answer is
+		// enumerable, so its catalog is authoritative and still narrows.
+		return len(mc.live.UnfilteredModelsForProvider(provider)) == 0 &&
+			len(mc.datasheet.DatasheetModelsForProvider(provider)) > 0
 	}
 	if allowedModels.IsEmpty() {
 		return false
--- a/framework/modelcatalog/wildcardcatalog_test.go
+++ b/framework/modelcatalog/wildcardcatalog_test.go
@@ -1,0 +1,132 @@
+package modelcatalog
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/maximhq/bifrost/core/schemas"
+	"github.com/maximhq/bifrost/framework/modelcatalog/datasheet"
+	"github.com/maximhq/bifrost/framework/modelcatalog/keyconfig"
+	"github.com/maximhq/bifrost/framework/modelcatalog/live"
+)
+
+// fireworksCatalog builds the catalog shape a Fireworks deployment actually has
+// (issue #6657):
+//
+//   - the datasheet carries Fireworks rows, because the public pricing sheet
+//     ships ~315 "fireworks_ai/..." entries;
+//   - the live store is empty for Fireworks, because Fireworks has no enumerable
+//     OpenAI-style list-models endpoint (see core/providers/fireworks: the
+//     provider builds its list from the key's configured models) and an operator
+//     who only pasted an API key configured none;
+//   - the provider key itself is unrestricted, i.e. an ordinary "just add the
+//     key" setup.
+//
+// The datasheet fixture mirrors the real feed: keys are "<litellm provider>/<model
+// id>" and the row's own "provider" field is "fireworks_ai", which normalizeProvider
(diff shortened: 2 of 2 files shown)


akshaydeo:
* **#6768** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/6768?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6767** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/6767?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/maximhq/bifrost/6767?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `dev`


<h2></h2>

Pratham-Mishra04 (approved):

akshaydeo:
### Merge activity

* **Sep 7, 9:49 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/6767).
* **Sep 7, 9:49 AM UTC**: @akshaydeo merged this pull request with [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/6767).

PR #6808, opened 2026-09-03
fix: encrypt rows one-per-transaction with cursor pagination and concurrent vault writes to prevent deadlocks and boot hangs

sammaji:
## Summary

The startup encryption backfill was prone to two failure modes when multiple nodes booted against the same database: nodes could deadlock by acquiring the same rows in opposite orders within a shared transaction, and the offset-based pagination could skip or repeat rows when rows were updated mid-walk. This PR fixes both by giving each row its own transaction with a single row lock, switching to a keyset cursor (`id > ?` with `ORDER BY id`) for stable pagination, and adding concurrent vault writes so boot time does not scale linearly with the number of secrets.

## Changes

- Added `encryptClaimedRows` generic helper that encrypts a slice of IDs concurrently, each in its own transaction that locks the row and re-checks it is still plaintext before the hook runs. A transaction holds exactly one row lock, so no deadlock cycle can form regardless of concurrency.
- Added `lockRow` helper that applies `SELECT ... FOR UPDATE` on Postgres and is a no-op on SQLite, preventing two nodes from encrypting the same row simultaneously.
- Added `encryptConcurrency` helper that returns `encryptionVaultConcurrency` (8) when vault writes are enabled and 1 otherwise, so OSS deployments keep the simple sequential path.
- Replaced offset-based pagination with a keyset cursor (`id > cursor` + `ORDER BY id`) across all ten `encryptPlaintext*` functions, eliminating the repeated-row and skipped-row bugs that offset pagination produces when rows are updated mid-walk.
- Changed all scan queries to `Pluck("id", &ids)` rather than loading full rows, so vault references in secret columns are not resolved over the network during the scan — only during the claim that actually encrypts the row.
- Changed the returned `count` to track only rows whose status actually advanced to `encrypted`, rather than all rows fetched, giving accurate telemetry when hooks decline to cipher a row.
- Fixed `TableOauthConfig.BeforeSave` to stamp `EncryptionStatus = encrypted` for env/vault reference secrets even though their value is not ciphered. Previously such rows were left as `plain_text`, causing the backfill to re-select and re-write them on every boot.
- Replaced the inline encrypt block in `TableOauthConfig.BeforeSave` with an `encryptSecretVar` call, making ref-passthrough behaviour consistent with other secret types.
- Added `TestTableOauthConfig_SecretRef_StampsStatus` to assert that env/vault references are stored verbatim but still leave `plain_text` after a save.
- Added `encryption_pagination_test.go` covering: multi-batch walks past `encryptionBatchSize`, all three "not yet encrypted" status sentinels (`plain_text`, `NULL`, `""`), termination when a hook never advances the status, and convergence of ref-only rows.
- Added `encryption_concurrency_postgres_test.go` covering: a second node blocking behind an open lock and then encrypting all rows after the lock is released, five concurrent nodes completing without deadlock or double-encryption, vault writes overlapping rather than queuing, and vault refs being resolved exactly once per row across concurrent nodes.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [x] Core (Go)
- [ ] Transports (HTTP)
- [ ] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

```sh
# Unit and integration tests (SQLite)
go test ./framework/configstore/... -v -run TestEncryptPlaintext

# Table-level hook tests
go test ./framework/configstore/tables/... -v -run TestTableOauthConfig

# Postgres concurrency tests (requires a running Postgres instance)
POSTGRES_DSN="host=localhost user=postgres password=postgres dbname=bifrost_test sslmode=disable" \
  go test ./framework/configstore/... -v -run TestEncryptPlaintextOAuthConfigs_WaitsForHolderThenConverges

POSTGRES_DSN="host=localhost user=postgres password=postgres dbname=bifrost_test sslmode=disable" \
  go test ./framework/configstore/... -v -run TestEncryptPlaintextOAuthConfigs_ConcurrentPodsDoNotDeadlock

POSTGRES_DSN="host=localhost user=postgres password=postgres dbname=bifrost_test sslmode=disable" \
  go test ./framework/configstore/... -v -run TestEncryptPlaintextKeys_OverlapsVaultWrites
```

## Breaking changes

- [ ] Yes
- [x] No

## Security considerations

Each sensitive row (API keys, OAuth secrets, session tokens, temp tokens, OAuth configs, MCP clients, provider proxies, vector store configs, and plugins) is now encrypted by exactly one node during startup. The row lock acquired before the hook runs prevents two nodes from both paying a row's vault write cost, and the re-check inside the transaction prevents a row from being written after another node already encrypted it. Fixing the status stamp on env/vault reference rows prevents those references from being repeatedly re-processed on every boot, reducing unnecessary reads of sensitive configuration.

## Checklist

### DIFF
--- a/framework/configstore/encryption.go
+++ b/framework/configstore/encryption.go
@@ -3,17 +3,94 @@ package configstore
 import (
 	"context"
 	"fmt"
+	"sync"
+	"sync/atomic"
 
+	"github.com/maximhq/bifrost/core/schemas"
 	"github.com/maximhq/bifrost/framework/configstore/tables"
 	"github.com/maximhq/bifrost/framework/encrypt"
 	"gorm.io/gorm"
+	"gorm.io/gorm/clause"
 )
 
 const (
 	encryptionStatusPlainText = "plain_text"
 	encryptionStatusEncrypted = "encrypted"
-	encryptionBatchSize       = 100
+	// encryptionBatchSize is how many candidate rows one scan reads ahead. The
+	// rows are still written one per transaction; this only bounds the read.
+	encryptionBatchSize = 100
+	// encryptionVaultConcurrency is how many rows are encrypted at once when the
+	// BeforeSave hook writes to the vault. Each row still gets its own
+	// transaction and its own single row lock, so this adds no deadlock risk; it
+	// only overlaps the vault round trips, which otherwise dominate the walk.
+	// Kept well under the runtime pool so a booting node cannot exhaust it.
+	encryptionVaultConcurrency = 8
 )
+
+// encryptConcurrency returns how many rows to encrypt in parallel. OSS, where vault
+// store is not enabled keeps concurrency to 1.
+func encryptConcurrency() int {
+	if schemas.VaultStoreWriteEnabled() {
+		return encryptionVaultConcurrency
+	}
+	return 1
+}
+
+// lockRow takes a row lock so a row claimed by one node is not re-encrypted by
+// another. Exactly one row is locked per transaction, so no transaction ever
+// holds one lock while waiting for a second and a deadlock cycle cannot form.
+// SQLite is single-writer and has no row locks, so the clause is omitted there.
+func lockRow(tx *gorm.DB) *gorm.DB {
+	if tx.Dialector.Name() == "sqlite" {
+		return tx
+	}
+	return tx.Clauses(clause.Locking{Strength: "UPDATE"})
+}
+
+// encryptClaimedRows encrypts the given ids, each in its own transaction that
+// locks the row and re-checks it is still plaintext before the hook runs. A
+// transaction holds exactly one row lock, so no deadlock cycle can form however
+// many run at once, and a row another node already encrypted is dropped before
+// its hook does any work — which matters because that work includes vault
+// writes. where clause must match a single row and take (status, id).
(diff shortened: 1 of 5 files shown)


sammaji:
* **#6833** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/6833?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6808** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/6808?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/maximhq/bifrost/6808?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `dev`


<h2></h2>

danpiths (approved):

akshaydeo:
### Merge activity

* **Sep 4, 3:22 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/6808).
* **Sep 4, 3:22 PM UTC**: @akshaydeo merged this pull request with [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/6808).

PR #6868, opened 2026-09-04
[fix]: preserve raw OpenAI Responses cache evidence

### LINKED ISSUE #6867: Preserve raw OpenAI Responses cache-counter presence for post hooks
## Problem

BifrostResponsesResponse.WithDefaults materializes InputTokensDetails whenever usage exists. After normalization, a post hook cannot distinguish an upstream response that omitted usage.input_tokens_details from one that explicitly supplied an empty or zero-valued object.

This matters for telemetry and promotion gates: pointer presence after defaults can incorrectly turn missing cache evidence into an observed zero.

The same distinction must be captured at three OpenAI Responses boundaries:

- ordinary unary parsing
- large-request unary passthrough before its early return
- the terminal completed or incomplete SSE event

## Proposed behavior

Record one payload-blind boolean in BifrostContext before typed normalization. The value is true only when the raw detail field exists as an object. Do not retain raw bodies, request content, response IDs, or arbitrary headers. A truncated stream should leave the value absent rather than claim false.

## Validation

A focused patch includes table tests for present, missing, null, and scalar shapes plus an httptest upstream that exercises both present-empty and omitted details through the real large-request passthrough. The OpenAI provider and schemas suites pass on current dev.

valentinyanakiev:
## Summary

Preserve whether raw OpenAI Responses usage included `input_tokens_details` before `WithDefaults` normalizes the response. This lets post hooks distinguish observed zero cache counters from missing telemetry.

Closes #6867.

## Changes

- Add one payload-blind `BifrostContext` boolean for raw Responses counter presence.
- Set it before normalization on ordinary unary responses.
- Set it before the large-request unary passthrough early return.
- Set it from terminal `response.completed` and `response.incomplete` SSE events.
- Leave the value absent when a stream truncates before a terminal event.
- Add parser table coverage for present-empty, missing, null, and scalar shapes.
- Add a real `httptest` upstream regression for the large-request boundary.
- Add table-driven streaming boundary coverage for completed-with-details, incomplete-without-details, and truncation.
- Add the required core changelog entry.

The implementation retains only a boolean. It does not retain raw bodies, prompts, outputs, response IDs, or arbitrary headers. A stream with no terminal usage event leaves the fact absent rather than reporting false.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [x] Core (Go)
- [ ] Transports (HTTP)
- [x] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

From `core`:

```sh
go test ./providers/openai -run "Test(RawResponsesCacheCountersKnown|ResponsesStreamCacheCounterKnownnessBoundary|LargeRequestPassthroughPreservesRawCacheCounterPresence)" -count=1
go test ./providers/openai -count=1
go test ./schemas -count=1
go vet ./providers/openai ./schemas
```

All commands pass on current `dev`. CodeRabbit approved current head `9f94efccc`.

## Breaking changes

- [ ] Yes
- [x] No

## Security considerations

The new context value is a closed boolean derived at the raw decoder boundary. Provider content, credentials, request content, response content, and response IDs are never copied into it.

## Checklist

- [x] I read the repository contribution guidance and followed it
- [x] I added tests for all changed execution paths
- [x] I updated the core changelog
- [x] I verified affected Go tests and vet locally

### DIFF
--- a/core/providers/openai/openai.go
+++ b/core/providers/openai/openai.go
@@ -19,6 +19,7 @@ import (
 
 	providerUtils "github.com/maximhq/bifrost/core/providers/utils"
 	schemas "github.com/maximhq/bifrost/core/schemas"
+	"github.com/tidwall/gjson"
 	"github.com/valyala/fasthttp"
 )
 
@@ -1637,6 +1638,20 @@ func (provider *OpenAIProvider) Responses(ctx *schemas.BifrostContext, key schem
 	)
 }
 
+// rawResponsesCacheCountersKnown preserves one fact that
+// BifrostResponsesResponse.WithDefaults otherwise destroys: whether the raw
+// provider response actually contained an input-token detail object. Only the
+// closed boolean is retained in request context; raw content never
+// leaves this parser boundary.
+func rawResponsesCacheCountersKnown(body []byte, streaming bool) bool {
+	path := "usage.input_tokens_details"
+	if streaming {
+		path = "response.usage.input_tokens_details"
+	}
+	detail := gjson.GetBytes(body, path)
+	return detail.Exists() && detail.IsObject()
+}
+
 // HandleOpenAIResponsesRequest handles a responses request to OpenAI's API.
 func HandleOpenAIResponsesRequest(
 	ctx *schemas.BifrostContext,
@@ -1683,6 +1698,12 @@ func HandleOpenAIResponsesRequest(
 			return nil, lpErr
 		}
 		if len(lpResult.ResponseBody) > 0 {
+			// The large-request passthrough returns before the normal response parser
+			// below. Preserve the same raw presence fact at this boundary before the
+			// returned typed response can reach WithDefaults in the caller.
+			ctx.SetValue(schemas.BifrostContextKeyRawResponsesCacheCountersKnown,
+				rawResponsesCacheCountersKnown(lpResult.ResponseBody, false))
+
 			response := &schemas.BifrostResponsesResponse{}
 			if err := sonic.Unmarshal(lpResult.ResponseBody, response); err != nil {
 				return nil, providerUtils.NewBifrostOperationError(schemas.ErrProviderResponseUnmarshal, err)
@@ -1750,6 +1771,11 @@ func HandleOpenAIResponsesRequest(
 			ExtraFields: schemas.BifrostResponseExtraFields{Latency: lpResult.Latency},
 		}, nil
 	}
+
+	// Preserve raw field presence before HandleProviderResponse / WithDefaults
+	// can materialize an empty InputTokensDetails object.
+	ctx.SetValue(schemas.BifrostContextKeyRawResponsesCacheCountersKnown,
+		rawResponsesCacheCountersKnown(body, false))
 
 	response := &schemas.BifrostResponsesResponse{}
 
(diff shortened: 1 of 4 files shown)


core/providers/openai/responses_cache_counter_presence_test.go
```diff
@@ -0,0 +1,104 @@
...
+		body      string
+		streaming bool
+		want      bool
+	}{
+		{"unary present even when empty", `{"usage":{"input_tokens_details":{}}}`, false, true},
+		{"unary missing", `{"usage":{"input_tokens":10}}`, false, false},
+		{"unary null", `{"usage":{"input_tokens_details":null}}`, false, false},
+		{"unary scalar is not a detail object", `{"usage":{"input_tokens_details":0}}`, false, false},
+		{"stream present", `{"type":"response.completed","response":{"usage":{"input_tokens_details":{"cached_tokens":0}}}}`, true, true},
+		{"stream missing", `{"type":"response.completed","response":{"usage":{"input_tokens":10}}}`, true, false},
```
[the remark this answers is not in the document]
valentinyanakiev: Addressed in 9f94efccc. The table-driven stream-boundary test now covers completed-with-details (key present/true), incomplete-without-details (key present/false), and truncation before a terminal event (key absent). Focused tests, the full OpenAI provider suite, and go vet all pass locally.

valentinyanakiev:
@akshaydeo @Pratham-Mishra04 @roroghost17 I know this one is only a day old, so no rush; just flagging that it is ready. Head `9f94efcc` is on current `dev`, all checks green, CodeRabbit approved after one round. It pairs with #6867 and is independent of the other open PRs from me. Happy to take feedback whenever you get to it.

core/providers/openai/openai.go
```diff
@@ -1683,6 +1698,12 @@ func HandleOpenAIResponsesRequest(
 			return nil, lpErr
 		}
 		if len(lpResult.ResponseBody) > 0 {
+			// The large-request passthrough returns before the normal response parser
+			// below. Preserve the same raw presence fact at this boundary before the
+			// returned typed response can reach WithDefaults in the caller.
+			ctx.SetValue(schemas.BifrostContextKeyRawResponsesCacheCountersKnown,
```
akshaydeo, line 1704: instead of storing it in context - we can extend the base BifrostResponse struct to carry these fields ? cc @Pratham-Mishra04 what do you think?

valentinyanakiev: Agreed — and having gone back through it, your way is better than I gave it credit for. Two reasons I got wrong when I chose context:

1. I assumed context was the safe place because `WithDefaults` is what destroys the evidence (it materializes `Usage.InputTokensDetails` into `&ResponsesResponseInputTokens{CachedReadTokens: 0, CachedWriteTokens: 0}` when the raw body had no such object, so "absent" becomes "observed zero"). But `WithDefaults` copies `ExtraFields` straight through — `result.ExtraFields = resp.ExtraFields` — so a struct field is equally safe from the exact normalization this PR exists to work around. That was the main thing holding me on context, and it doesn't hold.

2. A context value only reaches in-process Go plugins. `ExtraFields` serializes, so HTTP transport consumers get the fact too. For a field whose whole purpose is telling a downstream consumer "this zero is real" vs "this zero is manufactured", that's a straightforward win.

Concretely I'd suggest `BifrostResponseExtraFields` rather than a new top-level field on `BifrostResponse`, since this is provider-derived response metadata and that's what the struct is for. Two existing precedents make it fit cleanly:

- `UpstreamLatency *int64` — "Nil when the request never accumulated one; nil means unknown, not zero." That tri-state is exactly what this needs. A plain `bool` can't represent a stream that truncates before a terminal event, which is the case the current PR handles by leaving the context value unset. A `*bool` carries it honestly.
- `DroppedUnsupportedTools`, documented as populated only by Bedrock — so a field that only one provider path sets is already an accepted shape here, which was my other worry about putting an OpenAI-Responses concept on a shared struct.

So: `RawResponsesCacheCountersKnown *bool` on `BifrostResponseExtraFields`, nil meaning unknown, set at the same three raw-decoder boundaries this PR already uses (unary, the large-request passthrough early return, and the terminal `response.completed`/`response.incomplete` event).

One thing worth deciding explicitly rather than by accident. Context is request-scoped; `ExtraFields` is per-chunk. The fact is only knowable at the terminal event, so on streaming it would be non-nil on the terminal chunk and nil on every chunk before it. I think that's correct and reads naturally under "nil means unknown" — but if you'd rather it be stamped on every chunk once known, that's a different implementation and better decided now than after merge.

Happy to push that as a new commit here, dropping the `BifrostContextKey` entirely so there's one mechanism rather than two. Just say which way you want the streaming case and I'll do it. @Pratham-Mishra04 any preference?

PR #6942, opened 2026-09-07
fix: bedrock dropping reasoning sign

### LINKED ISSUE #6624: Bedrock reasoning signature field is dropped for Anthropic models, which require it present
## Summary

On the Bedrock Converse path, `reasoningSignatureForBedrock`
(`core/providers/bedrock/utils.go`) drops the reasoning `signature` field whenever the signature is
nil or empty. That is correct for Nova/MiniMax but wrong for Anthropic models, which require the
field to be **present** on a thinking block. The two families want opposite serialization for the
same input, and the decision is currently made without reference to the model.

Present on `main` (`d9b9c9a66`) and in `transports/v2.0.0`.

```go
func reasoningSignatureForBedrock(sig *string) *string {
	if sig == nil || *sig == "" {
		return nil
	}
	return sig
}
```

## The two behaviours

- **Anthropic/Claude** rejects a thinking block with the field absent:
  `messages.N.content.M.thinking.signature: Field required`
- **Nova/MiniMax** rejects the field present but empty:
  `This model doesn't support the reasoningContent.reasoningText.signature field`

So a single value cannot satisfy both, and the current `nil` result 400s Anthropic in the cases
below.

## When it bites

Only when a reasoning block is emitted with **no signature to put in it** — i.e. no thinking text
to sign. The paths we hit are replayed assistant turns where the reasoning came from a summary, or
from the streaming ingress (which can emit a reasoning detail carrying only a signature delta), and
the client faithfully replays it on the next turn.

We should be upfront that we do **not** have a minimal standalone reproduction. Hand-writing a
thinking block with real prose and an empty signature fails for a different reason — the signature
is a crypto token over the text, so Claude rejects the *value* (`Invalid signature in thinking
block`) regardless of how it is serialized. What we have is production evidence: our gateway sits in
front of Bedrock, and without a model-aware fix we see the `Field required` 400 on replayed
reasoning. If you can suggest a construction that reaches this deterministically we are glad to test
it.

## A trap for whoever fixes it

The same helper's `nil` return is used as **control flow**, not only for field serialization:

- `core/providers/bedrock/responses.go`: `signature := reasoningSignatureForBedrock(msg.ResponsesReasoning.EncryptedContent)`,
  then `if signature != nil { ... }` decides whether a signature-only block is emitted at all.
- `core/providers/bedrock/invoke.go`: `if sig := reasoningSignatureForBedrock(block.Signature); sig != nil { ... }`

Making that helper return `""` for Anthropic would therefore also make every Anthropic request
fabricate an empty thinking block, and would break the existing
`empty encrypted content string emits nothing` case in `reasoning_replay_test.go`. A fix needs to
keep "is this a real signature?" separate from "how do I serialize the field on a block I have
already decided to emit?".

## Suggested direction (and a question)

We have a working fix, but we are unsure it is the shape you want, which is why this is an issue
rather than a PR. Ours resolves the model family once per request and stashes a flag on the request
context, read back by a second helper at the field-serialization sites only.

Given that v2.0.0 moved this area toward datasheet-backed capability resolution (#6281, #6492), a
capability such as `requires_present_reasoning_signature` looks more in keeping with where the code
is going than a context flag — but we could not find an existing capability covering it.

Which would you prefer? We are happy to send a PR in whichever shape you want, with tests covering
both family branches and a guard against the control-flow trap above.

akshaydeo:
## Summary

Claude on Bedrock Converse verifies the signature on every thinking block it receives and rejects any block without one — field absent gives `"thinking.signature: Field required"`, field present but empty gives `"Invalid signature"`. There is no wire shape that makes an unsigned block acceptable to Claude. This means that when a conversation is replayed through Bedrock Converse to a Claude model (e.g. after a mid-conversation model switch, or from an agentic client that strips signatures), unsigned reasoning blocks must be dropped entirely rather than forwarded. Nova and MiniMax behave oppositely — they reject a present-but-empty signature — so they continue receiving unsigned blocks with the field omitted.

## Changes

- `convertBifrostReasoningToBedrockReasoning` gains a `requireSigned bool` parameter; when true, any `reasoningText` block without a valid signature is skipped rather than emitted, and multi-summary entries beyond the first (which carries the signature) are also dropped
- `converseRequiresSignedReasoning(model)` is introduced in `utils.go` to resolve the flag: the datasheet row wins when present, otherwise falls back to `IsAnthropicModel` family detection
- The same guard is applied in `convertBifrostResponsesMessageContentBlocksToBedrockContentBlocks` for the Responses API path and in `convertMessage` for the Chat Completions path
- `ModelCapabilities` gains `BedrockRequiresSignedReasoning *bool` (`bedrock_requires_signed_reasoning`) so individual model rows can override the family default in either direction
- `ModelCaps.BedrockRequiresSignedReasoning(fallback bool)` resolves the capability with the same pattern as `BedrockReasoningShape`
- The existing `TestReasoningSignatureEchoedOnlyWhenNonEmpty` test is moved to a Nova model ID, since on Claude an unsigned block is now dropped before the signature check is reached
- New test file `unsignedreasoning_test.go` pins the behaviour end-to-end for both the Chat and Responses paths across Claude and Nova, covering unsigned drop, signed keep, multi-summary truncation, and Nova pass-through
- New tests `TestConverseRequiresSignedReasoningFamilyFallback` and `TestConverseRequiresSignedReasoningDatasheetWinsOverFamily` cover the resolution logic

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [x] Core (Go)
- [ ] Transports (HTTP)
- [x] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

```sh
go test ./core/providers/bedrock/... -run "TestUnsignedReasoningReplay|TestConverseRequiresSignedReasoning|TestReasoningSignatureEchoedOnlyWhenNonEmpty"
go test ./core/...
```

The new tests in `unsignedreasoning_test.go` cover the regression directly. Probed live against `global.anthropic.claude-sonnet-4-6` on Converse on 2026-09-07 to confirm both rejection shapes.

## Breaking changes

- [ ] Yes
- [x] No

## Related issues

Closes #6624

## Security considerations

No auth, secrets, or PII implications. The change ensures that opaque encrypted thinking signatures Claude uses to verify its own output are not stripped or forwarded in an invalid state, which is consistent with Anthropic's documented security model for thinking blocks.

## Checklist


akshaydeo:
* **#6959** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/6959?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#6961](https://github.com/maximhq/bifrost/pull/6961) <a href="https://app.graphite.com/github/pr/maximhq/bifrost/6961?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#7140](https://github.com/maximhq/bifrost/pull/7140) <a href="https://app.graphite.com/github/pr/maximhq/bifrost/7140?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#6942** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/6942?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/maximhq/bifrost/6942?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `dev`


<h2></h2>

akshaydeo:
### Merge activity

* **Sep 8, 5:13 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/6942).
* **Sep 8, 5:14 AM UTC**: @akshaydeo merged this pull request with [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/6942).

PR #6948, opened 2026-09-07
fix: add custom provider toggle to differentiate in providers who send done marker

### LINKED ISSUE #6784: [Bug]: Chat completion stream hangs forever after finish_reason when upstream omits [DONE] but keeps sending heartbeats
### Prerequisites

- [x] I reproduced this against the latest `dev` commit `ece3ae0ca35c51814b7afce44a50a17f2d9a0e9e` (2026-09-03)

### Description

An OpenAI-compatible `/v1/chat/completions` stream can remain open indefinitely after Bifrost has already received a non-empty `finish_reason` if the upstream:

1. does not send `data: [DONE]`,
2. does not close the HTTP/SSE connection, and
3. continues sending SSE heartbeat comments often enough to stay below Bifrost's stream idle timeout.

Bifrost records the `finish_reason`, but for providers considered to send `[DONE]` it keeps reading until `[DONE]` or EOF. Upstream heartbeat bytes reset the raw-stream idle timer even though the SSE parser discards those comments. Meanwhile, the HTTP transport continues sending its own heartbeat to the downstream client, so both connections look alive while no terminal completion is delivered.

For tool-calling clients such as OpenCode, this can look like a permanently stuck tool call or sub-agent: the tool-call arguments are already complete upstream, but the final `finish_reason: "tool_calls"` chunk is not delivered because Bifrost is still waiting for `[DONE]`.

### Conditions required to reproduce

All of the following are required:

1. The request uses `/v1/chat/completions` with streaming enabled.
2. The upstream sends a non-empty finish reason, commonly `"tool_calls"` or `"stop"`.
3. The upstream never sends `data: [DONE]`.
4. The upstream keeps the HTTP/SSE connection open.
5. The upstream sends bytes such as `: ping\n\n` at an interval shorter than `stream_idle_timeout_in_seconds` (120 seconds by default).
6. The resolved provider name is treated as one that sends `[DONE]`.

Example upstream tail:

```text
data: {"choices":[{"delta":{"tool_calls":[{"index":0,"id":"call_1","type":"function","function":{"name":"read","arguments":"{"}}]},"finish_reason":null}]}

data: {"choices":[{"delta":{"tool_calls":[{"index":0,"function":{"arguments":"\"path\":\"a.txt\"}"}}]},"finish_reason":null}]}

data: {"choices":[{"delta":{},"finish_reason":"tool_calls"}]}

data: {"choices":[],"usage":{"prompt_tokens":1000,"completion_tokens":100,"total_tokens":1100}}

: ping

: ping

: ping

# continues forever; no data: [DONE], no EOF
```

### Steps to reproduce

Configure an OpenAI-compatible provider and point it at an SSE server that emits a completion and then heartbeats forever:

```go
func TestChatStreamFinishReasonThenHeartbeatHangs(t *testing.T) {
    server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
        flusher := w.(http.Flusher)

        fmt.Fprint(w, chatChunk("hello", nil))
        fmt.Fprint(w, chatChunk("", schemas.Ptr("tool_calls")))
        flusher.Flush()

        ticker := time.NewTicker(20 * time.Millisecond)
        defer ticker.Stop()
        for {
            select {
            case <-r.Context().Done():
                return
            case <-ticker.C:
                fmt.Fprint(w, ": ping\n\n")
                flusher.Flush()
            }
        }
    }))
    defer server.Close()

    provider := newStreamTestProvider(server.URL)
    ctx, cancel := schemas.NewBifrostContextWithCancel(context.Background())
    defer cancel()
    ctx.SetValue(schemas.BifrostContextKeyStreamIdleTimeout, 80*time.Millisecond)

    stream, bifrostErr := provider.ChatCompletionStream(
        ctx, passthroughPostHook, nil, testKey(), basicChatRequest(),
    )
    require.Nil(t, bifrostErr)

    // The content chunk arrives.
    <-stream

    // Current behavior: the stream is still open after more than 4x the configured
    // idle timeout because every heartbeat resets the raw read timer.
    select {
    case _, ok := <-stream:
        t.Fatalf("unexpected stream result: open=%v", ok)
    case <-time.After(350 * time.Millisecond):
        // Reproduced: no final finish_reason, no error, and the channel remains open.
    }
}
```

This was reproduced locally with an 80 ms idle timeout and 20 ms upstream heartbeats. The stream remained open after 350 ms and only terminated when the context was explicitly canceled.

#### 1. `finish_reason` does not terminate most provider streams

`HandleOpenAIChatCompletionStreaming` stores `finish_reason`, but only breaks immediately for providers explicitly registered as not sending `[DONE]`:

https://github.com/maximhq/bifrost/blob/ece3ae0ca35c51814b7afce44a50a17f2d9a0e9e/core/providers/openai/openai.go#L1441-L1482

```go
if choice.FinishReason != nil && *choice.FinishReason != "" {
    finishReason = choice.FinishReason
}

if !providerUtils.ProviderSendsDoneMarker(providerName) && finishReason != nil {
    break
}
```

For the common terminal chunk shape `delta: {}` plus `finish_reason`, the chunk is also not forwarded immediately. The final synthesized chunk is only emitted after the read loop exits.

#### 2. Unknown/custom providers default to requiring `[DONE]`

`ProviderSendsDoneMarker` only returns `false` for Cerebras, Perplexity, Bedrock, and Bedrock Mantle. Every other provider name defaults to `true`:

https://github.com/maximhq/bifrost/blob/ece3ae0ca35c51814b7afce44a50a17f2d9a0e9e/core/providers/utils/utils.go#L3488-L3499

This directly affects the built-in DeepSeek OpenAI-compatible path. It also affects custom OpenAI-compatible providers whose custom provider key is something like GLM, Qwen, HY, or ARouter, because the custom key is passed as `providerName` and falls through to the default branch.

The existing unit-test comment already describes the underlying failure mode:

https://github.com/maximhq/bifrost/blob/ece3ae0ca35c51814b7afce44a50a17f2d9a0e9e/core/providers/utils/utils_test.go#L2208-L2210

```go
// missing from the non-DONE case would make the stream loop in
// core/providers/openai/openai.go wait indefinitely for a [DONE] marker it never sends.
```

#### 3. Any upstream bytes reset the idle timeout

The idle timeout is reset before SSE framing is interpreted, whenever the underlying reader returns any bytes:

https://github.com/maximhq/bifrost/blob/ece3ae0ca35c51814b7afce44a50a17f2d9a0e9e/core/providers/utils/utils.go#L3157-L3183

```go
n, err = r.reader.Read(p)
if n > 0 {
    r.timer.Reset(r.timeout)
}
```

The SSE reader subsequently discards heartbeat comments:

https://github.com/maximhq/bifrost/blob/ece3ae0ca35c51814b7afce44a50a17f2d9a0e9e/core/providers/utils/sse.go#L131-L153

```go
if len(line) == 0 || line[0] == ':' {
    continue
}
```

Therefore heartbeat comments count as idle-timeout activity but do not make semantic stream progress.

#### 4. The downstream connection is also kept alive

The HTTP transport emits a downstream SSE heartbeat once per second while the provider stream channel remains open:

https://github.com/maximhq/bifrost/blob/ece3ae0ca35c51814b7afce44a50a17f2d9a0e9e/transports/bifrost-http/lib/streamreader.go#L224-L244

https://github.com/maximhq/bifrost/blob/ece3ae0ca35c51814b7afce44a50a17f2d9a0e9e/transports/bifrost-http/handlers/inference.go#L2081-L2097

The streaming HTTP client intentionally has no whole-response read/write timeout, so there is no independent maximum stream duration to break this loop:

https://github.com/maximhq/bifrost/blob/ece3ae0ca35c51814b7afce44a50a17f2d9a0e9e/core/providers/utils/utils.go#L1452-L1469

### Behavior matrix

| Upstream tail | Current result |
|---|---|
| `finish_reason`, then `[DONE]` | Normal completion |
| `finish_reason`, then clean EOF | Normal completion; `finish_reason` is accepted as a terminal signal after the loop exits |
| No `finish_reason`, no `[DONE]`, then EOF | Truncation error |
| `finish_reason`, then silence | Idle-timeout error after approximately `stream_idle_timeout_in_seconds` |
| `finish_reason`, then heartbeat forever | **Stream remains open indefinitely** |

An abrupt non-EOF connection failure after `finish_reason` can also produce a stream read error. A clean EOF after `finish_reason` does not produce a truncation error.

### Expected behavior

Once a complete semantic terminal signal (`finish_reason`) has been received, heartbeat comments must not be able to keep the request alive indefinitely.

Bifrost should still support trailing usage-only chunks, but it should finalize within a bounded period after semantic completion even if `[DONE]` is missing and the upstream connection remains open.

### Actual behavior

Bifrost keeps waiting for `[DONE]`. Upstream heartbeat bytes continually reset the 120-second idle timeout, Bifrost continually emits downstream heartbeats, and the request has no internal termination bound. The final `finish_reason`/usage chunk is never synthesized and the client can remain waiting forever.

### Suggested direction

Avoid changing all providers to immediately break on the first `finish_reason`, because some providers send a usage-only chunk between `finish_reason` and `[DONE]` (see #191).

A robust fix could:

1. Enter a `semantic completion seen` state after receiving all relevant non-empty finish reasons.
2. Continue reading for a short, bounded post-finish grace period to collect usage and `[DONE]`.
3. Do not let SSE comments/heartbeat frames reset that post-finish grace period.
4. Finalize normally when the grace period expires because a semantic terminal signal was already received.
5. Optionally expose a custom-provider capability such as `sends_done_marker`, while retaining the bounded safety mechanism.
6. Account for multiple choices if `n > 1` is supported, rather than treating only `choices[0]` as globally terminal.

This is separate from #6783: in the heartbeat case the provider read loop never ends, so no terminal error event is generated. #6783 only becomes relevant after an error path has actually been reached.

### Affected area(s)

Core (Go), Transports (HTTP)

### Version

Latest `dev`: `ece3ae0ca35c51814b7afce44a50a17f2d9a0e9e` (2026-09-03)

### Environment

```text
OS: macOS (darwin/arm64)
Go workspace requirement: Go 1.27.0
API: /v1/chat/completions streaming
Default stream idle timeout: 120 seconds
Affected provider behavior: OpenAI-compatible SSE stream, provider considered to send [DONE]
```

### Relevant logs/output

```shell
No error is logged while upstream heartbeats continue. The provider stream goroutine remains blocked in ReadDataLine(), and the downstream connection continues receiving heartbeat comments.
```

### Regression?

Unknown. The current behavior is related to the fix for #191, which intentionally waits for `[DONE]` for providers that may emit a trailing usage-only chunk after `finish_reason`.

### Severity

TejasGhatte:
## Summary

Fixes a stream hang ([#6784](https://github.com/maximhq/bifrost/issues/6784)) where an OpenAI-compatible upstream that ends generation on `finish_reason` — without ever sending `data: [DONE]` — then parks the connection open with SSE heartbeat comments. Because heartbeat bytes reset the raw-read idle timer before SSE framing is interpreted, the read loop waits indefinitely for a marker that never arrives. A new `does_not_send_done_marker` field on `custom_provider_config` lets operators declare this upstream behaviour; Bifrost then terminates the stream as soon as a non-empty `finish_reason` is received.

## Changes

- Added `DoesNotSendDoneMarker bool` to `CustomProviderConfig` (schema, JSON, OpenAPI, config JSON schema).
- Added `BifrostContextKeyDoesNotSendDoneMarker` context key; stamped per attempt in `requestWorker` and cleared on each retry so a fallback cannot inherit the previous provider's opt-in.
- `ProviderSendsDoneMarker` now accepts a `*BifrostContext` and returns `false` when the key is set, taking priority over the built-in provider list. The opt-in can only end the loop earlier — it cannot make a provider that already terminates on `finish_reason` start waiting for `[DONE]`.
- Both the chat-completion and text-completion OpenAI streaming loops pass `ctx` through to `ProviderSendsDoneMarker`.
- `ReleaseStreamingResponse` short-circuits body draining when the opt-in is active: instead of blocking on a connection that will heartbeat forever, it closes the body stream with a sentinel error (`errStreamParkedAfterFinish`) that takes fasthttp's `CloseConn` path and keeps the half-read stream out of the idle pool.
- UI: added a "Does Not Send [DONE] Marker?" toggle to the custom provider API structure form, wired through the Zod schemas and TypeScript types.
- Docs: added a "Streaming Termination" section to the custom providers page with a warning that trailing usage chunks are dropped when the opt-in is active.

## Type of change

- [x] Bug fix
- [x] Feature

## Affected areas

- [x] Core (Go)
- [x] Transports (HTTP)
- [x] Providers/Integrations
- [x] UI (React)
- [x] Docs

## How to test

```sh
# Core/Transports
go test ./core/... ./core/providers/openai/... ./core/providers/utils/... ./core/providers/huggingface/...

# UI
cd ui
pnpm i
pnpm build
```

To reproduce the original hang manually: configure a custom provider backed by `openai` pointing at an upstream that sends `finish_reason` and then streams SSE heartbeats without `[DONE]`. Without the fix the stream never closes. With `does_not_send_done_marker: true` the stream closes immediately after the `finish_reason` chunk.

## Breaking changes

- [x] No

`ProviderSendsDoneMarker` signature changed from `(providerName)` to `(ctx, providerName)`. All internal call sites are updated; this function is not part of the public API surface.

## Related issues

Closes #6784

## Security considerations

None. The new field is operator-configured and does not affect authentication or secret handling.

## Checklist

### DIFF
--- a/core/bifrost.go
+++ b/core/bifrost.go
@@ -6872,6 +6872,12 @@ func (bifrost *Bifrost) requestWorker(provider schemas.Provider, config *schemas
 		req.Context.SetValue(schemas.BifrostContextKeyIsCustomProvider, !IsStandardProvider(baseProvider))
 		// Lets downstream converters resolve a custom provider key back to the built-in provider it wraps.
 		req.Context.SetValue(schemas.BifrostContextKeyBaseProviderType, baseProvider)
+		// Re-stamped per attempt so a fallback cannot inherit the previous provider's opt-in.
+		if config.CustomProviderConfig != nil && config.CustomProviderConfig.DoesNotSendDoneMarker {
+			req.Context.SetValue(schemas.BifrostContextKeyDoesNotSendDoneMarker, true)
+		} else {
+			req.Context.ClearValue(schemas.BifrostContextKeyDoesNotSendDoneMarker)
+		}
 
 		bifrost.endCoreSpan(workerSetupSpan)
 
--- a/core/bifrost_test.go
+++ b/core/bifrost_test.go
@@ -620,6 +620,96 @@ func TestHandleProviderRequest_OCROperationNotAllowed(t *testing.T) {
 	}
 }
 
+// https://github.com/maximhq/bifrost/issues/6784: an OpenAI-compatible upstream that ends
+// generation on finish_reason, never sends [DONE] and then parks the connection behind SSE
+// heartbeats holds the stream open indefinitely — heartbeat bytes reset the idle timer without
+// making semantic progress. custom_provider_config.does_not_send_done_marker is the operator's
+// declaration that finish_reason is terminal for this upstream; this pins the whole path, from
+// the account config through the per-attempt context stamp to the provider's read loop.
+func TestCustomProviderDoesNotSendDoneMarkerEndsParkedStream(t *testing.T) {
+	const chunk = `data: {"id":"chatcmpl-repro","object":"chat.completion.chunk","created":1,"model":"repro-model",` +
+		`"choices":[{"index":0,"delta":{"content":"hello"},"finish_reason":null}]}` + "\n\n" +
+		`data: {"id":"chatcmpl-repro","object":"chat.completion.chunk","created":1,"model":"repro-model",` +
+		`"choices":[{"index":0,"delta":{},"finish_reason":"tool_calls"}]}` + "\n\n"
+
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Content-Type", "text/event-stream")
+		fl, ok := w.(http.Flusher)
+		if !ok {
+			return
+		}
+		if _, err := w.Write([]byte(chunk)); err != nil {
+			return
+		}
+		fl.Flush()
+
+		ticker := time.NewTicker(20 * time.Millisecond)
+		defer ticker.Stop()
+		for {
+			select {
+			case <-r.Context().Done():
+				return
+			case <-ticker.C:
+				if _, err := w.Write([]byte(": ping\n\n")); err != nil {
+					return
+				}
+				fl.Flush()
+			}
+		}
+	}))
+	defer server.Close()
+
+	const customProvider = schemas.ModelProvider("custom-openai")
+	account := NewMockAccount()
+	account.AddProviderWithBaseURL(customProvider, 1, 1, server.URL)
+	account.configs[customProvider].NetworkConfig.MaxRetries = 0
+	account.SetCustomProviderConfig(customProvider, &schemas.CustomProviderConfig{
+		BaseProviderType:      schemas.OpenAI,
+		DoesNotSendDoneMarker: true,
+	})
+	account.SetKeysForProvider(customProvider, []schemas.Key{
+		{ID: "custom-key", Value: *schemas.NewSecretVar("sk-custom"), Models: schemas.WhiteList{"*"}, Weight: 100},
+	})
+
+	client := newStreamTestClient(t, account)
+	ctx, cancel := schemas.NewBifrostContextWithCancel(context.Background())
+	defer cancel()
+
+	stream, bifrostErr := client.ChatCompletionStreamRequest(ctx, &schemas.BifrostChatRequest{
+		Provider: customProvider,
+		Model:    "repro-model",
+		Input: []schemas.ChatMessage{{
+			Role:    schemas.ChatMessageRoleUser,
+			Content: &schemas.ChatMessageContent{ContentStr: Ptr("hi")},
+		}},
+	})
+	if bifrostErr != nil {
+		t.Fatalf("stream request failed: %v", bifrostErr)
+	}
+
+	type drained struct {
+		content string
+		errs    []string
+	}
+	done := make(chan drained, 1)
+	go func() {
+		content, errs := drainChatStream(stream)
+		done <- drained{content: content, errs: errs}
+	}()
+
+	select {
+	case result := <-done:
+		if len(result.errs) > 0 {
+			t.Fatalf("stream carried errors: %v", result.errs)
+		}
+		if result.content != "hello" {
+			t.Errorf("expected the streamed content to survive the early break, got %q", result.content)
+		}
+	case <-time.After(30 * time.Second):
+		t.Fatal("stream never terminated: does_not_send_done_marker did not reach the provider read loop")
+	}
+}
+
 // Test that transientServerStatusCodes are properly defined.
 // These are upstream-side failures unrelated to the credential — the same key is retried.
 func TestTransientServerStatusCodes(t *testing.T) {
--- a/core/providers/huggingface/chat_test.go
+++ b/core/providers/huggingface/chat_test.go
@@ -205,6 +205,6 @@ func TestToHuggingFaceChatCompletionStreamRequest_StreamOptions(t *testing.T) {
 // (`choices: []` plus top-level `usage`), which several inference providers
 // emit after the finish chunk — the stream then completes with zero tokens.
 func TestHuggingFaceSendsDoneMarker(t *testing.T) {
-	assert.True(t, providerUtils.ProviderSendsDoneMarker(schemas.HuggingFace),
+	assert.True(t, providerUtils.ProviderSendsDoneMarker(nil, schemas.HuggingFace),
 		"HuggingFace sends [DONE]; breaking on finish_reason drops the trailing usage chunk")
 }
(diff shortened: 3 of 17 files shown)


TejasGhatte:
* **#6948** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/6948?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/maximhq/bifrost/6948?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `dev`


<h2></h2>

akshaydeo (approved):

akshaydeo:
### Merge activity

* **Sep 7, 6:02 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/6948).
* **Sep 7, 6:02 PM UTC**: @akshaydeo merged this pull request with [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/6948).

PR #6958, opened 2026-09-07
feat(databricks): add Claude reasoning translation, image inlining, and Responses API fallback

impoiler:
### TL;DR

Fixes three gaps in the Databricks provider: Claude endpoints now receive reasoning as a `thinking` object instead of the rejected `reasoning_effort`, remote image URLs are fetched and inlined before reaching Claude endpoints that reject them, and the Responses API now falls back to Chat Completions emulation when a pay-per-token endpoint declines the native surface.

### What changed?

**Reasoning on Claude endpoints (`reasoning.go`)**
Claude-backed Databricks endpoints reject `reasoning_effort` and instead accept Anthropic's `thinking` object. A new `thinkingParam` function translates Bifrost's neutral reasoning parameters — effort labels, explicit budgets, or `reasoning.max_tokens` — into the correct `thinking` wire shape. Adaptive-only models receive `{"type": "adaptive"}` with no budget. `max_completion_tokens` is raised above the budget when the caller has not set one, satisfying the endpoint's requirement that the ceiling exceed the budget. A new `chatResponseHandler` and `normalizeReasoningBlocks` function lift the reasoning content blocks the endpoint returns onto Bifrost's standard `reasoning` and `reasoning_details` fields, for both unary responses and streaming deltas.

**Image inlining (`imageinlining.go`)**
Claude endpoints on Databricks reject remote image URLs with "Http data URLs are not supported by Claude". A new `inlineChatImageURLs` / `inlineResponsesImageURLs` pair fetches any `http(s)` image URL and replaces it with a base64 data URL before the request leaves Bifrost, using the same SSRF-safe downloader as the Bedrock and Anthropic providers. A failed fetch aborts the request with a 400 rather than silently dropping the image. Requests with no remote images are returned as-is with no copy.

**Responses API fallback (`inference.go`)**
Pay-per-token Model Serving endpoints commonly decline the native `/serving-endpoints/responses` route with "Responses API passthrough is not supported". The provider now catches that specific 400, replays the request through Chat Completions, and records the endpoint in a `sync.Map` so subsequent requests for the same workspace host and model skip the failing round trip. Other 400 errors are surfaced to the caller unchanged. The Unity AI Gateway continues to be emulated unconditionally.

**`parallel_tool_calls` default changed**
Both Databricks surfaces reject `parallel_tool_calls` outright. The default for models with no datasheet row is now `false` (was `true`); a datasheet row with `supports_parallel_function_calling: true` still opts in.

### How to test?

- Run the new unit tests: `go test ./core/providers/databricks/...`
- `TestThinkingParam` covers the budget/effort translation matrix.
- `TestNormalizeReasoningBlocks` covers the content-block rewrite for unary and streaming shapes.
- `TestDatabricksClaudeReasoningOverTheWire` drives a stub server end-to-end for both unary and streaming reasoning.
- `TestInlineChatImageURLs` / `TestInlineResponsesImageURLs` cover inlining, passthrough of data URLs, fetch failures, and extension-based media type fallback.
- `TestDatabricksResponsesFallsBackWhenPassthroughUnsupported` / `TestDatabricksResponsesStreamFallsBackWhenPassthroughUnsupported` verify the fallback path and the skip-on-second-request behaviour.
- `TestDatabricksResponsesOtherErrorsAreNotRetried` pins that only the surface-not-supported rejection triggers emulation.
- The existing `TestDatabricks` and `TestDatabricksChatParamsFollowDatasheet` suites are updated to reflect the new `parallel_tool_calls` default.

### Why make this change?

Claude endpoints on Databricks have three behaviours that differ from the OpenAI-shaped contract Bifrost assumes: they reject `reasoning_effort` in favour of `thinking`, they reject remote image URLs, and many of them do not expose the Responses API surface at all. Without these fixes, reasoning requests silently lost their reasoning, image requests failed with an opaque upstream error, and Responses requests failed entirely on pay-per-token endpoints. The changes bring Claude-on-Databricks to feature parity with the native Anthropic and Bedrock providers for these capabilities.

### DIFF
--- a/core/internal/llmtests/account.go
+++ b/core/internal/llmtests/account.go
@@ -505,6 +505,10 @@ func (account *ComprehensiveTestAccount) GetKeysForProvider(ctx context.Context,
 				DatabricksKeyConfig: &schemas.DatabricksKeyConfig{
 					WorkspaceURL: *schemas.NewSecretVar("env.DATABRICKS_WORKSPACE_URL"),
 				},
+				// Databricks has no batch surface. The flag is still required so the
+				// router hands batch and file requests to the provider, whose
+				// unsupported_operation answer is what the harness asserts on.
+				UseForBatchAPI: bifrost.Ptr(true),
 			},
 		}, nil
 	case schemas.GithubCopilot:
--- a/core/providers/databricks/databricks.go
+++ b/core/providers/databricks/databricks.go
@@ -67,6 +67,11 @@ type DatabricksProvider struct {
 	networkConfig       schemas.NetworkConfig // Network configuration including extra headers
 	sendBackRawRequest  bool                  // Whether to include raw request in BifrostResponse
 	sendBackRawResponse bool                  // Whether to include raw response in BifrostResponse
+
+	// responsesUnsupported records "<workspace host>|<model>" pairs whose Model Serving
+	// endpoint has declined the native Responses API, so those requests go straight to
+	// chat-completion emulation. See inference.go.
+	responsesUnsupported sync.Map
 }
 
 // NewDatabricksProvider creates a new Databricks provider instance.
--- a/core/providers/databricks/databricks_test.go
+++ b/core/providers/databricks/databricks_test.go
@@ -57,8 +57,9 @@ func TestDatabricks(t *testing.T) {
 			ImageBase64:           true,
 			CompleteEnd2End:       true,
 			Embedding:             true,
-			ListModels:            true,
 			Reasoning:             true,
+			// listModelsByKey intentionally returns nothing; the datasheet is the catalog.
+			ListModels: false,
 			// Text completions are not exposed by either Databricks surface.
 			TextCompletion:       false,
 			TextCompletionStream: false,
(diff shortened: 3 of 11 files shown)


impoiler:
* **#6958** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/6958?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/maximhq/bifrost/6958?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `dev`


<h2></h2>

akshaydeo:
### Merge activity

* **Sep 7, 7:55 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/6958).
* **Sep 7, 7:56 PM UTC**: @akshaydeo merged this pull request with [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/6958).

PR #7013, opened 2026-09-09
fix: stop bifrost_active_requests drifting negative on streaming

### LINKED ISSUE #7005: [Bug]: bifrost_active_requests drifts negative on streaming — per-chunk post-hooks decrement on a sticky StreamEndIndicator
### Description

`bifrost_active_requests` drifts **negative** on streaming request types. Observed in production:

```
bifrost_active_requests{method="chat_completion_stream"} -9
```

An in-flight gauge cannot legitimately be negative. The cause is an increment/decrement asymmetry: the gauge is incremented **once per request**, but decremented **once per post-hook invocation that observes a sticky context flag** — and for streaming, post-hooks run **per chunk**.

**Increment — once per request:**

```go
// plugins/telemetry/main.go — PreLLMHook
func (p *PrometheusPlugin) PreLLMHook(ctx *schemas.BifrostContext, req *schemas.BifrostRequest) (...) {
	ctx.SetValue(startTimeKey, time.Now())
	ctx.SetValue(activeRequestTypeKey, req.RequestType)
	p.ActiveRequests.WithLabelValues(string(req.RequestType)).Inc()
	return req, nil, nil
}
```

**Decrement — every post-hook call where the flag is set:**

```go
// plugins/telemetry/main.go — PostLLMHook
streamEndIndicatorValue := ctx.Value(schemas.BifrostContextKeyStreamEndIndicator)
isFinalChunk, hasFinalChunkIndicator := streamEndIndicatorValue.(bool)

isStreamFinal := !bifrost.IsStreamRequestType(requestType) || (hasFinalChunkIndicator && isFinalChunk)
if isStreamFinal {
	if method, ok := ctx.Value(activeRequestTypeKey).(schemas.RequestType); ok {
		p.ActiveRequests.WithLabelValues(string(method)).Dec()
	}
}
```

**And post-hooks run per chunk:**

```go
// core/providers/utils/utils.go — ProcessAndSendResponse, called for every stream chunk
processedResponse, processedError := postHookRunner(ctx, response, nil)
```

The guard is correct only while `BifrostContextKeyStreamEndIndicator` is set exactly once, on the genuinely final chunk. But that key is **sticky on a shared per-request context**:

- It is **set** in 50+ places — every provider's terminal-chunk path, `tryStreamRequest`'s error branch, `HandleStreamCancellation`, `SendStreamTruncatedError`.
- It is **cleared in exactly one place repo-wide**: `clearCtxForFallback` (`core/utils.go`). Nowhere else.

Once set, every subsequent post-hook invocation on that context sees `hasFinalChunkIndicator && isFinalChunk` and decrements again.

### Two concrete paths that produce it

**1. Client disconnects mid-stream** (matches our production correlation: the negative gauge appears alongside ~1,275 HTTP 499s)

1. `PreLLMHook` → `Inc` (+1)
2. Client disconnects. `HandleStreamCancellation` sets `StreamEndIndicator = true` and runs terminal post-hooks → `Dec` (0) ✅
3. The provider goroutine is still producing. Each already-queued chunk calls `ProcessAndSendResponse` → `postHookRunner` → `PostLLMHook`, and the flag is **still true** → `Dec`, `Dec`, … (−1, −2, …) ❌

`HandleStreamCancellation` guards itself with `GetAndSetValue`, so it cannot double-fire — but nothing guards the ordinary per-chunk path against a flag another path already set.

**2. Streaming retry after a first-chunk error**

1. `PreLLMHook` → `Inc` (+1) — note `PreLLMHook` runs once per `tryRequest`, *not* per retry attempt
2. Attempt 1 fails on the first chunk. Its terminal handling sets `StreamEndIndicator = true` → `Dec` (0)
3. `executeRequestWithRetries` retries. It clears `BifrostContextKeyConnectionClosed` but **not** `StreamEndIndicator`:
   ```go
   bifrostError = firstChunkErr
   ```
4. The retry streams successfully — and **every one of its chunks** sees the stale flag → `Dec` per chunk ❌

Fallbacks are unaffected, because `clearCtxForFallback` does clear the key. That asymmetry — cleared for fallbacks, not for retries — looks like the oversight.

### Secondary: a leak in the opposite direction

`PostLLMHook` returns early without decrementing when `startTime` is missing:

```go
startTime, ok := ctx.Value(startTimeKey).(time.Time)
if !ok {
	p.logger.Warn("Warning: startTime not found in context for Prometheus PostLLMHook")
	return result, bifrostErr, nil          // ← ActiveRequests never decremented
}
```

Any path reaching `PostLLMHook` without `PreLLMHook` having run (a short-circuit before telemetry's pre-hook, a plugin-order change) leaks the gauge upward permanently.

### Steps to reproduce

1. Enable the telemetry plugin; scrape `/metrics`.
2. Start a streaming chat completion against a provider with a reasonably long generation.
3. Disconnect the client mid-generation (`curl --max-time 2`), so the provider goroutine still has chunks queued.
4. Repeat ~20×.
5. `curl -s localhost:8080/metrics | grep bifrost_active_requests`

Observed: the gauge for `method="chat_completion_stream"` goes and stays negative, drifting further with each repetition. Expected: it returns to 0 once no requests are in flight.

### Expected behavior

`bifrost_active_requests` returns to 0 when no requests are in flight, and is never negative.

### Actual behavior

Drifts negative on streaming paths; the value is unusable for capacity signalling, autoscaling or admission control.

### Affected area(s)

Plugins, Core (Go)

### Version

### Environment

```text
- Not platform-specific; observed in a production deployment and traced in code.
```

### Regression?

Not a regression as far as I can tell — appears to date from when the per-chunk `isStreamFinal` guard was introduced.

### Suggested fix

Make the decrement **idempotent per request** rather than conditional on a sticky shared flag — a one-shot latch on the context, so it cannot fire twice regardless of how many post-hooks observe a set indicator:

```go
if isStreamFinal {
	// Latch: only the first observer of the terminal state decrements.
	if prev := ctx.GetAndSetValue(activeRequestDecrementedKey, true); prev == nil {
		if method, ok := ctx.Value(activeRequestTypeKey).(schemas.RequestType); ok {
			p.ActiveRequests.WithLabelValues(string(method)).Dec()
		}
	}
}
```

`GetAndSetValue` is already used for exactly this purpose in `HandleStreamCancellation`, so the primitive and the precedent both exist.

Two smaller companion fixes worth considering:

- Clear `BifrostContextKeyStreamEndIndicator` on the streaming retry path alongside `ConnectionClosed`, matching what `clearCtxForFallback` already does for fallbacks. Independently of this gauge, a stale terminal flag on a retried stream is likely to confuse other per-chunk consumers.
- Decrement before the `startTime` early return (or latch the increment/decrement as a matched pair), so a missing `startTime` cannot leak the gauge upward.

Happy to open a PR.

siddiqueirshad:
## Summary

Fixes #7005 — `bifrost_active_requests` drifts negative on streaming request types because `PostLLMHook` runs per chunk while `BifrostContextKeyStreamEndIndicator` is sticky on the shared request context. After a client disconnect (or a first-chunk stream error before retry), every subsequent post-hook observes the terminal flag and `Dec()`s again.

**Root cause:** the gauge is incremented once in `PreLLMHook`, but decremented on every post-hook invocation that sees `StreamEndIndicator == true`. That flag is cleared for fallbacks (`clearCtxForFallback`) but not for same-provider streaming retries, and nothing made the telemetry `Dec` idempotent.

## Changes

- `plugins/telemetry/main.go`: one-shot latch (`GetAndSetValue`) around `ActiveRequests.Dec`; clear the latch on each `PreLLMHook` so fallback attempts can Inc/Dec correctly; also Dec on the missing-`startTime` early return so the gauge cannot leak upward.
- `core/utils.go`: add `clearCtxForStreamRetry` to wipe stream-lifecycle flags (`ConnectionClosed`, `StreamEndIndicator`, `StreamBodyExhausted`, `StreamParkedAfterFinish`).
- `core/bifrost.go`: call `clearCtxForStreamRetry` after a first-chunk stream error; clear `StreamEndIndicator` at the start of every streaming retry (not only the Azure-preamble path).
- Regression tests in `plugins/telemetry/main_test.go` and `core/utils_test.go`.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [x] Core (Go)
- [ ] Transports (HTTP)
- [ ] Providers/Integrations
- [x] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

```sh
# With go.work (or CI workspace) linking local core + plugins/telemetry:
go test -count=1 -run 'TestActiveRequests' ./plugins/telemetry/
go test -count=1 -run 'TestClearCtxForStreamRetry|TestClearCtxForFallback$' ./core/
```

Red-before-green: `TestActiveRequestsStreamEndIndicatorDoesNotDoubleDecrement` fails on `dev` (gauge goes negative after sticky-flag PostLLMHook repeats) and passes with this change.

## Breaking changes

- [ ] Yes
- [x] No

## Related issues

Closes #7005

## Security considerations

None. Metrics-only; no auth, PII, or request-path behavior change beyond clearing sticky stream flags that should not survive a retry (matching existing fallback cleanup).

## Checklist

### DIFF
--- a/core/bifrost.go
+++ b/core/bifrost.go
@@ -6434,10 +6434,15 @@ func executeRequestWithRetries[T any](
 		}
 
 		// The previous failed stream has drained before reaching this retry.
-		if checkAzurePreamble && attempts > 0 {
+		// Clear StreamEndIndicator for every streaming retry (not only Azure): a
+		// sticky terminal flag on the shared context makes per-chunk post-hooks
+		// treat every subsequent chunk as final (e.g. bifrost_active_requests).
+		if IsStreamRequestType(requestType) && attempts > 0 {
 			ctx.ClearValue(schemas.BifrostContextKeyStreamEndIndicator)
-			ctx.ClearValue(schemas.BifrostContextKeyStreamBodyExhausted)
-			ctx.ClearValue(schemas.BifrostContextKeyStreamParkedAfterFinish)
+			if checkAzurePreamble {
+				ctx.ClearValue(schemas.BifrostContextKeyStreamBodyExhausted)
+				ctx.ClearValue(schemas.BifrostContextKeyStreamParkedAfterFinish)
+			}
 		}
 
 		// Attempt the request
@@ -6463,12 +6468,10 @@ func executeRequestWithRetries[T any](
 				}
 				if firstChunkErr != nil {
 					<-drainDone
-					// The dead stream's teardown (ReleaseStreamingResponse) claimed the
-					// connection_closed flag on the shared context. That claim is scoped
-					// to the response it released; clear it so the retry or fallback
-					// attempt that follows doesn't see its own fresh stream as already
-					// closed and fail every read with ErrStreamClosed.
-					ctx.ClearValue(schemas.BifrostContextKeyConnectionClosed)
+					// The dead stream's teardown left stream-lifecycle flags on the
+					// shared context; wipe them so the retry does not inherit a
+					// sticky terminal / closed state (see clearCtxForStreamRetry).
+					clearCtxForStreamRetry(ctx)
 					bifrostError = firstChunkErr
 				} else if checkedStream == nil {
 					// Empty stream (zero chunks before close — includes the large-payload
--- a/core/utils.go
+++ b/core/utils.go
@@ -409,6 +409,19 @@ func clearCtxForFallback(ctx *schemas.BifrostContext) {
 	ctx.ClearValue(schemas.BifrostContextKeySupportsAssistantPrefill)
 }
 
+// clearCtxForStreamRetry clears stream-lifecycle flags left on the shared request
+// context by a failed streaming attempt before the same-provider retry starts.
+// Without this, a sticky StreamEndIndicator makes every subsequent chunk look
+// terminal to per-chunk post-hooks (e.g. telemetry's ActiveRequests Dec).
+// Mirrors the stream-key subset of clearCtxForFallback; ConnectionClosed is
+// included because ReleaseStreamingResponse on the dead attempt claims it.
+func clearCtxForStreamRetry(ctx *schemas.BifrostContext) {
+	ctx.ClearValue(schemas.BifrostContextKeyConnectionClosed)
+	ctx.ClearValue(schemas.BifrostContextKeyStreamEndIndicator)
+	ctx.ClearValue(schemas.BifrostContextKeyStreamBodyExhausted)
+	ctx.ClearValue(schemas.BifrostContextKeyStreamParkedAfterFinish)
+}
+
 // ClearContextForInternalRequest clears context state that is specific to the
 // caller's original request, so a context derived from it can carry an
 // internal sub-request (e.g. a plugin generating an embedding for its own
--- a/core/utils_test.go
+++ b/core/utils_test.go
@@ -417,6 +417,40 @@ func TestClearCtxForFallback(t *testing.T) {
 	}
 }
 
+// A same-provider streaming retry reuses the request context. Stream-lifecycle flags
+// claimed by the failed attempt must be wiped first; otherwise every chunk of the
+// retry looks terminal to per-chunk post-hooks (#7005).
+func TestClearCtxForStreamRetry(t *testing.T) {
+	cleared := []schemas.BifrostContextKey{
+		schemas.BifrostContextKeyConnectionClosed,
+		schemas.BifrostContextKeyStreamEndIndicator,
+		schemas.BifrostContextKeyStreamBodyExhausted,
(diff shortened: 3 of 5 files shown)


akshaydeo:
please sign the CLA @siddiqueirshad 

https://github.com/maximhq/bifrost/pull/7013#issuecomment-5598220040

siddiqueirshad:
@akshaydeo  sorry i was , i have signed please reveiw

siddiqueirshad:
Hi @G-XD, when you get a chance, could you please take a look at this PR? Thanks!

PR #7018, opened 2026-09-09
fix(telemetry): decrement active-requests gauge exactly once per request

### LINKED ISSUE #7005: [Bug]: bifrost_active_requests drifts negative on streaming — per-chunk post-hooks decrement on a sticky StreamEndIndicator

Huang-404-Q:
## Description

`bifrost_active_requests` drifted **negative** on streaming request types (observed in production alongside ~1,275 HTTP 499s: `-9` on `method="chat_completion_stream"`).

The gauge is incremented once per request in `PreLLMHook`, but `PostLLMHook` decrements it whenever it observes the stream-end indicator — and that indicator is **sticky** on the shared per-request context: it is set in 50+ terminal-chunk/cancellation/error paths and cleared in exactly one (`clearCtxForFallback`). Since `PostLLMHook` runs once per stream chunk, every already-queued chunk after the flag was set decremented again.

Two concrete paths:

1. **Client disconnects mid-stream**: `HandleStreamCancellation` sets the flag and decrements (correct), but the provider goroutine's remaining queued chunks each see the stale flag and decrement again.
2. **Streaming retry after a first-chunk error**: attempt 1 sets the flag and decrements; `executeRequestWithRetries` clears `ConnectionClosed` but **not** `StreamEndIndicator`, so every retry chunk sees the stale flag. (Fallbacks are unaffected because `clearCtxForFallback` clears the key — the asymmetry suggests the retry case was the oversight.)

### Fix

Make the decrement **idempotent** with a dedicated context flag flipped via `ctx.GetAndSetValue` (the same primitive `HandleStreamCancellation` already uses to guard its own decrement). The decrement now happens exactly once, at the top of `PostLLMHook`.

This also fixes a second leak in the opposite direction: the early return on missing `startTime` used to skip the decrement entirely, so any path reaching `PostLLMHook` without `PreLLMHook` (short-circuit, plugin-order change) leaked the gauge upward permanently.

### Trade-off note

The gauge now decrements at hook entry rather than after the request's terminal bookkeeping — meaning the "active" window for a request is measured up to the first `PostLLMHook` invocation rather than the stream's true end. For streaming that is per-chunk anyway (the old code decremented on the final chunk's hook, not after the stream closed), so the semantic change is minimal and strictly better than a negative gauge.

## Testing

- `TestPostLLMHookActiveRequestsIdempotentDecrement`: sticky flag set, then 1 final + 5 queued-chunk hook invocations → gauge lands at exactly `0` (was `-5` pre-fix; verified red-green).
- `TestPostLLMHookActiveRequestsDecrementOnMissingStartTime`: missing `startTime` no longer leaks the increment (verified red-green).
- Full `plugins/telemetry` suite: `ok`; `go vet` / `gofmt` clean.

Fixes #7005

### DIFF
--- a/plugins/telemetry/main.go
+++ b/plugins/telemetry/main.go
@@ -1006,12 +1006,21 @@ func canonicalEntitySet(ctx context.Context, idsKey, namesKey, scalarIDKey, scal
 func (p *PrometheusPlugin) PostLLMHook(ctx *schemas.BifrostContext, result *schemas.BifrostResponse, bifrostErr *schemas.BifrostError) (*schemas.BifrostResponse, *schemas.BifrostError, error) {
 	requestType, provider, originalModel, resolvedModel := bifrost.GetResponseFields(result, bifrostErr)
 
-	// Decrement the in-flight gauge exactly once, as early as possible: every
-	// later early return (pre-dispatch rejection, missing startTime) would
-	// otherwise leak the gauge upward (#7005).
+	// Decompose the stream-end state here so every early return below still
+	// sees it: non-stream requests decrement immediately, streaming requests
+	// only on the confirmed final chunk (#7005).
+	streamEndIndicatorValue := ctx.Value(schemas.BifrostContextKeyStreamEndIndicator)
+	isFinalChunk, hasFinalChunkIndicator := streamEndIndicatorValue.(bool)
+	isStreamType := bifrost.IsStreamRequestType(requestType)
+	isStreamFinal := !isStreamType || (hasFinalChunkIndicator && isFinalChunk)
 	if method, ok := ctx.Value(activeRequestTypeKey).(schemas.RequestType); ok {
-		if prev, _ := ctx.GetAndSetValue(activeRequestsDecrementedKey, true).(bool); !prev {
-			p.ActiveRequests.WithLabelValues(string(method)).Dec()
+		if isStreamFinal {
+			// GetAndSetValue makes the decrement idempotent: once the terminal
+			// path has decremented (final chunk, cancellation handler), later
+			// chunk hooks observing the sticky flag see prev == true and skip.
+			if prev, _ := ctx.GetAndSetValue(activeRequestsDecrementedKey, true).(bool); !prev {
+				p.ActiveRequests.WithLabelValues(string(method)).Dec()
+			}
 		}
 	}
 
@@ -1103,12 +1112,9 @@ func (p *PrometheusPlugin) PostLLMHook(ctx *schemas.BifrostContext, result *sche
 	// Get label values in the correct order (cache_type will be handled separately for cache hits)
 	promLabelValues := getPrometheusLabelValues(append(p.defaultBifrostLabels, p.customLabels...), labelValues)
 
-	// Extract stream end indicator BEFORE the goroutine
-	streamEndIndicatorValue := ctx.Value(schemas.BifrostContextKeyStreamEndIndicator)
-	isFinalChunk, hasFinalChunkIndicator := streamEndIndicatorValue.(bool)
-
-	// Decrement active requests on the final (or only) call for this request
-	isStreamFinal := !bifrost.IsStreamRequestType(requestType) || (hasFinalChunkIndicator && isFinalChunk)
+	// Stream end state (streamEndIndicatorValue / isFinalChunk /
+	// isStreamFinal) was already decomposed at the top of the hook, where the
+	// ActiveRequests decrement happens.
 
 	pricingScopes := modelcatalog.PricingLookupScopesFromContext(ctx, string(provider))
 
--- a/plugins/telemetry/main.go
+++ b/plugins/telemetry/main.go
@@ -32,9 +32,16 @@ const (
 const (
 	startTimeKey         schemas.BifrostContextKey = "bf-prom-start-time"
 	activeRequestTypeKey schemas.BifrostContextKey = "bf-prom-active-req-type"
-	mcpStartTimeKey      schemas.BifrostContextKey = "bf-prom-mcp-start-time"
-	mcpClientNameKey     schemas.BifrostContextKey = "bf-prom-mcp-client-name"
-	mcpToolNameKey       schemas.BifrostContextKey = "bf-prom-mcp-tool-name"
+	// activeRequestsDecrementedKey marks that the in-flight gauge has already
+	// been decremented for this request. PostLLMHook runs once per stream
+	// chunk, and the stream-end indicator is sticky once any path sets it, so
+	// a plain bool check decrements once per chunk and drives the gauge
+	// negative. This flag is set-and-cleared atomically via GetAndSetValue,
+	// making the decrement idempotent per request (#7005).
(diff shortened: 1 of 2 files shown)


Huang-404-Q:
Fixed in e02559d: the decrement now applies the idempotent latch only for a non-stream request or a confirmed final stream chunk — intermediate chunks leave `bifrost_active_requests` untouched. The missing-`startTime` early return still reaches the top-of-hook decrement, so that leak stays fixed.

Regression test added: gauge stays at `1` through two intermediate chunks, drops to `0` on the final chunk, and a stale trailing chunk (queued behind the terminal one) does not decrement again.

(Note: this PR also picked up a stray commit from a sibling branch during local rebasing; the branch history is now clean — only the two telemetry commits.)

akshaydeo:
merging #7013

PR #7021, opened 2026-09-09
fix(core): clear provider response headers on fallback boundaries

### LINKED ISSUE #6973: [Bug]: provider response headers leak across fallback boundaries (clearCtxForFallback misses ProviderResponseHeaders)
### Description

`clearCtxForFallback` exists to wipe context state that was resolved for the *previous* provider before a fallback attempt runs. It clears ten keys — but **not** `BifrostContextKeyProviderResponseHeaders`:

```go
// core/utils.go
func clearCtxForFallback(ctx *schemas.BifrostContext) {
	ctx.ClearValue(schemas.BifrostContextKeyAPIKeyID)
	ctx.ClearValue(schemas.BifrostContextKeyAPIKeyName)
	ctx.ClearValue(schemas.BifrostContextKeyDirectKey)
	ctx.ClearValue(schemas.BifrostContextKeyRoutingPinnedAPIKeyID)
	ctx.ClearValue(schemas.BifrostContextKeyGovernanceIncludeOnlyKeys)
	ctx.ClearValue(schemas.BifrostContextKeyChangeRequestType)
	ctx.ClearValue(schemas.BifrostContextKeyAttemptTrail)
	ctx.ClearValue(schemas.BifrostContextKeyStreamEndIndicator)
	ctx.ClearValue(schemas.BifrostContextKeySupportsAssistantPrefill)
	// ← BifrostContextKeyProviderResponseHeaders not cleared
}
```

Providers set that key from their own HTTP response, **before** the status check, so error paths can forward it:

```go
ctx.SetValue(schemas.BifrostContextKeyProviderResponseHeaders, providerUtils.ExtractProviderResponseHeaders(resp))
if resp.StatusCode() != fasthttp.StatusOK { ... }
```

Normally the fallback's own provider overwrites it. But when a fallback attempt fails **before** its provider issues an HTTP request, nothing overwrites it and the **primary's** headers are still on the context when the error is written out. The transport forwards them verbatim:

```go
// transports/bifrost-http/integrations/utils.go — sendError / sendStreamError
if headers, ok := bifrostCtx.Value(schemas.BifrostContextKeyProviderResponseHeaders).(map[string]string); ok {
	for key, value := range headers {
		ctx.Response.Header.Set(key, value)
	}
}
```

Reachable whenever a fallback fails pre-flight, e.g.:

- key selection fails for the fallback provider (`no keys found that support model: X`, all keys dead → `502 upstream_credentials_exhausted`)
- a plugin short-circuits the fallback (governance budget/rate-limit refusal → 402/429)
- the fallback provider's queue is retiring

**Consequence.** The client receives a response produced by provider B carrying provider A's `retry-after`, `x-ratelimit-remaining-*` and `x-ratelimit-reset-*`. Since the headers carry no provider attribution, a client that honours `Retry-After` — the correct thing to do — will wait according to a limit belonging to a provider that never served the request. `x-bifrost-routing-info-provider` on the same response names the fallback, so the two disagree with no way to reconcile them from the response alone.

Same class of staleness the function already guards against for key pins and the attempt trail; this key just seems to have been missed when it was added.

### Steps to reproduce

1. Configure `openai` as primary with a fallback to `anthropic`, and make the **fallback** unusable pre-flight — e.g. give its keys a `models` allowlist that excludes the fallback model, so key selection fails before any HTTP call.
2. Make the primary fail with a rate limit (invalid quota / mock 429 carrying `Retry-After: 60` and `x-ratelimit-remaining-tokens: 0`).
3. Send a chat completion.
4. Inspect the final error response headers.

Observed: `Retry-After: 60` and `x-ratelimit-*` from **openai**, alongside `x-bifrost-routing-info-provider: anthropic`.
Expected: no upstream rate-limit headers, since the provider that produced the terminal error never returned any.

### Expected behavior

Provider response headers do not leak across a fallback boundary. A response carries either the headers of the provider that actually served it, or none.

### Actual behavior

The previous provider's headers persist on the context and are forwarded on a response produced by a different provider.

### Affected area(s)

Core (Go), Transports (HTTP)

### Version

### Environment

```text
- Not platform-specific.
```

### Regression?

Not a regression; appears to date from when `ProviderResponseHeaders` context forwarding was introduced without a corresponding entry in `clearCtxForFallback`.

### Suggested fix

One line:

```go
	ctx.ClearValue(schemas.BifrostContextKeyProviderResponseHeaders)
```

Worth a look at whether the same reasoning applies to any other per-attempt provider state that has been added to the context since — `clearCtxForFallback` is the natural single point for it, and this key drifting out of the list suggests the list is not obviously discoverable from the set-site.

Happy to open a PR.

Huang-404-Q:
## Description

`clearCtxForFallback` wipes the context keys resolved for the *previous* provider before a fallback attempt runs — but it misses `BifrostContextKeyProviderResponseHeaders`.

Providers set that key from their own HTTP response **before** the status check, so error paths can forward it. Normally the fallback's own provider overwrites it, but when a fallback attempt fails **pre-flight** (key selection fails for the fallback provider, a plugin short-circuits it, the queue is retiring), nothing overwrites the key and the **primary's** headers survive on the context. The transport then forwards them verbatim on the fallback's error response.

**Consequence**: the client receives a response attributed to provider B (`x-bifrost-routing-info-provider: anthropic`) carrying provider A's `Retry-After: 60` and `x-ratelimit-remaining-*` — and a compliant client will wait according to a limit belonging to a provider that never served the request, with no way to reconcile the disagreement from the response alone.

This is the same class of staleness `clearCtxForFallback` already guards against for key pins and the attempt trail; this key just seems to have been missed when it was added.

## Fix

One line: clear `BifrostContextKeyProviderResponseHeaders` in `clearCtxForFallback`, with a comment explaining why.

## Testing

- Regression test added to the existing `clearCtxForFallback` coverage in `core/bifrost_test.go`: headers set to simulate the primary (`retry-after`, `x-ratelimit-remaining-tokens`) do not survive the call. Verified red-green (fails on current `dev` with the exact "survived clearCtxForFallback" message).
- Full `core` suite: only pre-existing failure is `TestContextSpanAttributesEmit` (fails identically on unmodified `dev`; unrelated to this change). `go vet` / `gofmt` clean.

Fixes #6973

### DIFF
--- a/core/bifrost_test.go
+++ b/core/bifrost_test.go
@@ -3345,6 +3345,19 @@ func TestClearCtxForFallback_DropsCallerSuppliedKey(t *testing.T) {
 		t.Fatalf("RoutingPinnedAPIKeyID survived clearCtxForFallback: %q", pin)
 	}
 
+	// #6973: provider response headers belong to the provider that produced
+	// them. If a fallback attempt fails pre-flight, the previous provider's
+	// headers must not survive on the context and be forwarded with the
+	// fallback's error response.
+	ctx.SetValue(schemas.BifrostContextKeyProviderResponseHeaders, map[string]string{
+		"retry-after":                  "60",
+		"x-ratelimit-remaining-tokens": "0",
+	})
+	clearCtxForFallback(ctx)
+	if headers, ok := ctx.Value(schemas.BifrostContextKeyProviderResponseHeaders).(map[string]string); ok {
+		t.Fatalf("ProviderResponseHeaders survived clearCtxForFallback: %v", headers)
+	}
+
 	keys, _, err := bifrost.selectKeyFromProviderForModelWithPool(ctx, schemas.ChatCompletionRequest, schemas.Anthropic, "claude-opus-4-5", schemas.Anthropic)
 	if err != nil {
 		t.Fatalf("selectKeyFromProviderForModelWithPool: %v", err)
--- a/core/utils.go
+++ b/core/utils.go
@@ -407,6 +407,11 @@ func clearCtxForFallback(ctx *schemas.BifrostContext) {
 	ctx.ClearValue(schemas.BifrostContextKeyStreamBodyExhausted)
 	ctx.ClearValue(schemas.BifrostContextKeyStreamParkedAfterFinish)
 	ctx.ClearValue(schemas.BifrostContextKeySupportsAssistantPrefill)
+	// Provider response headers belong to the provider that produced them.
+	// If a fallback attempt fails pre-flight (no HTTP request issued), the
+	// previous provider's headers would otherwise survive on the context and
+	// be forwarded with the fallback's error response (#6973).
+	ctx.ClearValue(schemas.BifrostContextKeyProviderResponseHeaders)
 }
 
 // ClearContextForInternalRequest clears context state that is specific to the


Huang-404-Q:
The telemetry comment on this PR belongs to a sibling branch — this PR's diff touches only `core/utils.go` and `core/bifrost_test.go` (the fallback-boundary header clearing), and `plugins/telemetry` isn't part of it. The three branches here share a base and the review stack appears to have mixed their diffs.

That telemetry finding itself is real and is already addressed in PR #7018 (commit e02559d: the decrement gates on a non-stream request or the confirmed final chunk, keeping the idempotent latch).

akshaydeo (approved):

PR #7027, opened 2026-09-09
fix(bedrock): assign unique names to untitled document blocks

### LINKED ISSUE #7003: [Bug]: Bedrock Converse assigns duplicate default name "document" to untitled document blocks → ValidationException
## Summary

When converting an Anthropic/unified request to the **Bedrock Converse** API, Bifrost assigns the **same default `name` (`"document"`) to every untitled document block**. Bedrock's Converse API requires document names to be unique within a request, so any request containing **two or more untitled documents** is rejected with:

```
ValidationException: Messages can't contain duplicate document names. Rename the document and retry your request.
```

This makes any multi-document (untitled) request fail 100% of the time when routed to a Bedrock provider.

## Version

- Bifrost image: `maximhq/bifrost:v1.6.11` (transports/v1.6.11)
- Route: `POST /anthropic/v1/messages` → Bedrock provider
- Model: `us.anthropic.claude-sonnet-5` (any Converse model with document support)

## Root cause

`core/providers/bedrock/utils.go` (~line 1209):

```go
documentSource := &BedrockDocumentSource{
    Name:   "document", // default
    ...
}
// Set filename (normalized for Bedrock)
if block.File.Filename != nil {
    documentSource.Name = normalizeBedrockFilename(*block.File.Filename)
}
```

Every document block without a filename/title falls through to the literal default `Name: "document"`. With two or more such blocks in one request, all share the name `"document"`, which Bedrock Converse rejects as duplicates.

The same default appears in the responses path at `core/providers/bedrock/responses.go` (~line 4741: `Name: "document" // Default`).

This matters in practice because upstream clients (e.g. the Claude Code CLI / Anthropic SDK) legitimately emit **bare** document blocks — `{"type":"document","source":{"type":"base64","media_type":"application/pdf",...}}` — with no `title`. Anthropic's native API and Bedrock's `invoke`/messages API accept these fine; only the Bedrock **Converse** conversion breaks, because Bifrost collapses them all to one name.

## Minimal reproduction

Direct Bedrock Converse call demonstrating the underlying constraint Bifrost trips:

```python
import boto3
br = boto3.client("bedrock-runtime", region_name="us-east-1")
pdf = open("any.pdf","rb").read()
def doc(name): return {"document":{"format":"pdf","name":name,"source":{"bytes":pdf}}}

# Fails: ValidationException "duplicate document names"
br.converse(modelId="us.anthropic.claude-sonnet-5",
    messages=[{"role":"user","content":[doc("document"), doc("document"), {"text":"hi"}]}],
    inferenceConfig={"maxTokens":16})

# Works: distinct names
br.converse(modelId="us.anthropic.claude-sonnet-5",
    messages=[{"role":"user","content":[doc("doc-1"), doc("doc-2"), {"text":"hi"}]}],
    inferenceConfig={"maxTokens":16})
```

Through Bifrost, send an `/anthropic/v1/messages` request with two title-less `document` blocks to a Bedrock provider — same failure.

## Suggested fix

Make the default document name unique per request, e.g. suffix an index when the name is missing/defaulted:

```go
Name: fmt.Sprintf("document-%d", docIndex),
```

(or de-duplicate names just before emitting the Converse payload). This keeps behavior unchanged for named documents while preventing the collision for unnamed ones.

## Impact

Any Bedrock-routed request carrying 2+ untitled documents (common with PDF-heavy assistant workflows) fails on every attempt, and because conversation history replays each turn, the whole conversation becomes unusable until the documents leave the context window.

Huang-404-Q:
## Description

When converting an Anthropic/unified request to the Bedrock Converse API, every untitled document block gets the same literal default name `"document"`. The Converse API rejects duplicate document names within a request, so **any request with two or more untitled documents fails unconditionally** when routed to a Bedrock provider:

```
ValidationException: Messages can't contain duplicate document names. Rename the document and retry your request.
```

## Fix

Thread a per-request document namer through the content-block conversion (`convertMessages` → `convertMessage` → `convertContent` → `convertContentBlock`). The namer normalizes filenames exactly as before and disambiguates collisions with numeric suffixes:

- untitled: `"document"`, `"document-2"`, `"document-3"`, …
- titled: keep the normalized filename; suffixed only on an actual collision

The Responses-API input replay path gets the same treatment scoped to its content-block list.

## Testing

- New `document_names_test.go`: two untitled documents produce unique names; explicitly titled documents keep their normalized names without suffixes.
- Full `core/providers/bedrock` suite: `ok`. `go vet` / `gofmt` clean.

Fixes #7003


Huang-404-Q:
Both points addressed in 6b7f2da:

1. **Request-scoped namer** — created once in `ConvertBifrostMessagesToBedrockMessages` and threaded through the per-message conversions, so documents in separate messages can no longer collide. Regression test with one untitled document in each of two messages.

2. **Generated-name collisions** — the namer now tracks emitted final names: emitting `"document-2"` reserves it, and a later explicit `"document-2"` gets `"document-2-2"`. Regression test: untitled / untitled / explicit `"document-2"` → three unique names.

(The telemetry comment on this PR belongs to the sibling PR #7018 — the three branches share a base and the reviewer stack appears to have mixed their diffs; that finding is already addressed there in e02559d.)

Huang-404-Q:
Follow-up: I've restructured the branches so each PR carries exactly one concern — this PR's diff is now Bedrock-only (the telemetry finding belonged to #7018, where it's already fixed in e02559d).

Also addressed this round's two findings on the document-naming path in e476ee42:
1. The Responses-input namer line had landed inside the error branch after an unconditional return — unreachable, and the success path kept emitting duplicate "document" names. Moved out of the error branch.
2. Suffix construction now reserves space within the 200-character document-name limit (base trimmed before appending "-N"), so a near-limit filename can't produce an over-limit duplicate name.

akshaydeo (approved):

PR #7052, opened 2026-09-10
feat: implement opencode session management with header resolution an…

### LINKED ISSUE #6815: [Feature]: forward x-opencode-session on opencode requests
### Prerequisites

### Problem to solve

Starting 09/06 OpenCode Go requires an x-opencode-session header (one stable ID per session) and may start rejecting requests without the header per their notice to subscribers.

Bifrost breaks this chain. The gateway drops the header on ingress and re-emits through fasthttp, so upstream sees a header-less request no matter what the client sent. That also wipes out client-side fixes: e.g. Hermes NousResearch/hermes-agent#101864 (merged, sends a per-conversation header past 0.21.0) still arrives with nothing when it goes through Bifrost. Anyone routing Go through Bifrost is exposed, and there is no workaround short of bypassing the gateway.

### Proposed solution

Resolve the session once per request in the HTTP transport and send it upstream on opencode calls (chat, responses, streaming; not ListModels):

1. A client-sent x-opencode-session goes through verbatim after a charset check that rejects CR/LF and control characters.
2. Without one, fall back to the request's own session identity (x-bf-session-id, then the harness headers), keeping the existing wins and poison-to-none rules.
3. With no signal at all, synthesize a per-request UUID. That keeps the request compliant but buys no affinity. Multi-turn tracking stays the client's job.
4. Scope the value per virtual key before sending, and only ever send it to the opencode family. I would send to both Zen and Go even though only Go enforces it: Zen ignores the header and uniform emission keeps zen-go fallbacks identical across retries.

I have this working locally with httptest-pinned tests (forwarding, fallback chain, poison rules, no leak to other providers, namespacing, streaming) and will open a PR against dev if maintainers like the shape.

### Alternatives considered

- x-bf-eh-x-opencode-session works today, but it forwards to every provider and puts the burden on each client to know the trick.
- A static value in network_config.extra_headers pins all conversations to one session, which doesn't follow OpenCode's "one stable ID per session"
- Doing nothing leaves Go traffic on cold backends now and facing rejections after 09/06.

### Area(s)

Core (Go), Transports (HTTP)

### Additional context

This is the email I and other subscribers recently received from OpenCode Go (updates@opencode.ai):

> 
> Hey there,
> 
> Some of your requests to OpenCode Go are missing an x-opencode-session header. If we don't have this we cannot properly optimize our service. Starting 09/06 requests missing this header may error.
> 
> Here are your useragents that are missing this header:
> 
> Fasthttp
> Add x-opencode-session (one stable ID per conversation) to your requests.
> 
> Thank you.

sstreichan:
## Summary

The OpenCode Zen/Go gateway requires the `x-opencode-session` header on inference calls since 2026-09-06 and may reject requests that omit it. This PR introduces session management for the OpenCode provider family: `ResolveOpencodeSession` resolves the session identity exactly once per request and attaches it to all upstream inference calls (chat, responses, including their streaming variants).

## Changes

- **New `ResolveOpencodeSession` function** (`core/providers/opencode/utils.go`): resolves the session identity in a fixed priority order:
  1. Client-sent `x-opencode-session` header (client wins), after charset validation
  2. The session identity already resolved by Bifrost (`BifrostContextKeySessionID`, `x-bf-session-id` or harness session headers)
  3. Fallback: a freshly generated per-request UUID so the request stays gateway-compliant

- **Resolved once per request**: the result is cached on the context (`BifrostContextKeyOpencodeSession`, a reserved key) via `BifrostContext.GetOrStoreReservedValue` — an atomic load-or-store that bypasses the restricted-writes guard, so concurrent callers on one context cannot each mint a different UUID and a retry after `BlockRestrictedWrites` still sees the original value. Retries, zen→go fallbacks, and MCP agent-loop turns of the same request see the same stable session identity.

- **Namespacing per virtual key**: resolved session values are qualified with the virtual key (`<vk>:<session>`), so sessions used under different virtual keys cannot collide upstream. Both components are escaped before joining — backslashes are doubled and colons prefixed — so the only literal `:` is the separator and distinct `(vk, value)` pairs always yield distinct namespaced strings.

- **Header-injection protection**: `isSafeOpencodeSessionValue` rejects values containing CR/LF or other control characters — unsafe values are treated as absent.

- **`inferenceExtraHeaders`** (`core/providers/opencode/opencode.go`): merges the static `ExtraHeaders` from the network config with the resolved session header (fresh map, config is never mutated). Every case-insensitive variant of the session header is dropped from the copied map before the resolved value is set, so the resolved session always wins regardless of map iteration order.

- **ListModels intentionally excluded**: `ListModels` requests do not carry the session header, since the gateway only requires it on inference calls.

- **Schemas**: new context key `BifrostContextKeyOpencodeSession` in `core/schemas/bifrost.go`, registered as a reserved key in `core/schemas/context.go` (DO NOT SET MANUALLY); new `GetOrStoreReservedValue` on `BifrostContext`.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [x] Core (Go)
- [ ] Transports (HTTP)
- [x] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

```sh
# Unit tests for session resolution (opencode provider)
cd core && go test ./providers/opencode/... -run OpencodeSession -v

# Provider tests incl. wire-level header checks (opencode)
make test-core PROVIDER=opencode
```

**Expected results:**

- `TestResolveOpencodeSessionClientHeaderWins` — client header wins and is namespaced
- `TestResolveOpencodeSessionFallbackChain` — fallback chain (header → session ID → UUID)
- `TestResolveOpencodeSessionSynthesizesUUID` — a UUID is generated when no signal is present
- `TestResolveOpencodeSessionNamespacesPerVirtualKey` — namespacing works; an unsafe virtual key leaves the value unqualified
- `TestResolveOpencodeSessionResolvesOncePerRequest` — resolution is cached exactly once per request
- `TestResolveOpencodeSessionConcurrentSameContext` — concurrent resolution on one context returns one identical value
- `TestResolveOpencodeSessionStableUnderBlockRestrictedWrites` — the cached value survives `BlockRestrictedWrites`
- `TestNamespaceOpencodeSessionNoCollision` — `("a","b:c")` and `("a:b","c")` do not collide
- `TestOpencodeSessionHeaderOnWire` — chat/responses calls carry the header
- `TestOpencodeSessionHeaderNotOnListModels` — ListModels does not carry it
- `TestOpencodeSessionHeaderNeverLeaksToOtherProviders` — the header never leaks to other providers
- `TestIsSafeOpencodeSessionValue` — header-injection values are rejected

## Screenshots/Recordings

No UI changes.

## Breaking changes

- [ ] Yes
- [x] No

## Related issues

Closes #6815

## Security considerations

- **Header-injection protection**: all forwarded session values (client header, session ID, virtual key) go through `isSafeOpencodeSessionValue` — values with control characters (CR/LF etc.) never reach the wire.
- **No session collision**: namespacing per virtual key with escaping on both components prevents identical client sessions under different virtual keys from colliding upstream, and prevents a crafted value from forging the namespace separator.
- **Bounded restricted-write bypass**: `GetOrStoreReservedValue` bypasses `BlockRestrictedWrites`, but only for the reserved `BifrostContextKeyOpencodeSession` key and only to cache a value this code derived itself.
- **No cross-provider leakage**: the header is sent exclusively to the OpenCode provider family, never to other providers.
- **No sensitive data**: only session IDs/UUIDs are transmitted, no PII.

## Checklist


core/providers/opencode/utils.go
```diff
@@ -0,0 +1,118 @@
+package network
+
+import (
+	"strings"
+
```
akshaydeo, line 5: these functions should be in providers/opencode/utils.go and

akshaydeo (changes requested):

core/providers/opencode/utils.go
sstreichan: @akshaydeo done — the session helpers are now in `core/providers/opencode/utils.go` and `utils_test.go` (package `opencode`), and the `core/network` import is gone from the opencode provider. Landed in commit 020e339e7; `core/network/opencodesession.go` and its test were removed. CodeRabbit re-reviewed the follow-up (1fa2c0a1e) and approved. Could you re-review / dismiss?

sstreichan:
@akshaydeo — your requested changes are addressed and every review thread is resolved. What changed since your review (2026-09-10):

- `020e339` — the session helpers moved out of `core/network/opencodesession.go` into `core/providers/opencode/{utils.go,utils_test.go}`; the old files are deleted. `core/network` is no longer imported by the provider.
- `1fa2c0a` — the resolved session now drops every case-insensitive variant of `x-opencode-session` from static `ExtraHeaders` before setting the canonical key, so the resolved value always wins regardless of map iteration order (CodeRabbit finding).
- `b25c9ab` — docstrings for the functions touched by the diff, clearing the failing docstring-coverage pre-merge check.

State now: CodeRabbit `APPROVED`, all threads resolved, `go build ./...`, `go vet` and `go test -race ./providers/opencode/... ./schemas/...` green.

Could you re-review or dismiss? Your review is the only thing holding `reviewDecision` at `CHANGES_REQUESTED`.

PR #7063, opened 2026-09-10
fix(bedrock-mantle): route gpt-6 to the openai/v1 path

arnzchen:
## Problem

Bedrock Mantle serves a closed frontier model on exactly one of its two OpenAI-compatible base paths and rejects it on the other. `mantleOpenAIURL` enumerates `gpt-5`, so `gpt-6` fell through to the bare `v1` path that gpt-oss uses, and every request failed:

```
POST bedrock-mantle.us-west-2.api.aws/v1/chat/completions
400 {"code":"validation_error","type":"invalid_request_error",
     "message":"model `openai.gpt-6-astra` isn't supported on this route"}
```

The identical request against `openai/v1` returns 200. The gate is duplicated, so both `providers/bedrockmantle` and the deprecated in-provider path in `providers/bedrock` were affected.

Two capability gates in `providers/openai/utils.go` missed the generation too, and both fail silently rather than erroring:

- `acceptsXHighEffort` omitted it while `acceptsMaxEffort` already lists `gpt-6-astra`. `defaultEffortControl` therefore built `low/medium/high/max`, skipping the tier between them, and a requested `xhigh` was downgraded to `high`.
- `IsOpenAIReasoningModel` ends in a `gpt-5` substring check, so the family defaulted to non-reasoning. That clears `reasoning.effort` on the OpenAI and Azure paths and drops replayed reasoning items.

## Changes

- `mantleOpenAIURL` (both copies) also routes `gpt-6` to `openai/v1`. gpt-oss and Gemma 3 keep the bare `v1` path.
- The OpenAI-compatible handlers (both provider copies) strip the optional `region/` addressing prefix before the model reaches the body. The native-Anthropic handlers already did; the OpenAI ones read the model straight off the request, so `us-west-2/openai.gpt-6-astra` reached the wire and mantle 404d it. Found by the harness run below.
- `acceptsXHighEffort` accepts `gpt-6-astra`, matching `acceptsMaxEffort`.
- `IsOpenAIReasoningModel` recognises `gpt-6`.

## Tests

Full provider-harness pass against `gpt-6-astra` on mantle (Go harness, `TestBedrockMantleOpenAICompatible`): **49 pass, 2 skip, 0 fail**. Details and the two model-behaviour findings it surfaced are in the comment thread. In short: Astra rejects function tools on `/v1/chat/completions` even with no `reasoning_effort` in the request (`param: reasoning_effort`; verified against the endpoint directly), and its chat stream carries no reasoning content while its Responses stream does. The new config scopes to what the model serves and says why.

Unit tests:

- `TestMantleOpenAIURL` in `providers/bedrock` gains gpt-6 cases, and `providers/bedrockmantle` gets the same table, which it had no coverage for. The function is duplicated, so a test in each place keeps the two from drifting.
- `TestEffortPredicatesAgainstCatalogIDs` gains bare, vendor-qualified and region-qualified Astra IDs.
- `TestIsOpenAIReasoningModel` is new; the predicate had no direct coverage.

`go test ./...` in `core` passes apart from `internal/mcptests`, which fails identically on an unmodified checkout because its test server binary is not built (`make setup-mcp-tests`).

## Note

The base-path gate names each closed generation explicitly, so the next one will need the same edit. Left as-is here to keep the change minimal: inverting it to "everything except gpt-oss" would move Gemma 3 off the bare path it currently requires.

### DIFF
--- a/core/providers/bedrockmantle/utils.go
+++ b/core/providers/bedrockmantle/utils.go
@@ -56,6 +56,33 @@ func parseBedrockRegionAndModel(model string) (region, bareModel string) {
 	return "", model
 }
 
+// withBareModel drops the region addressing prefix from the model a chat request
+// carries, so the body names the id mantle knows. The native-Anthropic handlers take
+// the bare model through their build config, but the OpenAI-compatible ones read it
+// off the request, and mantle 404s the prefixed form ("The model
+// 'us-west-2/openai.gpt-6-astra' does not exist"). Copied rather than mutated: the
+// caller's request is what the pipeline keys pricing and logging on.
+func withBareModel(request *schemas.BifrostChatRequest) *schemas.BifrostChatRequest {
+	_, bare := parseBedrockRegionAndModel(request.Model)
+	if bare == request.Model {
+		return request
+	}
+	clone := *request
+	clone.Model = bare
+	return &clone
+}
+
+// withBareResponsesModel is withBareModel for the Responses surface.
+func withBareResponsesModel(request *schemas.BifrostResponsesRequest) *schemas.BifrostResponsesRequest {
+	_, bare := parseBedrockRegionAndModel(request.Model)
+	if bare == request.Model {
+		return request
+	}
+	clone := *request
+	clone.Model = bare
+	return &clone
+}
+
 // resolveRegion returns the AWS region to use for a request.
 // Priority: model-string region prefix > alias-level Region > key-level
 // BedrockMantleKeyConfig.Region > defaultMantleRegion. The model-string prefix
--- a/core/providers/bedrockmantle/utils.go
+++ b/core/providers/bedrockmantle/utils.go
@@ -5,6 +5,7 @@ import (
 	"regexp"
 	"strings"
 
+	providerUtils "github.com/maximhq/bifrost/core/providers/utils"
 	schemas "github.com/maximhq/bifrost/core/schemas"
 )
 
@@ -56,31 +57,30 @@ func parseBedrockRegionAndModel(model string) (region, bareModel string) {
 	return "", model
 }
 
-// withBareModel drops the region addressing prefix from the model a chat request
-// carries, so the body names the id mantle knows. The native-Anthropic handlers take
-// the bare model through their build config, but the OpenAI-compatible ones read it
-// off the request, and mantle 404s the prefixed form ("The model
-// 'us-west-2/openai.gpt-6-astra' does not exist"). Copied rather than mutated: the
-// caller's request is what the pipeline keys pricing and logging on.
-func withBareModel(request *schemas.BifrostChatRequest) *schemas.BifrostChatRequest {
-	_, bare := parseBedrockRegionAndModel(request.Model)
-	if bare == request.Model {
-		return request
+// bareModelScope rewrites request.Model (and the model field of a raw passthrough
+// body) to the bare id for the duration of a handler call, and returns the func that
+// puts both back. Nothing is copied: the caller's request is what core hands to the
+// next retry and what pricing and logging read, so it has to leave with its prefix
+// intact. The native-Anthropic handlers take the bare model through their build
+// config; the OpenAI-compatible ones read it off the request, and mantle 404s the
+// prefixed form ("The model 'us-west-2/openai.gpt-6-astra' does not exist"). Costs
+// nothing when there is no prefix, which is every request that names a region on
+// the key or the alias rather than the model.
+func bareModelScope(model *string, raw *[]byte) func() {
+	region, bare := parseBedrockRegionAndModel(*model)
+	if region == "" {
+		return func() {}
 	}
-	clone := *request
-	clone.Model = bare
-	return &clone
-}
-
-// withBareResponsesModel is withBareModel for the Responses surface.
-func withBareResponsesModel(request *schemas.BifrostResponsesRequest) *schemas.BifrostResponsesRequest {
-	_, bare := parseBedrockRegionAndModel(request.Model)
-	if bare == request.Model {
-		return request
+	origModel, origRaw := *model, *raw
+	*model = bare
+	if len(origRaw) > 0 {
+		if edited, err := providerUtils.SetJSONField(origRaw, "model", bare); err == nil {
+			*raw = edited
+		}
 	}
-	clone := *request
-	clone.Model = bare
-	return &clone
+	return func() {
+		*model, *raw = origModel, origRaw
+	}
 }
 
 // resolveRegion returns the AWS region to use for a request.
--- a/core/internal/llmtests/account.go
+++ b/core/internal/llmtests/account.go
@@ -139,6 +139,7 @@ type ComprehensiveTestConfig struct {
 	FileStorageConfig        *schemas.FileStorageConfig // Typed storage config for file operations (e.g., GCS bucket for Vertex)
 	DisableParallelFor       []string                   // Test scenarios to disable parallel execution for (e.g., "Transcription" for rate-limited APIs)
 	ExpectRawRequestResponse bool                       // When true, validate rawRequest/rawResponse in ExtraFields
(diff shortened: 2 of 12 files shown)


akshaydeo (dismissed):

akshaydeo:
@arnzchen did you run actual LLM calls? could you share some reports here? 
Can you also update provider harness to have gpt-6-astra on mantle?

akshaydeo:
and please sign CLA

core/providers/bedrockmantle/utils.go
```diff
@@ -56,6 +56,33 @@ func parseBedrockRegionAndModel(model string) (region, bareModel string) {
 	return "", model
 }
 
+// withBareModel drops the region addressing prefix from the model a chat request
+// carries, so the body names the id mantle knows. The native-Anthropic handlers take
+// the bare model through their build config, but the OpenAI-compatible ones read it
+// off the request, and mantle 404s the prefixed form ("The model
+// 'us-west-2/openai.gpt-6-astra' does not exist"). Copied rather than mutated: the
+// caller's request is what the pipeline keys pricing and logging on.
+func withBareModel(request *schemas.BifrostChatRequest) *schemas.BifrostChatRequest {
```
akshaydeo: I dont think we should clone the request for this - too much of overhead on the hot path

arnzchen:
Yes — real calls throughout, and running them changed the PR twice. Latest push is `d2dfc7d`.

## Harness

Full pass against `gpt-6-astra` on mantle, via the Go provider harness: **49 pass, 2 skip, 0 fail**. Scenarios covered: SimpleChat, ChatCompletionStream, MultiTurnConversation, the reasoning set on both surfaces, StructuredOutputs on both surfaces plus streaming, ImageBase64, ResponsesStream and its lifecycle, ResponsesStreamWithTools, ToolCallsStreamingResponses, StreamErrorStatusCode, and the unsupported-operation stubs.

The first full pass was **52 pass / 25 fail**, and every one of the 25 failures had the same upstream cause:

```
invalid_request_error validation_error: Function tools with reasoning_effort are not
supported for gpt-6-astra in /v1/chat/completions. To use function tools, use /v1/responses.
```

Astra's thinking is always on, so there is no effort-free chat request to fall back to. Tools on `/v1/responses` pass (`ResponsesStreamWithTools`, `ToolCallsStreamingResponses`), tools on `/v1/chat/completions` cannot work. Since each tool scenario drives its chat and Responses variants off one flag, I disabled those flags and said so in the config rather than leave known-red rows. **This looks like a datasheet question for you:** if Astra's row marked the chat endpoint unsupported, `markForConversion` would route chat+tools to `/v1/responses` and those rows would pass instead of returning a raw upstream 400 to the caller. Happy to follow up separately if you want that.

Separately, Astra's chat stream carries no reasoning content (120–136 chunks, no reasoning indicators, ten attempts) while its Responses stream carries it. `ChatCompletionStreamWithReasoningValidated` only had a hardcoded OpenAI/Groq skip, so I added a `SkipChatReasoning` config field — disabling `Reasoning` outright would have dropped the five reasoning scenarios that do pass.

## A second bug the run found

Adding Astra to the harness meant addressing it as `us-west-2/openai.gpt-6-astra`, since mantle serves it in us-west-2 only while the key's region is us-east-1. That failed:

```
404 {"message":"The model 'us-west-2/openai.gpt-6-astra' does not exist"}
```

The package contract is that the id is sent verbatim save for an optional leading `region/` prefix, and the native-Anthropic handlers strip it at all four call sites — the OpenAI-compatible handlers read the model straight off the request, so the prefix reached the wire. Fixed in `50684b1`, copying the request rather than mutating it since the pipeline keys pricing and logging on the caller's string.

## Direct endpoint calls, before any of the above

| Request | Result |
|---|---|
| `openai/v1/chat/completions`, us-west-2 | 200, content returned with usage |
| bare `v1/chat/completions`, us-west-2 | 400 `isn't supported on this route` |
| either path, us-east-1 | 404 `model does not exist` |
| Converse `us.openai.gpt-6-astra`, us-west-2 | 200 |

## One question

You pointed me at `make run-provider-harness-test PROVIDER=bedrock_mantle`, which is the Postman collection against a live gateway — I updated the Go provider harness instead. Worth noting the Postman side has the same blind spot: `bifrost-v1-bedrock-mantle.postman_environment.json` sets every model to `anthropic.claude-opus-4-8` with `bedrock_mantle_region = us-east-1`, so it has never exercised the base-path split either. The prefix fix above is what lets it reach Astra without a region switch. Tell me which you meant and whether you want collection rows added, and I will follow your conventions there rather than guess.

core/providers/bedrockmantle/utils.go
arnzchen: Agreed, and done in 7d6f628. No clone: `bareModelScope` rewrites `request.Model` in place for the duration of the handler call and restores it on return. It returns a no-op when there is no prefix, so the common path (region on the key or alias, which is every production config I know of) pays a string scan and nothing else.

Restore rather than mutate-and-leave because core's retry loop hands the same `*BifrostChatRequest` back to the provider — a permanently stripped model would lose its region on attempt two and fall back to the key's region. Restoring on return is safe for streaming: both streaming handlers marshal the body before returning the channel, and the goroutine never reads the request afterwards (checked `HandleOpenAIChatCompletionStreaming` and `HandleOpenAIResponsesStreaming`).

Harness against Astra re-run on this commit: still 49 pass, 2 skip, 0 fail. Unit tests for the scope in both provider copies.

core/providers/bedrockmantle/utils.go
```diff
@@ -56,6 +56,33 @@ func parseBedrockRegionAndModel(model string) (region, bareModel string) {
...
+	_, bare := parseBedrockRegionAndModel(request.Model)
+	if bare == request.Model {
+		return request
+	}
+	clone := *request
+	clone.Model = bare
```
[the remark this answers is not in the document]
arnzchen: Done in 7d6f628 — `bareModelScope` now also rewrites the `model` field of `RawRequestBody` for the duration of the call (via `providerUtils.SetJSONField`, same idiom as `StripUnsupportedFieldsFromRawBody`) and restores the original slice afterwards, so passthrough sends the bare id and the caller's raw body is untouched. Covered by `TestBareModelScope` in both `providers/bedrockmantle` and `providers/bedrock` (prefixed model with raw body, prefixed without raw body, and the no-prefix no-op). The chat, Responses, and both streaming handlers all go through the same scope.

jakswa:
I am running into this as well.

core/providers/bedrockmantle/utils.go
```diff
@@ -56,6 +57,32 @@ func parseBedrockRegionAndModel(model string) (region, bareModel string) {
...
+// bareModelScope rewrites request.Model (and the model field of a raw passthrough
+// body) to the bare id for the duration of a handler call, and returns the func that
+// puts both back. Nothing is copied: the caller's request is what core hands to the
+// next retry and what pricing and logging read, so it has to leave with its prefix
+// intact. The native-Anthropic handlers take the bare model through their build
+// config; the OpenAI-compatible ones read it off the request, and mantle 404s the
+// prefixed form ("The model 'us-west-2/openai.gpt-6-astra' does not exist"). Costs
+// nothing when there is no prefix, which is every request that names a region on
+// the key or the alias rather than the model.
+func bareModelScope(model *string, raw *[]byte) func() {
```
akshaydeo, line 69: this i still think an be done in a better way - let me take a stab at it

arnzchen: Go ahead — I'll leave the helper alone so we don't collide. If you'd rather I take a different shape (e.g. pushing the bare model into the OpenAI handlers the way the Anthropic build config does), say which and I'll turn it around.

TejasGhatte:
Hey @arnzchen we are making some changes in bedrock and cutting a release so will make sure patch lands there. Closing this pr

PR #7066, opened 2026-09-11
[fix]: honor Retry-After in the core retry loop

### LINKED ISSUE #6971: [Bug]: retry scheduler ignores provider Retry-After / rate-limit headers it already captures and forwards
### Description

Bifrost **captures** upstream rate-limit headers, **forwards** them verbatim to the client, and then **ignores them entirely** when scheduling its own retries. `calculateBackoff` never sees the upstream response:

```go
// core/utils.go
func calculateBackoff(attempt int, config *schemas.ProviderConfig) time.Duration {
	backoff := min(config.NetworkConfig.RetryBackoffInitial*time.Duration(1<<uint(attempt)),
	               config.NetworkConfig.RetryBackoffMax)
	jitter := float64(backoff) * (0.8 + 0.4*rand.Float64())
	return min(time.Duration(jitter), config.NetworkConfig.RetryBackoffMax)
}
```

Its only inputs are the attempt index and static config. The `*schemas.BifrostError` that triggered the retry — which carries the status code, the message, and (via `ExtraFields`) the provider context — is in scope at the call site in `executeRequestWithRetries` but is not passed in.

A repo-wide search confirms nothing else fills the gap:

```
grep -rn "retry-after\|Retry-After\|retry_after" --include="*.go" core/ transports/ framework/ plugins/
→ core/internal/llmtests/test_retry_framework.go   (test harness only)
```

So on a 429 with `Retry-After: 60`, Bifrost sleeps ~500ms and retries into a window it has been told is closed — burning a retry from the budget, holding a worker slot for the duration (see #6964), and adding load to a provider that explicitly asked for quiet, which with most providers extends the throttle rather than shortening it.

### The odd part: the test harness already does this

`core/internal/llmtests/test_retry_framework.go` has a `parseRetryAfterFromError` helper, used in **10** places, that extracts a provider-specified delay and honours it:

```go
var retryAfterRegex = regexp.MustCompile(`(?i)retry after (\d+)\s*(seconds?|minutes?)`)

// ... at each call site:
if retryAfter := parseRetryAfterFromError(err); retryAfter > 0 {
	retryAfter += 2 * time.Second   // buffer
	retryReason := fmt.Sprintf("❌ rate limit error - waiting %v as specified by provider", retryAfter)
	time.Sleep(retryAfter)
	continue
}
```

To be precise about what this does and doesn't show: the harness regex-matches the **error message text** (`"Please retry after 30 seconds"`), not the HTTP `Retry-After` header. So it is not a case of "the tests parse the header and production doesn't".

What it does establish is that the project already treats *"wait as long as the provider said"* as the correct retry behaviour — that logic exists, is used in ten places, and is commented as such — while the production retry loop honours **neither** signal: not the message hint the harness reads, and not the header, which is the more reliable of the two and which Bifrost already has in hand.

### Bifrost already captures the header

Every provider stashes the response headers on the context **before** the status check, specifically so error paths can forward them:

```go
// core/providers/openai/openai.go and equivalents in every provider
// Store provider response headers in context before status check so error responses also forward them
ctx.SetValue(schemas.BifrostContextKeyProviderResponseHeaders, providerUtils.ExtractProviderResponseHeaders(resp))

if resp.StatusCode() != fasthttp.StatusOK {
	return nil, ParseOpenAIError(resp)
}
```

`providerResponseFilterHeaders` strips transport and credential headers but **not** `retry-after` or `x-ratelimit-*`, so the value reaches the client verbatim on both route families. The data is already captured, already filtered, already on the context at the moment the retry decision is made — it is simply never read by the retry loop.

### Previously requested and closed as completed

This was the second bullet of #3696 (*"`429` responses are retryable, but provider `Retry-After` hints do not appear to be honored in the retry scheduler"*), closed as **completed** on 2026-05-28. It does not appear to have landed — the grep above returns no production hits, and `calculateBackoff` still takes only `(attempt, config)`.

The first bullet of that same issue (uninterruptible `time.Sleep`) is likewise still present and is now tracked in #6964. Filing separately rather than commenting on a long-closed issue; happy to have this closed as a duplicate if reopening #3696 is preferred.

### Steps to reproduce

1. Configure a provider with `max_retries: 3`.
2. Have the upstream return `429` with `Retry-After: 60`.
3. Observe the debug log: `sleeping for 5xxms before retry`, then a retry well inside the 60s window — repeated until the budget is exhausted.

The forwarded response header on the eventual failure still shows `Retry-After: 60`, confirming Bifrost had the value the whole time.

### Expected behavior

When a retryable error carries a provider-specified delay, the backoff waits at least that long (clamped to something sane and to the remaining request deadline) instead of using the local exponential value.

### Actual behavior

Local exponential backoff only. The provider's instruction is forwarded to the client and otherwise discarded.

### Affected area(s)

Core (Go)

### Version

### Environment

```text
- Not platform-specific.
```

### Regression?

Not a regression — never implemented. Requested in #3696, which was closed as completed.

### Suggested fix

Parse the hint at error-construction time, where the `fasthttp.Response` is still available, and carry it on the error so the retry loop can use it without reaching back into transport types:

1. Add a field to `BifrostError.ExtraFields`, e.g. `RetryAfter *time.Duration`.
2. Populate it in `providerUtils.HandleProviderAPIError` / the per-provider error converters, from the response header. `Retry-After` is RFC 7231 dual-format — **delta-seconds** *or* an **HTTP-date** — so both forms need handling. Falling back to the harness's message regex when the header is absent would cover providers that only state it in prose.
3. Give `calculateBackoff` access to it:

```go
func calculateBackoff(attempt int, config *schemas.ProviderConfig, retryAfter *time.Duration) time.Duration {
	backoff := ... // existing exponential + jitter
	if retryAfter != nil && *retryAfter > backoff {
		// Honour the provider, but never exceed the ceiling the operator configured.
		return min(*retryAfter, config.NetworkConfig.RetryBackoffMax)
	}
	return backoff
}
```

Worth deciding explicitly: whether `retry_backoff_max` should cap a provider-specified delay. Clamping keeps operator control and bounds worst-case latency, but means a `Retry-After: 60` against a 5s cap still retries early. A separate `retry_after_max` (or an opt-in `honor_retry_after` flag) would let deployments choose. Happy to follow whichever direction maintainers prefer.

Also worth surfacing the value once it exists: `x-ratelimit-remaining-*` and the parsed `Retry-After` would make a useful gauge/log field for spotting throttling before it becomes failures — currently neither is persisted in the log row nor exported as a metric.

Happy to open a PR.

jjj-n:
## Summary

When an upstream returns `429` with `Retry-After: 2`, core currently retries after its local backoff, potentially well before the provider permits another request. Honor the standard header in the shared retry loop, including streaming startup failures and rate-limited key rotation.

Closes #6971.

## Changes

- Parse delta-seconds and HTTP-date values from the response headers already captured on the context. Preserve local backoff for missing, invalid, zero, or past hints; handle integer overflow without retrying early.
- Wait for the later of local backoff and the provider's advertised retry time. Make the wait interruptible by cancellation or deadline expiry.
- Clear response headers immediately before each provider attempt so a network failure cannot reuse a previous response's hint.
- Stop the current provider's retries and preserve its original error when the hint exceeds `retry_backoff_max` or reaches the request deadline. The existing fallback policy can then apply.
- Add deterministic Go regression tests, retry documentation, and four opt-in provider-harness cases backed by a local fixture.

**Policy for review:** a 60-second hint with a 5-second backoff cap stops retries rather than truncating the provider's minimum wait. A hint equal to the cap is allowed. Maintainer confirmation of this choice is still pending on #6971.

Reuses the cancellable retry wait and terminal cancellation/timeout classification now present on `dev` from #7105, removing this PR's duplicate wait helper. The cancellation audit regression now asserts the terminal routing-engine entry. Fixture commands use a per-run temporary directory. This log-only follow-up has no new wire-visible behavior; the four existing provider-harness cases remain the wire regressions. PR #7021 clears headers at fallback boundaries; this change clears them immediately before each actual provider attempt, including same-provider retries. Neither the jitter calculation nor provider-specific reset headers/error-message hints are changed.

## Type of change

- [x] Bug fix
- [x] Documentation

## Affected areas

- [x] Core (Go)
- [x] Docs
- Provider-harness regression tests

## How to test

Validated on 2026-09-21 against `dev` at `1a17949cd` with Linux / Go 1.27.1:

```sh
cd core
go test -race . -run 'TestRetryAfter|TestExecuteRequestWithRetries|TestCalculateBackoff|TestAzureResponsesRetriesAfterStartupEvents|TestWaitRetryBackoff' -count=1 -timeout=90s
go test -race . -count=1 -timeout=120s
```

Targeted retry/backoff/Azure startup tests pass with the race detector (4.473s). The full core package passes with the race detector (39.760s), with no tests excluded. The cancellation test `TestRotationWithoutBackoffStopsWhenRequestCancelled`, which previously timed out on the older upstream baseline, now passes as part of this full run.

This synchronization preserves upstream changes and keeps the existing Retry-After policy unchanged. The changelog retains only this unreleased fix after upstream cleared its prior entries. All 115 upstream provider-harness folders and collection metadata are preserved; the four existing Retry-After cases are appended as folder 105. No new wire behavior or regression cases are introduced by this merge.

The existing cancellation audit assertion was previously verified red against the pre-fix implementation. Original Retry-After regression coverage remains in place for seconds, HTTP dates, cancellation, deadlines, cap equality, stale headers, and streaming key rotation.

From the repository root:

```sh
node tests/e2e/api/runners/augment-provider-harness.mjs --source tests/e2e/api/collections/provider-harness.json --out tmp/harness-augmented.json
node tests/e2e/api/runners/filter-collection.mjs --source tmp/harness-augmented.json --out tmp/harness-filtered.json --provider openai --feature '#6971'
node --check tests/e2e/api/runners/retry-after-fixture.mjs
git diff --check
```

Structural validation passes (639 generated requests, 268 chained-variable guards) and selects exactly four cases. The three HTTP-status policy runner tests also pass, as do fixture syntax, Go formatting (normalizing Windows checkout line endings in the isolated Linux copy), and PR diff whitespace checks. They require `retryAfterFixture=1` and an isolated gateway using the local fixture; setup is documented in `tests/e2e/api/runners/retry-after-fixture.md`. Full gateway/Newman execution, paid provider tests, and multi-module `make test-all` were not run.

## Breaking changes

No API or configuration fields are added. Retryable responses with a valid `Retry-After` can now wait longer or stop retrying when the advertised wait exceeds the configured/request budget.

## Security considerations

No new external endpoints or credentials. The regression fixture uses dummy credentials and is opt-in.

## Checklist

- [x] Read the repository agent instructions and the contributing guide linked from README (the template's `docs/contributing/README.md` path is absent)
- [x] Added regression tests and provider-harness cases
- [x] Updated documentation and core changelog
- [x] Verified the core package builds and passes Linux race-enabled tests
- [ ] Full gateway/UI builds and full CI suite (not run)

### DIFF
--- a/core/bifrost.go
+++ b/core/bifrost.go
@@ -6178,6 +6178,7 @@ func executeRequestWithRetries[T any](
 	// True iff the previous attempt failed on rejected encrypted reasoning and we stripped
 	// it. Used to skip backoff: the payload changed, so there is nothing to wait out.
 	lastWasEncryptedContentStrip := false
+	var retryAfter time.Time
 
 	for attempts = 0; attempts <= config.NetworkConfig.MaxRetries+extraAttempts; attempts++ {
 		ctx.SetValue(schemas.BifrostContextKeyNumberOfRetries, attempts)
@@ -6345,8 +6346,27 @@ func executeRequestWithRetries[T any](
 
 			if !((lastWasPermanentKeyFailure && keyChanged) || lastWasEncryptedContentStrip) {
 				backoff := calculateBackoff(attempts-1, config)
+				if !retryAfter.IsZero() {
+					backoff = max(backoff, time.Until(retryAfter))
+				}
 				logger.Debug("sleeping for %s before retry", backoff)
-				time.Sleep(backoff)
+				if err := waitForRetry(ctx, backoff); err != nil {
+					var zero T
+					result = zero
+					bifrostError = &schemas.BifrostError{
+						IsBifrostError: true,
+						StatusCode:     Ptr(499),
+						Error: &schemas.ErrorField{
+							Type:    Ptr(schemas.RequestCancelled),
+							Message: "request cancelled during retry backoff",
+							Error:   err,
+						},
+					}
+					if errors.Is(err, context.DeadlineExceeded) {
+						bifrostError = providerUtils.NewBifrostTimeoutError("request timed out during retry backoff", err)
+					}
+					break
+				}
 			}
 		}
 
@@ -6441,6 +6461,9 @@ func executeRequestWithRetries[T any](
 		}
 
 		// Attempt the request
+		// Headers belong to one attempt; a failure before receiving a new response
+		// must not inherit Retry-After from an earlier retry or fallback provider.
+		ctx.ClearValue(schemas.BifrostContextKeyProviderResponseHeaders)
 		result, bifrostError = requestHandler(currentKey)
 
 		// Detect errors carried inside HTTP 200 streams before returning success.
@@ -6651,6 +6674,21 @@ func executeRequestWithRetries[T any](
 		}
 		lastWasPerKeyFailure = isPerKeyFailure
 		lastWasPermanentKeyFailure = isPermanentKeyFailure
+		retryAfter = time.Time{}
+		if !lastWasPermanentKeyFailure && !lastWasEncryptedContentStrip && attempts < config.NetworkConfig.MaxRetries+extraAttempts {
+			now := time.Now()
+			headers, _ := ctx.Value(schemas.BifrostContextKeyProviderResponseHeaders).(map[string]string)
+			retryAfter = retryAfterTime(headers, now)
+			if delay := retryAfter.Sub(now); delay > 0 {
+				deadline, hasDeadline := ctx.Deadline()
+				if delay > config.NetworkConfig.RetryBackoffMax || (hasDeadline && !retryAfter.Before(deadline)) {
+					// Do not shorten the provider's minimum wait or exceed the caller's
+					// budget. Preserve the upstream error for the existing fallback policy.
+					ctx.AppendRoutingEngineLog(schemas.RoutingEngineCore, schemas.LogLevelInfo, "Stopped retries: provider Retry-After exceeds the backoff or request budget")
+					break
+				}
+			}
+		}
 		// Remember the key used on this attempt so the next iteration's backoff check
 		// can detect whether key selection genuinely picked a different credential.
 		previousKeyID = currentKey.ID
--- a/core/changelog.md
+++ b/core/changelog.md
@@ -1,1 +1,2 @@
 - fix: drop namespace tools whose name Amazon Bedrock reserves (`web`, `image_gen`, `browser`, `python`) on the Bedrock and Bedrock Mantle Responses paths instead of forwarding them into a `tools.namespace` collision 400; a dropped Codex `web` namespace becomes the hosted `web_search` tool on Bedrock Mantle
+- [fix]: honor Retry-After within retry and request budgets [@jjj-n](https://github.com/jjj-n)
--- a/core/retryafter.go
+++ b/core/retryafter.go
@@ -1,0 +1,58 @@
+package bifrost
+
+import (
+	"context"
+	"math"
+	"net/http"
+	"strconv"
+	"strings"
+	"time"
+)
+
+// retryAfterTime returns the earliest retry time advertised by the current response.
+// A missing or malformed hint leaves the existing exponential backoff in control.
+func retryAfterTime(headers map[string]string, now time.Time) time.Time {
+	for name, value := range headers {
+		if !strings.EqualFold(name, "Retry-After") {
+			continue
+		}
+		value = strings.TrimSpace(value)
+		if value == "" {
+			return time.Time{}
+		}
+		isSeconds := true
+		for _, c := range value {
+			if c < '0' || c > '9' {
+				isSeconds = false
+				break
+			}
+		}
+		if isSeconds {
+			seconds, err := strconv.ParseUint(value, 10, 64)
+			// Saturate valid but enormous hints rather than overflowing or retrying early.
+			if err != nil || seconds > uint64(math.MaxInt64/int64(time.Second)) {
+				return now.Add(time.Duration(math.MaxInt64))
+			}
+			return now.Add(time.Duration(seconds) * time.Second)
+		}
+		if date, err := http.ParseTime(value); err == nil {
+			return date
+		}
+		return time.Time{}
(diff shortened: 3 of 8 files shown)


akshaydeo:
@Pratham-Mishra04 can you check this

jjj-n:
Hi @akshaydeo @Pratham-Mishra04 — gentle follow-up on this one when you have bandwidth.

Both CodeRabbit findings have been addressed (terminal error-classification order in the retry switch, per-run temp dir in the e2e fixture), and the latest bot review is clean.

The one item I'd appreciate a maintainer call on is the cap policy raised in #6971: when a Retry-After hint exceeds `retry_backoff_max` (e.g. a 60s hint with a 5s cap), this PR stops the provider's retries and preserves the original error so the fallback policy can apply, rather than retrying earlier than the provider permits. A hint equal to the cap is allowed. Happy to switch to clamping (wait the capped duration and retry anyway) if that matches the intended semantics better.

jjj-n:
Synced this branch with `dev` at `1a17949cd` and resolved the two append-only conflicts:

- preserved the upstream #7358 provider-harness group as folder 104 and moved the four Retry-After cases to folder 105
- retained all three unreleased changelog entries

Validation after the merge:

- targeted retry/backoff/Azure startup tests with `-race`: pass (4.473s)
- full core package with `-race`: pass (39.760s)
- provider-harness augmentation: 639 generated requests, 268 chained-variable guards
- `openai` + `#6971` filter: exactly 4 requests
- HTTP status-policy runner tests: 3/3 pass
- fixture syntax and `git diff --check`: pass

CodeRabbit's post-merge review is approved and the visible checks are green. This sync does not change the Retry-After behavior or the cap policy described above; the maintainer policy call remains the only open review question.

PR #7060, opened 2026-09-10
fix: responses event action string

TejasGhatte:
## Summary

Fixes a decode failure for `image_generation_call` items where OpenAI emits `action` as a bare JSON string (e.g. `"generate"`) rather than an object. The previous `UnmarshalJSON` implementation immediately tried to peek at a `.type` field, which cannot be read from a JSON string, causing the entire decode to fail with `"failed to peek at type field"`. This silently dropped the `response.output_item.done` and `response.completed` stream events carrying the image, leaving the stream without a terminal event and surfacing as a bogus `"provider closed the stream"` truncation error.

## Changes

- `ResponsesToolMessageActionStruct` now attempts to unmarshal the action as a bare string before falling back to the object type-peek, allowing `image_generation_call`'s `"generate"` (and similar) actions to decode correctly and round-trip through `MarshalJSON`.
- `ResponsesImageGenerationCall` gains fields for the generation settings OpenAI echoes back on completed items (`background`, `output_format`, `quality`, `revised_prompt`, `size`), which were previously dropped on the native `/v1` path.
- `ResponsesToolImageGeneration` gains an `action` field (`"generate"` | `"edit"` | `"auto"`) on the tool definition itself.
- Tests added to lock in the bare-string action fix, guard existing object action variants against regression, verify the new `ResponsesImageGenerationCall` settings fields round-trip correctly, and cover the `action` field on the tool definition.

## Type of change

- [x] Bug fix
- [x] Feature

## Affected areas

- [x] Core (Go)

## How to test

```sh
go test ./core/schemas/... -run TestResponsesToolMessageBareStringAction
go test ./core/schemas/... -run TestResponsesToolMessageObjectActionsUnchanged
go test ./core/schemas/... -run TestResponsesImageGenerationCallSettings
go test ./core/schemas/... -run TestResponsesToolImageGenerationAction
go test ./...
```

All four new tests should pass. Existing round-trip tests for computer use, web search, and local shell actions must continue to pass without capturing the string probe.

## Breaking changes

- [ ] Yes
- [x] No

## Security considerations

None. Changes are limited to JSON marshal/unmarshal logic for image generation tool call schemas.

## Checklist

- [ ] I read `docs/contributing/README.md` and followed the guidelines
- [x] I added/updated tests where appropriate
- [ ] I updated documentation where needed
- [x] I verified builds succeed (Go and UI)
- [ ] I verified the CI pipeline passes locally if applicable

PR #7067, opened 2026-09-11
[fix]: preserve bare-string tool call actions in streaming deep copy

### LINKED ISSUE #5834: [Bug]: Responses streaming drops image generation calls with string actions
## Prerequisites

- [x] I searched existing issues and discussions to avoid duplicates.
- [x] I reproduced the failure on the latest `dev` branch.

## Description

Bifrost drops valid OpenAI Responses image-generation events when an `image_generation_call` contains a string-valued `action`, for example `"action":"generate"`.

OpenAI models the image-generation call action as a JSON string (`"auto"`, `"generate"`, or `"edit"`). Bifrost currently models every `ResponsesToolMessage.action` as an object-only union and attempts to inspect an object `type` discriminator before selecting a variant. Decoding the valid string therefore fails before the event can be forwarded.

The streaming handler logs the decode error at warning level and continues to the next frame. This drops `response.output_item.added` / `response.output_item.done`, and it also drops `response.completed` when the terminal response contains the same image-generation item in `response.output[]`. Strict clients then wait for a terminal event until an idle or client timeout fires.

This is a schema compatibility failure. Increasing stream timeouts or adding keepalives does not make the rejected terminal frame parse successfully.

## Steps to reproduce

Run this regression against `core/schemas` on the current `dev` branch:

```go
func TestResponsesImageGenerationCallAcceptsStringAction(t *testing.T) {
    raw := []byte(`{"type":"response.output_item.done","sequence_number":7,"output_index":0,"item":{"id":"ig_123","type":"image_generation_call","status":"completed","action":"generate","result":"base64-image"}}`)

    var response schemas.BifrostResponsesStreamResponse
    if err := schemas.Unmarshal(raw, &response); err != nil {
        t.Fatalf("unmarshal image generation output item: %v", err)
    }
}
```

The same failure occurs for a terminal frame whose `response.output` contains that item:

```json
{
  "type": "response.completed",
  "response": {
    "id": "resp_123",
    "object": "response",
    "status": "completed",
    "model": "gpt-image-1",
    "created_at": 1785832000,
    "output": [
      {
        "id": "ig_123",
        "type": "image_generation_call",
        "status": "completed",
        "action": "generate",
        "result": "base64-image"
      }
    ]
  }
}
```

## Expected behavior

String-valued image-generation actions are preserved and relayed unchanged, including inside `response.completed`, while existing object-valued computer, web-search, web-fetch, local-shell, and MCP approval actions continue to work.

## Actual behavior

The valid event fails schema decoding and is skipped by the streaming loop, so clients may never receive `response.completed`.

## Affected area(s)

- Core (Go)
- Framework

## Version

- `dev` at `866527491e47f1cff4df90857210956e2d28e720`
- The same object-only schema is also present in the latest stable `transports/v1.6.7` source.

## Environment

```text
OS: macOS
Go: 1.26.3
Bifrost: latest upstream dev
Path: OpenAI-compatible Responses streaming
```

## Relevant logs/output

```shell
failed to peek at type field: Mismatch type struct { Type string "json:\"type\"" } with value "generate"
```

## Regression?

This appears to be a compatibility gap exposed by the newer string-valued image-generation action shape rather than a regression in an older supported shape.

## Severity

## Root cause and proposed direction

`ResponsesToolMessageActionStruct.UnmarshalJSON` assumes the wire value is always an object. The action union should retain all existing object variants and add a string variant for image-generation calls. The decoder should dispatch by JSON token kind, and unknown future string values should be preserved rather than rejected.

The request-side image-generation tool should also retain its optional `action`, and both core and framework deep-copy paths must preserve the new scalar variant.

## Related work

- #4713 and PR #4644 fixed an analogous strict-schema decode failure that silently dropped Responses events.
- #5010 and #5212 track SSE keepalives and intermediary idle timeouts. They can amplify the symptom but are not the schema root cause described here.

OpenAI reference: https://developers.openai.com/api/docs/guides/tools-image-generation#tool-options

citrocat:
## Summary

Follow-up to #7060, which taught `ResponsesToolMessageActionStruct` to decode bare-string actions (e.g. `image_generation_call`'s `"generate"`) into `ResponsesToolCallActionStr`. The Framework streaming accumulator's `deepCopyResponsesMessage` still copies only the object action variants, so a message carrying a bare-string action comes out of the copy with an all-nil action struct, and `MarshalJSON` on that copy fails with `responses tool message action struct is empty`. Anything consuming the accumulated stream through the Framework (plugins, aggregated responses) still loses the image-generation item for exactly the requests #7060 fixed on the decode side.

Together with #7060 this completes #5834. Supersedes #5835, which originally carried an independent fix for the decode side.

## Changes

- `framework/streaming/responses.go`: copy `ResponsesToolCallActionStr` in `deepCopyResponsesMessage`. Done by field name via the existing `copyOptionalStringFieldByName` helper (generalised to `any`) so the `framework` module keeps building against its pinned Core release, which predates the field.
- `framework/streaming/responses_test.go`: regression test `TestDeepCopyResponsesMessagePreservesBareStringAction`, asserting the value is preserved and not pointer-aliased. It skips when the pinned Core lacks the field.

## Type of change

- [x] Bug fix

## Affected areas

- [x] Framework (Go)

## How to test

```sh
# Against the local core (field present): must pass.
go work init ./core ./framework
cd framework && go test -race ./streaming/

# Against the pinned released core: framework must build, the new test skips.
cd framework && GOWORK=off go build ./... && \
  GOWORK=off go test ./streaming/ -run TestDeepCopyResponsesMessagePreservesBareStringAction -v
```

Without the `responses.go` change the new test fails with `bare-string action was not preserved`.

## Breaking changes

- [ ] Yes
- [x] No

## Related issues

Closes #5834
Related: #7060, #5835

### DIFF
--- a/framework/streaming/responses.go
+++ b/framework/streaming/responses.go
@@ -358,6 +358,10 @@ func deepCopyResponsesMessage(original schemas.ResponsesMessage) schemas.Respons
 				copyAction := *original.ResponsesToolMessage.Action.ResponsesMCPApprovalRequestAction
 				copy.ResponsesToolMessage.Action.ResponsesMCPApprovalRequestAction = &copyAction
 			}
+
+			// Bare-string actions (e.g. image_generation_call's "generate") live in a
+			// field the Framework's pinned Core release may predate; copy it by name.
+			copyOptionalStringFieldByName(copy.ResponsesToolMessage.Action, original.ResponsesToolMessage.Action, "ResponsesToolCallActionStr")
 		}
 
 		if original.ResponsesToolMessage.Caller != nil {
@@ -502,7 +506,7 @@ func copyRawMessageFieldByName(dst *schemas.ResponsesMessage, src schemas.Respon
 	}
 }
 
-func copyOptionalStringFieldByName(dst *schemas.ResponsesToolMessage, src *schemas.ResponsesToolMessage, fieldName string) {
+func copyOptionalStringFieldByName(dst, src any, fieldName string) {
 	srcField := reflect.ValueOf(src).Elem().FieldByName(fieldName)
 	if !srcField.IsValid() || srcField.IsNil() {
 		return
--- a/framework/streaming/responses_test.go
+++ b/framework/streaming/responses_test.go
@@ -2,6 +2,7 @@ package streaming
 
 import (
 	"fmt"
+	"reflect"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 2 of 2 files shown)


citrocat:
@TejasGhatte — follow-up to your #7060. `deepCopyResponsesMessage` in `framework/streaming` still copies only the object action variants, so a bare-string `ResponsesToolCallActionStr` comes out of the accumulator's copy as an all-nil struct and `MarshalJSON` then fails with "responses tool message action struct is empty" for exactly the image-generation items #7060 fixed on the decode side.

Single small commit with a regression test. The copy is done by field name so `framework` keeps building against its pinned Core release, which predates the field. Would appreciate a look when you get a chance.

PR #7118, opened 2026-09-13
image & video model fallbacks fixes

### LINKED ISSUE #6966: [Bug]: fallbacks silently re-target the primary provider for image edit / variation requests
### Description

`prepareFallbackRequest` rewrites `Provider` and `Model` for each request type it knows about, but **`ImageEditRequest` and `ImageVariationRequest` are missing from the list**. Both types declare a `Fallbacks []Fallback` field, so fallbacks are a supported, first-class feature for them.

```go
// core/bifrost.go — prepareFallbackRequest
fallbackReq := *req            // shallow struct copy: sub-request POINTERS are shared

if req.TextCompletionRequest  != nil { /* copy + set Provider/Model */ }
if req.ChatRequest            != nil { ... }
if req.ResponsesRequest       != nil { ... }
if req.CountTokensRequest     != nil { ... }
if req.CompactionRequest      != nil { ... }
if req.EmbeddingRequest       != nil { ... }
if req.RerankRequest          != nil { ... }
if req.OCRRequest             != nil { ... }
if req.SpeechRequest          != nil { ... }
if req.TranscriptionRequest   != nil { ... }
if req.ImageGenerationRequest != nil { ... }
if req.VideoGenerationRequest != nil { ... }
// ← no ImageEditRequest
// ← no ImageVariationRequest
return &fallbackReq
```

Because `fallbackReq := *req` is a shallow copy, an unhandled sub-request pointer is **shared with the original**, and that original still names the primary provider and model. So `GetRequestFields(fallbackReq)` returns the **primary** provider — and `tryRequest` routes the "fallback" straight back to the primary provider's queue.

This looks like an oversight rather than a deliberate exclusion: every sibling helper on `BifrostRequest` already handles both types — `GetRequestFields`, `SetProvider`, `SetModel` and `SetFallbacks` each have `case br.ImageEditRequest != nil:` / `case br.ImageVariationRequest != nil:` arms. `prepareFallbackRequest` is the only one that doesn't.

**Why this is worse than a no-op:** the fallback loop still runs to completion, and `SetFallbackRoutingInfo` still stamps `IsFallback: true`, `PrimaryProvider` and `PrimaryModel` onto the result. So a request that never left the primary provider is reported — in `RoutingInfo`, in the `x-bifrost-fallback-index` / `x-bifrost-routing-info-*` response headers, in the log row's `fallback_index`, and in the `fallback_index` metric label — as having been served by a fallback. The telemetry actively misreports what happened.

Net effect for an image-edit/variation request with N fallbacks configured: the primary is hit `N+1` times (each with its own full `max_retries` budget), the configured fallback providers are never contacted, and the failure is attributed to the wrong provider.

### Steps to reproduce

1. Configure two providers, e.g. `openai` and `azure`.
2. `POST /v1/images/edits` with a fallback:

```json
{
  "model": "gpt-image-1",
  "provider": "openai",
  "fallbacks": [{ "provider": "azure", "model": "gpt-image-1" }],
  "input": { "...": "..." }
}
```

3. Make the primary fail with something fallback-eligible (invalid key on `openai`, or a mocked 500).
4. Observe: the `azure` provider is never called. The request is retried against `openai` instead, and the response/log reports `fallback_index: 1` with `routing_info.is_fallback: true`.

Same with `POST /v1/images/variations`.

For contrast, the identical setup on `POST /v1/images/generations` (which *is* handled) correctly reaches `azure`.

### Expected behavior

The fallback attempt targets the configured fallback provider and model, exactly as it does for image *generation*.

### Actual behavior

The fallback attempt re-targets the **primary** provider and model, while telemetry reports it as a fallback.

### Affected area(s)

Core (Go)

### Version

### Environment

```text
- Not platform-specific; control-flow issue in core/bifrost.go.
```

### Regression?

Not a regression as far as I can tell — appears to date from when image edit/variation were added without a corresponding arm in prepareFallbackRequest.

### Suggested fix

Add the two missing blocks, mirroring the existing ones:

```go
if req.ImageEditRequest != nil {
	tmp := *req.ImageEditRequest
	tmp.Provider = fallback.Provider
	tmp.Model = fallback.Model
	fallbackReq.ImageEditRequest = &tmp
}

if req.ImageVariationRequest != nil {
	tmp := *req.ImageVariationRequest
	tmp.Provider = fallback.Provider
	tmp.Model = fallback.Model
	fallbackReq.ImageVariationRequest = &tmp
}
```

**Worth considering as a follow-up:** the current shape fails *silently* for any type it doesn't enumerate, which is what let this slip. Two options that would make it fail loudly instead:

- Build the fallback request through the existing `SetProvider` / `SetModel` methods on `BifrostRequest`, which already switch over every type — then the enumeration lives in one place instead of two that can drift.
- Or return `nil` (skip the fallback with the existing `Warn` log) when no sub-request arm matched, so an unhandled type is visible rather than silently re-running the primary.

Happy to open a PR for either.

akshaydeo:
## Summary

Fallback routing for image edit, image variation, and video edit requests was silently re-running the primary provider instead of the configured fallback. `prepareFallbackRequest` performed a shallow copy of `BifrostRequest` but had no arms for those three sub-request types, so the copy retained the primary's sub-request pointer. The fallback attempt was routed back to the primary while `RoutingInfo`, the `x-bifrost-routing-info-*` headers, log rows, and the `fallback_index` metric label all reported it as a legitimate fallback.

## Changes

- Added arms in `prepareFallbackRequest` for `VideoEditRequest`, `ImageEditRequest`, and `ImageVariationRequest`, each copying the sub-request and re-targeting provider and model to the fallback.
- Added a post-copy verification step that calls `GetRequestFields` on the prepared request and skips the fallback with a warning log if the result does not match the intended fallback provider and model. This ensures any future sub-request type added without a corresponding arm fails loudly rather than silently re-running the primary.
- Updated the nil-fallback log message in `handleRequest` and `handleStreamRequest` to read `missing provider config or unsupported request type` to reflect the new skip path.
- Added a reflection-driven unit test (`TestPrepareFallbackRequestRetargetsEveryFallbackCapableType`) that walks every pointer-to-struct field on `BifrostRequest` that declares a `Fallbacks` field, constructs a minimal primary request for each, calls `prepareFallbackRequest`, and asserts the result targets the fallback provider and model without mutating the original. The test fails automatically if a new fallback-capable type is added without a corresponding arm.
- Added an end-to-end Postman test case that posts a multipart image edit with an invalid primary model and a valid fallback, asserting the response is successful, the routing headers report the fallback, and an edited image asset is returned.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [x] Core (Go)
- [ ] Transports (HTTP)
- [ ] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

```sh
go test ./core/... -run TestPrepareFallbackRequestRetargetsEveryFallbackCapableType -v
go test ./core/... -v
```

For end-to-end validation, run the provider harness Postman collection and confirm test case **84** (`Fallback re-targets image edit requests`) passes: the response status is below 400, `x-bifrost-routing-info-is-fallback` is `true`, `x-bifrost-fallback-index` is `1`, and the response body contains a non-empty `data` array with a `url` or `b64_json` asset.

## Breaking changes

- [ ] Yes
- [x] No

## Related issues

Closes #6966

## Security considerations

None. The change only affects internal request routing logic and does not touch authentication, secrets, or user-facing data handling.

## Checklist


akshaydeo:
* **#7118** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/7118?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/maximhq/bifrost/7118?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `dev`


<h2></h2>

akshaydeo:
### Merge activity

* **Sep 13, 6:07 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/7118).

PR #7031, opened 2026-09-09
test(governance): add e2e coverage for regex entries in VK model allow/block lists

impoiler:

PR #6987, opened 2026-09-08
feat(schemas): add RE2 pattern lists for model allow/block rules on `Key` and `ProviderPermit`

impoiler:
## Summary

Model allow and block lists previously supported only exact names and the `*` wildcard. This PR adds RE2 pattern twins to both `Key` and `ProviderPermit`, enabling shape-based model access control without requiring exhaustive name enumeration. A new `ModelAccessRule` type centralises allow/block evaluation so all decision sites use consistent semantics.

## Changes

- Added `ModelPatternList` type holding RE2 patterns compiled as anchored, case-insensitive full matches (`(?i)^(?:pattern)$`). Patterns are cached after first compile via a `sync.Map`.
- `Key` gains `models_patterns` and `blacklisted_models_patterns` fields alongside the existing `models` and `blacklisted_models` exact lists.
- `ProviderPermit` gains `allowed_models_patterns` and `blacklisted_models_patterns` fields alongside its existing exact lists.
- `ModelAccessRule` struct unifies one allow list, one block list, and their pattern twins. Block always wins over allow; an empty allow side with no allow patterns admits nothing. `Key.ModelAccess()` and `ProviderPermit.ModelAccess()` return the composed rule.
- Patterns are tried against the bare model name and, when a provider is known, against `provider/model`, enabling provider-scoped patterns like `^openai/gpt-4.*`.
- `WhiteList` and `BlackList` remain exact-only; regex-looking literals in those lists match only themselves.
- The provider list-models pipeline accepts the `ModelAccessRule` so a patterns-only key can still enumerate the models it is permitted to serve.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [x] Core (Go)
- [ ] Transports (HTTP)
- [ ] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

```sh
go test ./core/schemas/...
```

New cases in `modelpattern_test.go` cover:

- `CompileModelPattern` caching and rejection of blank/invalid/lookbehind patterns.
- `ModelPatternList.Validate` rejects empty entries, the `*` wildcard, duplicates, and unparsable RE2.
- `ModelPatternList.Matches` verifies case-insensitivity, full-match anchoring, alternation, provider-qualified matching, and that invalid patterns never match.
[...]

PR #7130, opened 2026-09-14
fix(models): pre-feature parity for exact lists, plus review follow-ups on the regex pattern stack

impoiler:
Follow-up to the regex model pattern stack (#6987 to #7031). Goal: a deployment with no patterns configured behaves exactly as it did before the feature, the pattern side follows the same rules as the exact lists, and the naming feedback from the review on #6987 and #6988 is applied.

## Parity with the pre-feature code

- `IsModelAllowedForProvider` matches bare allow-list names with `slices.Contains` again. The feature had switched it to the case-insensitive `WhiteList.Contains`, which let a virtual key allowing `gpt-4o` admit a request for `GPT-4O` through the governance matcher and the enterprise load balancer. The test case that pinned the case-insensitive match now expects the old result.
- When keyconfig builds the provider-wide allow list it now skips entries the key blocks by pattern, the same way it already skipped entries blocked by the exact blocklist.

## Review follow-ups from #6987 and #6988

- `ModelAccessRule` is now `ModelRule`, and the accessors on `Key`, `ProviderPermit`, `TableKey`, `TableVirtualKeyProviderConfig` and `keyconfig.KeyEntry` are `ModelRule()`. "Access" already names the grant's access type, so the old names read as if they belonged to the grant. The catalog accessors are `ModelRuleFor` and `ModelRuleForProvider`, and the list-models pipeline field is `Rule`.
- The pipeline's `admitted` and `blocked` helpers duplicated `Admits` and `Blocks` with a different equality. `ModelRule` gains `AdmitsBy` and `BlocksBy` that take the equality function, `Admits` and `Blocks` call them with `strings.EqualFold`, and the pipeline passes its provider `MatchFns`. Same results, one place for the logic.
- `compileModelPattern` is unexported. Nothing outside `core/schemas` used it.
- Removed the unused `ModelAccessRule.IsRestricted`.

Two review comments are not applied here. Splitting `ModelPatternList` into allow and block types would produce two identical copies, since unlike `WhiteList` and `BlackList` the pattern lists have no allow or block semantics of their own. Folding `Aliases` into the rule would make a value type that answers "may this model be served" carry a name map that `Allows` never consults. Happy to do either as a separate change if you still want them.

## Audit

Every site where the old inline `Models.IsAllowed` / `BlacklistedModels.IsBlocked` checks were replaced by rule methods reduces to the original expression when the pattern lists are empty. The rename and the helper move are mechanical.

Tests: `core/schemas`, `core/providers/utils`, `core/providers/vertex`, `core/providers/huggingface`, `core/providers/runware`, `core/providers/openrouter`, `framework/modelcatalog/...`, `framework/grant`, `framework/configstore/tables`, `plugins/governance`, `transports/schema_test`.

Enterprise counterpart, which must merge after this one because it uses the new names: maximhq/bifrost-enterprise#2324.


impoiler:
Superseded. The pattern fields are being withdrawn in #7133 and replaced by a `regex:` prefix inside the existing lists in #7134.

PR #7132, opened 2026-09-14
fix(schemas): give an added message item an empty content array

### LINKED ISSUE #6487: [Bug]: response.output_item.added message items may omit required content array
### Prerequisites

- [x] I searched existing issues and pull requests for duplicates.
- [x] I reproduced the behavior on the current `dev` branch.

### Description

An OpenAI-compatible Responses upstream may omit `item.content` from the initial `response.output_item.added` event for an assistant message. Bifrost currently forwards the missing field unchanged.

Strict Responses clients expect message items to contain a content array from the moment the item is added, even before the first content part arrives. Grok Build 1.0.5 rejects the event when `content` is absent.

### Steps to reproduce

1. Stream a Responses request through Bifrost.
2. Have the upstream emit `response.output_item.added` with an item whose `type` is `message`, but without `content`.
3. Inspect the normalized event sent to the client.

Example upstream event:

```json
{"type":"response.output_item.added","item":{"id":"msg_1","type":"message","role":"assistant"}}
```

### Expected behavior

Bifrost emits the message item with `"content":[]` while preserving all other item fields and without mutating the source object.

### Actual behavior

The `content` field remains absent, and strict Responses clients reject the stream.

### Affected area(s)

Core (Go)

### Version

Reproduced on `transports/v1.6.11` and current `dev` (`0ac7f7142`).

### Environment

```text
Client: Grok Build 1.0.5
API: streaming Responses
Provider shape: OpenAI-compatible Responses
```

### Severity

Huang-404-Q:
Fixes #6487

An OpenAI-compatible Responses upstream may omit `item.content` from the initial `response.output_item.added` event for an assistant message. Bifrost forwarded the missing field unchanged, and strict Responses clients reject the whole stream over it — the reporter hits this with Grok Build 1.0.5, which expects a message item to carry a content array from the moment the item is added, before the first `content_part.added` arrives.

`WithDefaults()` already normalizes the sibling case: it defaults the item's `status` to `in_progress` on this event. The content array was simply not covered.

## The change

`core/schemas/responses.go` — in the `response.output_item.added` case of `WithDefaults()`, fill in an empty array when a **message** item arrives with no content:

```go
if result.Item != nil && result.Item.Content == nil &&
    result.Item.Type != nil && *result.Item.Type == ResponsesMessageTypeMessage {
    itemCopy := *result.Item
    itemCopy.Content = &ResponsesMessageContent{
        ContentBlocks: []ResponsesMessageContentBlock{},
    }
    result.Item = &itemCopy
}
```

Three details that are load-bearing:

- **`ContentBlocks` must be non-nil and empty.** `ResponsesMessageContent.MarshalJSON` emits `""` when both fields are nil (that is deliberate — OpenAI rejects a null `content` on input messages), so assigning a bare `&ResponsesMessageContent{}` would have sent `"content": ""`, which is not what a message item needs either. A non-nil empty slice marshals to `[]`.
- **Done on a copy.** `result.Item` aliases the source item (`result.Item = resp.Item`), so writing through it would mutate the caller's raw superset stream. The code-execution-carry strip a few lines above copies for the same reason.
- **Message items only.** `function_call`, `reasoning` and the other item types have their own shapes and must not gain a `content` field.

## Verification

New test `TestStreamOutputItemAddedMessageCarriesEmptyContentArray` in `core/schemas/responses_test.go`, three subtests:

1. `missing content becomes an empty array` — asserts `item.content` is present and exactly `[]`, that `id`/`type`/`role` survive on the item, and that the source item's `Content` is still nil. **Red before the change**: `item.content is absent; strict clients reject the event: map[id:"msg_1" role:"assistant" status:"in_progress" type:"message"]`.
2. `existing content is preserved` — a message item that already has a content block keeps it verbatim.
3. `non-message items are left alone` — a `function_call` item does not gain a `content` field.

```
go test ./schemas/    ok  github.com/maximhq/bifrost/core/schemas
go test .             ok  github.com/maximhq/bifrost/core
gofmt -l              clean
go vet ./schemas/     clean
```

`./internal/mcptests/` reports 35 failures here both with and without this change — the example MCP server binaries are not built on this machine (`make setup-mcp-tests` not run), so those are pre-existing and environmental, not related to this diff.


Huang-404-Q:
Fixed in 5bad18e — you were right, and it was the sharper half of the bug.

The content default was already on a copy, but the status default above it ran first against `result.Item`, which still aliased `resp.Item`. So a message item arriving with nil `Status` and nil `Content` had `"in_progress"` written into the caller's raw superset item — exactly the mutation the content default's own comment claimed to avoid.

Hoisted the copy above both defaults and applied both to it:

```go
if result.Item != nil {
    itemCopy := *result.Item
    result.Item = &itemCopy

    if result.Item.Status == nil {
        result.Item.Status = Ptr("in_progress")
    }
    if result.Item.Content == nil && ... {
        result.Item.Content = &ResponsesMessageContent{ContentBlocks: []ResponsesMessageContentBlock{}}
    }
}
```

Added the assertion you asked for. It has teeth — against the previous ordering it fails with:

```
responses_test.go:921: WithDefaults mutated the source item status: "in_progress"
```

`go test ./schemas/` and `go test .` both green; gofmt/vet clean.

PR #7135, opened 2026-09-14
fix(oauth): report an in-flight reauth as pending instead of stale authorized

### LINKED ISSUE #6477: [Bug]: MCP OAuth reauthorization prematurely closes popup on first poll tick and rejects completion with 409 (stale "authorized" status in oauth_configs)
### Prerequisites

### Description
When re-authorizing an existing OAuth-based MCP client (`auth_type='oauth'` or `per_user_oauth`), the frontend `<OAuth2Authorizer />` prematurely closes the browser authorization popup on the very first polling tick (~2s) and fires `complete-oauth`. The backend rejects `complete-oauth` with `HTTP 409 Conflict` because the in-flight authorization flow is still `pending`, but the UI misinterprets the 409 as a successful race condition and shows a false "Success" toast.

As a result:
1. The user's authorization popup is closed before they can approve consent at the upstream IdP.
2. No authorization code is exchanged, and the expired/revoked OAuth token is never updated.
3. The MCP client remains disconnected in `needs_reauth` status despite the UI indicating success.

### Root Cause
1. **Stale status polling in `getOAuthConfigStatus`**:
   `GET /api/oauth/config/{id}/status` (`transports/bifrost-http/handlers/mcpoauth2.go:175`) reads `oauth_configs.status`. For an existing client that was authorized in the past, `oauth_configs.status` is permanently `"authorized"` (it never regresses to `"pending"` during reauth).
2. **`OAuth2Authorizer.tsx` polling interval**:
   `startPolling` in `ui/app/workspace/mcp-registry/views/oauth2Authorizer.tsx:49-65` polls `getOAuthStatus(oauthConfigId)` every 2 seconds. On the first tick, it receives `status: "authorized"` from `oauth_configs`. It immediately calls `popupRef.current.close()` and triggers `handleOAuthComplete()`.
3. **Backend correctly detects premature completion**:
   `completeMCPClientOAuth` (`transports/bifrost-http/handlers/mcp.go:3321`) checks `isPrematureOAuthCompletion(pendingFlow, time.Now())`. Because the reauth flow in `mcp_oauth_flows` is still in status `"pending"`, it returns `409 Conflict`: `"Authorization has not completed yet: the browser flow may still be open..."`.
4. **UI false success masking**:
   `onConflict` in `ui/app/workspace/mcp-registry/views/mcpClientsTable.tsx:1178-1193` catches the 409 error, treats it as "flow completion raced / double submit", and shows a green success toast while dismissing the modal.

### Steps to reproduce
1. Configure an MCP client using OAuth authentication (e.g. Notion or any OAuth server) and complete initial setup.
2. Invalidate or let the credential expire so the client transitions to `needs_reauth`.
3. Click **Re-authenticate** in the MCP registry table.
4. Click **Continue** in the confirmation dialog to open the OAuth popup window.
5. Wait 2 seconds without interacting with the popup.
6. Observe that the popup window automatically closes and the UI shows a green success toast ("MCP client re-authorized successfully"), while the client remains in `needs_reauth` and cannot connect.

### Expected behavior
The status endpoint should reflect the active `mcp_oauth_flows` state (or the authorizer should poll the specific flow / await the `oauth_success` postMessage callback from the popup before completing), and the popup should remain open until authentication is completed by the user.

### Actual behavior
The popup closes after 2 seconds, `complete-oauth` returns 409, the UI displays a false success toast, and tokens are never refreshed.

### Affected area(s)
- Transports (HTTP)
- UI (React)
- Framework

### Version
dev @ `9dd8cfa16318433bd409c52c891caec551815dca` (2026-08-24)

### Severity

Huang-404-Q:
Fixes #6477

`GET /api/oauth/config/{id}/status` echoed `TableOauthConfig.Status`, which covers the **one-time bootstrap lifecycle only** (its own column doc says so). After a first successful auth that row reads `"authorized"` forever, and `InitiateUserOAuthFlow` never rewrites it when a reauth starts — it rotates the *flow* row to `pending` instead (framework/oauth2/main.go, the reauth path that reuses the existing row in place).

So during a reauth the authorizer UI's first 2s poll tick read the stale `"authorized"` and called `stopPolling()` + `handleOAuthComplete()`, closing the consent popup before the user could approve. `completeMCPClientOAuth` then rejected the premature completion with **409** — correctly, because `isPrematureOAuthCompletion` consults the still-pending flow row — and `onConflict` in the UI reads that 409 as a raced double-submit and shows a green success toast. Net result: the user is told the client was re-authorized while it stays in `needs_reauth` and the upstream token is never refreshed.

## The change

`transports/bifrost-http/handlers/mcpoauth2.go`, two pieces:

```go
func resolveOAuthConfigStatus(configStatus string, flow *configstoreTables.TableMCPOauthFlow, now time.Time) string {
	if configStatus == "authorized" && isPrematureOAuthCompletion(flow, now) {
		return flow.Status
	}
	return configStatus
}
```

plus a `currentAdminOauthFlow` lookup that maps the OAuth config to its MCP client (`GetMCPClientByOauthConfigID`) and reads that client's admin-mode flow row — the same pair of lookups `completeMCPClientOAuth` already performs.

Reusing `isPrematureOAuthCompletion` rather than open-coding the check is the point: the endpoint and the completion guard now read the same fact through the same predicate, so they cannot drift into disagreeing about whether a reauth is in flight.

The lookup is gated on `oauthConfig.Status == "authorized"`, the only status that can be stale this way — so a first auth (status `pending`) and the failure path (status `failed`) pay no extra query and are untouched.

The resulting lifecycle:

| state | flow row | reported status | UI |
|---|---|---|---|
| reauth in flight | `pending`, unexpired | `pending` | keeps polling — popup stays open |
| reauth succeeded | cleaned up by `CompleteOAuthFlow` | `authorized` | completes |
| reauth failed | cleaned up | `failed` | surfaces the failure |

## Verification

New `TestResolveOAuthConfigStatus` in `transports/bifrost-http/handlers/oauth_config_status_test.go`, six table cases. **Red before the change**, on exactly the reported case:

```
resolveOAuthConfigStatus("authorized", &{Status:pending, ExpiresAt:<future>}, now) = "authorized", want "pending"
```

Green after, with the boundaries that matter pinned alongside it: no flow row (a genuine completion cleaned it up) stays `authorized`; a `failed` flow row stays `authorized`; a first auth already reporting `pending` is unchanged; a `failed` config status is not masked by a flow row.

```
go test ./bifrost-http/handlers/   ok  12.3s
gofmt -l                           clean
go vet ./bifrost-http/handlers/    clean
go build ./...                     clean
```

## Not covered here

An **abandoned** attempt — the user opens the popup and never completes it — leaves a flow row that goes pending→expired rather than being cleaned up on a callback. Once it expires, `isPrematureOAuthCompletion` is false and this endpoint reports the config status again, so that path still ends in the old false-success rather than surfacing the expiry. Reporting `expired` there instead is a bigger semantic call (a stale expired row could also sit next to a legitimately authorized client), so I left it out of this fix — flagging it in case you want it handled separately.


Huang-404-Q:
Both addressed in f7a423bd7.

**Propagate OAuth flow lookup errors** — you are right, and the failure direction is what makes it matter: `nil` from that helper means "no reauth in flight", so folding a lookup error into it reported the config's own stale `authorized` and reopened the exact premature-completion path this PR exists to close. It now fails closed — the handler answers 500, which the authorizer UI's poll treats as transient (`catch { /* let polling continue */ }`) and retries. A slow-to-read status is a much better outcome than a popup that closes on a reauth nobody approved.

One case deliberately stays non-fatal: `ErrNotFound` from `GetMCPClientByOauthConfigID`. The admin-test popup validates an OAuth config *template* that may have no owning MCP client yet, and there is then no flow row to find — turning that into a 500 would break the template path. It returns `nil, nil` and the config status is reported as before. There is a test for it.

**Handler-level coverage** — added. The resolver tests could not reach the lookup or the response wiring, which is what you flagged:

- `TestGetOAuthConfigStatus_ReportsPendingFlowForAuthorizedConfig` — drives the handler over a mock store, asserts the response carries `pending`, and asserts both lookups actually happened (the client lookup runs with the right client ID, the flow lookup with `MCPAuthModeAdmin`).
- `TestGetOAuthConfigStatus_KeepsAuthorizedWithoutFlow` — the non-reauth path is unchanged.
- `TestGetOAuthConfigStatus_NoOwningClientIsNotAnError` — the template case above.
- `TestGetOAuthConfigStatus_FlowLookupFailureIs500` — both lookup failures, as a table.
- `TestGetOAuthConfigStatus_SkipsFlowLookupForNonAuthorizedConfig` — a first auth pays no extra query.

Verified to have teeth rather than assumed: folding the error back to `nil` fails the 500 test with `status = 200, want 500`.

```
go test ./bifrost-http/handlers/   ok  11.8s
gofmt -l / go vet ./bifrost-http/handlers/   clean
```

PR #7141, opened 2026-09-14
feat: add normalized `error_type` label to `bifrost_error_requests_total`

roroghost17:
## Summary

`bifrost_error_requests_total` previously carried only a raw `status_code`, which is insufficient to answer "whose fault was this?" — a 429 from a governance rate limit and a 429 from an upstream provider look identical, a 403 from a blocked model and one from a bad key are indistinguishable, and a 503 Bifrost shed under queue pressure reads the same as an upstream overload. This PR adds a normalized `error_type` label with a closed, prefix-structured vocabulary that separates fault domains without requiring per-provider error string parsing.

## Changes

- **`core/schemas/errortype.go`** — Introduces the `ErrorType` string type and its full vocabulary (`caller_*`, `policy_*`, `provider_*`, `bifrost_*`, `_OTHER`), plus `ErrorTypes` for conformance tests. Values are prefixed by fault domain so alarm expressions can match a family with a single regex rather than an enumerated list.

- **`core/schemas/errorclassify.go`** — Adds `ClassifyErrorType(err, requestType)` which resolves a failure to an `ErrorType` in priority order: a declared value on `ExtraFields.ErrorType` wins, then Bifrost's own string markers (`RequestCancelled`, `RequestDropped`, etc.), then the upstream status code. Status code classification deliberately ignores the provider's own `error.type`/`error.code` strings, which disagree across providers for the same condition and mislabel unrelated failures. Also adds `RequestTypeAddressesModel` and its inclusion map, so a 404 on a file or batch retrieval is not misattributed to a bad model name.

- **`core/schemas/bifrost.go`** — Adds `ExtraFields.ErrorType` to `BifrostErrorExtraFields` and the `NoKeySupportsModel` error type constant.

- **`core/bifrost.go`** — Declares `ErrorTypeProviderCredentialsExhausted` on the all-keys-exhausted 502 path (which would otherwise read as a generic upstream failure) and stamps `ErrorTypeCallerModelNotAvailable` plus a `StatusCode` and `Type` on the no-key-for-model path.

- **`core/providers/utils/utils.go`** — Declares `ErrorTypeCallerInvalidRequest` on `NewBifrostBadRequestError`, because OpenAI-family providers also send `invalid_request_error` (including for a 401), so only Bifrost's own use of that string is trustworthy.

- **`plugins/governance/main.go`** — Extracts a `refusal` helper and splits the previously merged `DecisionAccessBlocked/ModelBlocked/ProviderBlocked` case so each decision declares its specific `ErrorType`. A governance 403 is now distinguishable from a provider auth failure at the metric level.

- **`plugins/telemetry/main.go`** — Adds `error_type` as a label on `bifrost_error_requests_total`, populated by `ClassifyErrorType` using the same `requestType` that fills the `method` label.

- **`core/schemas/errorclassify_test.go`** — Full test suite: classification correctness for every path, a guard that every status 400–599 produces a non-empty label, a vocabulary conformance check, a declaration-wins property test, reachability of every vocabulary value, prefix structure enforcement, and `TestEveryRequestTypeIsCategorized` which parses `bifrost.go` via `go/ast` and fails if a new `RequestType` constant is added without a deliberate answer to "can a 404 here only mean the model?".

- **`plugins/governance/errortypedeclaration_test.go`** — Verifies that every refusal decision declares a non-empty `ErrorType`, that each declares the intended specific value (not a copy-paste of a single catch-all), and that the raw `Decision` string is still present in `Type` for wire compatibility.

- **Docs** — Updates `prometheus.mdx` with an Error Types reference section including the full vocabulary table, PromQL examples, precision caveats, and a note on `_OTHER`. Updates `telemetry.mdx` metric table and the example alert to filter `error_type!~"caller_.*"`. Updates the OpenAPI schema for `BifrostErrorExtraFields`.

**Design decisions:**
- Classification keys off status code rather than provider error strings. Per-provider normalization would let providers declare instead, shrinking the inference table rather than growing it.
- Declaration always beats inference. A Bifrost-produced error is never re-attributed to the upstream on a status it also carries.
- `_OTHER` uses the OTel catch-all name and should alarm at zero; a rise means a new refusal path was added without declaring its type.
- `error_type` is near-determined by `status_code` for upstream failures, so the new label adds minimal cardinality.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [x] Core (Go)
- [ ] Transports (HTTP)
- [ ] Providers/Integrations
- [x] Plugins
- [ ] UI (React)
- [x] Docs

## How to test

```sh
go test ./core/schemas/... ./plugins/governance/... ./plugins/telemetry/...
```

`TestEveryRequestTypeIsCategorized` will fail if a `RequestType` constant exists in `bifrost.go` without a corresponding entry in `modelAddressingByRequestType`. `TestEveryRefusalDeclaresAnErrorType` will fail if a governance decision produces an error without a declared `ErrorType`. `TestClassifyErrorTypeAlwaysLabelsAFailure` will fail if any status 400–599 produces an empty label.

To verify the metric label end-to-end, run Bifrost with Prometheus enabled and trigger a request that fails with a known status (e.g. a bad API key for a 401). Confirm `bifrost_error_requests_total` carries both `status_code="401"` and `error_type="provider_auth_failed"`. Trigger a governance rate-limit refusal and confirm `error_type="policy_rate_limited"` with `status_code="429"`.

## Breaking changes

- [x] Yes
- [ ] No

`bifrost_error_requests_total` gains a new `error_type` label. Any existing recording rules, dashboards, or alerts that query this metric without specifying `error_type` will continue to work (Prometheus aggregates over unspecified labels), but dashboards that hard-code the label set or use `without` clauses may need updating. The example alert in `telemetry.mdx` has been updated to include the recommended `error_type!~"caller_.*"` filter.

## Security considerations

None. The `error_type` label values are drawn from a closed vocabulary of string constants; no request content or credential material is included.

## Checklist


roroghost17:
* **#7141** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/7141?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/maximhq/bifrost/7141?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `dev`


<h2></h2>

core/schemas/bifrost.go
```diff
@@ -1910,6 +1910,10 @@ const (
 	RequestTimedOut          = "request_timed_out"
 	RequestDropped           = "request_dropped"
 	ProviderConnectionFailed = "provider_connection_failed"
+	// NoKeySupportsModel is stamped when no configured key serves the requested
+	// model. Classification does not read it — core declares ExtraFields.ErrorType at
+	// the same site — but it is part of the error contract.
+	NoKeySupportsModel = "no_key_supports_model"
```
Pratham-Mishra04: lets centralize these as well since everything is in errortype.go?

Pratham-Mishra04: for 5 of these

core/schemas/errortype.go
```diff
@@ -0,0 +1,70 @@
+package schemas
+
+// ErrorType is the normalized classification of why a request failed. StatusCode
+// stays the raw fact; this is the interpretation. Closed vocabulary — these are
+// metric label values.
+//
+// Prefixed by fault domain so alarms can match a family (error_type!~"caller_.*")
+// and so no value is ambiguous alone: a bare "rate_limited" would blend a governance
+// refusal with an upstream 429.
+type ErrorType string
```
Pratham-Mishra04: lets have a single errors.go which has both errortype.go and errorclassify.go contents

docs/features/observability/prometheus.mdx
```diff
@@ -248,6 +248,79 @@ The `path` label contains the matched route template (e.g. `/genai/v1beta/models
...
+### Error Types
+
+`bifrost_error_requests_total` carries two error dimensions. `status_code` is the raw fact. `error_type` is the normalized interpretation: a closed, low-cardinality vocabulary that answers **whose fault was it**, which is what alarms actually need.
+
+Values are prefixed by fault domain, so a success-rate alarm that should ignore caller mistakes is one clause rather than a list of reasons that grows over time:
+
+```promql
+# Failure rate, excluding faults the caller caused
+sum(rate(bifrost_error_requests_total{error_type!~"caller_.*"}[5m]))
+  / sum(rate(bifrost_upstream_requests_total[5m]))
```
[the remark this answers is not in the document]
roroghost17: The premise is right — `bifrost_upstream_requests_total` includes caller-faulted requests, since `UpstreamRequestsTotal.Inc()` (plugins/telemetry/main.go:1290) fires for everything reaching the LLM post-hook.

Keeping the query as-is though. It's labelled "failure rate, excluding faults the caller caused": the *failures* exclude caller faults, and a whole-traffic denominator is the usual SLO form — what fraction of everything we served failed for reasons that were ours. `success + filtered errors` answers a different question and, as noted on the earlier thread on this PR, brings its own imprecision: `bifrost_success_requests_total` counts semantic-cache hits that never reached a provider.

The real issue is that `bifrost_upstream_requests_total` counts requests that were never dispatched (policy refusals, cache hits), while the OTel exporter derives the same metric from attempt spans and excludes both. That parity bug is worth fixing, but redefining a metric that backs 11 documented ratios and existing dashboards belongs in its own PR, not one adding a label.

akshaydeo:
### Merge activity

* **Sep 14, 1:18 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/7141).
* **Sep 14, 1:19 PM UTC**: @akshaydeo merged this pull request with [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/7141).

PR #7145, opened 2026-09-14
mid-conversation switch cache checkpoint

### LINKED ISSUE #6825: [Bug]: Bedrock provider silently drops Anthropic compaction (compact_20260112) — capability matrix says supported, but Claude egress is Converse-only
### Prerequisites

### Description

Anthropic server-side compaction (`context_management.edits[].type = "compact_20260112"`, beta header `compact-2026-01-12`) is silently dropped when a Claude model is served by the Bedrock provider. The request succeeds, no warning is logged, the `anthropic-beta` header is forwarded, but no compaction ever happens and the conversation grows until it hits the model's context limit.

Root cause is a mismatch between the capability matrix and the Bedrock egress path:

1. `core/providers/anthropic/types.go` advertises compaction for Bedrock, so the request-body gate (`caps.SupportsCompaction`) keeps the `compact_20260112` edit and `AnthropicCompactionBetaHeader` is appended:
   https://github.com/maximhq/bifrost/blob/9537b2fadf42af90eb34ed47d3d4252e1beff4a0/core/providers/anthropic/types.go#L258-L260
   https://github.com/maximhq/bifrost/blob/9537b2fadf42af90eb34ed47d3d4252e1beff4a0/core/providers/anthropic/utils.go#L394
   https://github.com/maximhq/bifrost/blob/9537b2fadf42af90eb34ed47d3d4252e1beff4a0/core/providers/anthropic/utils.go#L1508
2. The Bedrock provider sends every Claude chat / responses request through the **Converse** API (`ChatCompletion`, `ChatCompletionStream`, `Responses`, `ResponsesStream` all comment "All other models (including Anthropic/Claude) use the Bedrock Converse API"):
   https://github.com/maximhq/bifrost/blob/9537b2fadf42af90eb34ed47d3d4252e1beff4a0/core/providers/bedrock/bedrock.go#L1239
   https://github.com/maximhq/bifrost/blob/9537b2fadf42af90eb34ed47d3d4252e1beff4a0/core/providers/bedrock/bedrock.go#L1430
   https://github.com/maximhq/bifrost/blob/9537b2fadf42af90eb34ed47d3d4252e1beff4a0/core/providers/bedrock/bedrock.go#L1752
3. AWS documents that compaction is **not supported by Converse**, only by `InvokeModel` / `InvokeModelWithResponseStream`: https://docs.aws.amazon.com/bedrock/latest/userguide/claude-messages-compaction.html ("Compaction is currently not supported by the Converse API, however it is supported with InvokeModel.")
4. `ToBedrockResponsesRequest` / `ToBedrockChatCompletionRequest` never read `ContextManagement`, so the edit is not even tunnelled through `additionalModelRequestFields`. The only reference to compaction in the Bedrock converter is the replay side, which flattens a previously returned `compaction` block into a plain text block:
   https://github.com/maximhq/bifrost/blob/9537b2fadf42af90eb34ed47d3d4252e1beff4a0/core/providers/bedrock/responses.go#L4837-L4838
5. `Passthrough` is unsupported on the Bedrock provider, so there is no raw-body escape hatch either:
   https://github.com/maximhq/bifrost/blob/9537b2fadf42af90eb34ed47d3d4252e1beff4a0/core/providers/bedrock/bedrock.go#L4072

Net effect: a client that enables compaction and points `bedrock/<claude-model>` (or a routing rule that targets Bedrock) gets the header, gets no error, and gets no compaction. The same request against `anthropic/<model>` compacts correctly (#6501 reproduces that path working).

Same state on `dev` at `c05bc0e2118532bc0a3aecc158b3431e36764206` (Converse-only comments at `bedrock.go` L1239/L1430/L1752, no `ContextManagement` handling in `bedrock/responses.go`).

**Production impact we observed (Bifrost `transports/v2.0.0`, `@ai-sdk/anthropic` 4.0.1 client on `/anthropic/v1/messages` with `bedrock/us.anthropic.claude-opus-4-8`):**

- Across ~290 long-running agent conversations over 45 days on the `bedrock/` path: **0** `compaction` blocks returned.
- The same workload on the direct Bedrock `InvokeModel` path (AI SDK's Bedrock provider, no gateway): ~24% of conversations compacted at least once.
- Weekly compaction count dropped from ~15-18 to ~2 the week we moved this traffic to the gateway.
- One conversation reached **865,793 input tokens per request** (cache read 865,581) with a 300,000-token compaction trigger configured — every turn re-bills ~865K cache-write/read tokens and the chat is a few turns from a hard context failure.

### Steps to reproduce

Needs ≥50k input tokens because that is Anthropic's minimum compaction trigger. Adapted from the repro in #6501; the only change is the model prefix.

```python
import os
from anthropic import Anthropic

client = Anthropic(
    base_url=os.environ["BIFROST_BASE_URL"].rstrip("/") + "/anthropic",
    api_key=os.environ["BIFROST_API_KEY"],
)

blob = "Warehouse aisle snapshot with SKU, units, and location. " * 80
messages = []
for i in range(30):
    messages.append({"role": "user", "content": f"Turn {i}. {blob}"})
    messages.append({"role": "assistant", "content": f"Ack {i}. {blob}"})
messages.append({"role": "user", "content": "Reply with just the word ok."})

for model in ["anthropic/claude-sonnet-4-6", "bedrock/us.anthropic.claude-sonnet-4-6"]:
    with client.beta.messages.stream(
        model=model,
        max_tokens=256,
        betas=["compact-2026-01-12"],
        context_management={
            "edits": [{
                "type": "compact_20260112",
                "trigger": {"type": "input_tokens", "value": 50_000},
            }]
        },
        messages=messages,
    ) as stream:
        msg = stream.get_final_message()

    kinds = [b.type for b in msg.content]
    print(model, "content block types:", kinds)
    print(model, "usage.iterations:", getattr(msg.usage, "iterations", None))
```

1. Configure Bifrost with both an `anthropic` provider and a `bedrock` provider (Claude Sonnet 4.6 or Opus 4.6+ enabled; both are listed as compaction-capable in the AWS doc above).
2. Run the script.
3. Compare the two outputs.

### Expected behavior

Both models return a `compaction` block first, `usage.iterations` contains a `compaction` iteration, and subsequent requests that replay that block are served from the summary. Alternatively, if Bedrock egress cannot support it, the edit should be stripped by the capability gate (`Compaction: false` for `schemas.Bedrock`) and the request should fail loudly or log a warning, rather than silently succeeding.

### Actual behavior

`anthropic/…` returns `['compaction', 'text']` with a compaction iteration. `bedrock/…` returns `['text']` only, no `iterations`, no compaction, no error, no log line, and the `anthropic-beta: compact-2026-01-12` header is forwarded to a Converse call that ignores it.

### Affected area(s)

Core (Go)

### Version

transports/v2.0.0 (tag `9537b2fadf42af90eb34ed47d3d4252e1beff4a0`); verified same code on `dev` @ `c05bc0e2118532bc0a3aecc158b3431e36764206`

### Environment

```text
- Bifrost: maximhq/bifrost:v2.0.0 docker image, config-as-code (config.json), single replica on ECS Fargate
- Client: @ai-sdk/anthropic 4.0.1 -> /anthropic/v1/messages (streaming), model "bedrock/us.anthropic.claude-opus-4-8"
- Also reproducible with a routing rule whose cel_expression matches provider == 'anthropic' && model == 'claude-opus-4-8' and targets provider "bedrock"
- Bedrock auth: ECS task-role credential chain; regions us-west-2 / us-west-1 / us-east-1
- Request type shown in the Bifrost log UI: "Responses Stream"
```

### Relevant logs/output

```text
Bifrost log entry for one affected request (Opus 4.8 via bedrock/, compact_20260112 with trigger 300000 in the request body):

  PROVIDER            BEDROCK
  MODEL               us.anthropic.claude-opus-4-8
  TYPE                Responses Stream
  INPUT TOKENS        865793
  OUTPUT TOKENS       596
  CACHE READ TOKENS   865581
  CACHE WRITE TOKENS  210
  (no compaction block in the response, no usage.iterations)
```

### Proposed fix

Either of these would resolve it; (1) is what we actually need.

1. **Route Claude requests that carry a `compact_20260112` edit to `InvokeModel` / `InvokeModelWithResponseStream`** with the native Anthropic Messages body (`anthropic_version`, `anthropic_beta: ["compact-2026-01-12"]`, `context_management`, `messages`, `system`, `tools`), and parse the native Anthropic response (including the `compaction` content block and `usage.iterations`) back into the Bifrost schema. `invoke.go` already has the Anthropic-shaped `BedrockInvokeRequest` type and the response converters (`toBedrockInvokeAnthropicResponse`, `toAnthropicInvokeStreamBytes`) that could be reused in the other direction. Gating this on "request has a compaction edit" keeps every other Claude request on Converse and limits blast radius.
2. **At minimum, make the capability matrix truthful**: set `Compaction: false` for `schemas.Bedrock` (and `bedrockmantle`) so the edit is stripped and the beta header is not forwarded, and emit a warning that compaction was dropped. This is the same class of problem as #6471 — a provider advertising a `context_management` capability it does not actually deliver — just in the opposite direction (silent no-op instead of a 400).

Happy to test a fix against our staging gateway with real Bedrock credentials.

### Regression?

Not a regression as far as we can tell — Claude-on-Bedrock appears to have always been Converse-only. We noticed it when we moved traffic from the AI SDK's direct Bedrock provider (InvokeModel, compaction works) onto the gateway.

### Severity

akshaydeo:
## Summary

Bedrock Converse requests now inline mid-conversation `role: "system"` messages in place for **every model family**, not just Claude. Claude Code appends a trailing `<total_tokens>` system reminder after every turn; previously, non-Anthropic models hoisted these into the top-level `system` block, which grew the prefix front on each turn. Because Bedrock's prompt cache is prefix-based for every model that has one (explicit `cachePoint` for Claude/Nova, implicit exact-prefix matching for OpenAI models on Converse), this caused a full cache write and zero cache reads on every turn — observed in the field against `global.openai.gpt-5.6-luna`.

## Changes

- `ToBedrockResponsesRequest` now unconditionally passes `inlineSystemReminders: true` to `ConvertBifrostMessagesToBedrockMessages` instead of gating on `IsAnthropicModelFamily`. The `false` path (hoist-everything) is retained only for rendering a stored response back into a Converse shape.
- Comments and doc strings updated to reflect that the model-family gate is gone and `false` is response-rendering-only.
- The test previously named `TestMidConversationSystemReminderHoistedForNonAnthropic` is renamed to `TestHoistEverythingModeStillHoistsAllSystemMessages` and its assertions updated to match the new semantics — the test still pins the `false` mode, but no longer implies it is the live non-Anthropic path.
- A new unit test `TestMidconvNonAnthropicWireShapeKeepsConversePrefixStable` verifies that for a GPT-5.6 model on Converse: the `system` block contains only the leading prompt and does not grow between turns, turn N's messages are a byte-identical prefix of turn N+1's, and every trailing `<total_tokens>` reminder is inlined as a user turn.
- New e2e growing-turn cells added for both Claude and `bedrock/us.openai.gpt-5.6-luna`: turn 1 writes the cache, turn 2 appends a new trailing `role: "system"` reminder and asserts `raw_request.system` did not grow and that turn 1's prefix was read back. These cells expose the defect that a byte-identical replay cannot detect.
- Prompt-caching docs updated to clarify that `global.openai.gpt-5.6-luna`-style IDs served by Converse receive Bedrock's implicit exact-prefix caching, and that mid-conversation system messages are now inlined for every model family.

## Type of change

- [x] Bug fix

## Affected areas

- [x] Core (Go)
- [x] Providers/Integrations
- [x] Docs

## How to test

```sh
go test ./core/providers/bedrock/... -run TestMidConversationSystemReminderStaysInline
go test ./core/providers/bedrock/... -run TestHoistEverythingModeStillHoistsAllSystemMessages
go test ./core/providers/bedrock/... -run TestMidconvNonAnthropicWireShapeKeepsConversePrefixStable
go test ./core/providers/bedrock/... -run TestMidconvAnthropicWireShapeReachesBedrockWithAllBreakpoints
go test ./...
```

For the e2e growing-turn cells, run the generated Postman collection against a Bifrost instance with `send_back_raw_request` enabled and `bedrockDirectAccessKeyId`/`bedrockDirectSecretAccessKey`/`bedrockDirectRegion` configured. Expect `CACHE_ANCHOR_REPORT` lines showing `rawSystemLen` equal between round 1 and round 2, and `hitRate >= 0.5` on round 2 for both Claude and `us.openai.gpt-5.6-luna`.

## Breaking changes

- [x] No

## Related issues

Closes #6825 (partial — Converse system-block growth for non-Anthropic models)

## Security considerations

None.

## Checklist


akshaydeo:
* **#7145** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/7145?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/maximhq/bifrost/7145?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#7142** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/7142?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `dev`


<h2></h2>

akshaydeo:
### Merge activity

* **Sep 14, 2:08 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/7145).
* **Sep 14, 2:09 PM UTC**: @akshaydeo merged this pull request with [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/7145).

PR #7187, opened 2026-09-15
wait_for_usage flag for custom providers

### LINKED ISSUE #7143: [Bug]: does_not_send_done_marker drops trailing Chat Completions usage and records zero cost
### Prerequisites

- [x] I reproduced this against the latest `dev` commit `3ebaefa00d29857c58fa1f28b5ac0efb2805c3f6` (2026-09-14)

### Description

For an OpenAI-compatible custom provider, enabling `custom_provider_config.does_not_send_done_marker` makes Bifrost stop reading the upstream SSE stream immediately after the first non-empty `finish_reason`.

This silently drops a trailing usage-only chunk when the upstream emits the common sequence:

```text
data: {"choices":[{"delta":{},"finish_reason":"stop"}],"usage":null}

data: {"choices":[],"usage":{"prompt_tokens":1000,"completion_tokens":100,"total_tokens":1100}}

# no data: [DONE]
: ping
: ping
```

The synthesized final Bifrost chunk then carries zero token counts, so logging, pricing, virtual-key usage, budgets, and cost attribution cannot account for the request correctly.

This is specifically the custom-provider opt-in path introduced by #6948. It is closely related to the historical failure fixed in #191 and the built-in provider classification fixed in #7065, but neither fix covers `does_not_send_done_marker=true`.

### Steps to reproduce

1. Configure an OpenAI-compatible custom provider:

```json
{
  "custom_provider_config": {
    "base_provider_type": "openai",
    "does_not_send_done_marker": true
  }
}
```

2. Have the upstream return a streaming response whose terminal events are:

```text
data: {"id":"chatcmpl-repro","object":"chat.completion.chunk","model":"repro-model","choices":[{"index":0,"delta":{},"finish_reason":"stop"}],"usage":null}

data: {"id":"chatcmpl-repro","object":"chat.completion.chunk","model":"repro-model","choices":[],"usage":{"prompt_tokens":1000,"completion_tokens":100,"total_tokens":1100}}

: ping
: ping
```

3. Send a streaming `/v1/chat/completions` request through Bifrost.
4. Inspect the synthesized final chunk and the stored usage/cost.

A temporary characterization test against latest `dev` sets the same per-attempt context key used by core and observes `final.Usage.TotalTokens == 0`:

```go
ctx.SetValue(schemas.BifrostContextKeyDoesNotSendDoneMarker, true)
stream, err := provider.ChatCompletionStream(ctx, passthroughPostHook, nil, testKey(), request)
// upstream sends finish_reason, then usage-only chunk
chunks := collectChunks(t, stream)
final := chunks[len(chunks)-1].BifrostChatResponse
require.Equal(t, 0, final.Usage.TotalTokens) // current behavior; expected 1100
```

Executed with:

```shell
cd core
go test ./providers/openai/... -run TestReproDoesNotSendDoneMarkerDropsTrailingUsage -count=1 -v
```

The characterization test passes, confirming the zero-usage behavior on `3ebaefa00`.

### Expected behavior

`does_not_send_done_marker` should allow a stream to terminate safely without requiring `[DONE]`, while still collecting valid trailing usage-only chunks sent after `finish_reason`.

### Actual behavior

Bifrost breaks at the `finish_reason` chunk, never reads the following usage-only chunk, and records zero tokens/cost.

### Root cause / code path

1. Core stamps the custom-provider option onto the request context:
   https://github.com/maximhq/bifrost/blob/3ebaefa00d29857c58fa1f28b5ac0efb2805c3f6/core/bifrost.go#L6986-L6994
2. `ProviderSendsDoneMarker` returns `false` whenever that context flag is set:
   https://github.com/maximhq/bifrost/blob/3ebaefa00d29857c58fa1f28b5ac0efb2805c3f6/core/providers/utils/utils.go#L3788-L3809
3. Usage is only accumulated if the usage event is actually read:
   https://github.com/maximhq/bifrost/blob/3ebaefa00d29857c58fa1f28b5ac0efb2805c3f6/core/providers/openai/openai.go#L1500-L1529
4. The read loop breaks immediately after `finish_reason` for providers classified as not sending `[DONE]`:
   https://github.com/maximhq/bifrost/blob/3ebaefa00d29857c58fa1f28b5ac0efb2805c3f6/core/providers/openai/openai.go#L1540-L1583
5. The final synthetic response therefore receives the still-zero accumulator:
   https://github.com/maximhq/bifrost/blob/3ebaefa00d29857c58fa1f28b5ac0efb2805c3f6/core/providers/openai/openai.go#L1620-L1639

The current tests demonstrate the discrepancy directly: the no-opt-in heartbeat path preserves trailing usage, while the opt-in test only checks termination/finish reason and does not include a trailing usage event:

- Opt-in test: https://github.com/maximhq/bifrost/blob/3ebaefa00d29857c58fa1f28b5ac0efb2805c3f6/core/providers/openai/streamtruncation_test.go#L305-L343
- No-opt-in trailing-usage test: https://github.com/maximhq/bifrost/blob/3ebaefa00d29857c58fa1f28b5ac0efb2805c3f6/core/providers/openai/streamtruncation_test.go#L380-L407

Latest `dev` already handles all three no-`[DONE]` termination shapes without this immediate break: EOF, consecutive post-finish heartbeat comments, and post-finish idle timeout (#6960 and #7115). The opt-in bypasses those safer termination paths and recreates the usage-loss problem.

### Suggested direction

Treat the option as "do not require `[DONE]`" rather than "break immediately at `finish_reason`". For example:

1. After `finish_reason`, continue accepting data events so a usage-only chunk can arrive.
2. End on EOF, `[DONE]`, the existing consecutive-heartbeat detector, or the existing post-finish idle timeout.
3. Alternatively, retire/deprecate the option now that latest `dev` handles no-`[DONE]` streams automatically.
4. Add a regression test with `does_not_send_done_marker=true` and the exact `finish_reason -> usage-only -> heartbeat` ordering.

### Affected area(s)

Core (Go), Providers/Integrations, Plugins

### Version

Latest `dev`: `3ebaefa00d29857c58fa1f28b5ac0efb2805c3f6` (2026-09-14). The behavior was introduced with PR #6948.

### Environment

```text
OS: macOS / darwin arm64
Go: go1.27.0
API: /v1/chat/completions streaming
Provider: custom provider with base_provider_type=openai
Config: does_not_send_done_marker=true
Upstream sequence: finish_reason -> usage-only chunk -> no [DONE]
```

### Relevant logs/output

```shell
=== RUN   TestReproDoesNotSendDoneMarkerDropsTrailingUsage
--- PASS: TestReproDoesNotSendDoneMarkerDropsTrailingUsage (0.01s)
PASS
```

The test passes because it characterizes the current incorrect result: the final usage object exists but `TotalTokens` is zero instead of 1100.

### Regression?

Yes. The generic early-exit behavior was fixed by #191. PR #6948 reintroduced the same data-loss mode for custom providers when `does_not_send_done_marker` is enabled.

### Severity

High (streaming token usage and cost accounting are silently incorrect)

akshaydeo:
## Summary

Fixes #7143: when `does_not_send_done_marker` is set on a custom provider, the SSE read loop broke immediately on `finish_reason`. Because OpenAI-compatible upstreams send the usage-only frame *after* that chunk (and before `[DONE]`), the loop never read it. The synthesized terminal chunk carried a zero-valued usage object, causing logging, pricing, virtual-key usage, budgets, and cost attribution to record the request at zero tokens and zero cost.

A new `wait_for_usage` flag on `custom_provider_config` tells Bifrost that the upstream does send that trailing usage-only frame, so the read loop should stay open past `finish_reason` until it arrives. Termination remains bounded: the usage chunk, two consecutive post-finish heartbeat comments, EOF, or `stream_idle_timeout_in_seconds`.

## Changes

- Added `WaitForUsage bool` to `CustomProviderConfig` and a corresponding `BifrostContextKeyWaitForUsage` context key. The key is set or cleared in `requestWorker` with the same set-or-clear discipline as `DoesNotSendDoneMarker`, so it never leaks onto a fallback provider that did not declare it.
- Added `WaitForStreamUsage(ctx)` helper in provider utils.
- Modified the `finish_reason` break condition in both the chat-completion and text-completion OpenAI streaming loops: the loop now holds open when `wait_for_usage` is true and the usage frame has not yet been seen (`usageSeen` bool), then breaks as soon as it arrives.
- `usageSeen` is set to `true` in the usage-accumulation branch of each loop, keyed on the presence of a usage object rather than on `TotalTokens > 0`, so an upstream that legitimately reports zero tokens still terminates correctly.
- Added four unit tests covering: opt-in with `wait_for_usage` collects trailing usage; opt-in without `wait_for_usage` still breaks on `finish_reason` (pinned default); a silent upstream with `wait_for_usage` ends on the idle timeout with no error chunk; and the text-completion loop receives the same fix.
- Added E2E Postman collection items (92.1–92.3) and two integration config entries (`openai_dnsdm`, `openai_dnsdm_wait`) to exercise the regression and its documented default against real OpenAI.
- Updated `CustomProviderConfig` type, Zod schemas, and form state in the UI so `wait_for_usage` is persisted. The toggle is nested under `does_not_send_done_marker` in both the add-provider sheet and the API-structure edit fragment, and is cleared automatically when its parent flag is disabled or turned off.
- Updated OpenAPI spec, Helm values schema, transport config schema, and `custom-providers.mdx` documentation with a warning that `does_not_send_done_marker` alone discards the usage frame, and guidance on pairing it with `wait_for_usage` and a low `stream_idle_timeout_in_seconds`.

## Type of change

- [x] Bug fix
- [x] Feature
- [ ] Refactor
- [x] Documentation
- [ ] Chore/CI

## Affected areas

- [x] Core (Go)
- [ ] Transports (HTTP)
- [x] Providers/Integrations
- [ ] Plugins
- [x] UI (React)
- [x] Docs

## How to test

```sh
# Core/Transports
go test ./core/providers/openai/... -run TestChat -v
go test ./core/providers/openai/... -run TestText -v

# UI
cd ui
pnpm i
pnpm build
```

**Integration (requires `OPENAI_API_KEY`):**

Register `openai_dnsdm` and `openai_dnsdm_wait` from `tests/integrations/python/config.json`, then run the Postman collection items 92.1–92.3 against a running Bifrost instance:

- 92.1 (`openai_dnsdm_wait/gpt-4o-mini`): `total_tokens` must be > 0.
- 92.2 (`openai_dnsdm/gpt-4o-mini`): `total_tokens` must be 0 (documented default, pinned).
- 92.3 (`openai/gpt-4o-mini`): `total_tokens` must be > 0 (control).

**New config fields:**

| Field | Type | Scope | Effect |
|---|---|---|---|
| `custom_provider_config.wait_for_usage` | `bool` | custom providers with `does_not_send_done_marker: true` | Keeps the SSE read loop open past `finish_reason` until the trailing usage-only chunk arrives. A silent upstream then ends on `network_config.stream_idle_timeout_in_seconds`. |

## Breaking changes

- [x] No

`wait_for_usage` defaults to `false`. Existing `does_not_send_done_marker` behaviour is unchanged and is now explicitly pinned by a unit test and an E2E item.

## Related issues

Closes #7143

## Security considerations

No new auth, secrets, or PII surface. The idle-timeout bound on `wait_for_usage` prevents a malicious or misbehaving upstream from holding a stream open indefinitely.

## Checklist


akshaydeo:
* **#7187** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/7187?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/maximhq/bifrost/7187?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#7184** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/7184?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `dev`


<h2></h2>

akshaydeo:
### Merge activity

* **Sep 15, 12:22 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/7187).
* **Sep 15, 12:22 PM UTC**: @akshaydeo merged this pull request with [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/7187).

PR #7196, opened 2026-09-15
adds make recipie to run integration tests locally

akshaydeo:
## Summary

Adds a `make test-integrations` target that runs both the Python and TypeScript SDK integration suites in parallel (one process per test file) against a single shared gateway, and improves the underlying `test-integrations.sh` script with live output streaming, a structured markdown failure report, `INTEGRATION_TEST_FILTER` support for narrowing to a single provider, and `SKIP_GATEWAY_START` support for attaching to an already-running gateway.

## Changes

- **`make test-integrations`**: New Makefile target that delegates to `test-integrations.sh --parallel-files`. Detects whether a gateway is already serving `HOST:PORT` and sets `SKIP_GATEWAY_START=1` if so, reuses `tmp/bifrost-http` by default (skipping the full UI build), accepts `JOBS=N` and `INTEGRATION=<name>` overrides, and validates bash 5.1+ before spending time on secrets or builds.

- **`SKIP_GATEWAY_START`**: When set, the script skips building the binary, skips starting the MCP fixture, and re-probes `/health` on the existing server rather than launching its own. `cleanup()` leaves the reused server running since it did not start it.

- **`INTEGRATION_TEST_FILTER` / `--parallel-files` narrowing**: The filter is validated to `[A-Za-z0-9_-]` only (interpolated into a glob, so metacharacters would escape the test directories). In the parallel path it filters both `test_<name>.py` and `test-<name>.test.ts` lists after globbing, errors if nothing matches, and lists available integrations to help correct typos. In the sequential path it warns that the filter is ignored rather than silently running the full suite.

- **Live output streaming**: `launch_test_file` now pipes each job through `tee` so lines appear as they happen. When more than one file runs concurrently, an `awk` prefix (`[label] line`) keeps concurrent output attributable. `PYTHONUNBUFFERED=1` is exported so Python does not block-buffer into the pipe. Only failed jobs are replayed at the end; passing jobs are not echoed a second time.

- **Markdown failure report**: `write_failure_report` writes `test-reports/integration-failures.md` on every run (overwriting a stale red report on a green run). It extracts failed case names from pytest (`::.*FAILED`) and vitest (`× suite > case`) output after stripping ANSI escapes, copies each failed job's full log to `test-reports/integration-<slug>.log`, and includes a collapsible failure-output excerpt per file. `TEST_FINAL_STATUS[]` is introduced alongside `TEST_STATUSES[]` because the throttle's bookkeeping array is empty for jobs the final `wait` reaps.

- **`mktemp` portability**: Removed `.log` suffixes from `mktemp` templates. BSD/macOS `mktemp` only substitutes trailing `X`s, so `name.XXXXXX.log` is created literally and the second run fails with "File exists". GNU `mktemp` accepts a suffix flag, which is why this only manifests outside CI.

- **bash version hint**: The "bash 5.1 required" error now also prints the macOS remedy (`brew install bash`).

- **Trailing whitespace**: Removed a stray trailing space in `cleanup()`.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Documentation
- [x] Chore/CI

## Affected areas

- [ ] Core (Go)
- [ ] Transports (HTTP)
- [x] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

```sh
# Run all integration tests (starts its own gateway if none is running)
make test-integrations

# Run only the OpenAI integration
make test-integrations INTEGRATION=openai

# Attach to an already-running gateway on a non-default port
make test-integrations PORT=8081

# Force a fresh binary build before running
make test-integrations BUILD=1

# Limit concurrency (e.g. to avoid 429s under quota pressure)
make test-integrations JOBS=4
```

After a run, inspect `test-reports/integration-failures.md` for a structured digest and `test-reports/integration-<slug>.log` for each failed file's full output.

Set `INTEGRATION_TEST_FILTER=openai` (or `INTEGRATION=openai` via the Makefile) to confirm the filter rejects invalid characters, errors on an unmatched name with a list of available integrations, and runs only the matching Python and TypeScript files.

## Breaking changes

- [ ] Yes
- [x] No

## Security considerations

`INTEGRATION_TEST_FILTER` is validated to `[A-Za-z0-9_-]` before being interpolated into a glob. A path separator, glob metacharacter, or `..` sequence would be rejected, preventing the filter from reaching files outside the two `tests/` directories. `mktemp` paths use random suffixes rather than predictable names to avoid symlink-based truncation attacks in world-writable `/tmp`.

## Checklist


akshaydeo:
* **#7196** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/7196?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/maximhq/bifrost/7196?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `dev`


<h2></h2>

akshaydeo:
### Merge activity

* **Sep 15, 4:49 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/7196).
* **Sep 15, 4:49 PM UTC**: @akshaydeo merged this pull request with [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/7196).

PR #7246, opened 2026-09-17
fix: guard WeightedRandom against a weight sum that rand.Intn rejects

cpsc:
### The bug

`Key.Weight` is an unvalidated `float64`. The provider-key HTTP handlers decode straight into `schemas.Key` — `transports/bifrost-http/handlers/provider_keys.go:111` (create) and `:211` (update) — and neither the `schemas` package nor the configstore GORM hooks compare it against anything. So `{"weight": -1}` is accepted, stored, and loaded back.

`WeightedRandom` sums `int(key.Weight * 100)` and falls back to uniform selection only when that total is **exactly** 0:

```go
totalWeight := 0
for _, key := range keys {
    totalWeight += int(key.Weight * 100)
}
if totalWeight == 0 {
    return keys[rand.Intn(len(keys))], nil
}
randomValue := rand.Intn(totalWeight)   // panics for n <= 0
```

A negative weight drives the total below zero instead, and `rand.Intn` panics for `n <= 0`. Conversions that do not fit in an `int` are implementation-defined in Go and on amd64 yield the most negative `int`, so NaN, both infinities, and overflowing weights all reach the same place.

### Why it ends the process

`WeightedRandom` is the default selector (`core/bifrost.go:299`) and is called from three places: `core/bifrost.go:4712`, `:7187` and `:9318`. The one at `:7187` runs inside `requestWorker`, which is launched as a bare goroutine at `:3935` and `:4632`. `core/bifrost.go` contains no `recover()`. A panic there is not a failed request — it takes down the process.

Reaching `:7187` requires `canRotate`, which per its own doc comment (`core/bifrost.go:9163`) is true only with two or more eligible keys, so this needs a provider with at least two keys configured, one of them carrying an unusable weight.

### Reproduction

Two keys on one provider, weights `-2` and `1`. Any request routed to that provider:

```
panic: invalid argument to Intn
```

The added test reproduces this — on `dev` without the source change it fails with exactly that panic.

### The fix

Clamp each key's contribution at 0, and widen the uniform fallback to any non-positive total.

Treating an unusable weight as 0 keeps the consequence on the key that carries it — it simply loses to a weighted sibling — rather than discarding the whole configured split, which is what returning early on the whole set would do.

### Tests

New `core/keyselectors/weightedrandom_test.go` covers a negative weight, per-key overflow, sum overflow, NaN, both infinities, that ordinary weighting still follows the configured ratio, and that the pre-existing all-zero uniform fallback is unchanged.

`gofmt`, `go vet ./keyselectors/` and `go test ./keyselectors/` are clean on top of `dev` at `fe1abf56b`.


cpsc:
Thanks — the riscv64 point is correct and I have taken it, plus a second case it led me to.

**Validate before converting.** The old guard tested the *converted* value for a negative, which only works where an out-of-range float-to-int conversion yields the most negative int. Go leaves that implementation-defined, and the implementations disagree — amd64 gives the most negative int, riscv64's `FCVT.L.D` saturates to the most positive one. So on riscv64 the old check accepted `+Inf` and handed an unusable key an enormous selection range. `weightUnits` now rejects NaN, both infinities, and anything outside `int`'s range before the conversion, so the result no longer depends on the target.

**Saturate the sum.** Clamping each key individually still leaves the total able to overflow: several in-range keys can push it past `math.MaxInt`, and a plain `+=` wraps. Wrapping to a negative is caught by the `<= 0` fallback, but wrapping far enough lands back on a *positive* number that nothing catches, and `rand.Intn` then gets a bound unrelated to the configured weights. `addWeightUnits` saturates instead, so the sum stays monotonic.

**Tests.** Table-driven coverage for NaN, both infinities, negative and out-of-range weights, plus the accepted ordinary/fractional/sub-unit/zero cases; a direct saturation table for `addWeightUnits`; and a selector-level case for sums that overflow at 2, 3, 4 and 8 keys. The saturation table does fail against a plain `+` — it reports the wrapped `-2` and `-9223372036854775808`.

**What the tests do not show.** The `weightUnits` table documents the contract and makes the behaviour target-independent by construction, but it does **not** reproduce the riscv64 failure on amd64 — with the old post-conversion check it passes there, because amd64's conversion happens to produce exactly the negative that check was looking for. I have no riscv64 machine to run it on, so that half of the argument rests on the conversion being implementation-defined plus RISC-V's documented saturating semantics rather than on an observed failure. Worth knowing when weighing the change.

`gofmt`, `go vet ./keyselectors/` and `go test ./keyselectors/` are clean.

PR #7269, opened 2026-09-17
fix(ui): key picker options by id so duplicate labels stop highlighting together

impoiler:
## What

Picking a user under Routing rules highlighted two rows at once when two users shared a name. The user could still select the right one, but the list looked like both were active.

`SearchSelect` gave each cmdk item `value={option.label}`. cmdk identifies the active row by that string, so two entries with the same name were one row as far as cmdk was concerned, and both got `data-[selected=true]` whenever either was hovered or arrow-keyed. The check mark was always right, because it reads the id.

## Fix

Items are keyed by `option.value` (the id) and carry the label and description as cmdk `keywords`. Async mode already bypasses cmdk filtering (`filter={() => 1}`), so search behaviour there is unchanged. Sync mode now filters through a small `filterByText` that scores the keywords rather than the value, since the value is an id and would otherwise fuzzy-match unrelated input.

## Audit

Checked every list in the app built on cmdk:

- `components/ui/multiSelect.tsx` had the same bug in a quieter form. Its items passed no `value` at all, so cmdk fell back to the rendered text and two options with the same label highlighted together. Fixed the same way in the grouped and flat branches. Checkbox state was already id-driven, so only the highlight changes.
- `components/ui/combobox.tsx` already keys items by value, in both `ComboboxItem` and the standalone select. No change.
- `skillVersionDialog`, `promptsViewHeader` and `logsHeaderView` use ids, id-prefixed composites, or static action rows. No change.
- The react-select surfaces (`asyncMultiselect`, `modelMultiselect`, `entityAssociationSelect`) identify options by `value`, and `EntitySelector` multi mode keys off the id. No change.
- Radix `SelectItem` lists pass provider, model and role names as values, but those names are the stored value and unique by definition, so they are not the same class of bug.

## Testing

`tsc --noEmit` passes, oxfmt applied, oxlint clean on `searchSelect.tsx`. The four oxlint warnings in `multiSelect.tsx` (lines 290-294, 441) predate this change.


impoiler:
* **#7269** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/7269?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/maximhq/bifrost/7269?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `dev`


<h2></h2>

akshaydeo (approved):

akshaydeo:
### Merge activity

* **Sep 17, 2:53 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/7269).
* **Sep 17, 2:54 PM UTC**: @akshaydeo merged this pull request with [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/7269).

PR #7309, opened 2026-09-18
core: send claimed results without racing ctx.Done

### LINKED ISSUE #7308: [Bug]: non-streaming caller hangs forever — worker keeps a ctx.Done() arm on a claimed delivery send (regression from #6972)
### Description

The delivery handoff added for #6972 makes a non-streaming caller **commit** to receiving once the worker wins `claimDelivery`. But the worker's send still races its own `<-req.Context.Done()` arm, so when the client has disconnected it discards the value roughly half the time — and the committed caller then blocks forever.

The caller's side is correct: after a failed `abandonDelivery` it waits in an inner `select` over `msg.Response` / `msg.Err` with **no ctx arm** (`core/bifrost.go:5837-5849`).

The worker's side is not:

```go
// core/bifrost.go:7444-7457 — requestWorker, result path (:7392-7405 is the same shape for errors)
deliveryTimer.Reset(5 * time.Second)
select {
case req.Response <- result:
	// Response sent successfully
case <-req.Context.Done():
	// Only reachable if the send could block, which a cap-1 channel drained
	// on acquire never does. A claimed caller is receiving; nothing to bill.
	bifrost.logger.Debug("Client context cancelled while sending response")
case <-deliveryTimer.C:
	...
}
```

That comment is the mistaken premise. The context is already done, and the cap-1 channel makes the send ready too — both arms are ready, and `select` picks uniformly at random. A **ready** send does not take precedence over a **ready** receive.

`sendWorkerError` (`core/bifrost.go:7481-7488`) already has the right shape, and its comment is accurate:

```go
// Claimed: the caller is committed to receiving and req.Err is a cap-1 channel
// drained on acquire, so this never blocks.
req.Err <- bifrostError
```

The two `requestWorker` sites are simply inconsistent with it.

### Version

### Affected paths

Any non-streaming entry point — `ChatCompletionRequest`, `ImageGenerationRequest`, etc., all of which reach `tryRequest` via `handleRequest`. Streaming is unaffected: it never goes through the claim, and its sends keep the ctx escape legitimately.

### Steps to reproduce

- One provider with `ConcurrencyAndBufferSize{Concurrency: 1}`, `MaxRetries: 0`, `DefaultRequestTimeoutInSeconds: 600`, pointed at an upstream that accepts the request, sends no response headers, and parks until its request context ends.
- Issue a non-streaming request with a cancellable context. Cancel it (or let a deadline expire) as soon as the upstream reports it has the request — i.e. while the attempt is in the header wait. The context transport closes the socket and the provider produces its terminal error at that same instant; that is the window.

### Observed

The caller never returns. A goroutine dump shows the caller parked at `core/bifrost.go:5843` while the provider worker sits idle at `:6942` waiting for its next request.

Reproduces in roughly 1 attempt in 40 under load (measured: 7 in 280 attempts, with six parallel `go test -race` processes on one machine). It is deterministically forceable by pausing the worker between `claimDelivery()` and its send while the caller's context ends.

### Expected

The caller returns with the terminal value it was committed to receiving, as `sendWorkerError` already guarantees on its path.

### Impact

In a gateway this leaks one handler goroutine per occurrence, permanently, on ordinary client disconnects — along with whatever per-request state that handler holds (in our case a distributed budget reservation that a background refresher then keeps alive indefinitely).

### Suggested fix

Make both claimed sends unconditional, matching `sendWorkerError`:

- `core/bifrost.go:7392-7405` (error path) → `req.Err <- *bifrostError`
- `core/bifrost.go:7444-7457` (result path) → `req.Response <- result`

Both are safe for the reason `sendWorkerError` already states: past a successful claim the caller is committed, and the channel is cap-1 and drained on acquire, so the send cannot block. `deliveryTimer` then has only its streaming use left (`:7419-7432`), where the ctx arm is legitimate because streaming never claims.

team-humaki:
After `claimDelivery` the caller is committed to receiving. The claimed send still selected on `ctx.Done`, so a ready cap-1 send raced a ready cancel and dropped the value — hanging `tryRequest`'s inner receive.

Match `sendWorkerError`: send unconditionally on the claimed path. Streaming keeps the ctx escape.

Fixes #7308


akshaydeo:
hey we are already on it and PR is ready - closing this

Issue #6972, opened 2026-09-08
[Bug]: abandoned-request billing is a ~50% coin flip when a client disconnects mid-request


Adarsh-jaiss:
### Description

When a client disconnects while a non-streaming request is in flight, whether Bifrost bills and logs the completed work is decided by **Go's random `select` case choice** — roughly a coin flip per request.

The worker's delivery handoff:

```go
// core/bifrost.go — requestWorker
req.sentAt = time.Now()
deliveryTimer.Reset(5 * time.Second)
select {
case req.Response <- result:
	// Response sent successfully
case <-req.Context.Done():
	bifrost.logger.Debug("Client context cancelled while sending response")
	// The provider already produced this non-streaming result
	// (consuming tokens). tryRequest returned on ctx.Done and will never
	// receive it, so bill/log it here.
	bifrost.billAbandonedTerminal(req, result, nil)
case <-deliveryTimer.C:
	bifrost.logger.Warn("Timeout while sending response, client may have disconnected")
}
```

`msg.Response` is a **buffered channel of capacity 1**, freshly drained on both acquire (`getChannelMessage`) and release (`releaseChannelMessage`). Nothing has been written to it at this point, so the send case is **always immediately ready**.

If the client has already disconnected, `req.Context.Done()` is *also* ready. The Go spec is explicit that when more than one case can proceed, `select` chooses **uniformly at random**. So:

- ~50% → the `ctx.Done()` branch runs `billAbandonedTerminal`, terminal post-hooks fire, governance bills and the logging plugin writes its row. **Correct.**
- ~50% → the send succeeds into a buffer nobody will ever read. `tryRequest` already returned on `ctx.Done` and deliberately does **not** release the message (to avoid corrupting another request's pooled channels), so the value is simply garbage-collected. No post-hooks, **no billing, no log row.**

The comment on the `ctx.Done()` branch shows the intent clearly — *"tryRequest returned on ctx.Done and will never receive it, so bill/log it here"* — but the branch only executes about half the time it should.

Net effect: for client-disconnect-during-flight, roughly half of the affected requests consume upstream tokens that are never billed to any budget or rate limit, never counted in cost metrics, and never appear in the logs. The provider charged for all of them.

### Steps to reproduce

1. Point Bifrost at a provider with a noticeable latency (or a mock that sleeps ~2s before responding).
2. Issue N requests that disconnect before the response returns:
   ```bash
[...]

PR #7312, opened 2026-09-18
agent and skill updates

akshaydeo:
## Summary

Regression tests — both unit tests and provider-harness cases — were previously required only for bug fixes in `core/`. This PR extends that requirement to every issue type (bug, feature, refactor) and every layer of the request path (`core/`, `framework/`, `transports/bifrost-http/`, `plugins/`). It also formalises the harness case as a named, mandatory output of the investigation and implementation workflow, not an optional recommendation.

## Changes

- **AGENTS.md**: Added a new top-level rule ("Every fix and every change ships with regression tests") that applies to all issue types and all request-path layers. Broadened the existing `core/`-only harness rule to cover `framework/`, `transports/bifrost-http/`, and `plugins/`. Added explicit exemption language (comments, internal renames, log lines) and updated the common-task checklists to include unit tests and a harness case as numbered steps.
- **`investigate-issue/SKILL.md`**: Renamed Step 5 from "Test Recommendations" to "Required Regression Tests" and rewrote it to mandate named tests rather than suggestions. Added Step 5g, which defines the harness regression case as a mandatory deliverable for any wire-visible change, including the structural validation commands, surgical-insert requirement, and exemption language. Added a "Regression Tests to Add" table to the Step 6 report template. Updated the implementation task list to include unit regression tests and the harness case as separate numbered pending items. Updated Step 7d references to reflect the broadened harness rule.
- **`resolve-pr-comments/SKILL.md`**: Extended the test-running guidance to cover all request-path layers, not only `core/` and provider packages.
- **AGENTS.md skills index**: Documented the `/harness-test-writer` skill, which was previously unregistered. Updated the `/investigate-issue` description to reflect that it now plans required regression tests rather than suggesting them. Changed "Four skills" to "Skills" to avoid a stale count.

## Type of change

- [ ] Bug fix
- [ ] Feature
- [ ] Refactor
- [x] Documentation
- [ ] Chore/CI

## Affected areas

- [ ] Core (Go)
- [ ] Transports (HTTP)
- [ ] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [x] Docs

## How to test

These are documentation and agent-skill changes with no runtime code. Validate by reviewing that:

1. The `investigate-issue` skill report template includes the "Regression Tests to Add" table and Step 5g harness case section.
2. The AGENTS.md "Testing" section opens with the new universal regression-test rule before the existing bug-fix and harness sections.
[...]

PR #7313, opened 2026-09-18
fixes abandoned connection handling for non-streaming connection

### LINKED ISSUE #7308: [Bug]: non-streaming caller hangs forever — worker keeps a ctx.Done() arm on a claimed delivery send (regression from #6972)

akshaydeo:
## Summary

Fixes a race condition (#7308) where the `requestWorker`, after winning the claim/abandon CAS race, could still discard a result or error by selecting the `ctx.Done()` arm in a `select` statement. Because both the cap-1 channel send and `ctx.Done()` are simultaneously ready when the caller's context is cancelled, Go's scheduler picks uniformly between them — meaning roughly half of claimed deliveries could be silently dropped, permanently stranding the committed caller.

## Changes

- On the non-streaming error path and the result path, once `claimDelivery()` succeeds, the send to `req.Err` or `req.Response` is now unconditional (a bare channel send with no `select`). A claimed caller is guaranteed to be blocked on the receive side, and both channels are cap-1 and drained on acquire, so the send never blocks.
- The streaming error path retains the `select` with `ctx.Done()` and timer arms because streaming callers keep a context escape and may have already released the message back to the pool.
- The `deliveryTimer` and its associated `select` block are removed from the claimed non-streaming paths entirely, eliminating the coin-flip hazard.
- Added `runClaimedDeliveriesWithDeadCaller` and two new test cases (`TestRequestWorkerDeliversClaimedResultAfterCallerCtxEnds`, `TestRequestWorkerDeliversClaimedErrorAfterCallerCtxEnds`) that deterministically reproduce the race: they cancel the caller's context at the exact moment the upstream completes, spin until the worker claims delivery, then commit to a blocking receive with no context escape — mirroring the `tryRequest` production code path. Running 24 iterations exposes a discard with probability `1 - 2^-24` if the bug is present.

## Type of change

- [x] Bug fix

## Affected areas

- [x] Core (Go)

## How to test

```sh
go test ./core/... -run TestRequestWorkerDeliversClaimed -v -count=5
go test ./...
```

Both `TestRequestWorkerDeliversClaimedResultAfterCallerCtxEnds` and `TestRequestWorkerDeliversClaimedErrorAfterCallerCtxEnds` should pass consistently. Before this fix, they would intermittently fatal with `"worker claimed delivery, then discarded the value; the committed caller would block forever (#7308)"`.

## Screenshots/Recordings

N/A

## Breaking changes

- [x] No

## Related issues

Closes #7308
Related: #6972

## Security considerations

None. This is a concurrency correctness fix with no changes to auth, secrets, or data handling.

## Checklist


akshaydeo:
* **#7318** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/7318?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7313** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/7313?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/maximhq/bifrost/7313?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#7312** <a href="https://app.graphite.com/github/pr/maximhq/bifrost/7312?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `dev`


<h2></h2>

akshaydeo:
### Merge activity

* **Sep 18, 1:29 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/7313).
* **Sep 18, 1:30 PM UTC**: @akshaydeo merged this pull request with [Graphite](https://app.graphite.com/github/pr/maximhq/bifrost/7313).

Issue #7337, opened 2026-09-19
[Feature]: PostgreSQL LISTEN/NOTIFY-based config sync across pods


ParthGupta-28:
### Prerequisites

- [x] I have searched existing issues and discussions to avoid duplicates

### Problem to solve

Bifrost's config store (providers, keys, virtual keys, routing rules, teams,
customers, budgets, model configs, pricing overrides, plugins, MCP clients,
webhook endpoints, etc.) is cached in memory per pod for fast request-path
access. In a multi-pod Kubernetes deployment backed by a shared PostgreSQL
config store, a write made through one pod's API (e.g. adding a provider key,
updating a virtual key's budget, disabling a routing rule) is only visible on
the pod that served the request. Every other pod keeps serving the old
in-memory state until it happens to restart.

This is silently incorrect in production: a revoked virtual key stays usable
on other pods, a newly added provider key isn't picked up cluster-wide, a
routing rule change or budget update doesn't take effect everywhere. There is
currently no mechanism to propagate a config mutation from one pod to its
siblings without a rolling restart.

### Proposed solution

Add a PostgreSQL LISTEN/NOTIFY based pub/sub layer to RDBConfigStore, scoped
to Postgres-backed deployments (SQLite is unaffected — single-process only).

- After every config-mutating write, RDBConfigStore publishes a pg_notify
  payload on a dedicated channel (bifrost_config_sync) describing the
  entity, action, and affected ID (providers, keys, virtual keys, VK
  provider/MCP configs, client config, MCP clients, plugins, routing rules,
  teams, customers, budgets, model configs, pricing overrides, virtual MCPs,
  webhook endpoints).
- Each pod's HTTP server holds a dedicated pgx connection issuing LISTEN on
  that channel, with exponential-backoff reconnect on connection loss.
- Received events are dispatched to the same in-memory reload/remove methods
  the REST handlers already call after a local write (ReloadProvider,
  ReloadClientConfigFromConfigStore, ReloadRoutingRule, ReloadTeam,
  ReloadCustomer, ReloadModelConfig, ForceReloadPricing, and their Remove*
  counterparts), so no new cache-invalidation logic is introduced — pods
  reuse the existing reload path.
- A synthetic full_reload event fires on every (re)connect to catch up on
  any NOTIFY missed while disconnected (LISTEN/NOTIFY has no replay or
  durability), covering pod startup and network blips.
- Explicitly out of scope: hot-path usage counters (budget/rate-limit spend)
[...]

PR #7338, opened 2026-09-19
Feat/pg notify config sync

ParthGupta-28:
## Summary

Adds PostgreSQL LISTEN/NOTIFY-based config sync so that a config change
made on one pod (provider/key, virtual key, routing rule, team, customer,
budget, model config, pricing override, etc.) propagates to every other
pod in a multi-pod deployment without a restart. Previously each pod only
saw its own in-memory config state, so writes made through one pod's API
were invisible on its siblings until they restarted. Scoped to the
Postgres-backed config store — SQLite is single-process only and was never
a valid multi-pod backing store, so it's unaffected.

## Changes

- `framework/configstore/pgnotify.go` (new): `ConfigChangeEvent` payload,
  `ConfigChangeListenable` interface, `pgNotifier` (publishes via
  `pg_notify` through the existing GORM pool, following the atomic
  pool-swap pointer), `pgListener` (dedicated `pgx` connection issuing
  `LISTEN`, exponential-backoff reconnect with jitter, synthetic
  `full_reload` event on every (re)connect to catch up on anything missed
  while disconnected — LISTEN/NOTIFY has no replay/durability).
- `framework/configstore/rdb.go`: `notifyChange(...)` call added after
  every config-mutating CRUD method (providers, keys, virtual keys, VK
  provider/MCP configs, client config, MCP clients, plugins, routing
  rules, teams, customers, budgets, model configs, pricing overrides,
  virtual MCPs, webhook endpoints).
- `framework/configstore/postgres.go`: wires the notifier/listener into
  `newPostgresConfigStore`; no-op on SQLite.
- `transports/bifrost-http/server/configlistener.go` (new): starts the
  listener at the end of `Bootstrap`, dispatches received events to the
  same in-memory reload/remove methods the REST handlers already call
  after a local write (`ReloadProvider`/`RemoveProvider`,
  `ReloadClientConfigFromConfigStore`, `ReloadRoutingRule`/
  `RemoveRoutingRule`, `ReloadTeam`/`RemoveTeam`, `ReloadCustomer`/
  `RemoveCustomer`, `ReloadModelConfig`/`RemoveModelConfig`,
  `ForceReloadPricing`) — no new cache-invalidation logic, pods reuse the
  existing reload path.
- `transports/bifrost-http/server/server.go`: lifecycle wiring — starts
  the listener after `Bootstrap`, stops it during graceful shutdown.

**Design decisions / trade-offs:**
- Notifier errors are logged and swallowed, never returned — a failed
  `pg_notify` must not fail the already-committed DB write that triggered
  it; sync degrades to eventual (via `full_reload` on next reconnect)
  rather than the write failing.
- Hot-path usage counters (budget/rate-limit spend) are deliberately
  excluded from notify — they already converge via their own periodic
  dump cycle, and notifying on every usage update would cause a
  notification storm.

## Type of change

- [x] Feature

## Affected areas

- [x] Core (Go)
- [x] Transports (HTTP)

## How to test

```bash
# Core/Transports
go version
go test ./...

# UI
cd ui
pnpm i || npm i
pnpm test || npm test
pnpm build || npm run build
 ```                                                                                                                                                                                
                                                                                                                                                                                    
 Verification done so far: go build ./... clean for both framework and                                                                                                              
 transports modules. Not yet verified: no unit tests added yet, no live                                                                                                             
 multi-pod Postgres LISTEN/NOTIFY smoke test performed. Implementation is                                                                                                           
 still in progress — see linked issue for remaining items.                                                                                                                          
                                                                                                                                                                                    
 Relates to #7337


akshaydeo:
Hey @ParthGupta-28 thanks for the PR, but adding this feature will break the gossip-based flow we have on enterprise. If you really want this feature to be there, you can create a fork of Bifrost and maintain that feature on your side. We'll be happy to pin this fork on our README so that, if folks are interested in that particular feature, they can use your fork. 

Thank you. 
Closing this PR.

ParthGupta-28:
> Hey @ParthGupta-28 thanks for the PR, but adding this feature will break the gossip-based flow we have on enterprise. If you really want this feature to be there, you can create a fork of Bifrost and maintain that feature on your side. We'll be happy to pin this fork on our README so that, if folks are interested in that particular feature, they can use your fork.
> 
> Thank you. Closing this PR.

Thanks for taking the time to review this, @akshaydeo — totally understand closing it.

One thing worth flagging before I go fork this: I'd already started a follow-up
(stacked on this branch, not yet opened as its own PR) adding a manual
`enable_cross_pod_sync` toggle specifically because I was worried about exactly this
overlap with Enterprise's gossip-based sync. I can take it a step further and make
it safe by default instead of opt-out:

- Auto-disable the LISTEN/NOTIFY notifier/listener whenever `cluster_config.enabled`
  is true, since that's the actual signal that a gossip-based sync layer is running.
  This would make it a strict no-op for every existing Enterprise cluster deployment
  with zero config changes required on their end — it would only ever activate for a
  Postgres-backed config store running *without* cluster mode enabled (which, as far
  as I can tell from the OSS-visible schema, is the case this was built for: a plain
  multi-pod OSS deployment with no other sync mechanism at all).
- Keep the manual `enable_cross_pod_sync: false` toggle too, as an explicit escape
  hatch for any edge case the auto-detection doesn't catch.

I don't have visibility into the gossip implementation itself, so I want to ask
directly rather than assume: is the concern specifically that both mechanisms would
redundantly fire on the same config change (extra DB/network load, but each side
still converges to correct state), or is there an actual correctness/state issue
where the two would conflict rather than just duplicate work? If it's the latter I'd
like to understand the failure mode so the auto-disable above actually closes the
gap rather than papering over it.

If a cluster_config.enabled-gated auto-disable addresses the concern, happy to reopen
with that change. If not — understood, and thanks for the offer to pin a fork; I may
take you up on it.

akshaydeo:
The issue is the feature conflicts with the flow we have on Enterprise side. Even if you enable or disable, or mistakenly enable it along with the gossip protocol, that will have a lot of cascading effects:

1. When we accept an external PR for doing something as particular as syncing the configuration, we end up maintaining that, not you.
2. Something that is not in our roadmap for OSS, we do not tend to accept PRs for those features purely because it becomes our overhead to maintain these two splits between OSS and enterprise.

ParthGupta-28:
That makes sense, thanks for the detail — especially the maintenance-overhead point.
A safer toggle only addresses the "used correctly" case; it doesn't change that
you'd be signing up to maintain compatibility with a system you can't fully expose
here, indefinitely, for something outside the OSS roadmap. Appreciate you explaining
the reasoning rather than just closing it.

I'll take you up on the fork offer if I end up needing this for my own deployment —
thanks again for the review.

Issue #7344, opened 2026-09-19
[Bug]: strict, allowed_tools, parallel_tool_calls and allowed_callers are dropped in provider conversion


Android-Login-Analysis:
### Prerequisites

- [x] I have searched existing issues and discussions to avoid duplicates
- [x] I am using the latest version (or have tested against main/nightly)

### Description

Eight tool fields a caller sets do not reach the provider. Each request succeeds, so the caller has no way to learn the restriction is not in force.

| # | route | caller sent | forwarded |
|---|---|---|---|
| 1 | Anthropic to Gemini | tool `strict: true` | declaration with no strict, no tool config |
| 2 | OpenAI to Gemini | tool `strict: true` | declaration with no strict, no tool config |
| 3 | OpenAI to Anthropic | `parallel_tool_calls: false` | `{"type":"auto"}`, no parallel control |
| 4 | Anthropic to Gemini | `disable_parallel_tool_use: true` | no tool config at all |
| 5 | OpenAI to Gemini | `parallel_tool_calls: false` | no tool config at all |
| 6 | OpenAI to Anthropic | `tool_choice.allowed_tools` with `tool_a` | `{"type":"any"}`, both tools kept |
| 7 | OpenAI to Gemini | `tool_choice.allowed_tools` with `tool_a` | no tool config, both tools kept |
| 8 | Anthropic to Gemini | tool `allowed_callers` | plain function declaration |

Case 6 is the clearest. `core/providers/anthropic/chat.go:681` maps the whole subset to `any`:

```go
case schemas.ChatToolChoiceTypeAllowedTools:
    toolChoice.Type = "any"
```

The caller said "you may call only `tool_a`". The upstream is told "call any of them".

Case 3 is a one-line gap. `core/providers/anthropic/types.go:1855` already has the target field:

```go
DisableParallelToolUse *bool  `json:"disable_parallel_tool_use,omitempty"`
```

`core/providers/anthropic/chat.go` never sets it. The identifier does not appear in that file.

Cases 1, 2, 4, 5 and 7 all land on the Gemini side, where the forwarded request carries `functionDeclarations` and no `toolConfig`. Gemini does define `functionCallingConfig.mode: "VALIDATED"`, and the constant exists at
`core/providers/gemini/types.go:315`, but it is only reachable from `applyServerSideToolInvocations`, which promotes AUTO when Google Search runs alongside function declarations. Nothing routes a caller's `strict` to it.

[...]

PR #7348, opened 2026-09-20
fix(anthropic): carry allowed_tools and parallel_tool_calls into the request

basil-k-aji-dev:
Fixes two of the cases in #7344 — cases 3 and 6, both OpenAI → Anthropic.

Both restrictions were dropped silently. The requests still succeeded, so a caller had no way to learn their restriction was not in force.

## `allowed_tools` became "call anything"

`core/providers/anthropic/chat.go:681` mapped the whole subset to one word:

```go
case schemas.ChatToolChoiceTypeAllowedTools:
    toolChoice.Type = "any"
```

A caller saying *"you may call only `tool_a`"* had **every** declared tool forwarded and the upstream told it could call any of them. The mode was lost as well, so an `auto` subset became a mandatory call.

Anthropic's `tool_choice` has no subset form — `auto`, `any`, `tool`, `none` — so the restriction is expressed where it can be: **a tool that is not declared cannot be called.** The declarations are now filtered to the allowed names, and the mode maps to `auto` or `any` rather than always `any`.

An allowed list naming tools that are not in the request keeps the original declarations. Forwarding none would turn a restriction into a request that cannot call anything, which is a larger change than the caller asked for.

## `parallel_tool_calls` had nowhere to go

`AnthropicToolChoice.DisableParallelToolUse` has existed in `types.go:1855` all along. Nothing ever set it — `ParallelToolCalls` appears nowhere in this provider:

```console
$ grep -rn "ParallelToolCalls" core/providers/anthropic/ | grep -v _test
(no output)
```

It is now carried across with the sense inverted. Anthropic accepts the field on `tool_choice` only, so when a caller disables parallel calls without expressing a preference about *which* tool runs, an explicit `auto` is emitted to hold it. `auto` is Anthropic's own default, so this adds no restriction that was not already there.

## Scope

Cases 3 and 6 only. The Gemini cases in #7344 are a different provider and a different conversion path; I have not touched them.

## Tests

Six cases in `core/providers/anthropic/toolchoicerestrictions_test.go`. All fail against the previous revision:

```
--- FAIL: TestToAnthropicChatRequest_AllowedToolsRestrictsDeclarations
--- FAIL: TestToAnthropicChatRequest_AllowedToolsAutoModeDoesNotForceACall
--- FAIL: TestToAnthropicChatRequest_ParallelToolCallsFalseDisablesParallelUse/with_an_explicit_tool_choice
--- FAIL: TestToAnthropicChatRequest_ParallelToolCallsFalseDisablesParallelUse/with_no_tool_choice_of_its_own
```

They cover: the subset is enforced, `auto` mode does not force a call, unknown names keep the declarations, the parallel flag is carried both with and without an existing tool choice, and `parallel_tool_calls: true` still sends nothing because it is already the default.

`go build`, `go vet` and `go test ./providers/anthropic/` all pass. The package was green before this change too, so the delta is only the new cases.

## Not verified

No live Anthropic call. These are pure transformation functions per `AGENTS.md`, so the tests assert the converted request shape rather than upstream behaviour. I have not exercised the Bedrock path that shares this converter.


basil-k-aji-dev:
@coderabbitai you are right, and the reasoning I gave in the comment was the wrong way round. Fixed in `bcdb52d`.

The case is an allowed list naming only tools the request does not declare. The intersection is empty, and I returned the original declarations — so a caller who restricted the model to a tool that is not present had **every tool they excluded** forwarded instead. That is the same failure this PR exists to stop, arrived at from the other direction.

My justification was that a request which cannot call anything is a larger change than the caller asked for. That does not hold. The caller asked for exactly one set; the intersection of that set with the declarations is empty; "nothing" is the faithful answer, not "everything".

The intersection is now honoured literally, and the tool choice says so:

```go
switch {
case len(anthropicReq.Tools) == 0:
    toolChoice.Type = "none"
case allowed.Mode == "required":
    toolChoice.Type = "any"
default:
    toolChoice.Type = "auto"
}
```

`"any"` or `"auto"` with no declarations is a request Anthropic rejects, so an empty result has to map to `none` rather than being left as a forced call over nothing.

`TestToAnthropicChatRequest_AllowedToolsUnknownNamesKeepDeclarations` asserted the old behaviour, so it is replaced by `...UnknownNamesPermitNothing`, which checks both halves — no tool forwarded, and the choice is `none`.

`go vet` and `go test ./providers/anthropic/` pass.

On your other two comments: `docs/providers/session-affinity.mdx` and `plugins/routing/complexitysession.go` are not touched by this PR, so I have left them alone. The session-affinity wording point looks reasonable on its own terms and would be better as its own issue where the people who own that feature can weigh in.

basil-k-aji-dev:
Correction, and an apology for the noise: this PR was carrying changes that are not mine, and my previous reply to @coderabbitai was wrong because of it.

I cloned with `--depth 1` and branched from whatever the tip was at that moment, which turned out to be `94fb9ee` — a commit that is **not on `main`**:

```console
$ gh api repos/maximhq/bifrost/compare/94fb9ee...main
status=diverged  ahead_by=3  behind_by=12
```

So the diff included the whole session-affinity change: 27 files and around 2,000 lines of work by @Pratham-Mishra04 and @akshaydeo. That is also why the CLA check wanted three signatures rather than one.

I told @coderabbitai that `docs/providers/session-affinity.mdx` and `plugins/routing/complexitysession.go` were "not touched by this PR". **That was false** — they were in my diff, and the bot was reviewing what I had actually pushed. The finding it raised on my own code was correct and is fixed; the other two it raised were correct to raise, and I dismissed them on a mistaken premise.

The branch is rebuilt on the current `main` (`a9937c9`) with only my two commits cherry-picked. It is now:

```
core/providers/anthropic/chat.go                          +87/-1
core/providers/anthropic/toolchoicerestrictions_test.go  +132/-0
```

`go build`, `go vet` and `go test ./providers/anthropic/` all pass against the real `main`.

Nothing about the fix itself changed — the allowed-tools filtering and the `parallel_tool_calls` carry-across are as described in the PR body, including the empty-intersection correction. Only the base is different.

basil-k-aji-dev:
@coderabbitai review

The branch has been rebased onto current `main` since your review - it was based on `94fb9ee`, which is not on `main`, so the diff previously carried the session-affinity work as well. It is now the two anthropic files only, and the empty-intersection finding you raised is fixed.

basil-k-aji-dev:
Found a flaw in my own change while re-reading it, fixed in `90b336d`. Flagging it rather than folding it in silently, because it is the same category of bug this PR exists to fix.

`anthropicReq.Tools` can hold **server tools** — MCP toolsets, web search, computer use — built by `convertServerToolToAnthropic` a few lines above the tool-choice block. My filter compared every declaration against the allowed names.

An OpenAI `allowed_tools` list describes function tools; its entries are `{type: "function", function: {name}}`. A server tool can never appear in one. So matching it on a name it cannot have removed it **every time a caller set `allowed_tools`** — restricting function calling to `tool_a` silently switched off a web search the caller had enabled elsewhere in the same request.

That is a stated intent quietly discarded, which is exactly what the PR is meant to stop, arrived at from the other side. Twice now on this change, which is worth noting.

The discriminator is already in the type: `convertFunctionToolToAnthropic` leaves `Type` nil, and server conversions set it (`mcp_toolset` and so on). Server tools now pass through untouched, and only nameable declarations are subject to the allowed list.

The empty-intersection case follows from that. "None of the allowed tools are declared" is a statement about *nameable* tools, so the choice becomes `none` when no function tool survives rather than when the list is empty — otherwise a request carrying a server tool would keep `any` over nothing the model could actually be told to call.

Two tests, both failing against the previous revision:

```
--- FAIL: TestToAnthropicChatRequest_AllowedToolsLeavesServerToolsAlone
    a server tool must survive an allowed_tools filter
--- FAIL: TestToAnthropicChatRequest_AllowedToolsNoneNameableStillMeansNone
```

`go vet` and `go test ./providers/anthropic/` pass.

akshaydeo:
dont add new test files please - move these tests into existing test files and update the PR

core/providers/anthropic/toolchoicerestrictions_test.go
```diff
@@ -0,0 +1,177 @@
+package anthropic
+
+import (
+	"testing"
+
+	schemas "github.com/maximhq/bifrost/core/schemas"
```
akshaydeo: move these tests to existing test files

akshaydeo (changes requested):

basil-k-aji-dev:
Done — the seven cases now live in `core/providers/anthropic/chat_test.go` alongside the existing `TestToAnthropicChatRequest_*` tests, and `toolchoicerestrictions_test.go` is gone. No test content changed; testify was added to that file's imports and is already used elsewhere in the package.

The PR is two files now:

```
core/providers/anthropic/chat.go       +111/-1
core/providers/anthropic/chat_test.go  +170/-0
```

`go vet` and the full anthropic package pass.

Worth flagging one thing you may not have seen, since it landed just before your review: I found a flaw in my own change and fixed it in `90b336d`. The allowed-tools filter was stripping **server tools** — MCP toolsets, web search, computer use — because they can never be named in an OpenAI `allowed_tools` list, which is function-shaped. So restricting function calling to one tool silently switched off a web search the caller had enabled elsewhere in the same request. Server tools now pass through untouched.

basil-k-aji-dev:
@coderabbitai right again, and this is the third finding on the same change. Fixed in `7dce6e5`.

I used `Type != nil` to recognise a server tool. That holds for web search, web fetch and the computer-use family, and not for an MCP toolset:

```go
return AnthropicTool{MCPToolset: toolset}, true
```

Type nil, Name empty. The struct records it explicitly, which makes the miss worse rather than better:

```go
// MCP toolset (mcp-client-2025-11-20 format) — embedded when Type is nil and MCPToolset is set
MCPToolset *AnthropicMCPToolsetTool `json:"-"`
```

So the filter matched an MCP toolset's empty name against the allowlist and removed it — the same silent capability loss the previous commit set out to fix, still happening for the one shape I did not check. I had read `Type: "mcp_toolset"` a few lines above and not registered that it sits on the nested type, not on `AnthropicTool`.

Both shapes are preserved now, and `countFunctionTools` excludes both, so a request whose nameable tools are all excluded still yields `none` rather than `any` over nothing callable.

The test fixture is the part worth calling out. It used `Type` to stand in for a server tool, so it *could not* have caught an MCP toolset — it encoded the same assumption as the code it was checking. It now carries both shapes, and both cases fail against the previous revision:

```
--- FAIL: TestToAnthropicChatRequest_AllowedToolsLeavesServerToolsAlone
    an MCP toolset has Type nil and an empty Name; it must not be matched against the allowlist
--- FAIL: TestToAnthropicChatRequest_AllowedToolsNoneNameableStillMeansNone
    only tool_a is nameable; neither server shape counts
```

`go vet` and the full anthropic package pass.

akshaydeo (approved):

PR #7357, opened 2026-09-20
[fix]: preserve Gemini image signatures in Responses streams

### LINKED ISSUE #7342: [Bug]: Gemini Responses stream drops a generated image part that carries a thoughtSignature
### Description

On the Gemini Responses **streaming** path, a model-generated image part that also carries a `thoughtSignature` is dropped: only the signature is processed, and the image never reaches the client.

`processGeminiPart` dispatches on a `switch` whose `ThoughtSignature` case comes before the `InlineData` case:

```go
// core/providers/gemini/responses.go:1522 processGeminiPart
switch {
...
case part.ThoughtSignature != nil:        // ~:1557
	// thought signature
case part.FunctionResponse != nil:
	...
case part.InlineData != nil:              // ~:1564
	responses = append(responses, processGeminiInlineDataPart(part, state, sequenceNumber)...)
...
}
```

A part with both `thoughtSignature` and `inlineData` matches the first case only, so `processGeminiInlineDataPart` never runs for it.

### Version

### Impact

Gemini image-generation models sign their image parts, so on the Responses streaming path a generated image is silently replaced by an encrypted reasoning item, and the client gets an empty answer with `stop`. The Chat Completions streaming converter does not have this problem: there the signature and the image travel in the same delta (`providers/gemini/chat.go:670-677`).

### Steps to reproduce

Point a Gemini provider at an upstream whose `streamGenerateContent` frame contains a part with both `"thoughtSignature":"..."` and `"inlineData":{"mimeType":"image/png","data":"..."}`. Issue a streaming Responses request. No `output_item.added` / `content_part` event carries the image.

### Suggested fix

Process the signature and the inline data independently rather than as exclusive `switch` arms — e.g. handle `ThoughtSignature` as an additional step (attach it to the item the part produces) and then fall through to the content cases, mirroring what the Chat converter does.

axelray-dev:
## Summary

Gemini Responses streaming drops a generated image when the same Gemini part includes `inlineData` and `thoughtSignature`. Preserve both fields in the streamed Responses output.

## Changes

- Dispatch media parts before standalone thought signatures.
- Attach a media part signature to its output content block and restore it on Gemini conversion.
- Add regression coverage for the streamed image, signature, and round trip.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Chore/CI

## Affected areas

- [x] Core (Go)
- [ ] Transports (HTTP)
- [x] Providers/Integrations
- [ ] Plugins
- [ ] UI (React)
- [ ] Docs

## How to test

From `core/`:

```sh
GOTOOLCHAIN=local GOSUMDB=off GOCACHE=$PWD/../.gocache GOMODCACHE=$PWD/../.gomodcache GOTMPDIR=$PWD/../.go-tmp TMPDIR=$PWD/../.tmp /home/ubuntu/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.27.0.linux-amd64/bin/go test ./providers/gemini/ -count=1 -run TestGeminiResponsesStreamInlineDataWithThoughtSignatureSurvives
GOTOOLCHAIN=local GOSUMDB=off GOCACHE=$PWD/../.gocache GOMODCACHE=$PWD/../.gomodcache GOTMPDIR=$PWD/../.go-tmp TMPDIR=$PWD/../.tmp /home/ubuntu/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.27.0.linux-amd64/bin/go test ./providers/gemini/ -count=1
GOTOOLCHAIN=local GOSUMDB=off GOCACHE=$PWD/../.gocache GOMODCACHE=$PWD/../.gomodcache GOTMPDIR=$PWD/../.go-tmp TMPDIR=$PWD/../.tmp /home/ubuntu/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.27.0.linux-amd64/bin/go vet ./providers/gemini/
```

The focused test passed with `ok github.com/maximhq/bifrost/core/providers/gemini 0.012s`. The full Gemini package tests passed with `ok github.com/maximhq/bifrost/core/providers/gemini 0.383s`, and `go vet` completed without diagnostics. The regression verifies that the image block is emitted, its base64 signature remains attached, no standalone reasoning item is emitted, and Gemini stream conversion round trips the original signature.

## Breaking changes

- [ ] Yes
- [x] No

## Related issues

Fixes #7342

## Security considerations

None.

## Checklist


axelray-dev:
I tightened the PR description and verified the fix on the current head. From core/, the focused regression, full providers/gemini package tests, and go vet all pass. The regression confirms the image remains in the streamed output with its thought signature attached and round-trips back to Gemini. Please re-review when convenient.

PR #6932, opened 2026-09-07
fix: preserve budget state across edits

danpiths:
## Summary

Fix two governance budget edit paths that could silently lose persisted state:

1. Editing a team budget's reset frequency could replace the stored budget
   instead of updating it, resetting accumulated usage even when the user chose
   to preserve usage.
2. Editing a provider budget could discard a quarterly budget's
   `quarter_start_month`, changing its reset calendar to the default January
   schedule.

## Changes

- Include persisted team budget IDs in update requests. Temporary IDs used for
  newly added UI rows are still omitted.
- Reconcile team budgets by stable ID first, matching the pattern used by other
  governance entities.
- Retain `reset_duration` matching as a compatibility fallback for clients that
  do not send budget IDs.
- Preserve and submit provider budget `reset_config` values.
- Use the shared budget signature when detecting provider budget changes so
  fiscal-quarter changes also trigger the preserve/reset usage choice.
- Add focused coverage for preserving a team budget's identity, usage, and last
  reset when its frequency changes.

### Team budget example

Before this change, the team form omitted the persisted budget ID:

```text
stored budget
┌──────────┬─────────┬───────────┬────────┐
│ ID       │ limit   │ frequency │ usage  │
├──────────┼─────────┼───────────┼────────┤
│ budget-a │ $100    │ monthly   │ $40    │
└──────────┴─────────┴───────────┴────────┘
                         │
                         │ user changes monthly → weekly
                         ▼
request: { limit: $150, frequency: weekly }
                         │
                         │ no ID; weekly does not match monthly
                         ▼
delete budget-a + create budget-b
┌──────────┬─────────┬───────────┬────────┐
│ budget-b │ $150    │ weekly    │ $0     │
└──────────┴─────────┴───────────┴────────┘
```

The backend was not explicitly clearing usage. It interpreted the changed
frequency as a removed monthly budget and a new weekly budget. The replacement
budget naturally started at zero.

The request now includes the persisted ID:

```text
request: { ID: budget-a, limit: $150, frequency: weekly }
                         │
                         │ stable ID matches the existing row
                         ▼
update budget-a in place
┌──────────┬─────────┬───────────┬────────┐
│ budget-a │ $150    │ weekly    │ $40    │
└──────────┴─────────┴───────────┴────────┘
```

The reconciliation order is now:

```text
request contains an ID?
├─ yes → match the existing budget by ID
└─ no  → fall back to matching by reset duration

matched → update the existing row and preserve usage unless reset was requested
[...]

PR #7362, opened 2026-09-20
fix(governance): preserve VK provider-config key associations when PUT omits key_ids

### LINKED ISSUE #7347: [Bug]: PUT /api/governance/virtual-keys/{id} silently sets allow_all_keys=false and drops key associations when key_ids is omitted (GET does not return key_ids)
### Prerequisites

- [x] I have searched existing issues and discussions to avoid duplicates (searched `allow_all_keys`, `key_ids`, `virtual key provider config PUT`, `no keys found for provider`; closest hits are #4640, #5672, #6620, #5683 — see "Related" below, none cover this PUT round-trip behaviour)
- [x] I am using the latest version (reproduced on `maximhq/bifrost:v2.1.1`, tag `transports/v2.1.1`, commit `c193745d`)

### Description

On v2.1.1 (`transports/v2.1.1`, `c193745d`), `GET /api/governance/virtual-keys/{id}` returns each provider config with `allow_all_keys` and `keys`, but **not** `key_ids`. `PUT /api/governance/virtual-keys/{id}` (`transports/bifrost-http/handlers/governance.go` ~L2234-2248) recomputes `AllowAllKeys` and `Keys` solely from the request's `key_ids`: absent/empty -> `allowAllKeys=false`, `keys=nil`.

So a read-modify-write of the GET body (changing only `description`, `allowed_models`, etc.) silently turns every provider config into deny-all-keys. Every subsequent request through that VK fails with `no keys found for provider: X and model: Y` (`core/bifrost.go`) and, if a routing rule with a fallback matches, is silently routed to the fallback provider. The PUT returns 200 and the response shows `allow_all_keys:false`, but nothing flags the change.

### Steps to reproduce

Setup: sqlite config store, `source_of_truth: "config.json"`, one VK with `provider_configs: [{provider: openai, key_ids: ["*"]}]`.

1. Boot; `SELECT provider,allow_all_keys FROM governance_virtual_key_provider_configs` -> `openai|1`.
2. `GET /api/governance/virtual-keys/<id>` -> `provider_configs[0] = {id:1, provider:openai, allow_all_keys:true, keys:[], allowed_models:["*"]}` (no `key_ids` in the response).
3. `PUT /api/governance/virtual-keys/<id>` with `{name, is_active, provider_configs: <exactly as returned by GET>}` -> **200**; response now shows `allow_all_keys:false`; DB row is now `openai|0`. With an explicit-key provider config (`key_ids:["key-anthropic-1"]`), the join rows in `governance_virtual_key_provider_config_keys` are also deleted.
4. `POST /v1/chat/completions` with the VK -> `{"error":"no keys found for provider: openai and model: gpt-4o-mini"}`.

With `source_of_truth=config.json` a restart restores the file values via `reconcileVirtualKeyAssociations` (confirmed), so this is API-side breakage until the next restart; in `split` mode it is permanent.

### Expected behavior

Either (a) an absent `key_ids` in a PUT `provider_config` leaves `allow_all_keys`/`keys` unchanged (treat omitted as "unchanged", consistent with how a nil `budget`/`rate_limit` is handled a few lines below in the same handler), or (b) the GET response includes `key_ids` (`["*"]` when `allow_all_keys`, else the key UUID list) so a GET->PUT round-trip is lossless. Ideally both.

### Actual behavior

Omitted `key_ids` is interpreted as "no keys"; the VK is silently disabled for that provider (`allow_all_keys` -> `false`, key associations deleted), with a 200 response.

### Impact

Any client/UI/automation that does GET->modify->PUT on a virtual key breaks inference for that VK with no error from the admin API. We hit this in production: "local" traffic fell into a routing-rule fallback to a cloud provider for ~19h before the `no keys found` error was noticed.

### Related (not duplicates)

- #4640 (closed, fixed by #4641): file-sync not re-syncing `provider_configs` from `config.json` — different code path (this issue is the admin API PUT handler).
- #5672 / #6620 (comment): `allow_all_keys: true` sent explicitly in the request body is not persisted. Likely the same root cause (handler derives `AllowAllKeys` only from `key_ids` and ignores the `allow_all_keys` field), but this report is specifically about the GET->PUT round-trip losing key associations when `key_ids` is omitted.

### Affected area(s)

Transports (HTTP), Framework (configstore)

### Version

transports v2.1.1 (image `maximhq/bifrost:v2.1.1`, commit `c193745d`)

### Environment

```text
Docker, maximhq/bifrost:v2.1.1, sqlite config store, source_of_truth=config.json
```

### Relevant logs/output

```shell
{"error":"no keys found for provider: openai and model: gpt-4o-mini"}
```

xiechimon:
## Summary

PUT /api/governance/virtual-keys/{vk_id} silently dropped provider-config key associations whenever the request body omitted `key_ids`. The GET response exposes `allow_all_keys` and `keys` but never `key_ids`, so the standard GET -> edit -> PUT round-trip (change a description, toggle active, etc.) flipped `AllowAllKeys` to false and detached every key. Inference through the virtual key then failed with `no keys found for provider`, with no hint that the PUT had rewritten the associations. Reproduces on v2.1.1 and current `dev` (see #7347 for the full transcript).

## Changes

- `transports/bifrost-http/handlers/governance.go`: in the existing-config branch of `updateVirtualKey`, only touch the key associations when the request actually carries `key_ids`. `schemas.WhiteList` is a slice, so after unmarshal `nil` (field absent) is distinguishable from `[]` (explicit deny-all): absent now means "unchanged", an explicit empty list still clears, an explicit list still replaces. Create-path semantics are untouched. The existing-config rows are now loaded with `Preload("Keys")` because the store update replaces the association with whatever the row carries; without the preload the preserved branch would write back an empty key set.
- `transports/bifrost-http/handlers/governance_test.go`: regression test covering all four transitions (allow-all survives omitted `key_ids`, explicit key list survives omitted `key_ids`, explicit `[]` still clears, explicit list still replaces), following the existing-file convention in AGENTS.md.
- `transports/changelog.md`: entry under the transports changelog.

## Type of change

- [x] Bug fix

## Affected areas

- [x] Transports (HTTP)

## How to test

```sh
cd transports/bifrost-http
go test ./handlers/ -run TestUpdateVirtualKey_PutWithoutKeyIDsPreservesKeyAssociations -count=1
go test ./handlers/ -count=1   # full suite, all green
go vet ./handlers/
```

Red/green verified: with only the test added on top of `dev`, the test fails at `omitted key_ids wiped allow_all_keys`; after the handler change it passes, and reverting each of the two hunks separately reproduces a distinct failure (the `Preload("Keys")` hunk is pinned by the "detached the existing key" assertion).

## Breaking changes

None. Clients that always send `key_ids` see identical behavior.

## Harness-case note (AGENTS.md)

This change is wire-visible only on the management API (`/api/governance/*`), which the provider-harness collection does not exercise at all (zero `api/governance` requests in `provider-harness.json`; governance flows live in `bifrost-v1-mgmt-flows.postman_collection.json`, run via `make run-e2e-api`). The Go regression test above pins the round-trip through the real handler and a real SQLite store, and the merged governance precedents (#7291, #6932) shipped the same shape: handler + `governance_test.go`, no provider-harness case. If you would still like a newman case in the mgmt-flows collection (create VK -> PUT without `key_ids` -> GET -> assert `allow_all_keys` survives), say the word and I will add it.

Fixes #7347


xiechimon:
recheck

akshaydeo (approved):

PR #7365, opened 2026-09-21
[fix]: Anthropic passthrough - narrow caller Accept-Encoding instead of dropping it

### LINKED ISSUE #7364: Anthropic passthrough drops caller Accept-Encoding, forcing identity on the upstream hop
Following `docs/contributing/raising-a-pr.mdx` ("create an issue first"). I have a patch ready and will open a PR against `dev` referencing this issue.

## What happens

`handlePassthrough` (`transports/bifrost-http/integrations/router.go`) lists `accept-encoding` among the headers it drops:

```go
case "api-key", "x-api-key", "x-goog-api-key",
    "host", "connection", "transfer-encoding", "cookie", "set-cookie", "proxy-authorization", "accept-encoding":
```

So every passthrough request reaches the provider with no `Accept-Encoding` at all, and the upstream answers with identity.

Observed with Anthropic OAuth passthrough: a Claude Code client advertises `gzip, deflate, br, zstd`, and the gateway-to-Anthropic hop carries an uncompressed body (roughly 3.8x the bytes on our measurements). Nothing about the protocol requires this — it is purely the dropped header.

## Why it was dropped (this part looks deliberate)

Passthrough writes `SafeHeaders` onto the upstream request verbatim (`core/providers/anthropic/anthropic.go`). Forwarding the caller header as-is would let a streaming caller ask for `br` or `zstd`; `DecompressStreamBody` cannot decode those, so the SSE parser would see raw compressed bytes. Dropping the header is the simplest way to guarantee a decodable response.

## Why it can be fixed rather than just documented

The codebase already solves this exact problem for the non-passthrough OAuth path. `setPassthroughHeaders` in `core/providers/utils/utils.go` intersects the caller list with `supportedBufferedContentEncodings` / `supportedStreamingContentEncodings`, and pins `identity` when the intersection is empty — deleting the header would mean "any coding is acceptable" (RFC 9110 12.5.3), which is wider, not narrower.

Passthrough appears to be the only path that does not reuse that mechanism.

## Proposed fix

Export that logic as `PinAcceptEncodingForPassthrough(safeHeaders, streaming)`, call it from both Anthropic passthrough paths (the streaming one with the gzip-only set), and stop dropping the header in the router.

Happy to adjust the shape if you would rather see it live somewhere else.

## Environment

Reproduced against `dev` at `94fb9ee`. Deployed build is v2.2.1.

senlin1003:
Closes #7364.

## Problem

`handlePassthrough` drops `accept-encoding` from the caller's headers, so every passthrough request reaches the provider with no `Accept-Encoding` at all and the upstream answers with identity. Observed with Anthropic OAuth passthrough: a Claude Code client advertises `gzip, deflate, br, zstd`, and the gateway-to-Anthropic hop carries an uncompressed body for no protocol reason.

## Why it was dropped

This looks deliberate rather than an oversight. Passthrough writes `SafeHeaders` onto the upstream request verbatim, so forwarding the caller header as-is would let a streaming caller ask for `br` or `zstd`, which `DecompressStreamBody` cannot decode — the SSE parser would then see raw compressed bytes. Dropping the header is the simplest way to guarantee a decodable response.

## Why this is a gap rather than a design choice

The codebase already solves this exact problem for the non-passthrough OAuth path. `setPassthroughHeaders` intersects the caller list with `supportedBufferedContentEncodings` / `supportedStreamingContentEncodings` and pins `identity` when the intersection is empty — deleting the header would mean "any coding is acceptable" (RFC 9110 12.5.3), which is wider rather than narrower. Passthrough is the one path that never reused it.

## Change

- `core/providers/utils/utils.go` — expose the narrowing as `PinAcceptEncodingForPassthrough(safeHeaders, streaming)`, reusing the existing `filterSupportedAcceptEncodings` and the same identity-pin rule.
- `core/providers/anthropic/anthropic.go` — call it from both passthrough paths, the streaming one with the gzip-only set.
- `transports/bifrost-http/integrations/router.go` — stop dropping the header, with a comment pointing at where it is now narrowed.

## Tests

`core/providers/utils/passthrough_accept_encoding_test.go` covers the streaming/buffered split, the identity pin on an empty intersection, case-insensitive header matching, q-value preservation, and a nil map.

`go build ./...` and `go vet` pass in `core`; `go test ./providers/anthropic/...` passes.

Two pre-existing failures on my machine, unrelated to this change and reproduced on `94fb9ee` before it:
- `TestStreamingPoolSurvivesCancelledStreams` (440 concurrent connections against a local listener, Windows).
- `transports` does not build against the local `core` at this commit; its `go.mod` pins a published version predating `BifrostContextKeySessionAffinity`.

Happy to move the helper elsewhere or reshape the API if you would prefer a different arrangement.


senlin1003:
Closing this on my side — withdrawing the proposal for now, not because of anything about the change itself. Apologies for the noise.

Leaving the analysis in #7364 in case it is useful to anyone: passthrough is the one path that does not reuse the existing `filterSupportedAcceptEncodings` narrowing, and the identity-pin rule there already handles the streaming decode concern. Anyone is welcome to take it from there.

